Issue #83, opened 2024-08-08
Provide better API for `libclang`'s command line arguments


edsko:
We currently configure `libclang` by passing a raw set of command line arguments (of type `[String]`). We should have a proper Haskell datatype that lists the `libclang` options we are interested in (and then internally translate this to `[String]`).

This is part of #10 and #71.

edsko:
Perhaps some of the options we should support are the ones mentioned in the docs of 'CXDiagnosticDisplayOptions' (added in #176). However, this is only useful if we don't write our own renderer for `libclang` options (#174), and instead use `diagnosticFormatted`/`clang_formatDiagnostic`.

edsko:
I'm going to define a record with some fields as a starting point for this.

edsko:
Closing; we now have [`ClangArgs`](https://github.com/well-typed/hs-bindgen/blob/b6dc7224cd08376e4d7a0e5e4a6d11e261a27a2b/clang/src/Clang/Args.hs#L31). If it is missing specific arguments, we can open individual tickets for those.

PR #98, opened 2024-08-19
Issue #83: Better API for libclang command line arguments

phadej:
Technically this solves #83, as nothing breaks.

I think we should have some tests first. `tasty-golden` seems like a good idea.


edsko:
Stretching the definition of "technically solves" here :grimacing: Agreed that starting on a test harness is useful. Perhaps an initial way (and most important) to actually use the command line arguments is to configure a target platform. One thing that this should affect is the `sizeof` and `alignof` fields of the parsed structs.

edsko:
Obsoleted by #196, which takes a first step towards defining a record type for `ClangArgs` (by no means exhaustive).

PR #100, opened 2024-08-21
Configure

### LINKED ISSUE #78: Automatically discover location of `libclang`
Unfortunately, `libclang` does not support `pkg-config` (https://github.com/llvm/llvm-project/issues/9777). [`clang-pure`](https://hackage.haskell.org/package/clang-pure) uses a [custom setup script](https://hackage.haskell.org/package/clang-pure-0.2.0.6/src/Setup.hs) which calls [`llvm-config`](https://llvm.org/docs/CommandGuide/llvm-config.html), we might consider doing something similar. For now, we simply depend on something like this in `cabal.project`:

```
package hs-bindgen
  extra-lib-dirs:
    /usr/lib/llvm-18/lib
  extra-include-dirs:
    /usr/lib/llvm-18/include
```

phadej:
Add hs-bindgen-libclang with build-type: Configure

Resolves #78.

Having C-dependencies in a separate package work-around unclear
behavior of `build-type: Configure` which seems to add extra definitions
to the main library only (`a-lib.buildinfo`; possibly we can have extra
buildinfos for other components too, but the file naming scheme is
mysterious).

On the other hand, separate package for `libclang` bindings might be a
good idea anyway. It's a separate self-contained functionality,
and e.g. `clang-tutorial` depends on that only (and could be pulled into
separate package as well, not to pollute the main thing).


edsko:
I'll merge this as is, but I don't  fully understand how it works. I guess the `configure` script is generated, but I don't know what it was generated _from_. What if we want to update it?

It's a bit unfortunate we now depend on all of `llvm`, but it's not a big deal; have opened a separate issue about this (#102). 

I'm not sure I understand this comment:

> Having C-dependencies in a separate package work-around unclear behavior of build-type: Configure which seems to add extra definitions to the main library only (a-lib.buildinfo; possibly we can have extra buildinfos for other components too, but the file naming scheme is mysterious). On the other hand, separate package for libclang bindings might be a
good idea anyway. It's a separate self-contained functionality

I certainly don't object to moving the libclang stuff to `hs-bindgen-libclang` (and then I like the name you chose, we should avoid falling into the trap of trying to write general purpose `libclang` bindings). What I am unsure about is whether that will re-introduce the problem you mention about the extra definitions? 

> and e.g. clang-tutorial depends on that only (and could be pulled into separate package as well, not to pollute the main thing).

Agreed, although we can also just remove it entirely. It was primarily useful to check my initial `libclang` bindings, it's served it's purpose now,  although we could also keep it as a demonstration of how to use `hs-bindgen-libclang`, as it is a direct re-implementation of the "official" `libclang` tutorial.

phadej:
> but I don't know what it was generated from

`configure.ac`. With `autoreconf -i`

> What I am unsure about is whether that will re-introduce the problem you mention about the extra definitions?

The problem with `hs-bindgen` package was that `clang-tutorial` didn't depend on the `hs-bindgen` library, so the C dependency  info (includes, linkage) was not propagated/shared. As long as other components depend on the main library of the package, there won't be any problems.

edsko:
> > but I don't know what it was generated from
> 
> `configure.ac`. With `autoreconf -i`

Is that `configure.ac` then something we should add to the repo? Or am I just overlooking it?

> > What I am unsure about is whether that will re-introduce the problem you mention about the extra definitions?
> 
> The problem with `hs-bindgen` package was that `clang-tutorial` didn't depend on the `hs-bindgen` library, so the C dependency info (includes, linkage) was not propagated/shared. As long as other components depend on the main library of the package, there won't be any problems.

Ah! Gotcha. I think it was using it because the `libclang` stuff isn't really part of the `hs-bindgen` API, which is also solved by moving the bindings.  I'll do that.

edsko:
By the way, I see

```
Configuring hs-bindgen-libclang-0.1.0...
configure: WARNING: unrecognized options: --with-compiler
```

Can we avoid this?

PR #103, opened 2024-08-22
Only build main in CI

### LINKED ISSUE #101: Build only main branch in CI
Otherwise each PR will build the branch and merge commit, i.e. everything twice.

... and if you need to  build a branch, opening a (draft) PR is always possible.

phadej:
Resolves #101

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -14,8 +14,12 @@
 #
 name: Haskell-CI
 on:
-  - push
-  - pull_request
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
 jobs:
   linux:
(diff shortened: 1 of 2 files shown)


edsko:
Perfect, thank you sir :)

phadej:
FWIW and FYI the `haskell-ci` doesn't do that by default because
- some people don't like it, and I don't care about default that much
- the *main* branch is not always `main` (it was `master`, but could also be `develop` or ...); in fact there could be multiple main tips

edsko:
> FWIW and FYI the `haskell-ci` doesn't do that by default because
> 
> * some people don't like it, and I don't care about default that much
> * the _main_ branch is not always `main` (it was `master`, but could also be `develop` or ...); in fact there could be multiple main tips

Gotcha. I think what you've done here makes sense, I will do the same in `grapesy`.

PR #104, opened 2024-08-22
My bad, forgot configure.ac

phadej:


edsko:
Oh, wow, this is still non-trivial. Thanks for doing this!

PR #107, opened 2024-08-22
Make clang-tutorial into test-suite

phadej:
This is pattern I use often, make example executables into test-suites. This way they are not affecting cabal-install solver, so we don't need a flag. We only need to make sure they do something sensible without any arguments.

Another pattern I use is to name executable entry points as `executable.hs` (i.e. not `Main.hs`); as it's easy to end up with dozens of `Main.hs` files.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,6 +1,3 @@
 packages: hs-bindgen-patterns
 packages: hs-bindgen
 packages: hs-bindgen-libclang
-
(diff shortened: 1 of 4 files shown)


edsko:
Works for me. I think in the past HLS didn't like `Main` modules living in files not called `Main.hs`, but we can cross that bridge if and when it comes up.

PR #109, opened 2024-08-22
Simple golden test

phadej:
I changed `other-modules` to `exposed-modules`. For a time being we don't know what the public interface should look like, so hiding stuff only gets in a way of doing stuff.


edsko:
I'll merge this (and will then delete `foldDumpAST` in favour of your much nicer `foldShowAST`). If we can find a way to keep the exposed/non-exposed distinction I'd prefer it though. I agree with you that the public API isn't really known yet and we'll definitely have to adjust as we go, but I worry that if we do this after the fact, we're not thinking enough about what it should look like and might make different decisions. Like you pointed out earlier, it's important that we think carefully about `hs-bindgen`-as-a-library.

PR #110, opened 2024-08-23
Enable tests

edsko:


edsko:
So here is something interesting: these tests fail on my machine:

```diff
-"" :: "Record"
+"S3_t" :: "Record"
   "a" :: "Char_S"
 "S3_t" :: "Typedef"
-  "" :: "Record"
+  "S3_t" :: "Record"
     "a" :: "Char_S"
```

This seems related to this comment I added to `parseStruct`:

```haskell
-- Implementation note: It seems libclang will give us a name for the struct if
-- the struct it a tag, but also when it's anonymous but the surrounding typedef
-- has a name.
```

Apparently this behaviour is not the same for older versions of `llvm`. 

@phadej do you have an opinion on how to address this?

phadej:
> how to address this?

I don't have good ideas. Probably we need to support *and test* various llvm versions. (`configure.ac` has a comment to test version, its relatively easy to add). Ubuntu 22.04 provides LLVM-11..15 and 14 is the default atm.  Ubuntu 24.04 provides LLVM-14..18. (https://www.haskell.org/ghc/download_ghc_9_10_1.html) supports 11-15. 

Recent macOS/XCode seems to have 16 or 17 based on https://en.wikipedia.org/wiki/Xcode#Xcode_15.0_-_(since_visionOS_support)_2

... so it looks like there is no reasonable overlap to support all scenarios.

phadej:
`configure` could also generate some header with version CPP defines, so the checks can be made in code.

edsko:
This _particular_ problem with `llvm-14` versus `llvm-18` is resolved in https://github.com/well-typed/hs-bindgen/pull/120 . We may well run into others, of course.

PR #111, opened 2024-08-23
Delete `foldDumpAST` in favour of `foldShowAST`

edsko:


edsko:
@phadej I've dropped `foldDumpAST` in favour of your `foldShowAST`, and now export this as part of the public API. Since it was easy, I've changed the golden tests to use `HsBindgen.Lib` again and made the rest un-exposed; if that gets annoying again, feel free to change again.

hs-bindgen/src/HsBindgen/Lib.hs
```diff
@@ -92,11 +93,11 @@ parseCHeader ::
...
     WrapCHeader . C.Header <$> C.parseHeaderWith args fp (C.foldDecls tracer)
 
--- | Dump the raw @libclang@ AST
+-- | Show the raw @libclang@ AST
 --
 -- This is primarily for debugging.
-dumpClangAST :: ClangArgs -> FilePath -> IO ()
-dumpClangAST args fp = const () <$> C.parseHeaderWith args fp C.foldDumpAST
+showClangAST :: ClangArgs -> FilePath -> IO (Forest String)
+showClangAST args fp = C.parseHeaderWith args fp C.foldShowAST
```
phadej, line 100: Arguably `parseHeaderWith` and `foldShowAST`  should be part of public API. Maybe not as part of the main module, but these are kind of internal-ish but still generally useful functions (which you need to make a slightly different behavior from what the library provides) that people argue that internal modules should always be exposed.

PR #112, opened 2024-08-23
Basic comment parsing support

edsko:
We can probably do better than this for Doxygen specifically.

@phadej The fold that returns the raw `libclang` AST now returns a `Forest Element`, where `Element` is defined as

```haskell
data Element = Element {
      elementDisplayName      :: Strict.ByteString
    , elementTypeKind         :: SimpleEnum CXTypeKind
    , elementTypeKindSpelling :: Strict.ByteString
    , elementRawComment       :: Strict.ByteString
    , elementBriefComment     :: Strict.ByteString
    }
  deriving (Show)
```

I've added a `showElem` function to the golden tests that recovers the existing output, so the `fixtures` are not (yet) updated.

### DIFF
--- a/hs-bindgen/src/HsBindgen/C/Parser.hs
+++ b/hs-bindgen/src/HsBindgen/C/Parser.hs
@@ -6,7 +6,10 @@
 module HsBindgen.C.Parser (
     parseHeaderWith
   , foldDecls
-  , foldShowAST
+    -- * Debugging
+  , Element(..)
+  , foldClangAST
+    -- * Logging
   , ParseMsg(..)
   ) where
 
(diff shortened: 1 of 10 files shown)


hs-bindgen/src/HsBindgen/C/Parser.hs
```diff
@@ -139,6 +142,16 @@ primType = either (const Nothing) aux . fromSimpleEnum
   Debugging
 -------------------------------------------------------------------------------}
 
+-- | An element in the @libclang@ AST
+data Element = Element {
+      elementDisplayName      :: Strict.ByteString
```
phadej, line 147: Should these be strict fields?

hs-bindgen/src/HsBindgen/Lib.hs
```diff
@@ -138,3 +135,13 @@ preprocess tracer clangArgs inp modOpts renderOpts out = do
     modl <- genModule modOpts <$> parseCHeader tracer clangArgs inp
     prettyHs renderOpts out modl
 
+{-------------------------------------------------------------------------------
+  Debugging
+-------------------------------------------------------------------------------}
+
+-- | Show the raw @libclang@ AST
```
phadej, line 142: This is not true anymore.

hs-bindgen/src/HsBindgen/C/Parser.hs
edsko: I generally don't bother with making things strict unless there is a good reason.

hs-bindgen/src/HsBindgen/Lib.hs
edsko: Good point, will fix.

hs-bindgen/src/HsBindgen/C/Parser.hs
phadej: I rather do opposite. Every field is strict, except if there is a reason for it to be lazy. Strict data, lazy functions.

edsko: That works for me, though it gets annoying when it involves things like lists, `Maybe`, etc. etc.

PR #118, opened 2024-08-28
More libclang bindings

edsko:
This is a failed attempt at two things:

1. Distinguish between the _tag_ of a struct (which may be absent) and the name of the surrounding typedef. In llvm18 the struct tag (at least when using `clang_getCursorSpelling` or `clang_getCursorDisplayName`) is set to the name of the surrounding typedef, which is causing some compatibility trouble against different llvm vesions (https://github.com/well-typed/hs-bindgen/pull/110#issuecomment-2306844665). So far the only way that I've found to figure out if this name is a "fake" name is by asking for the source location, then looking up that source location in the file, and looking at what's there; for "fake" names, the source location will point to the `struct` keyword instead of the actual name. But this feels a bit ugly. Neither `clang_Type_isTransparentTagTypedef` nor `clang_Cursor_isAnonymous` seems to help here; since the docs of the former refer to `NS_ENUM`, perhaps that is an Objective C thing; the latter refers to structs that are truly anonymous (see `examples/anonymous.h` for an example).
2. Distinguish between the definition of a struct, and its reference in a typedef, for example as in

    ```c
    typedef struct {
        char a;
    } S3_t;
    ```
  
    When we fold over this AST, we see this struct definition twice; first as a standalone thing, and then repeated inside the body of the `typedef`. I was hoping that we'd be able to use `clang_isCursorDefinition` to distinguish between these two cases, but it seems to report `True` for absolutely everything.

### DIFF
--- a/hs-bindgen/examples/anonymous.h
+++ b/hs-bindgen/examples/anonymous.h
@@ -1,0 +1,9 @@
+// Struct containing an anonymous inner struct
+struct S1 {
+  struct {
+    int a;
+    int b;
+  } c;
+
+  int c;
+};
--- a/hs-bindgen-libclang/cbits/clang_wrappers.c
+++ b/hs-bindgen-libclang/cbits/clang_wrappers.c
@@ -61,6 +61,16 @@ CXString* wrap_malloc_Cursor_getBriefCommentText(CXCursor* C) {
     return result;
 }
 
+unsigned wrap_isCursorDefinition(CXCursor *C) {
+    return clang_isCursorDefinition(*C);
+}
+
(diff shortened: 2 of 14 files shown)


hs-bindgen/examples/anonymous.h
```diff
@@ -0,0 +1,9 @@
+// Struct containing an anonymous inner struct
```
phadej, line 1: Not added to test-suite on purpose?

edsko: Yeah, I guess I could. Right now they are mostly tiny examples that I'm using for ad-hoc tests for the additional bindings. Probably actually that `anonymous` thing should just be yet another example in the example structs, with a separate file for testing all the Doxygen stuff.

PR #120, opened 2024-08-29
Introduce `getUserSuppliedName`

edsko:
Consider

```c
// anonymous struct with typedef
typedef struct {
    char a;
} S3_t;
```

Previously we were returning `"S3_t"` for the tag of this struct, because that was the value supplied by `libclang` -- that is, with `llvm-18`. With `llvm-14`, this tag was actually reported as `""`. Of course, in actuality no tag was provided at all, but the `libclang` API did not make that very easy to figure out. We now provide `getUserSuppliedName`, which compares the name as provided by `libclang` to the actual token in the tree, and if they don't match, we conclude that the user provided no name at all. Perhaps there is a better way to do this, but if so, we can just change the implementation, and keep the contract.

This still leaves the problem that when we traverse the AST for this struct, we see the strict twice: once without the `typedef`, and then again as a child _of_ the `typedef`. I don't yet know how to address that; perhaps we should just ignore a struct definition with no tag: after all, such a struct is _only_ useable inside a `typedef`, and therefore we know that we will see it again. That feels like a reasonable option, but haven't explored that in this PR yet.


edsko:
@phadej do you happen to know what's going on with the Windows failure?

hs-bindgen-libclang/src/HsBindgen/Clang/Aux/Fold.hs
```diff
@@ -1,13 +1,12 @@
 -- | Higher-level bindings for traversing the API
-module HsBindgen.Clang.Core.Util.Fold (
+module HsBindgen.Clang.Aux.Fold (
```
phadej, line 2: Are you going to rename this module in every patch? :)

edsko: Possibly? :)

edsko: Ah, sorry, I'm being slow on the uptake. Thank you for the clue :) Will fix.

edsko: Have renamed `Aux` back to `Util`, to avoid the name `Aux` which is illegal on Windows. The fact that only the `Classification.hs` was reported as a problem threw me off the scent (but of course in hindsight it was just the _first_ problematic file).

phadej: I opened a cabal-install issue https://github.com/haskell/cabal/issues/10295, that would helped figure out the issue. (Without CI we wouldn't find it until possibly too late).

edsko: Ah, good idea, thanks :)

PR #125, opened 2024-09-01
Initial enum support

phadej:


edsko:
I've merged his but I'm not sure I fully understand: what is the meaning of `sizeof` and `alignof` for `enum` members?

PR #127, opened 2024-09-01
Primitive types "failing test"

phadej:


phadej:
The failing examples are interesting. We probably need to make the golden tests somehow machine dependent; their generation on development machines will become tricky though. On a "positive" side, we not have a first example where targets behave differently.

phadej:
One option is to actually try do cross compilation stuff now. That would help move #83 / #98 forward, as cross compilation flags were the ones we care about in `ClangArgs` AFAIU

edsko:
I'd certainly be in favour of starting to look at cross compilation now; it's a critical requirement, and taking it into account sooner rather than later may well help avoid some design mistakes.

PR #131, opened 2024-09-01
Stdint

phadej:


edsko:
The segfault _happens_ to be gone on `master`, but only because we do `SelectFromMainFile`; when changing the golden tests to use `SelectAll` instead, the segfault re-appears. Will dig a bit more.

edsko:
@phadej I believe the segfault is resolved in #147 , so you can rebase this PR (and add the new fixtures).

phadej:
Rebased

phadej:
Oh, interesting. The target triple is not enough. We also need the corresponding (standard) headers! Obvious in hindsight.

edsko:
> Oh, interesting. The target triple is not enough. We also need the corresponding (standard) headers! Obvious in hindsight. Not sure what to do though.

I am not sure if I fully understand you here. Is the problem that on different architectures we need to include different headers (`stdint.h` on Linux, `foo.h` on OSX..), or rather that the header is the same on all architectures but it's simply not available on CI? I guess it's the latter rather than the former; if true, what makes it difficutl to make these headers available?

phadej:
From failing osx job

```diff
      -  "FieldDecl" "sixty_four" :: "Typedef"
      -    "TypeRef" "uint64_t" :: "Typedef"
      -  "FieldDecl" "thirty_two" :: "Typedef"
      -    "TypeRef" "uint32_t" :: "Typedef"
      +  "FieldDecl" "sixty_four" :: "Int"
      +  "FieldDecl" "thirty_two" :: "Int"
```

The way `uint64_t` / `uint32_t` are defined seems to be different on different arch.

edsko:
My plan is that one I can generate the prelude stuff, `uint32_t` and `uint64_t` will be recognized as such, and appear as such, in the AST, and so the platform specific differences in how they are represented will not be visible in the AST anymore. Working on this now.

Issue #10, opened 2024-08-01
Support cross-compilation


edsko:
We should not assume that the machine we are generating bindings for is the same as the host machine. Things like `alignment` requirements for `Storable` instance may for example be different. This will probably primarily mean passing the right flags to `libclang`.

PR #137, opened 2024-09-01
Add target info bindings

phadej:


phadej:
FWIW, `llvm-config --host-target` returns the host target triple. If we'd always require llvm-config (we don't cause Windows), we could always be explicit (re #10) and pass that. Maybe there's some way to dig out host target triple from LLVM lib, but I cannot find a way from  `libclang`.

PR #141, opened 2024-09-03
Add enum values

phadej:


edsko:
I'll merge this as-is, but the sections in `.Core` are meant to correspond to the various sections of the `libclang` docs (see module header at the top, which explains the general structure). Don't worry about fixing it though, I'll submit a separate PR to do so.

PR #146, opened 2024-09-04
Consistent error handling in the libclang bindings

edsko:


hs-bindgen-libclang/src/HsBindgen/Clang/Core.hs
```diff
@@ -688,17 +692,26 @@ clang_Cursor_isAnonymous cursor =
...
+-- declaration.
 --
 -- <https://clang.llvm.org/doxygen/group__CINDEX__TYPES.html#ga6b8585818420e7512feb4c9d209b4f4d>
-clang_getEnumConstantDeclValue :: CXCursor -> IO CLLong
-clang_getEnumConstantDeclValue cursor = do
+clang_getEnumConstantDeclValue ::
+     HasCallStack
+  => CXCursor -- ^ Parent
+  -> CXCursor -> IO CLLong
+clang_getEnumConstantDeclValue parent cursor = do
```
phadej, line 703: I don't like this
- Now the *unchecked* precondition is that `parent` is indeed parent of a `cursor`. 
- The binding is not low-level anymore (not even "low-level with some extra checks").
    - I'd rather experiment with clang_getCursorSemanticParent()

edsko: I don't mind, if `clang_getCursorSemanticParent` works, also fine with me, but "parents" are not my invention, they are provided explicitly by [`clang_visitChildren`](https://clang.llvm.org/doxygen/group__CINDEX__CURSOR__TRAVERSAL.html#ga5d0a813d937e1a7dcc35f206ad1f7a91), so in that sense it's a natural use of the API that libclang provides.

phadej: Having parent in `Fold` is ok. But I'd try to keep the actual bindings as close original bindings as possible. `clang_getEnumConstantDeclValue` doesn't take a parent argument.

edsko: Ok, I will try `clang_getCursorSemanticParent` (but first looking into the segfault in #131).

edsko: Done in #148.

PR #149, opened 2024-09-04
Avoid leaks due to async exceptions

### LINKED ISSUE #136: packCXString =<< foo pattern may leak
`foo` allocates with malloc. `packCXString` disposes and frees. If async exception comes in between `foo` and `packCXString` some memory is leaked.

edsko:
@phadej I'd appreciate your eyes on this. 

Closes #136.

### DIFF
--- a/hs-bindgen-libclang/src/HsBindgen/Clang/Core.hs
+++ b/hs-bindgen-libclang/src/HsBindgen/Clang/Core.hs
@@ -133,11 +133,13 @@ module HsBindgen.Clang.Core (
   , CXCursor_(..)
   ) where
 
+import Control.Exception
 import Control.Monad
 import Data.ByteString (ByteString)
 import Data.ByteString qualified as BS
 import Foreign
 import Foreign.C
+import Foreign.Concurrent qualified as Concurrent
 import GHC.Stack
 import System.IO.Unsafe (unsafePerformIO)
 
@@ -295,7 +297,7 @@ clang_parseTranslationUnit cIdx src args options =
 clang_TargetInfo_getTriple :: HasCallStack => CXTargetInfo -> IO ByteString
 clang_TargetInfo_getTriple info =
     ensure (not . BS.null) $
-      packCXString =<< wrap_malloc_TargetInfo_getTriple info
+      packCXString $ wrap_malloc_TargetInfo_getTriple info
 
 {-------------------------------------------------------------------------------
   Cursor manipulations
@@ -325,7 +327,7 @@ newtype CXCursor_ = CXCursor_ (Ptr ())
 --   AST.
 --
 -- <https://clang.llvm.org/doxygen/structCXCursor.html>
-newtype CXCursor = CXCursor (ForeignPtr CXCursor_)
+newtype CXCursor = CXCursor (ForeignPtr ())
   deriving IsForeignPtr via DeriveIsForeignPtr CXCursor_ CXCursor
 
 foreign import capi unsafe "clang_wrappers.h wrap_malloc_getTranslationUnitCursor"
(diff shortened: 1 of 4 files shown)


edsko:
Critical changes are: 

* definition of `packCXString`
* definition of `wrapForeignPtr` (in `IsForeignPtr` instance for `DeriveIsForeignPtr`)
* definition of `clang_visitChildren`

edsko:
Perhaps we should instead of masking async exceptions _preallocate_ the buffer Haskell side. Especially during the traversal (maybe only there?) that might be easier. I need to think about this some more.

phadej:
> Perhaps we should instead of masking async exceptions preallocate the buffer Haskell side.

Yes, one could have `newtype CXCursor = CXCursor ByteArray`. I do something like that e.g. in `hashable`'s `xxhash` bindings for hash state (https://github.com/haskell-unordered-containers/hashable/blob/57e0d06623d9f2c8a0c5ff13e04b5a4002d699f5/src/Data/Hashable/XXH3.hs#L72), though there state is mutable the concept is the same.

hs-bindgen-libclang/src/HsBindgen/Clang/Core.hs
```diff
@@ -325,7 +327,7 @@ newtype CXCursor_ = CXCursor_ (Ptr ())
 --   AST.
 --
 -- <https://clang.llvm.org/doxygen/structCXCursor.html>
-newtype CXCursor = CXCursor (ForeignPtr CXCursor_)
+newtype CXCursor = CXCursor (ForeignPtr ())
```
phadej, line 330: is this change necessary?

edsko: It brings this in line with all the other definitions, this was just a bug.

phadej: It feels that *others* are buggy. We have

```haskell
insance IsForeignPtr CXCursor where
  type UnderlyingPtr CXCursor :: Ptr CXCursor_
  
   unwrapForeignPtr :: forall a. CXCursor -> (Ptr CXCursor_ -> IO a) -> IO a  
```

but we `coerce :: Ptr () -> Ptr CXCursor`, this is ok, there is `castPtr` and the `Ptr` (and `ForeignPtr` arguments are phantom, but it feels wrong.

edsko: I don't think so..? 

```haskell
newtype DeriveIsForeignPtr u p = DeriveIsForeignPtr p

instance (Coercible p (ForeignPtr a), Coercible u (Ptr a))
      => IsForeignPtr (DeriveIsForeignPtr u p) where
  type UnderlyingPtr (DeriveIsForeignPtr u p) = u
```

so

```haskell
newtype CXCursor = CXCursor (ForeignPtr CXCursor_)
  deriving IsForeignPtr via DeriveIsForeignPtr CXCursor_ CXCursor
```

results in

```haskell
instance IsForeignPtr CXCursor where
  type UnderlyingPtr CXCursor = CXCursor_
```

not `Ptr CXCursor_`; note that `CXCursor_` is _already_ a `Ptr ()`; `Ptr CXCursor_` would be a pointer to a pointer. 

Am I wrong..?

(I'll experiment with your `ByteArray` suggestion though, which will change this anyway.)

phadej: I'd argue that we should have `data CXCursor_` be a tag for the struct. And use `Ptr CXCursor_` to indicate pointer to that struct. `Ptr ()` is writing all bindings with `void *`. Fine, but not precise. `CApiFFI` has less to work with as `void *` pointers can be freely cast.

edsko: I'd be okay with that also. I'll look at the `ByteArray` thing and also reconsider this.

PR #150, opened 2024-09-04
Introduce `clang_getCursorKind`

edsko:
We were using `clang_getCursorType` to determine what kind of AST element we were dealing with, but that's not quite right; we should instead be using `clang_getCursorKind`. For example, this tells us that we are dealing with an enum declaration (which happens to be reported as type enum); this matters in some cases, for example, a macro definition does not have any associated type (reported as "invalid"), but is reported as cursor kind "macro definition".


hs-bindgen/fixtures/enums.dump.txt
```diff
@@ -1,20 +1,20 @@
...
-  "SAME_B" :: "Int"
-    _ :: "Int"
-"packad" :: "Enum"
-  "packed" :: "Invalid"
-  "PACKED_A" :: "Int"
-  "PACKED_B" :: "Int"
-  "PACKED_C" :: "Int"
+"macro definition" "ENUMS_H" :: "Invalid"
+"EnumDecl" "first" :: "Enum"
+  "EnumConstantDecl" "FIRST1" :: "Int"
```
phadej, line 3: does this mean that we don't need to check parent when getting enum value, checking the kind of the cursor is enough?

edsko: Ah, yes, well-spotted. Will fix.

PR #154, opened 2024-09-05
Preallocate in Haskell rather than allocate in C

edsko:
This avoids the possibility of an async exception resulting in a finalizer not being attached. 

This also adds a `render-comments` mode to the command line, which is (for now anyway) useful for debugging dealing with comments.

Obsoletes #149.


hs-bindgen-libclang/cbits/Makefile
```diff
@@ -6,8 +6,8 @@ all: clang_wrappers.o doxygen_wrappers.o
 clang_wrappers.o: clang_wrappers.c clang_wrappers.h
 	gcc -c -Wall clang_wrappers.c -I/usr/lib/llvm-14/include
 
-doxygen_wrappers.o: doxygen_wrappers.c doxygen_wrappers.h
-	gcc -c -Wall doxygen_wrappers.c -I/usr/lib/llvm-14/include
+doxygen_wrappers.o: doxygen_wrappers.h
+	gcc -c -Wall doxygen_wrappers.h -I/usr/lib/llvm-14/include
```
phadej, line 10: This doesn't look right.

edsko: Yeah, it will, but the only purpose of this `Makefile` is for me to make it a bit easier to get all compiler warnings.

Issue #133, opened 2024-09-01
StructField should include an offset


phadej:
... even if we don't yet support bitnarrowed fields. See #132 

```c
struct S4 {
    char b;
    int a;
};
```

```haskell
      DeclStruct
        Struct {
          structTag = Just "S4",
          structSizeof = 8,
          structAlignment = 4,
          structFields = [
            StructField {
              fieldName = "b",
              fieldType = PrimChar},
            StructField {
              fieldName = "a",
              fieldType = PrimInt}]}])
```

the total size is 8 not 1+4=5.

PR #157, opened 2024-09-06
Haskell AST

edsko:
This is just a first sketch so far.


hs-bindgen/src/HsBindgen/Hs/AST.hs
```diff
@@ -0,0 +1,85 @@
...
+
+{-------------------------------------------------------------------------------
+  Structs
+-------------------------------------------------------------------------------}
+
+-- | Construct value of a struct
+data IntroStruct f = IntroStruct Free [Expr f]
+
+-- | Pattern match on a struct
+data ElimStruct f a = ElimStruct Free ([f Bound] -> a)
```
phadej, line 41: I'm not sure I like HOAS. It's not `Show`able or generally inspectable. Also here, the arity is not encoded.

edsko: This is not HOAS, it's PHOAS, and therefore it is in fact showable. And yes, arity isn't encoded; I'm not sure exactly how precise we want to be in this AST. I nearly have a TH translation for this done (along with some cleanup). We can discuss details then.

phadej: derive `Show`. I'd like to write tests comparing the ASTs (i.e. golden tests) before backends. I only see that possible using a "mock" backend.

edsko: Yes, I agree that we should have `Show` instances.

edsko: I think actually we should just have `ToExpr` instances in the main library, rather than declaring them as orphans.

phadej: For users of the `hs-bindgen` as library (if they use TH generation), each extra dependency is an extra dependency.

edsko: Yeah, that's a fair point.

phadej: Especially as `tree-diff` is not particularly lean, it pulls in `aeson` e.g. because it's a batteries-included library. Having batteries is nice in testing (to avoid dzillion of packages, c.f. `data-default`), but not so for dependency footprint.

edsko: Perhaps we can actually flip it around: use a first order representation but provide a HOAS API. I'll try that.

edsko: Although I find the translation to TH in this PHOAS form quite pleasant.

edsko: On second thoughts, I'll leave it like this for now (but will provide a `Show` instance for closed forms).

hs-bindgen/src/HsBindgen/Hs/AST/ToTH.hs
```diff
@@ -0,0 +1,149 @@
+{-# LANGUAGE TemplateHaskell #-}
```
phadej, line 1: `TemplateHaskellQuotes`

hs-bindgen/src/HsBindgen/Hs/AST/ToTH.hs
```diff
@@ -0,0 +1,149 @@
...
+simpleMatch pat b = TH.match pat (TH.normalB b) []
+
+-- | Applicative call
+--
+-- f <$> e1 <*> .. <*> eN
+apE :: TH.ExpQ -> [TH.ExpQ] -> TH.ExpQ
+apE = \f es ->
+    case es of
+      []    -> f
+      e:es' -> go [| $f <$> $e |] es'
```
phadej, line 145: this is `foldl'` isn't it (except you wrote` foldl`).

hs-bindgen/test-th/HsBindgen/TestTH/Examples.hs
```diff
@@ -0,0 +1,58 @@
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
```
phadej, line 1: blargh. Typed interface, yeah, but with partial bindings. Safe, but not safe.

edsko: This is the eternal tradeoff: how much information do we want to include in types?

phadej: I'd argue we do, as we won't get "partial pattern match" exceptions.

I really want the compiler to be able to help me, to construct and consume valid structures.

edsko: Yes, true, there is a benefit to it, I agree. Perhaps I should rephrase my question: is that benefit worth the cost of tracking bound vars at the type level? We'd have to have type-level naturals, type-level addition (so that we can track how many variables we bind when we match on a struct), we'd have to record at the type-level what the size of structs are, we'd then have to introduce those as existentials when we do the translation from C, ... It's possible, but quite a bit of machinery. As it stands, we get _some_ type safety (more than in a first-order representation), and I'd argue some convenience also, without too much cost. 

But if you think we can track the vars at the type-level without too much cost, I'd love to see. I do agree that it would be nicer.

hs-bindgen/src/HsBindgen/Hs/AST.hs
```diff
@@ -0,0 +1,110 @@
+-- | Haskell AST
+--
+-- This is a bespoke AST, specifically targetted at the hs-bindgen use case. We
+-- do not attempt to cover all of Haskell, and we provide a more strongly typed
```
phadej, line 4: I'm not sure I like this.

From the backend perspective it would be nicer to have as small AST as possible to work with, not dozens of possible case splits, instance declarations etc. But from constructing, frontend perspective I understand that it's nicer to have more "sugar".

I'd argue that there should be concrete low-level AST. Whether the high-level is also data-types with conversion functions to low-level AST, or directly functions to construct low-level AST, I don't have good opinion about, both have own merits.

phadej: An argument for having low-level (close to syntax) haskell AST is that in perfect case, when adding new frontend functionality we wouldn't need to extend all backends (we only have two, but still). Now touching both backends will be in most cases necessary.

edsko: Yeah, changing some of these to be shallow rather than deep embeddings might be a better option, that's certainly something we can tweak.

hs-bindgen/test-th/HsBindgen/TestTH/Examples.hs
edsko: Thinking about this some more: one benefit actually of having the size available at the type level is that the `Show` instances also become easier (it's a bit tricky to `Show` open terms right now; possible, but a bit ugly). Maybe I will experiment with a richer representation that does in fact track bound variables at the type-level; the case that case me most concerned is pattern matching on structs, but perhaps we can avoid type-level addition by instead having a type-level list instead of a single `Nat`, so that we can "index" variables. Something like

```haskell
'[ 'Z, 'Z, 'Z ]
```

would then just to equivalent to 3 (three bound variables), but if we have

```haskell
'[ 'S 'S 'Z, 'Z, 'Z ]
```

then perhaps we have three bound variables, `x`, `y`, and `z.0`, `z.1`, `z.2`. Maybe this way the overhead isn't too large. I'll experiment a bit with this.

edsko: Experimented a tiny bit, I think it might be workable, having terms indexed by `[Nat]` rather than `Nat`, for available "subscripted" variables; however, if we do only that, we're back to essentially dealing with DeBruijn indices when constructing terms, which is unpleasant. Perhaps that we can restore the higher-order interface using something like [I am not a number, I am a classy hack](https://mazzo.li/epilogue/index.html%3Fp=773.html) but I am out of time for now. To be continued.

edsko: Still not convinced it's worth it though.

phadej: You overly complicated this. 

```haskell
variableLambda :: Size n -> (Vec n Bound -> Term) -> Term
```

or

```haskell
variableLambda :: Traversable t
  => t Bound -- ^ names
  -> (t Bound -> Term) -- ^ body
  -> Term
```

which also couples the arity of binder to the body. (The consumer cannot do anything else with names `t Bound` other than `traverse` - so they cannot come up with wrong arity container to pass to the body-function).

no need to track the arity outside of a term.

edsko: Ah! Good idea. Ok, will do. I like the `Vec` approach better, it allows to continue to write clean-looking manual examples; I think the `Traversable` thing will still be tricky (how do you write somehing like

```haskell
    , storablePoke = Hs.Lambda $ \ptr -> Hs.Lambda $ \x ->
        Hs.ElimStruct myStruct x $ \[f1, f2] -> Hs.Seq [
            Hs.PokeByteOff ptr 0 f1
          , Hs.PokeByteOff ptr 4 f2
          ]
```

in a type-safe way if all you have is `Traversable`? Much easier with the `Vec`-like approach. Will do (perhaps `NP` instead of `Vec`).

hs-bindgen/src/HsBindgen/Hs/AST.hs
edsko: Will simplify this.

edsko: I've thought about this some more. We currently have two kinds of specific "kinds" of concepts in the AST:

* `StorableInstance`. We could replace this with a general `Instance` concept.
* `PeekByteOff`, `PokeByteOff`. We could replace this with general `App` expression.

In both of these cases, we have the problem that we'd still like to refer to specific concepts in the `base` libraries, such as `Foreign.Storable.sizeOf`. We could represent these simply as strings, and then match against those strings in the TH backend to resolve to `'Foreign.Storable.sizeOf'`, but this feels quite unpleasant. We could instead TH-side use `mkName` and rely on imports, but that doesn't feel quite right either? 

We could resolve this by introducing something like `data Name = Generated String | Global GlobalName`, where `Generated` would correspond to TH `mkName` (for types we generate during generation), and `data GlobalName = ForeignStorableSizeOf | ForeignStorableAlignment | ...`, where we could then resolve the latter probably.

I was halfway implementing this, but then stopped again. There is another benefit to the AST representation as it is now: it can only represented Well-Typed Haskell programs! This is something that I think is actually quite a useful benefit. I don't to worry that the code I generate in the AST might be ill-typed, only to realize this when I try it out. (Provided no bugs in the TH generation, of course.) 

Since we will generate AST values probably in more places than we modify the backend, this feels like a useful benefit at not too much cost? So let's keep it for now (but still changing the `Vec` thing). We can always reconsider again if it turns out the benefit isn't worth the extra cost in the backends.

edsko: Sorry for going back and forth on this, but of course we could still achieve the type safety through the high-level API. Let me actually write some more code, experiment a bit more, and then come to a conclusion.

hs-bindgen/src/HsBindgen/Hs/AST/ToTH.hs
edsko: Ish. I can write it as

```haskell
apE :: TH.ExpQ -> [TH.ExpQ] -> TH.ExpQ
apE = \f es ->
    case es of
      []    -> f
      e:es' -> foldl' (\x y -> [| $x (<*>) $y |]) [| $f <$> $e |] es'
```

Not sure if that is clearer, but I don't mind.

hs-bindgen/src/HsBindgen/Hs/AST/ToTH.hs
edsko: TIL about [TemplateHaskellQuotes](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/template_haskell.html#extension-TemplateHaskellQuotes) :)

hs-bindgen/src/HsBindgen/Hs/AST.hs
phadej: > it can only represented well-typed Haskell programs! 

Surely it can if it cannot even enforce arity of functions! Missing instances for `peekByteOff` etc. It's quite a loose guarante.

But I guess we'll just make enough helpers function to generate case binds etc. from each specific construct  (as we already have helpers like `simpleMatch`).

hs-bindgen/test-th/HsBindgen/TestTH/Examples.hs
phadej: You can instantiate `t` with some `Vec`, but you don't need to.

hs-bindgen/src/HsBindgen/Hs/AST.hs
edsko: True :)

By the end of today I hope to have a version which reduces duplication between the two backends as well as uses vectors, which I hope you will be happier with :)

edsko:
Ok, this is ready. To quote the commit message:

    Introduce Haskell AST and multiple backends
    
    This also
    
    * Adds a struct offset field (closes #133)
    * Starts the translation from C to Haskell for structs
    * Provide infrastructure for PHOAS, including generics for
      deriving lawful (parseable) `Show`.

The two backends (TH and haskell-src-exts) _only_ need to define basic translations; they do not need to deal with all the details of our precise Haskell AST, that is done only once in a backend-independent way. 

As the commit message mentions, we can show these instances (using SOP generics to derive `ShowOpen` instances). I have replaced the TH dump golden tests with "internal Haskell representation" golden tests. I'm not sure if we want to revive the TH golden tests; I'm not sure it adds much (we should just test the TH generation, we don't need to check it for all examples). It's a bit trickier to add golden tests for this now because the generation of TH now really does depend on Q (for fresh name generation), so we can't easily `ppr` them anymore; I don't think this is a great loss.

There are no `ToExpr` instances for these anymore, but I don't think there's any reason why wouldn't introduce some generic machinery to define them if we need them.

edsko:
After this PR, we are also quite close to being able to generate Haskell modules with Storable instances for simple structs (I think the only thing that might still be missing is proper imports).

phadej:
> the generation of TH now really does depend on Q

It doesn't need to, it should work in any `Quote m => ...` context (which can be provided by `State Int`.

PR #164, opened 2024-09-13
More `libclang` bindings for dealing with types

edsko:
This isn't _quite_ right yet, as `clang_getUnqualifiedType` is only available on `libclang-16` and higher. We have three options here:

1. Don't bind to it at all, side-step the problem. Not a great option, it's useful to be able to experiment with this at least when trying to figure out how to get more info out of `libclang`.
2. Define the wrapper `wrap_getUnqualifiedType` even if `clang_getUnqualifiedType` is not defined, and have it return `NULL` (or some other error value), and throw a runtime exception if `clang_getUnqualifiedType` is called but `libclang` is too old. Doesn't feel terribly Haskell-y, but may have some advantages.
3. Somehow make the Haskell API dependent on the `libclang` version. I'm not totally sure how to do this; we can use `CPP` obviously in the Haskell module, but how do we make (some version of) `CINDEX_VERSION_MINOR` available?

@phadej Do you have an opinion on this? And if you feel that (3) is the best option, do you know how to actually do it?

### DIFF
--- a/hs-bindgen-libclang/cbits/clang_wrappers.h
+++ b/hs-bindgen-libclang/cbits/clang_wrappers.h
@@ -10,6 +10,47 @@
 
 #include <clang-c/Index.h>
 
+#include <stdio.h>
+
+enum WrapperResult {
+    /** The function is supported, but the call failed.
+     *
+     * This can be due to any number of reasons; a typical one is that an
+     * argument @CXType@ is invalid.
+     */
+    WrapperFailed = 0,
+
+    /** Wrapper was successful */
+    WrapperOk = 1,
+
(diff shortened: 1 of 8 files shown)


phadej:
(4). Require libclang 16

edsko:
We could, and if we _need_ this function I agree that's the best option, but so far anyway I don't think we do; I added it in a failed attempt to coax some info out of `libclang`, so it's useful to have an experiment with, but until we really need it, I think I'd like to avoid (4).

phadej:
Then I'd do (2) as it's less intrusive than (3).

PR #179, opened 2024-09-14
Exception unrecognized type

phadej:
We cannot (silently or not) ignore unrecognised types


hs-bindgen/src/HsBindgen/C/Parser.hs
```diff
@@ -230,22 +236,24 @@ foldEnumValues tracer _parent current = do
 -- If we encounter an unrecognized type, we return 'Nothing' and issue a
 -- warning. The warning is ussed here, rather than at the call site, because
 -- 'parseType' is recursive.
-parseType :: HasCallStack => Tracer IO ParseMsg -> CXType -> IO (Maybe C.Typ)
-parseType tracer = go
+parseType :: Tracer IO ParseMsg -> CXType -> IO C.Typ
```
phadej, line 239: Not having `Maybe` will help with recursive types too (`MonadFix IO` works, `MonadFix (MaybeT IO)` is tricky).

PR #187, opened 2024-09-20
More macro parsing

edsko:
Slowly making my way through the macros in the stdlib.


phadej:
I think the default should be `False`. And parse the opposite. A quite common mistake is to forget to commit the outputs, so CI will "silently" succeed.

edsko:
Should the flag be called `--create-file` in that case?

phadej:
I'll make a patch myself.

PR #196, opened 2024-09-24
Parser improvements

edsko:


edsko:
Golden fixtures are wrong; I think we need to specify a C standard to `libclang`.

hs-bindgen/src/HsBindgen/C/AST/Literal.hs
```diff
@@ -0,0 +1,156 @@
...
+-------------------------------------------------------------------------------}
+
+type Parser = Parsec Text ()
+
+-- | Case-insensitive match
+--
+-- Does not consume any input unless the full string matches.
+caseInsensitive :: String -> Parser ()
+caseInsensitive = \str ->
+    try (mapM_ aux str <?> str)
```
phadej, line 145: https://hackage.haskell.org/package/parsec-3.1.17.0/docs/Text-Parsec-Char.html#v:string-39-

edsko: Yeah, but I wasn't sure how to avoid it here. I looked at `string'`, but it's implemented in terms of `tokens'`, which is not general enough (it uses equality on tokens, which is not what I want), and I can't copy and paste it to generalize it (I tried), because it uses features that aren't in the public API. So I wan't sure what to do; do you have a suggestion?

phadej: Don't treat `parsec` as a context-free grammar. It's not. It's PEG.

```
0 -> (x -> hex | b -> bin | 1 .. 9 -> oct)
1 -> dec
```

Not as pretty as what you have now, but that is the price to pay.

phadej: And for the future there is https://hackage.haskell.org/package/parsec-3.1.17.0/docs/Text-Parsec-Prim.html#v:mkPT which is close enough to using `ParsecT` constructor directly

edsko: I will take a look at this `try` thing, but going to merge this PR to avoid blocking other work.

edsko: Addressed in #197.

Issue #202, opened 2024-10-04
Refactor `clang_getUnqualifiedType`: check for errors on Haskell side


TravisCardwell:
The current implementation checks `libclang` version compatibility on the C side ([code](https://github.com/well-typed/hs-bindgen/blob/487b3457c07040faaac6131f81c74126d3bfb472/hs-bindgen-libclang/cbits/clang_wrappers.h#L336-L348)), using [`WrapperResult`](https://github.com/well-typed/hs-bindgen/blob/487b3457c07040faaac6131f81c74126d3bfb472/hs-bindgen-libclang/cbits/clang_wrappers.h#L113-L150).  Discussing with @edsko, we decided that it is probably a better idea to perform such checks on the Haskell side when possible.

TravisCardwell:
`WrapperResult` indicates success (`WrapperOK`) and failure (`WrapperFailed`), not just version incompatibility.

`WrapperResult` values are included in [`Unsupported`](https://github.com/well-typed/hs-bindgen/blob/487b3457c07040faaac6131f81c74126d3bfb472/hs-bindgen-libclang/src/HsBindgen/Clang/Internal/Results.hs#L110-L112) exceptions.  It is currently only used by `*_getUnqualifiedType`, where an `Unsupported WrapperFailed _` exception is raised when the type is invalid.  Perhaps calling this `Unsupported` is mismatched semantics?

The goal is to implement what we can in Haskell to take advantage of Haskell type checking.  I will go ahead and factor `WrapperResult` out of the C side completely.  The Haskell side can check version compatibility as well as if the CT is invalid.  A `WrapperResult` (with a different name) still needs to be defined on the Haskell side for the exception.

Also, is `LLVM` terminology used deliberately?  We use `Clang` elsewhere.  (Example: `Clang11` version and `WrapperNeedsLlvm11` wrapper result)  I will (tentatively) create the new type using `Clang` terminology for consistency.

TravisCardwell:
I ran across the following [comment](https://github.com/well-typed/hs-bindgen/blob/487b3457c07040faaac6131f81c74126d3bfb472/hs-bindgen-libclang/src/HsBindgen/Clang/Core.hs#L1170):

> Throws `CallFailed` if the argument is an invalid type.

This type wraps C results, which is probably why `WrapperResult` was used instead.  (This case is caught in Haskell to avoid segfaults.)  I will (tentatively) create a new `InvalidCXTypeError` exception instead.

[...]

PR #203, opened 2024-10-04
Refactor clang_getUnqualifiedType (#202)

TravisCardwell:
Motivation: perform checks in Haskell to take advantage of type safety.

* `WrapperResult` is completely factored out.
* `ClangVersionError` is added to indicate version support errors.
* `InvalidCXTypeError` is added to indicate Haskell-side invalid type errors.
* `clang_getUnqualifiedType` (Haskell) is refactored to check the `libclang` version and type validity, removing the checks from `wrap_getUnqualifiedType` (C).


TravisCardwell:
Build failures!  Interesting.  Builds and tests pass locally, and I purposefully made the build fail before changing the C code, so I do not think it is a local cache issue.  Investigating...

hs-bindgen-libclang/src/HsBindgen/Clang/Internal/Results.hs
```diff
@@ -104,28 +106,35 @@ ensureNotInRange = ensureOn conv (not . simpleEnumInRange)
...
+  | Clang13Required -- ^ We need at least Clang version 13
+  | Clang14Required -- ^ We need at least Clang version 14
+  | Clang15Required -- ^ We need at least Clang version 15
+  | Clang16Required -- ^ We need at least Clang version 16
+  | Clang17Required -- ^ We need at least Clang version 17
+  | Clang18Required -- ^ We need at least Clang version 18
+  deriving stock (Show, Eq, Enum, Bounded, Generic)
+
+-- | The version of Clang being used is not supported
+data ClangVersionError = ClangVersionError ClangVersionRequirement Backtrace
```
edsko, line 126: Can we re-use `ClangVersion` instead of re-enumerating the types in `ClangVersionRequirement`? We cannot make more fine-grained distinctions than `ClangVersion`.

hs-bindgen-libclang/src/HsBindgen/Clang/Internal/Results.hs
```diff
@@ -104,28 +106,35 @@ ensureNotInRange = ensureOn conv (not . simpleEnumInRange)
...
-        stack <- collectBacktrace
-        throwIO $ Unsupported mSupported stack
+{-------------------------------------------------------------------------------
+  Invalid CXType error
+-------------------------------------------------------------------------------}
+
+-- | Haskell-side invalid type error
+data InvalidCXTypeError = InvalidCXTypeError Backtrace
+  deriving stock (Show)
+  deriving Exception via CollectedBacktrace InvalidCXTypeError
```
edsko, line 137: It's a bit unfortunate that we have a new exception type now, while we use `CallFailed` everywhere else. I can see why you did it, but perhaps we can avoid it by changing `data CallFailed result` to `data CallFailed hint` instead? The only purpose for that argument really is to give a clue about what went wrong. What do you think?

hs-bindgen-libclang/src/HsBindgen/Clang/Internal/Results.hs
TravisCardwell: Yes, that should not be a problem.  I rebased to include merges to `main` and made the change.

Details:

(Some) `ClangVersion` constructors map to more than one Clang version, but not being able to specify requirements for intermediate versions should never be a problem because `CINDEX_VERSION_MINOR` is bumped on API additions.

We use `Show` instances when displaying exceptions, so we can simply wrap a `ClangVersion` value with a type that indicates the requirement, being sure to derive with the `stock` strategy.  Example: `Requires Clang17_or_18_or_19` would indicate that Clang 17 or newer is required.

I went ahead and used `CallFailed` for this as well.  The implementation is in `HsBindgen.Clang.Version` to avoid an import cycle.  I think that this (exposed module) is a good place for it, as users may want to use `requireClangVersion` themselves (when using a higher-level API).

hs-bindgen-libclang/src/HsBindgen/Clang/Internal/Results.hs
TravisCardwell: Sure, that sounds good.  Thanks!

Details:

When checking results, the `CallFailed` "hint" is the result value property that is checked, and the backtrace provides the context required to understand the exception.  I refactored to do the same in this case as well: `fromSimpleEnum (cxtKind typ)` is passed as the "hint."

`fromSimpleEnum` can return a `Left` value, so I went ahead and treat that case as an error as well.

If we need to do this anywhere else, it may be worth creating a function.  We use terminology `ensure` to check the result of an action "call," so perhaps different terminology (such as `assert`) could be appropriate.

```haskell
assertTypeValid :: HasCallStack => CXType -> IO ()
assertTypeValid typ =
    case fromSimpleEnum (cxtKind typ) of
      e@Left{}                 -> callFailed e
      e@(Right CXType_Invalid) -> callFailed e
      Right{}                  -> pure ()
```

This cannot be done (like this) in module `HsBindgen.Clang.Internal.Results` because `CXType` is defined in `HsBindgen.Clang.Core`.

TravisCardwell:
I am currently using Arch.  [`llvm14`](https://archlinux.org/packages/extra/x86_64/llvm14/) is available, so I installed it confirmed that putting the directory at the beginning of my `PATH` works.  Thank you!

```
$ export PATH=/usr/lib/llvm14/bin:$PATH
$ rm -rf dist-newstyle
$ cabal build all
...
checking for llvm-config... /usr/lib/llvm14/bin/llvm-config
checking LLVM version... 14.0.6
checking LLVM library directory... /usr/lib/llvm14/lib
checking LLVM include directory... /usr/lib/llvm14/include
...
```

</details>

At some point, we might want to test against more versions.  Debian has packages for many versions (on different distributions), and it should not be difficult to test locally within containers.

| `llvm` | Version    | Distribution            |
| ------ | ---------- | ----------------------- |
| [6][]  | `6.0.1`    | `buster` (oldoldstable) |
| [7][]  | `7.0.1`    | `buster` (oldoldstable) |
| [9][]  | `9.0.1`    | `bullseye` (oldstable)  |
| [11][] | `11.0.1`   | `bullseye` (oldstable)  |
| [13][] | `13.0.1`   | `bookworm` (stable)     |
| [14][] | `14.0.6`   | `bookworm` (stable)     |
| [15][] | `15.0.6`   | `bookworm` (stable)     |
| [16][] | `16.0.6`   | `bookworm` (stable)     |
| [17][] | `17.0.6`   | `sid` (unstable)        |
| [18][] | `18.1.8`   | `sid` (unstable)        |
| [19][] | `19.1.1`   | `sid` (unstable)        |
| [20][] | `20240921` | (experimental)          |

[6]: <https://packages.debian.org/buster/llvm-6.0>
[7]: <https://packages.debian.org/buster/llvm-7>
[9]: <https://packages.debian.org/bullseye/llvm-9>
[11]: <https://packages.debian.org/bullseye/llvm-11>
[13]: <https://packages.debian.org/bookworm/llvm-13>
[14]: <https://packages.debian.org/bookworm/llvm-14>
[15]: <https://packages.debian.org/bookworm/llvm-15>
[16]: <https://packages.debian.org/bookworm/llvm-16>
[17]: <https://packages.debian.org/sid/llvm-17>
[18]: <https://packages.debian.org/sid/llvm-18>
[19]: <https://packages.debian.org/sid/llvm-19>
[20]: <https://packages.debian.org/experimental/llvm-20>

I am not sure what the situation is on Ubuntu, used in CI, as [Ubuntu Packages Search][] is returning internal server errors when searching.

[Ubuntu Packages Search]: <https://packages.ubuntu.com/>

hs-bindgen-libclang/src/HsBindgen/Clang/Internal/Results.hs
edsko: Ok, this looks great now, thank you :) And indeed, `ClangVersion` isn't as granular as one would like, but it is as granular as we can test for, so I think it makes sense to reuse it here.

By the way, thank you for sticking to local coding style, that is much appreciated!

hs-bindgen-libclang/src/HsBindgen/Clang/Internal/Results.hs
edsko: Yeah, this would make sense alongside `ensureTypeValid`. We don't do much input value checking at the moment, but indeed, something to keep an eye on.

edsko:
Yes, testing with some older versions especially would be useful, to see if there are any assumptions that we are making that aren't warranted.

Issue #189, opened 2024-09-21
Name mangling


edsko:
We need to be careful translating C names to Haskell names.

edsko:
As of #190 , basic infra for this is now in place (see `HsBindgen.Hs.AST.Name`), but the actual mangling still needs to happen.

TravisCardwell:
I thought that C identifiers are pretty restricted, but that knowledge is outdated: C does indeed support Unicode now!  From the C23 working draft ([N3096][]):

[N3096]: <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3096.pdf>

* `identifier`:
    * `identifier-start`
    * `identifier` `identifier-continue`
* `identifier-start`:
    * `nondigit`
    * XID_Start character
    * universal-character-name of class XID_Start
* `identifier-continue`:
    * `digit`
    * `nondigit`
    * XID_Continue character
    * universal-character-name of class XID_Continue
* `nondigit`: one of `/_a-zA-Z/`
* `digit`: one of `/0-9/`

Package [`text-icu`][] could be used to check such Unicode character properties correctly.  (One would use [`property`][] to query constructors `XidStart` and `XidContinue` of type [`Bool_`][].)  Perhaps there is no need to do so, however, since we are creating Haskell identifiers.

[`text-icu`]: <https://hackage.haskell.org/package/text-icu>
[`property`]: <https://hackage.haskell.org/package/text-icu-0.8.0.5/docs/Data-Text-ICU-Char.html#v:property>
[`Bool_`]: <https://hackage.haskell.org/package/text-icu-0.8.0.5/docs/Data-Text-ICU-Char.html#t:Bool_>

A Haskell identifier "consists of a letter followed by zero or more letters, digits, underscores, and single quotes" ([Haskell 2010 Language Report][]).  Some details:

* "Small" letters include the underscore and Unicode lowercase letters.
* "Large" letters include Unicode uppercase and titlecase letters.
* Digits include Unicode digits.

[Haskell 2010 Language Report]: <https://www.haskell.org/onlinereport/haskell2010/haskellch2.html#x7-180002.4>

Perhaps we can remove any characters that are not valid in Haskell names, as well as change the case of the first character according to the type of name as appropriate.  There are some edge cases:

* The name may start with an underscore, which is only valid for variable-like names.
    * For other names, perhaps we can drop leading underscores.
* The filtered C name may be empty.
* Multiple C names could result in the same Haskell name.
[...]

PR #213, opened 2024-10-11
Adjust case of first letter for namespace (#189)

TravisCardwell:
See https://github.com/well-typed/hs-bindgen/issues/189#issuecomment-2406412261

I did this minimal change first in case this is blocking anybody, and I will add the context and options next.


edsko:
Thanks for this, this should at least unblock initial problems.

TravisCardwell:
BTW, there is one implementation note.

The implementation includes a `Char.isLetter` guard.  My intuition is that this is unnecessary.  `Char.toLower` and `Char.toUpper` *should* inherently perform this check, passing through any character that is not a letter.  I referenced the source for [`toTitle`][] before implementing, however, and decided to include it in the quick fix to be safe.  I will look a little deeper when I work on the next iteration of the implementation.

[`toTitle`]: <https://hackage.haskell.org/package/text-2.1.1/docs/Data-Text-Internal-Fusion-Common.html#v:toTitle>

PR #217, opened 2024-10-11
Add test for fixedwidth types

edsko:
This rebases #131 now that we have support for macros (and adds one required case to the fold function to ignore inclusion directives).


edsko:
Ok, wow, both OSX and Windows fail with

```
fatal error: 'stdint.h' file not found"
```

(in #131 that error wasn't visible). Perhaps the whole `#define`/`typedef` was not in fact related at all to this problem? Oh well, that was still useful to address even if not, but it does mean that this PR is once again blocked.

However, I don't understand why; `stdint.h` is one of the standard headers; it's available on OSX (I asked Alfredo to check), and it's also [available on Windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/standard-types?view=msvc-170); as it should, it's one of the standard headers. So it looks like CI is simply missing some setup? @phadej , any thoughts?

phadej:
I don't know. Something weird is happening, as https://github.com/well-typed/hs-bindgen/pull/218 fails differently (also on Linux), i.e. if I add `#include <stdio.h>` it fails with

```
UnrecognizedCursor {unrecognizedCursorKind = simpleEnum CXCursor_InclusionDirective, unrecognizedCursorLoc = "examples/enums.h:4:1", unrecognizedCursorTrace = CallStack (from HasCallStack):
  collectBacktrace, called at src/HsBindgen/C/Fold/Common.hs:84:32 in hs-bindgen-0.1.0-inplace:HsBindgen.C.Fold.Common
  unrecognizedCursor, called at src/HsBindgen/C/Fold/Decl.hs:58:9 in hs-bindgen-0.1.0-inplace:HsBindgen.C.Fold.Decl
  foldDecls, called at src/HsBindgen/Lib.hs:143:10 in hs-bindgen-0.1.0-inplace:HsBindgen.Lib}
    treediff:    Test suite golden: FAIL
```

is `stdint.h` special? I have no idea.

phadej:
Ah, you add

```
      Right CXCursor_InclusionDirective ->
        -- Ignore
        return $ Continue Nothing
```

here. Is it correct thing to do? Is the include contents still traversed? It's really worth a comment.

phadej:
> it's available on OSX

Yes, but Windows and OSX GitHub builders are piles of software. Some users machine may look like that as well, so I have no other suggestion then adding `find`, `ls` etc debug steps into Ci configuration to figure out what's happening there.

phadej:
Ah, so my `stdint` branch failing with

```diff
-  "FieldDecl" "sixty_four" :: "Typedef"
-    "TypeRef" "uint64_t" :: "Typedef"
-  "FieldDecl" "thirty_two" :: "Typedef"
-    "TypeRef" "uint32_t" :: "Typedef"
+  "FieldDecl" "sixty_four" :: "Int"
+  "FieldDecl" "thirty_two" :: "Int"
```

might have been because there are no `stdint` include found, and unbound types defaulting to `int`. That's a good explanation for that at least.

phadej:
The https://github.com/well-typed/hs-bindgen/pull/219 fails with not finding `time.h` either. So we can extrapolate that *all* standard library are missing. Can that happen, I guess. `libclang` can exist without any of headers being present. (Or what is worse, there can be multiple different ones, even bogus ones).

phadej:
Using

```diff
diff --git a/hs-bindgen/tests/TH.hs b/hs-bindgen/tests/TH.hs
index 6c24d32..5a8a3e8 100644
--- a/hs-bindgen/tests/TH.hs
+++ b/hs-bindgen/tests/TH.hs
@@ -60,7 +60,10 @@ goldenTh name = goldenVsStringDiff_ "th" ("fixtures" </> (name ++ ".th.txt")) $
     return $ unlines $ map (show . TH.ppr) $ unqualNames $ runQu decls
 
 clangArgs :: ClangArgs
-clangArgs = defaultClangArgs{clangTarget = Just "x86_64-pc-linux-gnu"}
+clangArgs = defaultClangArgs
+    { clangTarget = Just "x86_64-pc-linux-gnu"
+    , clangOtherArgs = ["-nobuiltininc"]
+    }
 
 -- | Deterministic monad with TH.Quote instance
 newtype Qu a = Qu (State Integer a)
```

makes tests fail on linux too. `-nostdinc` is also an option. (See https://clang.llvm.org/docs/ClangCommandLineReference.html)

I do strongly think that using `-nostdinc` would be a good idea, and instead be explicit on what's included.

edsko:
Obsoleted by #219. Agreed that I should add a comment about the include directive.

PR #220, opened 2024-10-14
Add `--all` flag to `clang-ast-dump`

TravisCardwell:
This flag makes it very easy to get maximum verbosity, removing the need to list all flags on the command line.


TravisCardwell:
While mentally reviewing code in the shower, I realized that I should not even include `optAll` in `Options`.  I am already using `ApplicativeDo` after all, so it was trivial to refactor.

PR #228, opened 2024-10-18
Implement name mangling (#189)

### LINKED ISSUE #189: Name mangling

TravisCardwell:
Please see https://github.com/well-typed/hs-bindgen/issues/189#issuecomment-2421706983 for notes about the latest changes.

If anything needs to be changed or improved, please let me know.

When merged, perhaps this closes #189.


hs-bindgen/src/HsBindgen/Hs/AST/Name.hs
```diff
@@ -29,50 +62,564 @@ data Namespace =
...
+appendSingleQuote :: Text -> Text
+appendSingleQuote = (<> "'")
+
+-- | Haskell keywords
+--
+-- <https://wiki.haskell.org/Keywords>
+keywords :: [Text]
+keywords =
+    [ "as"
+    , "case"
```
phadej, line 365: cases

hs-bindgen/src/HsBindgen/Hs/AST/Name.hs
```diff
@@ -29,50 +62,564 @@ data Namespace =
...
+    , "infix"
+    , "infixl"
+    , "infixr"
+    , "instance"
+    , "let"
+    , "mdo"
+    , "module"
+    , "newtype"
+    , "of"
+    , "proc"
```
phadej, line 388: also `role`

phadej: and `pattern` and `static`

hs-bindgen/src/HsBindgen/Hs/AST/Name.hs
TravisCardwell: Thank you very much for the review!

According to the [proposal][]:

> Unlike `case`, `cases` is not a keyword.  Only the token sequence `\` `cases` is treated specially (more precisely, lexically, `cases` is a *varid* rather than a *reservedid*, and it may be used as a variable).

[proposal]: <https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0302-cases.rst>

A test indicates that we can indeed uses `cases` as a variable or type variable.

<details><summary>Code</summary>

```haskell
#!/usr/bin/env cabal
{- cabal:
build-depends:
    base
-}
{- project:
with-compiler: ghc-9.6.6
-}

{-# LANGUAGE ImportQualifiedPost #-}
{-# LANGUAGE LambdaCase #-}

module Main (main) where

-- https://hackage.haskell.org/package/base
import Data.Char qualified as Char

------------------------------------------------------------------------------

newtype Foo cases = Foo { unFoo :: String }

cases :: Foo ()
cases = Foo "hHeElLlLoO"

------------------------------------------------------------------------------

filter' :: (a -> Bool) -> [a] -> [a]
filter' = \cases
    _ []          -> []
    p (x:xs)
      | p x       -> x : filter' p xs
      | otherwise ->     filter' p xs

------------------------------------------------------------------------------

main :: IO ()
main = putStrLn $ filter' Char.isLower (unFoo cases)
```

</details>

Perhaps there is a different extension that causes trouble?

hs-bindgen/src/HsBindgen/Hs/AST/Name.hs
TravisCardwell: Thank you!  I (rebased and) added these.

> This commit replaces `keywords` with separate `varReservedNames` and `typeVarReservedNames`, for variables and type variables, respectively.
>
> Reserved names `pattern` and `static` are added to `varReservedNames`.
>
> List `typeVarReservedNames` contains reserved name `role` as well as all `varReservedNames`.  `role` is only problematic when used as a type variable.
>
> Documentation is added to reference these added reserved names.

edsko (approved):
@TravisCardwell feel free to merge this once the changes that we discussed are in.

PR #229, opened 2024-10-18
Type inference for macros

sheaf:
This PR implements a simple type-inference pass for macros,
see `HsBindgen.C.Tc.Macro`. It implements a type system with:

  - variables
  - function types
  - type constructors
  - quantified types

Type inference for a macro function definition proceeds in the
following stages:

  1. constraint generation following Algorithm W, creating new
     metavariables and carrying a substitution,
  2. constraint solving: constraint simplification and defaulting
  3. quantification over remaining (undefaulted) type variables
     and remaining (unsolved) constraints.


hs-bindgen/src/HsBindgen/C/Tc/Macro.hs
```diff
@@ -0,0 +1,1121 @@
...
+      let !tvs' = goTy bound tvs argTy
+      in case NE.nonEmpty mbArgTys of
+        Nothing     -> goTy bound tvs resTy
+        Just argTys -> goFunTy bound tvs' argTys resTy
+
+tyVarSet :: [ TyVar ] -> IntSet
+tyVarSet = IntSet.fromList . map ( uniqueInt . tyVarUnique )
+
+newtype Subst tv = Subst { substMap :: IntMap ( tv, Type Ty ) }
+  deriving newtype Monoid
```
phadej, line 345: this makes `mappend` not equal to your `<>`.

hs-bindgen/src/HsBindgen/C/Tc/Macro.hs
```diff
@@ -0,0 +1,1121 @@
...
+  | InstOrigin !FunName
+  | IntLitOrigin !( Literal Integer )
+  | FloatLitOrigin !Double
+  deriving stock Generic
+  deriving anyclass PrettyVal
+
+instance Show CtOrigin where
+  show = \case
+    AppOrigin fun ->
+      "In an application of '" ++ showFunName fun ++ "'."
```
phadej, line 417: I'm not a fan of non-derived like `Show` instances. ping @edsko

hs-bindgen/src/HsBindgen/Pretty/Orphans.hs
```diff
@@ -0,0 +1,31 @@
...
+
+--------------------------------------------------------------------------------
+
+deriving anyclass instance PrettyVal a => PrettyVal ( NE.NonEmpty a )
+
+instance PrettyVal a => PrettyVal ( Vec n a ) where
+  prettyVal v = Pretty.Con "Vec" [ prettyVal ( toList v ) ]
+
+instance PrettyVal a => PrettyVal ( IntMap a ) where
+  prettyVal v = Pretty.Con "IntMap" [ prettyVal ( toList v ) ]
```
phadej, line 31: this strips keys, is it intentional?

hs-bindgen/src/HsBindgen/C/Tc/Macro.hs
sheaf: Fair point. I made these into separate functions and derived the Show instance.

hs-bindgen/src/HsBindgen/C/Tc/Macro.hs
sheaf: Yes indeed, thanks. Fixed.

PR #232, opened 2024-10-23
Nested anonymous record

phadej:


phadej:
The `struct ex4` example with inner struct which refers to itself now works as it should. The `struct ex3` example however doesn't: that's the case when the struct doesn't have name nor field name (from C usage perspective it looks like inner structs had been spliced). The `ex3` example is what I will look next.

PR #233, opened 2024-10-23
Add an example of macro defining a type

phadej:
In this example we lose the information that `foo` is a type "defined" by a library.

I honesty don't think we can do better.


phadej:
We don't get `newtype Foo = Foo CLong`, as we would if library had `typedef long foo;`.

phadej:
Oh wait, we do get

```
              fieldType = TypElaborated
                (CName "foo")},
```

hmm.

phadej:
Ok, so simple example works, but if I do

```
#define foo long

struct bar {
	signed foo x;
	signed foo y;
};
```

then I get an error

```
mkStructField: ReparseError {reparseError = "\"examples/macrodeftype.h\" (line 4, column 9):\nunexpected Unexpected primitive type \"signed\"", reparseErrorTokens = [Token {tokenKind = simpleEnum CXToken_Keyword, tokenSpelling = TokenSpelling {getTokenSpelling = "signed"}, tokenExtent = "<examples/macrodeftype.h:4:2-4:8>", tokenCursorKind = simpleEnum CXCursor_FieldDecl},Token {tokenKind = simpleEnum CXToken_Identifier, tokenSpelling = TokenSpelling {getTokenSpelling = "foo"}, tokenExtent = "<examples/macrodeftype.h:4:9-4:12>", tokenCursorKind = simpleEnum CXCursor_MacroExpansion},Token {tokenKind = simpleEnum CXToken_Identifier, tokenSpelling = TokenSpelling {getTokenSpelling = "x"}, tokenExtent = "<examples/macrodeftype.h:4:13-4:14>", tokenCursorKind = simpleEnum CXCursor_FieldDecl}]}
```

apparently our parser has only partial support.

edsko:
> apparently our parser has only partial support.

Indeed it does, the parser itself is very simple right now. With a small amount of work we should be able to do much better and cover the most common use cases. 

The way that it works is that we keep track of where macro expansion takes place, and if we see a field declaration that has a macro expansion, we try to reparse, and fail if that fails.

phadej:
On the other hand, we have `TypElaborated foo`; and we don't have type definition for `foo`. So we need to figure out that

```
+      DeclMacro
+        (Right
+          Macro {
+            macroLoc = MultiLoc {
+              multiLocExpansion = SingleLoc {
+                singleLocPath = [
+                  "examples",
+                  "macrodeftype.h"],
+                singleLocLine = 1,
+                singleLocColumn = 9},
+              multiLocPresumed = Nothing,
+              multiLocSpelling = Nothing,
+              multiLocFile = Nothing},
+            macroName = CName "foo",
+            macroArgs = [],
+            macroBody = MTerm
+              (MType (PrimLong Signed))}),
```

is a type definition.

edsko:
> which doesn't immediately look like type definition anymore.

Yeah, that's fair. Perhaps @sheaf 's work can help with this.

phadej:
What's unfortunate is that 

```c
#include <stdint.h>

struct bar {
    int uint32_t;
	int uint64_t;
};

int main() { return 0; }
```

is valid C.

PR #239, opened 2024-10-25
Re-parsing for floating-point literals

sheaf:
NB: this PR sits on top of #229.

This PR adds the ability to reparse floating-point literals, for use in macros.

The implementation supports all the syntaxes described in [the C++ reference](https://en.cppreference.com/w/cpp/language/floating_literal), with the exception of the C++23 suffixes `f16, f32, f64, f128, bf16, F16, F32, F64, F128, BF16`


ansi-diff/src/AnsiDiff.hs
```diff
@@ -16,7 +16,8 @@ import Text.EditDistance qualified as ED
 ansidiff :: String -> String -> String
 ansidiff old new = runStringWriter $ do
     writeStr ansiReset
-    let linesDiff = Diff.getDiff (lines old) (lines new)
+    let normalisedLines = lines . filter ( /= '\r' )
```
phadej, line 19: Why this, where form `\r` comes.

The issue is somewhere else.

PR #243, opened 2024-10-29
Add HsTypRef for handling TypElaborated

phadej:


hs-bindgen/fixtures/nested_types.th.txt
```diff
@@ -6,7 +6,9 @@ instance Storable CFoo
            poke = \x_1 -> \x_2 -> case x_2 of
                                   {MkCFoo cFoo_i_3
                                           cFoo_c_4 -> pokeByteOff x_1 0 cFoo_i_3 >> pokeByteOff x_1 32 cFoo_c_4}}
-data CBar = MkCBar {cBar_foo1 :: Void, cBar_foo2 :: Void}
+data CBar
+    = MkCBar {cBar_foo1 :: CStruct'0020foo,
```
phadej, line 10: TypElaborated and name manging is broken. The symbol here is `struct foo`; which gets manged to something different than a standalone `foo` in `struct foo`.

edsko: Yes, that is little more than a stand-in at the moment. We need to do this properly.

edsko: Have opened https://github.com/well-typed/hs-bindgen/issues/244 to track this.

PR #246, opened 2024-10-30
goldenVsStringDiff_: normalise line endings

sheaf:
This commit normalises line endings before comparing, and ensures that we always commit `\n` line endings to the golden files. This avoids spurious test failures on Windows due to `\r\n` line endings.

### DIFF
--- a/hs-bindgen/tests/Misc.hs
+++ b/hs-bindgen/tests/Misc.hs
@@ -21,12 +21,15 @@ import HsBindgen.Lib (ClangArgs (..), defaultClangArgs, CStandard (..))
 
 -- | Like goldenVsString but using our own diff function.
 goldenVsStringDiff_ :: TestName -> FilePath -> ((String -> IO ()) -> IO String) -> TestTree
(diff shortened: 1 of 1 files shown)


hs-bindgen/tests/Misc.hs
```diff
@@ -36,6 +39,11 @@ goldenVsStringDiff_ name fp action = goldenTestSteps name correct action cmp upd
         | xss == yss = return Nothing
         | otherwise  = return $ Just $ ansidiff xss yss
 
+    normalise :: String -> String
+    normalise [] = []
+    normalise ('\r':'\n':r) = '\n' : normalise r
```
phadej, line 44: The question is still, *what* generates `\r\n` endings, even on Windows. We *should not* generate  them in the first place. That will  be a problem for users of  the tool too.

phadej (changes requested):

sheaf:
Maybe this is no longer needed. I re-ran the test-suite locally without this change and it passed, whereas in the past all the TH tests failed due to line-ending differences.

PR #249, opened 2024-10-30
git: set line endings to lf by default

sheaf:
This ensures that Git doesn't try to check out the golden tests from the testsuite with CRLF line-endings, which would cause the tests to fail.

### DIFF
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,0 +1,3 @@
+[core]
+  eol = lf
+  autocrlf = false


sheaf:
I now realise that `.gitconfig` does not change the default settings.

edsko:
If we ever get contributors who are bitten by this, we might want to reconsider something like https://github.com/well-typed/hs-bindgen/pull/246 again.

PR #254, opened 2024-11-01
Update name mangling

### LINKED ISSUE #189: Name mangling

TravisCardwell:
This PR adds support for anonymous structures/unions for named fields, removes the `ToHsName` class, renames the options to `NameMangler`, and refactors the name mangling contexts.

In the previous implementation, name mangling functions were always passed a `CName`.  Some Haskell name creation does not use a `CName`, however, so this resulted in mismatched semantics that could be a bit difficult to understand.

This commit refactors contexts to include the `CName`s when needed, and the `CName` argument is removed from all name mangling functions.  The contexts therefore define and document the whole context that is required to create a Haskell name.

Various notes:

* Contexts are changed to *not* include Haskell names.  They instead include the context for creating Haskell names.  I am not sure if this is necessary yet, but I suspect that it will help with translation if we use a C IR.
* In the default name manglers, enumeration type fields are now created by simply prepending the type constructor name with `un` (and changing the namespace).  This changes the generated code; for example, we now generate `unCFoo` instead of `un_foo`.  This is not the same as how record fields are created, but I think that it is better (and preferred) for this special case.
* In the default name manglers, constructor names are now created by simply prepending the type constructor name with `Mk` (and changing the namespace).  This does not result in any change in the generated code.
* Context names are refactored for consistency.
* Code is rearranged to consistently work with names from large (module) to small (variable) namespaces.  This order works well because smaller namespaces often reference larger namespaces.

Closes #189


TravisCardwell:
@edsko This PR implements the changes that we discussed yesterday.  I do not think that it is worth splitting the module, however.  If this looks fine, it is ready to merge.

edsko:
Thanks!

PR #252, opened 2024-10-31
Typedef decl

phadej:

PR #259, opened 2024-11-07
Add golden preprocessor tests

phadej:
I had to expose some "internals". I'm not sure how this functionality should be exposed, but for now the priority is to get tests


phadej:
This will "conflict" with #252, merge either one first; second one would need to be rebased.

PR #267, opened 2024-11-11
Improve reserved words (#258)

### LINKED ISSUE #258: Define list of "reserved" Haskell identifiers
This avoids overly many qualified names in the generated code. Currently we are reserving `pure` and `return` (these could conceivably clash with C global variables or C functions). We'll also want to reserve names such as `CInt` (a type called `Int` in a C header would become `CInt` with the default name mangler, which would be incredibly confusing, and so we should treat that as a reserved keyboard); note that `Int` does not need to be reserved as the default name mangler will never generate it.

If people use a different name mangler, it is their responsibility to think about such issues (we should provide some guidance on what to think about).

TravisCardwell:
Please see https://github.com/well-typed/hs-bindgen/issues/258#issuecomment-2467416060.

This change reserves the names of all types in `Foreign.C.Types` (which all start with `C` and could therefore be created by the default name manglers).

It should be easy to add new reserved words as we develop.  If there are any more that we should go ahead and add now, please let me know.

Closes #258

### DIFF
--- a/hs-bindgen/src/HsBindgen/Hs/AST/Name.hs
+++ b/hs-bindgen/src/HsBindgen/Hs/AST/Name.hs
@@ -28,15 +28,23 @@ module HsBindgen.Hs.AST.Name (
   , handleReservedNone
   , handleReservedNames
   , appendSingleQuote
-  , varReservedNames
-  , typeVarReservedNames
   , handleModuleNameParent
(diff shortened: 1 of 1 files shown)


edsko:
I like the reorganization, feels quite clean now.

PR #269, opened 2024-11-11
Constant size arrays

phadej:


hs-bindgen/fixtures/fixedarray.pp.hs
```diff
@@ -0,0 +1,13 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Example where
+
+import qualified Foreign as F
+import qualified Foreign.C as FC
+import qualified HsBindgen.ConstantArray
+
+newtype CTriple = MkCTriple
+  { unCTriple :: (HsBindgen.ConstantArray.ConstantArray 3) FC.CInt
```
edsko, line 10: @TravisCardwell I wonder if this is another example of where some explicit precedence/associativity would result in more readable output.

PR #270, opened 2024-11-11
Generate Haskell code for function-like macros

sheaf:
This commit implements a translation layer which transpiles simple function-like C macros into Haskell function definitions, e.g.

  #define Foo(X,Y) X + Y

will be compiled into

  foo :: Num a => a -> a -> a
  foo x y = x + y


hs-bindgen/src/HsBindgen/Backend/TH.hs
```diff
@@ -66,11 +101,23 @@ instance TH.Quote q => BackendRep (BE q) where
...
 
+  resolveLocal _ nm = TH.mkName (Text.unpack $ getHsName nm)
   mkExpr be = \case
       EGlobal n     -> TH.varE (resolve be n)
       EVar x        -> TH.varE (getFresh x)
       ECon n        -> hsConE n
-      EInt i        -> TH.litE (TH.IntegerL $ fromIntegral i)
+      EIntegral i _ -> TH.litE (TH.IntegerL i)
+      -- TH doesn't have floating-point literals, because it represents them
+      -- using the Rational type, which is incorrect.
```
phadej, line 111: is there a GHC issue about this?

hs-bindgen/src/HsBindgen/Hs/AST.hs
```diff
@@ -115,6 +124,58 @@ data DataDecl n f = MkDataDecl
...
+    , varDeclType :: VarDeclType f
+    -- | RHS of variable/function.
+    , varDeclBody :: VarDeclRHS f
+    }
+  deriving stock (GHC.Generic)
+  deriving anyclass (SOP.Generic, SOP.HasDatatypeInfo)
+
+type VarDeclType :: PHOAS
+data VarDeclType f
+  = TyLamTy (HsName NsTypeVar) (VarDeclType f)
```
phadej, line 143: This is wrong. We have PHOAS, but you ahve explicit name for local binding. This is combining worst of both.

phadej: And also this is forall-binder, not type-level lambda. Is it?

hs-bindgen/src/HsBindgen/Backend/TH.hs
sheaf: I found [#13124](https://gitlab.haskell.org/ghc/ghc/-/issues/13124). I added a reference.

hs-bindgen/src/HsBindgen/Hs/AST.hs
sheaf: Thanks, fixed now. However, I wasn't sure how to deal with macros referring to other macros; ideally we should be able to resolve them directly but currently in the backends I just use plain names (e.g. in the TH backend I use `mkName`).

phadej: > macros referring to other macros; 

We simply cannot have "fresh" names for top-level definitions. But that's not a real problem, as top-level names are part of API, so they cannot  really be fresh anyway.

hs-bindgen/src/HsBindgen/Backend/PP.hs
```diff
@@ -111,6 +118,31 @@ resolveGlobal = \case
...
+    Bits_and        -> importQ iBits    ".&."
+    Bits_xor        -> importQ iBits    "xor"
+    Bits_or         -> importQ iBits    ".|."
+    Bits_complement -> importQ iBits    "complement"
+    Num_negate      -> importQ iPrelude "negate"
+    Num_add         -> importQ iPrelude "+"
+    Num_minus       -> importQ iPrelude "-"
+    Num_times       -> importQ iPrelude "*"
+    Fractional_div  -> importQ iPrelude "/"
+    Integral_rem    -> importQ iPrelude "rem"
```
phadej, line 143: I think we are missing integral `quot`. Maybe there's a note somehwere, but `#define DIV(x,y) x / y` is a valid in integral context in C too. And often is.

hs-bindgen/src/HsBindgen/Backend/Common.hs
```diff
@@ -30,10 +33,16 @@ class BackendRep be where
...
   type Ty   be :: Type
 
-  resolve :: be -> Global   -> Name be  -- ^ Resolve name
-  mkExpr  :: be -> SExpr be -> Expr be  -- ^ Construct expression
-  mkDecl  :: be -> SDecl be -> Decl be  -- ^ Construct declaration
-  mkType  :: be -> SType be -> Ty   be  -- ^ Construct type
+  -- | Resolve a global name.
+  resolve      :: be -> Global -> Name be
+  -- | Resolve a local name.
+  resolveLocal :: be -> HsName NsVar -> Name be
```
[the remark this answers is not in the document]
sheaf: OK but that was my question: what do I use for e.g. the name of a user-defined macro, e.g. the occurence of `FOO` within the RHS of `BAR` in:

```c
#define FOO(X,Y) X + Y
#define BAR(X,Y) 2 * FOO(X,Y)
```

Do I add another field to `Global`:

```hs
data Global
  = ...
  | UserGlobal (HsName NsVar)
```

or something?

hs-bindgen/src/HsBindgen/Backend/PP.hs
sheaf: Very good point, I hadn't considered that at all. I will have to think about what the type of `/` should be.

hs-bindgen/src/HsBindgen/Backend/Common.hs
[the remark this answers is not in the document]
phadej: AFAICT, (@edsko should verify, as it's his creation, I'm only assuming) we use constructors elsewhere:

```
data SExpr be =
    EGlobal Global
  | EVar (Fresh be Bound)  -- ^ local
  | ECon (HsName NsConstr) -- ^ kind of global, but refers to "our" names.
```

I.e. `Global` are valid in all contexts / modules. More like built-ins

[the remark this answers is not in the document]
phadej: And to add, it's not sure why we have `Backend` class.

In `Backend.Common.Translation`, we genetate `SDecl`, `SExpr` etc. See e.g

```
instance Backend be => ToBE be Hs.Decl where
  type Rep be Hs.Decl = Decl be
  toBE be (Hs.DeclData d) = mkDecl be <$> toBE be d
  toBE be (Hs.DeclNewtype n) = mkDecl be <$> return (newtypeToBE be n)
  toBE be (Hs.DeclInstance i) = inst be <$> toBE be i
```

`inst` is also essentially a `mkDecl`. Why we don't simply have a `Hs.Decl -> SDecl` function escapes me.

The only truly backend specific thing we do at that step is generate names, but I'd argue we could delay that too, making `SDecl`, `SExpr` backend independent (Haskell AST).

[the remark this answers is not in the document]
edsko: Yes, I think I agree with Oleg's understanding in https://github.com/well-typed/hs-bindgen/pull/270#discussion_r1837007175 .

As for the simplification proposed in https://github.com/well-typed/hs-bindgen/pull/270#discussion_r1837874453 , I'm not opposed to that. I went through a few iterations of this setup, perhaps I ended up on this one more due to historical accident than anything else. If it can be made simpler, let's make it simpler.

hs-bindgen/tests/Misc.hs
```diff
@@ -36,6 +39,11 @@ goldenVsStringDiff_ name fp action = goldenTestSteps name correct action cmp upd
         | xss == yss = return Nothing
         | otherwise  = return $ Just $ ansidiff xss yss
 
+    normalise :: String -> String
+    normalise [] = []
+    normalise ('\r':'\n':r) = '\n' : normalise r
```
phadej, line 44: again?

hs-bindgen/src/HsBindgen/Hs/AST/Type.hs
```diff
@@ -21,6 +21,7 @@ data HsPrimType
     | HsPrimCUShort
     | HsPrimCLong
     | HsPrimCULong
+    | HsPrimCString
```
phadej, line 24: CString is `Ptr Char`. I'm not sure we need to introduce another way to encode it. (It's not even a newtype).

hs-bindgen/src/HsBindgen/Translation/LowLevel.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
```
phadej, line 3: Unnecessary, part of GHC2021

hs-bindgen/tests/Misc.hs
sheaf: I just cherry-picked the commit for updating the test-suite, I will drop it before merging

hs-bindgen/fixtures/macros.th.txt
```diff
@@ -0,0 +1,62 @@
...
+fLT5_1 :: forall a_0 . Fractional a_0 => a_0
+fLT5_1 = castWord64ToDouble 4635400285215260672
+fLT5_2 :: CFloat
+fLT5_2 = castWord32ToFloat 1118044160
+fLT6_1 :: forall a_0 . Fractional a_0 => a_0
+fLT6_1 = castWord64ToDouble 4669757824560005120
+fLT6_2 :: forall a_0 . Fractional a_0 => a_0
+fLT6_2 = castWord64ToDouble 4636596553866280960
+fLT6_3 :: CFloat
+fLT6_3 = castWord32ToFloat 1008874396
```
phadej, line 62: This is type-incorrect, `CFLoat` and `Float`.

hs-bindgen/src/HsBindgen/Backend/PP.hs
sheaf: I will tackle this separately; see #277.

hs-bindgen/src/HsBindgen/Backend/PP/Render.hs
```diff
@@ -9,16 +9,20 @@ module HsBindgen.Backend.PP.Render (
   , renderIO
   ) where
 
+import Data.Maybe (maybeToList)
```
phadej, line 12: FYI, `maybeToList` is the same as `Foldable.toList` we have in `HsBindgen.Imports`.

PR #272, opened 2024-11-11
Add HsBindgen.Imports module with some common imports

phadej:
... the project become big enough, that managing imports becomes a chore.

Having more utilityies easily available without needing to adjust imports (even if tool does that) allows to concentrate on actual problems.

The controversial change may be the `type Star = Data.Kind.Type`. IMHO, the C (or/and generate Hs) `Type` is more common than host Haskell `Type`, so that has to give up it's valuable name. I have written `Typ` or `Ty` way too many times, these should be `Type`.


edsko:
Fine with me. I always preferred `* -> *` over `Type -> Type`, so I don't mind `Star -> Star`. Agreed that `Type` is too overloaded especially in this context.

PR #274, opened 2024-11-12
(well-scoped) DeBruijn representation

phadej:
Changes high level and simplified Haskell representations to use DeBruijn indices.

Uses https://hackage.haskell.org/package/debruijn

Introduce `CHs` module namespace (previously `Backend.Common`).

I propose that we rearrange modules so that

- we start with `C.AST` 
- then we go to `Hs.AST` using `Hs.Translation`
- then to `CHs.AST` using `CHs.Translation`
- then to `Backend.TH` using `Backend.TH.Translation`; or `Backend.PP` using `Backend.PP.Translation`

This is almost what we have today, except we have `Translation.LowLevel` instead of `Hs.Translation`, and `Backend.Common` instead of `CHs`.

After this we'll introduce `IR` for "simplified" C in betweren `C` and `Hs`.


phadej:
I haven't yet updated the backends, will do later.

phadej:
This is ready sans some cleanup.

`debruijn` doesn't yet support GHC-9.10.1 as I haven't released the next version (there's nothing which affects the usage as in this PR). I'll do a release if this all looks ok.

phadej:
Ready for review

phadej:
Rebased

Issue #22, opened 2024-08-01
Generate test-suite for the low-level `Storable` instances


edsko:
When we generate low-level data types with `Storable` instances, we should generate a test suite that verifies that does a "self-test": generate some C code that accepts a pointer to memory, fills it with some values, then on the Haskell side use the `Storable` instance to verify that all values are in the right place, and construct an error message if not. 

Note that such a test failure would be a bug in `hs-bindgen`, so users would not be able to do much with it other than report it (or try to fix it), but at least we'd catch the problem.

(Rust `bindgen` [generates tests also](https://rust-lang.github.io/rust-bindgen/tutorial-4.html); it might be useful to take a closer look at exactly they are testing, so see if we can/should do the same.)

edsko:
The cross-compilation story here is more complicated, _if_ we want to be able to generate bindings for target B but run the test suite on target A; probably we simply don't want to support that, at least not initially.

TravisCardwell:
I am working with the [`simple_structs`](https://github.com/well-typed/hs-bindgen/blob/7ab2aa3b482743038e40a44c94a16b9cc7f8e18d/hs-bindgen/examples/simple_structs.h) example, and the fields offsets are not what I expect.

I expect `S2` to have a size of 12 bytes with an alignment of 4 bytes, with offsets `0`, `4`, and `8` for the respective fields.

[`clang_Cursor_getOffsetOfField`](https://clang.llvm.org/doxygen/group__CINDEX__TYPES.html#gaa7e0f0ec320c645e971168ac39aa0cab) is returning offsets of `0`, `32`, and `64` ([generated code](https://github.com/well-typed/hs-bindgen/blob/7ab2aa3b482743038e40a44c94a16b9cc7f8e18d/hs-bindgen/fixtures/simple_structs.pp.hs#L50-L52)) even though the size and alignment of the structure are as expected.

* Experimenting in C, the offsets are as I expect.
* Experimenting with `hsc2hs`, the offsets are as I expect.
* Experimenting with the generated offsets, my tests fail.

Is this an issue, or am I doing something wrong?

TravisCardwell:
The Rust `bindgen` tutorial is out of date.  It no longer generates tests, but it runs "build-time tests" instead.  Doing the tutorial, the tests are in the generated `bindings.rs` file and look like this:

```rust
const _: () = {
    ["Size of bz_stream"][::std::mem::size_of::<bz_stream>() - 80usize];
[...]

PR #275, opened 2024-11-13
Fix `struct` field offsets

TravisCardwell:
`libclang` reports offsets in *bits*, while `Storable` methods use offsets in *bytes*.  We do not support bit fields yet, so this PR adds an assertion that all offsets are byte-aligned, converts from bits to bytes where necessary, and documents the unit.

In addition, documentation is added to `clang_Cursor_getOffsetOfField`, functions `clang_Cursor_isBitField` and `clang_getFieldDeclBitWidth` are added, and `clang-ast-dump` is updated to dump more information.


hs-bindgen/src/HsBindgen/C/Fold/Type.hs
```diff
@@ -124,6 +127,8 @@ mkStructField unit current = do
       fieldOffset <- fromIntegral <$> clang_Cursor_getOffsetOfField current
       fieldName   <- CName <$> clang_getCursorDisplayName current
 
+      unless (fieldOffset `mod` 8 == 0) $ error "bit-fields not supported yet"
```
phadej, line 130: `fail`. Don't use `error` in `IO`.

TravisCardwell: @edsko requested that I use `error` for such assertions, only using `fail` for things that may be handled.  (I put similar assertions in the documentation reification, also running in `IO`.)

edsko:
Thanks for doing this. I'll merge this as-is, but I still think https://github.com/well-typed/hs-bindgen/issues/22#issuecomment-2470302620 is worth considering, a check such as "is `n divisible by 8" is too easy to forget.

PR #281, opened 2024-11-13
macros: support integer quotient

sheaf:
With this commit, we give the C division operator / a more general type when typechecking macros:

```hs
  (/) :: forall a. Div a => a -> a -> a
```

where `Div` is a custom typeclass defined in hs-bindgen-patterns which has two sets of instances, one for integral types where the method uses `Prelude.div`, and one for floating-point types where the method uses `(Prelude./)`.


hs-bindgen/src/HsBindgen/C/Tc/Macro.hs
```diff
@@ -296,6 +296,8 @@ data ClassTyCon n where
   IntegralTyCon   :: ClassTyCon ( S Z )
   -- | Class type constructor for Fractional
   FractionalTyCon :: ClassTyCon ( S Z )
```
phadej, line 298: is `Fractional` stuff used for anything anymore?

sheaf: It's used for literals, e.g. `0.123 :: forall a. Fractional a => a`, but probably that can be removed as the C spec says that should just be a double.

PR #283, opened 2024-11-15
Add unnamed struct example

phadej:
clang warns "declaration does not declare anything"

Arguably, we shouldn't generate anything for this declaration; but currently we do.


phadej:
I "found" this example while working on C IR.

edsko:
I'll merge this, though I agree we should probably not generate anything else.

In that light, I'm wondering if the `fromMaybe "X"` cases in `HsBindgen.Hs.Translation` , as well as the empty case in `IsStirng NameHint`, should be reconsidered. @TravisCardwell thoughts?

phadej:
>  I'm wondering if the fromMaybe "X" cases in HsBindgen.Hs.Translation ... should be reconsidered

They could be (quite naturally) removed as part of #282

TravisCardwell:
Regarding `fromMaybe "X"`, I agree that it is better to avoid using such constant names.  They could be confusing to users as well as lead to collisions when there is more than one.

Note that I followed the `fromMaybe "X"` convention in an early design of name mangling, but it is *not* used in the current design.

Regarding `NameHint`, I see that it is used with macro arguments.  Is it even possible to have an empty argument name?  The GCC [documentation][] states that parameters must be valid C identifiers (which must not be empty).  I see that it is also used with structure fields (in the simplified Haskell).  I think that all fields must have valid names.  Perhaps `NameHint` can be changed so that it is always non-empty?

[documentation]: <https://gcc.gnu.org/onlinedocs/cpp/Macro-Arguments.html>

PR #284, opened 2024-11-15
Weird example

phadej:
Arguably this is a weird example, but we shouldn't fail on it nevertheless.


phadej:
This example works now. It's weird, but it works.

Issue #262, opened 2024-11-08
Support tuples in macros


edsko:
Example:

```c
#define TWO_ARGS   0x3456, 0x789A
```

edsko:
This has been implemented.

PR #285, opened 2024-11-16
Parser for tuples in macros

edsko:
This is a first step towards #262.


sheaf:
Made redundant by #409.

PR #287, opened 2024-11-18
Add example with opaque declration

phadej:


phadej:
@TravisCardwell 

```
cabal run clang-ast-dump -- --type hs-bindgen/examples/opaque_declaration.h
## `hs-bindgen/examples/opaque_declaration.h`

* "foo"
    * cursor kind: simpleEnum CXCursor_StructDecl
    * cursor type: "struct foo"
        * kind: Right CXType_Record
            * spelling: "Record"
        * canonical: "struct foo"
clang-ast-dump: CallFailed (simpleEnum CXTypeLayoutError_Incomplete) CallStack (from HasCallStack):
  collectBacktrace, called at src/HsBindgen/Clang/Internal/Results.hs:41:14 in hs-bindgen-libclang-0.1.0-inplace:HsBindgen.Clang.Internal.Results
  callFailed, called at src/HsBindgen/Clang/Internal/Results.hs:68:12 in hs-bindgen-libclang-0.1.0-inplace:HsBindgen.Clang.Internal.Results
  ensureOn, called at src/HsBindgen/Clang/Internal/Results.hs:97:20 in hs-bindgen-libclang-0.1.0-inplace:HsBindgen.Clang.Internal.Results
  ensureNotInRange, called at src/HsBindgen/Clang/LowLevel/Core.hs:1147:34 in hs-bindgen-libclang-0.1.0-inplace:HsBindgen.Clang.LowLevel.Core
  clang_Type_getSizeOf, called at clang-ast-dump/Main.hs:93:37 in hs-bindgen-libclang-0.1.0-inplace-clang-ast-dump:Main
```

phadej:
https://joshpeterson.github.io/identifying-a-forward-declaration-with-libclang

TravisCardwell:
Both [`clang_Type_getSizeOf`][] and [`clang_Type_getAlignOf`][] are returning `CXTypeLayoutError_Invalid` for the opaque type, which the documentation states occurs when the "type declaration is invalid."  Our error handling then raises a `CallFailed` exception.

[`clang_Type_getSizeOf`]: <https://clang.llvm.org/doxygen/group__CINDEX__TYPES.html#ga027abe334546e80931905f31399d0a8b>
[`clang_Type_getAlignOf`]: <https://clang.llvm.org/doxygen/group__CINDEX__TYPES.html#gaee56de66c69ab5605fe47e7c52497e31>

I do not see a different way to detect this, so we currently need to catch the exception to handle it.  This is no problem in `clang-ast-dump`, as long as we add `CXTypeLayoutError(..)` to the exports of `HsBindgen.Clang.LowLevel.Core`.  I prepared PR #289 with this change.  Please feel free to merge or cherry-pick as necessary.

If we always need to handle this, it may be preferable to change the API to return an `Either` type instead of raise an exception.

edsko:
> Both [clang_Type_getSizeOf](https://clang.llvm.org/doxygen/group__CINDEX__TYPES.html#ga027abe334546e80931905f31399d0a8b) and [clang_Type_getAlignOf](https://clang.llvm.org/doxygen/group__CINDEX__TYPES.html#gaee56de66c69ab5605fe47e7c52497e31) are returning CXTypeLayoutError_Invalid for the opaque type, which the documentation states occurs when the "type declaration is invalid." Our error handling then raises a CallFailed exception.

I guess the question is: is it consider programmer error to call these functions when they shouldn't be (e.g., lots of functions in `libclang` are for specific use cases, "get me the alignment of this struct", which should only be called if it's in fact a struct), or are these genuine error cases that the caller cannot know about ahead of time? If it's the former, I think the exception is appropriate; f it's the latter, we should change the signature.

phadej:
>  f it's the latter, we should change the signature

See the linked blog post

TravisCardwell:
Ah, "invalid layout" is only an issue for forward declarations!  I did not realize that; sorry about that.

In the `is-forward-declaration` branch, I made the following changes:

* A new `HsBindgen.Clang.HighLevel.Extras` defines `isForwardDeclaration` like the final example of the blog post.
* That function is re-exported from `HsBindgen.Clang.HighLevel`.
* When a cursor is a forward declaration, `clang-ast-dump` shows it.  It only shows the size and alignment when not a forward declaration.
* `tryCallFailed` is moved into the library (renaming it to better match exceptions terminology in this more general context).

PR #292, opened 2024-11-20
Monomorphize `CallFailed`

edsko:
We weren't really benefiting from the polymorphic type and it was making exception handling more difficult.


TravisCardwell:
Looks good to me!

As we discussed in the call yesterday:

* `CallFailed` is meant to never be handled.
* The exception is `clang-ast-dump`, which handles the exception (printing the hint) in order to avoid program termination.
* Any error that we need to handle must be implemented using an `Either` type.

I am merging, so that I can rebase and update the `is-forward-declaration` branch (#291) accordingly.

Thank you!

PR #298, opened 2024-11-22
Add rust's bindgen outputs to fixtures

phadej:
I'm curious what Rust's bindgen generates for our examples, so lets have it in the tests as well.


phadej:
On top of #300, to include `distilled_lib_1`

edsko:
This passes with `llvm-14` but fails with `lvvm-15`. I will dig a bit.

edsko:
I'm now realizing that `lvvm-15` is causing trouble as of #300 and is unrelated to this PR. Will fix that separately.

PR #299, opened 2024-11-22
Macro type-checking: support arithmetic conversions

sheaf:
This PR generalises macro type-checking to handle C arithmetic
conversion rules as per the C standard.

What this means is that in a C expression such as

```c
  a + b
```

the values `a` and `b` are allowed to have different types, with
implicit conversions being inserted. For example:

  - `a :: CFloat`, `b :: CInt` => convert `b` to `CFloat` before adding
  - `a :: Int`, `b :: UInt` => convert `a` to `UInt` before adding

This is achieved as follows:

  - a new package, `c-expr`, with the following components:

    - `c-expr-core`, a public sublibrary which:

        - defines standard C types and C operators,
        - implements the C arithmetic conversion rules
        - defines a collection of type-classes that can be used for
          a DSL for C arithmetic expressions
        - uses Template Haskell to define a function that can, given
          a platform, implement all the instances of these typeclasses
          for types such as `CInt`, `CUInt`, `CFloat`, pointers, etc

    - `c-expr`, the main library, which exports **platform-dependent**
      modules such as `C.Expr.Posix32` or `C.Expr.Win64` with
      type-class instances for all the C arithmetic typeclasses

    - a test-suite, which uses hsbindgen-libclang to invoke Clang on
      test programs in order to check that the implementation
      of C arithmetic conversion rules in `c-expr-core` are
      compatible with Clang

  - the C macro typechecker imports `c-expr-core`, which it uses to
    compute type family reduction (e.g. the `SubRes` type family
    which computes the return type of the subtraction operator given
    its argument types)

  - `hs-bindgen` generates programs that will depend on `c-expr`, e.g.
    to turn a C macro into a Haskell function we will generate a module
    that imports the C arithmetic operator DSL provided by `c-expr`

To do this, we also needed to significantly generalise the macro
typechecker:

  - Make class constraints more general by allowing what GHC calls
    FlexibleContexts and FlexibleInstances,

  - Use a special monad for constraint solving, `TcSolveM`.
    This monad keeps track of a work list and a set of
    inert (= fully processed) constraints.

  - Solving a constraint using a top-level instance may now add
    additional constraints to the work list.

  - Instances are keyed using a TrieMap, similar to GHC's RoughMap,
    which avoids traversing all instances when doing instance matching.


typing-c/src/C/Typing/Clang.hs
```diff
@@ -0,0 +1,315 @@
...
+        Clang.CXType_VariableArray       -> return Nothing
+        Clang.CXType_DependentSizedArray -> return Nothing
+        Clang.CXType_MemberPointer       -> return Nothing
+        Clang.CXType_Auto                -> do { canTy <- Clang.clang_getCanonicalType cxTy
+                                               ; parseClangType canTy }
+        Clang.CXType_Elaborated          -> do { namedTy <- Clang.clang_Type_getNamedType cxTy
+                                               ; parseClangType namedTy }
+
+-- | Query @clang@ for canonical names for types.
+getCanonicalTypeMapping :: [ Type ] -> IO ( Map Type Type )
```
sheaf, line 141:
```suggestion
getExpansionTypeMapping :: [ Type ] -> IO ( Map Type Type )
```

typing-c/src/C/Typing/Clang.hs
```diff
@@ -0,0 +1,315 @@
...
+    expectJust cxTy =
+      \case
+        Nothing -> do
+          tyNm <- Clang.clang_getTypeSpelling cxTy
+          error $ unlines
+            [ "getCanonicalTypeMapping: could not parse CXType " ++ show cxTy
+            , Text.unpack tyNm ]
+        Just ty -> return ty
+
+queryClangForResultType :: forall n. Vec n Type -> ( Vec n String -> String ) -> IO ( Maybe Type )
```
sheaf, line 204: Can we use `Clang.CXType` instead of `String` for the `Vec n String` argument please?

typing-c/src/C/Typing/Clang.hs
```diff
@@ -0,0 +1,315 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module C.Typing.Clang
```
sheaf, line 4: Does this need to be in the library component of `typing-c`, or could it just be in the testsuite?

typing-c/typing-c.cabal
```diff
@@ -0,0 +1,80 @@
...
+        >= 0.3.2 && < 0.4
+    , temporary
+        >= 1.2   && < 1.4
+    , text
+        >= 1.2   && < 2.2
+    , vec
+        >= 0.5   && < 0.6
+
+    , hs-bindgen-libclang
+    , hs-bindgen-patterns
```
sheaf, line 49: Remove this by moving the `Clang` module to the test-suite component.

typing-c/src/C/Type.hs
```diff
@@ -0,0 +1,294 @@
...
+          4 -> WordWidth32
+          8 -> WordWidth64
+          w -> error $ "hostPlatform, unsupported word width " ++ show (8 * w)
+    , platformOS =
+        case System.Info.os of
+          "mingw32" -> Windows
+          _         -> Posix
+    }
+
+data Type
```
sheaf, line 77:
```hs
data Type a
  = Void
  | Arithmetic !ArithmeticType
  | Ptr        !a
```

so that we can use e.g. `Type (Maybe Int)` for generating type family equations.

sheaf (commented):
For platform-dependent instances, we should provide separate modules with the instances for all supported platforms. Possibly also a `Host` platform that uses CPP to re-export the appropriate module.

hs-bindgen/src/HsBindgen/C/Tc/Macro.hs
```diff
@@ -815,47 +1032,195 @@ unifyFunTys orig swapped ( arg1 NE.:| args1 ) res1 ( arg2 NE.:| args2 ) res2 = d
...
+  BinResTyCon        -> binary C.Type.integralBinaryType
+  ComplementResTyCon -> unary  C.Type.integralUnaryType
+  ShiftResTyCon      -> \ ( a ::: VNil ) -> do
+                          a' <- fromMacroType a
+                          ( r, _ ) <-
+                            C.Type.shiftType platform a'
+                              ( C.Type.Arithmetic $ C.Type.Integral $ C.Type.IntLike $ C.Type.Int C.Type.Signed )
+                          toMacroType r
+  where
+    platform = C.Type.hostPlatform
```
sheaf, line 1086: TODO: parametrise over the platform.

hs-bindgen/src/HsBindgen/C/Tc/Macro.hs
```diff
@@ -1231,93 +2048,114 @@ combineSubsts ( Subst s1 ) ( Subst s2 ) =
...
-          , quantTyBody  =       applySubst quantSubst   ty
-          }
+  where
+    solveOne = solveCt Defer ( DefaultTyVarsExcept quantTvs ) classInstancesWithDefaults
+
+{-------------------------------------------------------------------------------
+  Typechecking macros: generalisation
+-------------------------------------------------------------------------------}
+
+tcMacro :: Map CName ( Quant ( Type Ty ) ) -> CName -> [ CName ] -> MExpr -> Either TcMacroError ( Quant ( Type Ty ) )
```
sheaf, line 2112:
```suggestion
tcMacro :: Platform -> Map CName ( Quant ( Type Ty ) ) -> CName -> [ CName ] -> MExpr -> Either TcMacroError ( Quant ( Type Ty ) )
```

sheaf: This change will propagate but we don't have the infrastructure in place to pass this right now, so leave a TODO.

hs-bindgen/src/HsBindgen/C/Tc/Macro.hs
```diff
@@ -245,172 +300,238 @@ data MetaTyVar
...
+  IntLikeTyCon   :: DataTyCon ( S Z )
   -- | Type constructor for floating-point types, such as 'Float' or 'Double'.
-  FloatLikeTyCon :: !PrimFloatType -> DataTyCon Z
+  FloatLikeTyCon :: DataTyCon ( S Z )
+  -- | Type constructor for pointers
+  PtrTyCon       :: DataTyCon ( S Z )
+
+  -- | Type constructor for the type of primitive integral types,
+  -- which appear as arguments to 'IntLikeTyCon'.
+  PrimIntTyCon   :: ( PrimIntType, PrimSign ) -> DataTyCon Z
```
sheaf, line 339:
```suggestion
  PrimIntTyCon   :: !Typing.C.IntegralType -> DataTyCon Z
```

sheaf, line 339: Use a different name for `PrimIntTyCon` e.g. `PrimIntShape` or something.

Also update the comment: "Type constructor for the type-level pair of a width and sign, intended for use with `IntLike`"

typing-c/typing-c.cabal
```diff
@@ -0,0 +1,80 @@
+cabal-version:      3.0
+name:               typing-c
```
sheaf, line 2:
```suggestion
name:               c-arithmetic-expressions
```

sheaf, line 2:
```suggestion
name:               c-expr
```

typing-c/typing-c.cabal
```diff
@@ -0,0 +1,80 @@
...
+version:            0.1.0.0
+license:            BSD-3-Clause
+author:             sheaf
+maintainer:         sheaf
```
sheaf, line 6: copy the license etc from bindgen

typing-c/src/C/Typing/Arithmetic.hs
```diff
@@ -0,0 +1,455 @@
...
+import Data.Vec.Lazy ( Vec(..) )
+
+-- typing-c
+import C.Type
+
+--------------------------------------------------------------------------------
+
+-- | How is a C operator, when instantiated at a particular type, implemented?
+type OpImpl :: Nat -> Hs.Type
+data OpImpl n
```
sheaf, line 25: This should be private.

typing-c/typing-c.cabal
```diff
@@ -0,0 +1,80 @@
...
+    src
+  exposed-modules:
+    C.Type
+    C.Typing
+    C.Typing.Arithmetic
+    C.Typing.Arithmetic.Classes
+    C.Typing.Arithmetic.Instances
+    C.Typing.Arithmetic.TH
+    C.Typing.Arithmetic.TyFams
+    C.Typing.Clang
```
sheaf, line 64: Separate out a simple "C.Expr" (platform-dependent) module for the C-like DSL, and then maybe a separate sublibrary for the stuff that hs-bindgen needs to import (or could be exposed `.Internal` modules)

edsko:
Big chunk of work! Great to get this in.

PR #303, opened 2024-11-24
Define Fold operations using MonadUnliftIO, rename FoldM to Eff

phadej:
Move the FoldM (renamed to Eff) "effect" monad to hs-bindgen from hs-bindgen-libclang

There weren't anything folding specific in that monad, nor folding needs to be specialised to that  monad either.


phadej:
Probably makes sense to merge only after everything else is merged.

edsko:
Makes sense. I wonder if we should just move to `effectful` or something?

PR #315, opened 2024-11-28
Test with LLVM 14, 16 and 18

phadej:


phadej:
Interesting, I get

```
[2 of 2] Linking D:\\a\\hs-bindgen\\hs-bindgen\\dist-newstyle\\build\\x86_64-windows\\ghc-9.4.8\\hs-bindgen-libclang-0.1.0\\build\\clang-ast-dump\\clang-ast-dump.exe
ld.lld: error: undefined symbol: __declspec(dllimport) clang_getUnqualifiedType
```

but only on windows with LLVM 16 and 18.

edsko:
Nice, I like that we are testing with multiple llvm versions now.

Issue #26, opened 2024-08-01
Include Haddocks for exported (low-level) bindings


edsko:
When we generate foreign imports from C headers, _if_ the declarations in those headers have documentation attached, then we should include that documentation in the generated bindings also.

Ideally we would be using Haddock formatting here, but given the absence of a general agreed standard for the documentation of C headers, that might be hard to do in general. Nonetheless, we could at least support common formats, most importantly [Doxygen](https://www.doxygen.org/).

edsko:
As mentioned in #25, it might also be useful to include C variable names in as part of the Haddocks.

TravisCardwell:
I defined the wrappers in `doxygen_wrappers.h` as well as the corresponding Haskell API in `HsBindgen.Clang.Doxygen` and the necessary enumerations in `HsBindgen.Clang.Doxygen.Enums`, in the `doxygen` branch.

Of the content in `clang-c/Documentation.h`, the following is *not* (yet) implemented:

* Struct `CXComment` attributes `ASTNode` and `TranslationUnit` are not implemented.  I think that we will traverse the AST top-down, so I doubt we will need these.
* Typedef `CXAPISet`, as well as related functions `clang_createAPISet` and `clang_disposeAPISet`, are not implemented.  I do not think that we will need these for our purposes.
* Functions `clang_getSymbolGraphForUSR` and `clang_getSymbolGraphForCursor` are not implemented.  I do not think that we will need these for our purposes.

Please feel free to let me know if we will need something that I have not implemented, though I will no doubt find out when I proceed with implementation.

TravisCardwell:
I am currently implementing an `ast-dump` utility that walks the AST and dumps information so that I can see exactly how the AST is structured.  The LLVM tool works fine, of course; I want to see details via our library.

Just to get started quickly, I started implementing it as a `hs-bindgen-libclang` test suite, like the tutorial.  It could be useful in the long run, though.  Perhaps it should be put in a separate package, so that more dependencies can be used.  For example, it would probably be worthwhile to implement a CLI parser using `optparse-applicative` to provide options to select what information is displayed.  I am being *very* verbose with documentation output since that is what I am working on, but such information would get in the way when working on something else.

BTW, output is currently formatted in Markdown

I have not committed the code yet, but I should be able to do so tomorrow.

TravisCardwell:
I wonder if it is worth adding `HsBindgen.Clang.Util.Classification` predicates for the documentation API.

[...]

PR #318, opened 2024-11-29
Add clang-syntax

phadej:
Our `clang-ast-dump` is nice, but `clang` has this feature built in:

image

### DIFF
--- a/clang-syntax
+++ b/clang-syntax
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+clang -Xclang -ast-dump -fsyntax-only "$@"


TravisCardwell:
Indeed; the `clang` command is documented in the `clang-ast-dump` README.

The purpose of `clang-ast-dump` is to explore the `libclang` API.  While we can see the structure using the `clang` command, it does not show how the information is retrieved via the API.  `clang-ast-dump` allows us to experiment with the implementation, easily trying it on any number of examples.

An additional purpose of `clang-ast-dump` is to explore the documentation AST, which the `clang` output does not include.

Since it is a development tool, we may remove `clang-ast-dump` when it is no longer useful.  Please do not do so while #26 is still open, however, as I use it when developing documentation translation.

PR #326, opened 2024-12-03
Handle anonymous field structs

phadej:


hs-bindgen/src/HsBindgen/C/Fold/Type.hs
```diff
@@ -44,9 +44,26 @@ processTypeDecl unit ty = do
...
+    -- TODO: | PathPtr Path
+    -- TODO: | PathConstArray Natural Path
   deriving Show
 
+isPathTop :: Path -> Bool
+isPathTop PathTop = True
+isPathTop _       = False
+
+-- | Make definition name for anonymous structures from their path
+mkDefnName :: Path -> DefnName
```
phadej, line 58: I'm not 100% happy with the `Path` structure, it allows combinations which should't occur. But I'd like to first add more features to see what's there before putting too much work into thinking up front.

PR #334, opened 2024-12-05
Foreign import declarations

phadej:


hs-bindgen/fixtures/distilled_lib_1.pp.hs
```diff
@@ -30,6 +30,8 @@ a_DEFINE_2 = 2
 tWO_ARGS :: forall a0. P.Integral a0 => a0
 tWO_ARGS = 13398
 
+foreign import capi safe "distilled_lib_1.h some_fun" some_fun :: (F.Ptr CATypeT) -> CUint32T -> Void -> IO CInt32T
```
TravisCardwell, line 33: I noticed that parameter `uint8_t k[]` is being translated do `Void`.  Array parameters are rewritten as pointers, so perhaps we would like to translate to `F.Ptr CUint8T` if not `[CUint8T]`?  I think that the list translation would be preferred, however, so that a null pointer cannot be used.

Perhaps this is a detail that will be addresses in a separate/future PR?

hs-bindgen/src/HsBindgen/C/AST.hs
```diff
@@ -103,7 +103,10 @@ data MacroDecl
 data Function = Function
     { functionName :: CName
     , functionType :: Type
-    -- TODO: add location
+
+    -- TODO: we might not need functionHeader field,
+    -- https://github.com/well-typed/hs-bindgen/issues/333
+    , functionHeader :: String
```
TravisCardwell, line 109: Perhaps `FilePath` type is appropriate here?

TravisCardwell (commented):
Looks good!

I commented on one issue that we may want to address separately as well as one trivial change.

hs-bindgen/fixtures/distilled_lib_1.pp.hs
phadej: https://github.com/well-typed/hs-bindgen/blob/main/hs-bindgen/src/HsBindgen/C/Fold/Type.hs#L290 added in #330

TravisCardwell (approved):
Looks good to me!

PR #337, opened 2024-12-13
Preprocessor backend fixes

TravisCardwell:
This PR fixes some preprocessor backend issues that I spotted while working on other things this week.

GHC pragmas are now resolved based on the declarations generated.  The following extensions are used when needed:

* `CApiFFI`
* `DerivingStrategies`
* `GeneralizedNewtypeDeriving`
* `PatternSynonyms`
* `StandaloneDeriving`

Support for resolved names without imports is added.  This is required for `()`.

Negative integers are now parenthesized.

### DIFF
--- a/hs-bindgen/src/HsBindgen/Backend/PP/Render.hs
+++ b/hs-bindgen/src/HsBindgen/Backend/PP/Render.hs
@@ -55,10 +55,17 @@ renderIO opts (Just fp) modl = withFile fp WriteMode $ \h ->
 
 instance Pretty HsModule where
   pretty HsModule{..} = vsep $
-      "{-# LANGUAGE NoImplicitPrelude #-}"
+      vcat (map pretty hsModulePragmas)
     : hsep ["module", string hsModuleName, "where"]
     : vcat (map pretty hsModuleImports)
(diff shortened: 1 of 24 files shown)


hs-bindgen/src/HsBindgen/Backend/PP/Render.hs
```diff
@@ -173,7 +181,8 @@ prettyExpr env prec = \case
     EFree x  -> pretty x
     ECon n   -> pretty n
 
-    EIntegral i _ -> showToCtxDoc i -- TODO: why we have type annotation if we don't use it?
+    -- TODO: why we have type annotation if we don't use it?
```
phadej, line 184: I keep TODO's on the line they referring to, so `git grep TODO` shows the context. Don't change.

TravisCardwell: I see!  Thanks for letting me know.  I shall put this comment back on the same line.

I generally stay within the maximum line length (80 characters for this project), so I have likely moved such comments before when changing the code resulted in a line that exceeded the limit.  From now, I will remember this exception and try to keep such `TODO` on a relevant line in such cases.

Issue #331, opened 2024-12-04
Minimal default name mangling


TravisCardwell:
In https://github.com/well-typed/hs-bindgen/issues/312#issuecomment-2510114083 @phadej wrote:

> IMHO, the default name mangling should do as little as necessary to make valid Haskell names.
>
> Doing underscores to camelCase like translations feels a bit tricky. Perfectly we will be able to describe the mangling rules in few sentences. Users should be able to predict what the Haskell names will be. IMHO it doesn't matter what are the "idiomatic" names in Haskell.
>
> In particular, recently I started to dislike the `C` prefix. Why we have it. (`Foreign.C.Types` has it for a reason, but we don't share the same reasoning: there are no types to clash with).

In https://github.com/well-typed/hs-bindgen/issues/312#issuecomment-2510253092 @TravisCardwell replied:

> This is a question for @edsko, but I think it is just a matter of taste/preference.
>
> A few technical notes:
>
> * A prefix is required for type constructors to preserve leading underscores without translation or escaping, since type constructor names may not start with an underscore.
>
> * When preserving C names as much as possible, a prefix makes it possible to avoid converting the case of the first letter of the C name.

I discussed this with @edsko yesterday.

We currently provide two sets of name manglers.  The default (`defaultNameMangler`) creates Haskell-style type/constructor names while variables are minimally mangled.  It is safer than `haskellNameMangler`, which creates Haskell-style variable names as well.  By "safety," I mean risk of name collision.  "Safe" means relatively low risk.  In addition, the API enables developers to create their own name manglers.

@edsko is fine with making the default do as little as necessary to make valid Haskell names.  Users who want Haskell-style names can use the `haskellNameMangler`.

@phadej Do you think it is worth adding a prefix in order to avoid case conversion?  Example: C types `foo` and `Foo` may both exist.

If we do not prefix every type constructor name, we have to decide what to do about leading underscores.  Should we include a prefix only in that case?

phadej:
> Example: C types foo and Foo may both exist.

Would they really in any reasonable library?

phadej:
> we could generate the following, dropping the underscore and treating it like the C names are the same 

We can't. Firstly it's quite tricky to implement. Currently translation only considers the C names (and those only in quite special cases). The name mangling is an utility.

Again, the less surprising (i.e. more predictable) the pipeline is the better.

TL;DR lets keep it simple

TravisCardwell:
[...]

Issue #316, opened 2024-11-28
Add C source information to `Hs` and `SHs` ASTs


TravisCardwell:
We would like to track how various parts of our generated ASTs are created, referencing the C source.

One motivation for this is test generation (#22), which requires generating both C and Haskell code for testing.  Generating the C test code from a C `Header` is not a good option because we would need to reimplement a lot of the logic for translating from C to Haskell.  If the Haskell AST includes C source information, we could traverse the Haskell AST and determine exactly what test functions are required, perhaps referencing the C `Header` to get C details.

For example, for a given `data` declaration (called `Struct` in `Hs` and `Record` in `SHs`), it is useful to know the name of the corresponding C type, including the C namespace.  When generating C code, the namespace determines how an identifier is written.

* Ordinary namespace: `foo`
* `struct` tag namespace: `struct foo`
* `union` tag namespace: `union foo`
* `enum` namespace: `enum foo`

C source information can also be used to improve the generated Haddock documentation (#26).  For example, we could output corresponding C names to help users understand/confirm which Haskell maps to which C.

We should include source locations, which may optionally be output in `LINE` pragmas (#74).  We could even consider including source location information in generated Haddock documentation.

Related to #23 (which is for the high-level API)

TravisCardwell:
One idea is to change [`CName`](https://github.com/well-typed/hs-bindgen/blob/85914954e05b4e5c215b3f66bb2003810ee5ee8a/hs-bindgen/src/HsBindgen/C/AST/Name.hs#L14) to be a phantom type with a `Namespace` parameter, like `HsName`.

```haskell
data Namespace =
    NsOrdinary
  | NsStructTag
  | NsUnionTag
  | NsEnumTag

newtype CName (ns :: Namespace) = CName { getCName :: Text }
  ...
```

This type does not include *member* namespaces, which are separate per structure/union.  Discussing with @edsko yesterday, he suggested that we may want to model member namespaces as well, perhaps using a GADT.

phadej:
> struct tag namespace: struct foo

it doesn't work in general. Counter example

```c
struct foo {
  struct { int x; int y } bar;
  int z;
}
```

we will create `CFooBar` which doesn't correspond to any type we could reference in C.

More generally, what you proposes is essentially delaying name mangling.

TravisCardwell:
Thanks for the example!

[...]

PR #338, opened 2024-12-17
Add C source information to ASTs

TravisCardwell:
The primary goal of this PR is to add C source information to the ASTs (#316), for use in test and documentation generation.  This is the initial attempt at implementing this so that we have something to discuss (https://github.com/well-typed/hs-bindgen/issues/316#issuecomment-2516549401).  We plan to discuss during the call on Thursday (December 19th).

This needed to build upon the new minimal default name mangling (#331), so the single commit from that branch is cherry-picked.  Since a primary motivation is to support test generation (#22), my initial implementation of test generation is also implemented in this branch.  Some commits are cherry-picked, some commits are rewritten from scratch with this different design, and new commits are added.

This comment serves as a guide to the various [commits](https://github.com/well-typed/hs-bindgen/compare/main...source-info).  Some commits have descriptions, and links to related comments are included below.

1. Prepare types for adding source information
    * Add `Field` type (cherry-picked)
    * Add `OpaqueStruct` and `OpaqueEnum`
2. Add source locations
    * Add source locations to C AST
    * Make `typedef` source locations relative (https://github.com/well-typed/hs-bindgen/issues/316#issuecomment-2526657759)
3. Name mangling
    * Implement minimal default name mangling (cherry-picked)
    * Change name mangling overrides (https://github.com/well-typed/hs-bindgen/issues/331#issuecomment-2530703068)
    * Refactor `struct` name mangling
4. Add origin information (https://github.com/well-typed/hs-bindgen/issues/316#issuecomment-2533229352)
    * Add origin info to `Hs` AST
    * Add origin info to `SHs` AST
5. Test generation (cherry-picked)
    * Add `hs-bindgen-testlib` (WIP)
    * Continue `hs-bindgen-testlib` (WIP)
    * Rename to `SameSemantics`
    * Update `SameSemantics` properties/assertions
    * Update `Storable` properties/assertions
    * Add `sizeof` and `alignof` assertions/tests
    * Audit C types and headers
    * Rename to `Preturb`
    * Add size to `Preturb`
    * Add missing comments
    * Rewrite `GenSeq`
    * Rename to `HsGenSeq1SameSemanticsCGenSeq1`
    * Format C using `clang-format`
    * Apply C linter suggestions
    * Move `Text.SimplePrettyPrint`
    * Refactor `Text.SimplePrettyPrint`
    * Sort lists
    * Implement test generation (WIP)
6. Test generation (new) (https://github.com/well-typed/hs-bindgen/issues/316#issuecomment-2538066984)
    * Fix typos
    * Use `Hs` AST
    * Add `DeclName` (https://github.com/well-typed/hs-bindgen/issues/316#issuecomment-2537444410)
    * Implement C header and source generation (WIP)
    * Add `hangs`
    * Implement Haskell test module generation (WIP)
    * Generate `Spec` and `Main` modules
    * Define/use common context
    * Add documentation
    * Add `Main` module to sample Cabal config
7. Fix source location normalization (https://github.com/well-typed/hs-bindgen/pull/338#issuecomment-2552893328)
    * Render `Hs` AST using `Data.TreeDiff.Expr`
    * Revert make `typedef` source locations relative
    * Make `SourcePath` relative in `toExpr` instance


TravisCardwell:
Tests fail on Windows because the source paths in the `.hs` fixtures are displayed as `FilePath` and Windows uses backslash directory separators that do not match the forward slash directory separators in the fixtures.  The declarations are rendered using the `Show` instance.

Note that this issue is avoided in the `.tree-diff.txt` fixtures by displaying source paths as a list instead.

One way to resolve this issue would be to change the `Show SourcePath` instance to render source paths in a platform-independent manner.  Rendering them as lists like in the `.tree-diff.txt` fixtures is *not* a straightforward solution, however, because we render them in strings that specify the line and column (such as `examples/simple_func.h:3:22`) in `Show` instances of other types.

We *could* mangle source path strings, converting Windows directory separators to POSIX directory separators.  Perhaps it is better to display source paths using actual/native directory separators to users, though.  I can implement it if we decide to go this route.

My inclination is to not do special rendering in `Show` instances.  The `Show SourcePath` instance could then render as lists, while the instances for the types that include `SourePath`s can derive `Show` instances.  Special rendering of source paths can be implemented using separate functions, optionally using type classes (like TTC).

edsko:
Re rendering of paths, see https://github.com/well-typed/hs-bindgen/blob/be144fed4f0a10db5ba6453a8d98b9e95d5ea820/hs-bindgen-libclang/src/HsBindgen/Clang/HighLevel/SourceLoc.hs#L53 which I introduced for that precise reason.

TravisCardwell:
Indeed, that `newtype` allows us to define our own instances.  The two options suggested in the above comment:

* Mangle `SourcePath` paths so that they are consistent across platforms for the purposes of our tests.  Displayed paths would use POSIX directory separators even on Windows.
* Use `Show` instances for debugging and testing only, rendering `SourcePath` as a list like in `.tree-diff.txt` fixtures.  Use separate functions to render output, rendering `SourcePath` with the actual directory separators.

Note that this branch already includes a change that makes the paths relative to the package directory when under that directory (https://github.com/well-typed/hs-bindgen/issues/316#issuecomment-2526657759).

edsko:
Ah, sorry. Yes, I think we should go for something again to what we do with the `tree-diff` tests, so that any massaging we do, is confined to the tests only.

TravisCardwell:
I changed the `hs` golden tests to pretty-print `Data.TreeDiff.Expr` representations of declarations instead of using `Show`.  This allows us to use the same instance for `SourcePath` that is used in the `tree-diff` fixtures.

Note that I did not use `ediffGolden1` because we do not have `Eq` instances for the `Hs` AST.  We could use it if we decide to implement `Eq` instances.  For now, per-line string comparisons are made.  This is still an improvement over the previous implementation using `Show` because the `Expr`s are pretty-printed across multiple lines.

I imagine that we would like to make source locations relative in the tests only as well.

I planned on doing this by moving the type class to the tests package.  Doing so at this level requires a lot more instances, of course, since the AST has to be traversed.  Unfortunately, I ran into the age-old Haskell issue of public API vs testing.  It worked great for the Haskell representations because we export some modules specifically for testing, though I had to add some.  It did not work for the C API, however, because our `HsBindgen.Lib.CHeader` type is explicitly opaque.

The Cabal file includes the following comment.  Perhaps we can use an internal library to expose more things for testing?

```haskell
-- Exposed for the sake of tests
-- TODO: We should reconsider the proper way to export these.
```

Making paths relative in general requires context (current working directory), but we do not have to worry about the general case since it is specific to our tests.  I implemented a fix in the `ToExpr` instance for `SourcePath`.  It should work as long as we keep our include directories under `hs-bindgen*` package directories.

Alas, my initial implementation did not work in CI!  Here is a diff:

```diff
- ["hs-bindgen", "musl-include", "bits", "alltypes.h"],
+ ["hs-bindgen-0.1.0", "musl-include", "bits", "alltypes.h"],
```

In the CI environment, Cabal appends the version number to the package directory names!  I resolved this issue by normalizing package directory names.

edsko (approved):
We discussed this in a meet. Conclusions:

- Name mangling will get a `CName -> HsName` map in addition to the `HsName -> HsName` map, or alternatively a `Maybe CName -> HsName -> HsName` map, where the latter allows the user to specify the `CName` _if_ they so desire (useful to state intent and useful for disambiguation).
- We will make paths relative to the `cwd` as we construct them, rather than doing this in tests.
- We will use the `tree-diff` infrastructure for comparison, adding missing `Eq` instances (though it's not totally clear what they are required for).

Feel free to merge when ready.

Issue #340, opened 2024-12-19
Fail gracefully


phadej:
Currently we don't fail gracefully, thus e.g. when we run tests we get situations like

```
golden
  bitfields
    treediff: golden: bit-fields not supported yet
CallStack (from HasCallStack):
  error, called at src/HsBindgen/C/Fold/Type.hs:391:47 in hs-bindgen-0.1.0-inplace:HsBindgen.C.Fold.Type
```

There's two part, the `tree-diff` probably should catch exceptions and fail more gracefully (report test as failed, not abort everything);
But also we should fail more gracefully to begin with, I'm sure there will be unsupported features for a long time.

PR #344, opened 2024-12-20
More graceful failure

phadej:
`error` is simply wrong in `IO`-ish contexts. Use fail or throwIO.

cc @edsko @TravisCardwell 

Related to #340 (but doesn't fix it, as we have more `error`s in later stages).


phadej:
rebased

Issue #19, opened 2024-08-01
Support flexible array members in low-level API


edsko:
See https://en.wikipedia.org/wiki/Flexible_array_member .

edsko:
Discussed this with the client. We propose something along the following lines:

Given

```c
struct foo {
        int len;
        char data[];
};
```

(which potentially has many more fields than just these two, of course),  generate something like

```haskell
data PartialFoo = PartialFoo {
      len :: Int
     -- plus other fields, if any, but without the flexible array member
    }

-- generated instance, as usual
instance Storable PartialFoo where
```

In addition, we should also generate an instance of this class:

```haskell
class SupportsFlexibleArray a where
  arrayMemberOffset :: Proxy a -> Int
```

Moreover, the _user_ must provide an instance of this class:

```haskell
class SupportsFlexibleArray a => HasFlexibleArrayLength a where
  arrayMemberLength :: a -> Int
```

Finally, we provide the following infrastructure in `hs-bindgen-patterns`:

```haskell
newtype WithFlexbileArrayMember a = WithFlexbileArrayMember a

-- Single instance, defined once and for all 
instance (Storable a, HasFlexibleArrayLength a) => Storable (WithFlexibleArrayMember a) where 
  ...  
```

phadej:
```haskell
instance (Storable a, HasFlexibleArrayLength a) => Storable (WithFlexibleArrayMember a) where
```

cannot be defined. Even we have `sizeOf :: a -> Int`; the `a` there is really meant to be `Proxy a` (but this is pre-Proxy) interface.

One way to understand this is that `Storable` types are of fixed size; consider usage in `Storable.Vector`.

We can have Storable-like class though, with peek and poke.

edsko:
Sorry, I should have answered here. Good point regarding `Storable` implying a fixed size. Yes, let's introduce a custom class, and perhaps we should then derive instances for that custom class for _all_ types, and `Storable` instances _additionally_ for _some_ types?

PR #345, opened 2024-12-20
Recognise flexible array members

phadej:


phadej:
Builds on top of #344 
In this PR we recognise flexible array members, but doesn't yet generate any particular code for them. The design in #19 has still answered questions.

hs-bindgen/src/HsBindgen/C/Fold/Type.hs
```diff
@@ -204,15 +205,21 @@ processTypeDecl' relPath path unit ty = case fromSimpleEnum $ cxtKind ty of
                     sizeof    <- liftIO (clang_Type_getSizeOf  ty)
                     alignment <- liftIO (clang_Type_getAlignOf ty)
 
-                    fields <- HighLevel.clang_visitChildren decl $ \cursor -> do
+                    fields' <- HighLevel.clang_visitChildren decl $ \cursor -> do
                         mfield <- mkStructField relPath unit declPath cursor
                         return $ Continue mfield
 
+                    (flam, fields) <- case partitionEithers fields' of
```
edsko, line 212: Surprised to see that the flam does not have to be the last field.

phadej: It has to be:

```
flam.c:2:7: error: flexible array member 'bar' with type 'int[]' is not at the end of struct
  int bar[];
      ^
flam.c:3:7: note: next field declaration is here
  int len;
      ^
1 error generated.
```

but libclang AST doesn't really enforce it. So I found it easier just to treat it as it could occur anywhere in folds, and simply rely on the fact that libclang rules out the invalid structs for us. In other words, I don't see a point to do checks that there is at most one flam at the end of a struct.

edsko: Ah, I see, makes sense.

Issue #293, opened 2024-11-20
Generate "Prelude" of standard C types


edsko:
In the repository there is a file `standard_headers.h` which imports all of the C standard library. We also have a skeleton implementation of some code (`bootstrapPrelude`) that trawls through these definitions; currently this is only being used to check if the macro parser is failing to parse some macros that we should be parsing (just to have a source of examples). We should extend this to make a list of all standard types in the standard library, and ensure that for all of these standard C types we have a well-defined translation to standard Haskell types (either from `base` or from `hs-bindgen-patterns`).

TravisCardwell:
Directory `hs-bindgen/musl-include` is not searched unless we specify it as an option as follows:

```
$ cabal run hs-bindgen -- \
    --clang-option "-I$(pwd)/hs-bindgen/musl-include" \
    dev prelude
```

Do we want to always/only use that directory, or might we want to run the command using different include directories?  Would it be worthwhile to either always inject that include directory or expose our own `-I` option that defaults to that directory accordingly?

TravisCardwell:
I get a fatal error that `stdatomic.h` is not found.  It is included by `standard_headers.h`, in the C11 section.  Support for atomic primitives and types is an [optional feature][], and it sounds like Musl does not support it ([reference](https://github.com/unikraft/lib-musl/issues/36)).

[optional feature]: <https://en.wikipedia.org/wiki/C11_(C_standard_revision)#Optional_features>

What should we do?

TravisCardwell:
I think that we should document the exact [`VERSION`][] of `musl-include` that we vendor.  We also need to include the [`COPYRIGHT`][] file.

[`VERSION`]: <https://git.musl-libc.org/cgit/musl/tree/VERSION>
[`COPYRIGHT`]: <https://git.musl-libc.org/cgit/musl/tree/COPYRIGHT>

My suggestion is to vendor the following:

```
hs-bindgen/
  musl/
    COPYRIGHT
    VERSION
    include/
      ...
```

Sound good?

TravisCardwell:
Assuming that we want to update the vendored Musl headers as new versions are released, I would like to make doing so trivial by using a script.  Would it be acceptable to add a Bash script to do this named `scripts/musl-update.sh`?

phadej:
>  include the generic architecture headers?

No, they include x86_64 build.

I fail to understand how updating musl is related to generating prelude of standard C types.

phadej:
[...]

PR #347, opened 2024-12-27
Add minimal `LibC` module

TravisCardwell:
The purpose of this PR is to merge a minimal `LibC` implementation so that we can start thinking about how to integrate it into `hs-bindgen`.

Here is the relevant commit message:

> This commit includes the "easy" types.  Once merged, we can start work on integrating using of the library while simultaneously filling in the details.
>
> Notes:
>
> * The `x86_64` architecture `Arch` module is put in place, though no architecture-specific types that require it are implemented yet.  TODO comments marked "`arch`" need to be implemented in `Arch` modules.
> * As decided on in our last call, integral types with a specific bit width are implemented using corresponding Haskell types.  This allows them to be implemented in general, not in `Arch` modules.
> * Documentation indicates that `stdint.h` should define types `int_fast8_t`, `int_fast16_t`, `int_fast32_t`, and `int_fast64_t` (as well as the unsigned versions), but Musl does *not* define `int_fast16_t` or `int_fast32_t` (or the unsigned versions).  I went ahead and implemented them in the module anyway.
> * No types from `stdatomic.h`, `stdbit.h`, or `stdckdint.h` are implemented yet.  These are not (yet) supported by Musl.
> * The types implemented use names that match those in `base`.  I can rename them to use our default name mangling rules if we decide to do that instead.  I did not implement any structures or enumerations yet; I will do so once we decide on naming.

Other changes included in this PR:

* `bootstrap/standard_headers.h` is documented, with links to references.
* Vendored Musl headers are now stored per architecture.  Bash script `musl-update.sh` can be used to easily add architectures as well as update to new versions, without requiring any cross-compilation toolchains.
* The Musl `VERSION` and `COPYRIGHT` files are now included.
* Executable `hs-bindgen` top-level options are added:
    * `--standard` allows selection of the C standard.
    * `--gnu` enables GNU extensions.
    * `--target` allows specification of the target architecture (triple).
* We now use `c17` by default, instead of `gnu17` (the Clang default).
* The `prelude` command no longer appends to the log file; it instead overwrites any existing log file with a new one.
* The `prelude` command now has an `--include` option for specifying the header directory to use.  When Musl headers for the host architecture are vendored, it defaults using them.  Otherwise, the user must specify a directory.


TravisCardwell:
Changes:

* Rebased
* Removed re-exports (https://github.com/well-typed/hs-bindgen/issues/293#issuecomment-2578906412)
* Removed `stdint` wrappers (https://github.com/well-typed/hs-bindgen/issues/293#issuecomment-2578906412)

We will do doubt make further changes as we refine the design, but I am merging this as an initial implementation.

PR #348, opened 2025-01-06
Improve `Eq` `Hs` instances

TravisCardwell:
I received feedback about how `Vec.withDict` can be used to recover the `SNatI` dictionary of a `Vec`, and using it makes adding `SNatI` many constraints unnecessary.  Thank you @sheaf!

I also received feedback about using `GEq`.  Thank you @phadej!

This commit refactors the changes made to commit "Add Hs AST Eq instances, test using tree-diff" (`916ab6db`).  It removed all added `SNatI` constraints and uses `GEq` instances.

An unsightly aspect of the refactored code is the use of auxiliary functions just to bring type variables that are not part of the type signature in scope, but I do not know of a way to do this otherwise.


hs-bindgen/src/HsBindgen/Hs/AST.hs
```diff
@@ -211,8 +222,7 @@ data VarDecl =
 type SigmaType :: Star
 data SigmaType where
   ForallTy ::
-       SNatI n
-    => { forallTySize    :: Size n
+       { forallTySize    :: Size n
```
sheaf, line 225: I would remove this field as it can be deduced from the `Vec n NameHint` field.

hs-bindgen/src/HsBindgen/Hs/AST.hs
```diff
@@ -221,8 +231,20 @@ data SigmaType where
...
+        -> Vec nL NameHint
+        -> PhiType nL
+        -> Size nR
+        -> Vec nR NameHint
+        -> PhiType nR
+        -> Bool
+      aux sL vL pL sR vR pR = Vec.withDict vL $ Vec.withDict vR $
+        case Nat.eqNat @nL @nR of
+          Just Refl -> sL == sR && vL == vR && pL == pR
+          Nothing   -> False
```
sheaf, line 247: I think this implementation (and similar ones below, e.g. `TyConAppTy`, `ClassTy`, `ElimStruct` are needlessly complicated. I have suggested something for `ElimStruct` (which seems to be the most complicated), please let me know what you think.

hs-bindgen/src/HsBindgen/Hs/AST.hs
```diff
@@ -242,30 +264,42 @@ data TauType ctx
...
+      aux :: forall nL nR.
+           C.DataTyCon nL
+        -> Vec nL (TauType ctx)
+        -> C.DataTyCon nR
+        -> Vec nR (TauType ctx)
+        -> Bool
+      aux dL vL dR vR = Vec.withDict vL $ Vec.withDict vR $
+        case Nat.eqNat @nL @nR of
+          Just Refl -> dL == dR && vL == vR
+          Nothing   -> False
```
sheaf, line 283: See comment on `instance Eq SigmaType`.

hs-bindgen/src/HsBindgen/Hs/AST.hs
```diff
@@ -242,30 +264,42 @@ data TauType ctx
...
+      aux :: forall nL nR.
+           C.ClassTyCon nL
+        -> Vec nL (TauType ctx)
+        -> C.ClassTyCon nR
+        -> Vec nR (TauType ctx)
+        -> Bool
+      aux cL vL cR vR = Vec.withDict vL $ Vec.withDict vR $
+        case Nat.eqNat @nL @nR of
+          Just Refl -> cL == cR && vL == vR
+          Nothing   -> False
```
sheaf, line 302: See comment on `instance Eq SigmaType`.

hs-bindgen/src/HsBindgen/Hs/AST.hs
```diff
@@ -388,60 +421,57 @@ data ElimStruct t ctx where
...
+
+      eqAdds ::
+           (nL ~ nR, mL ~ mR)
+        => Add nL mL pL
+        -> Add nR mR pR
+        -> (pL :~: pR)
+      eqAdds l r = case (l, r) of
+          (AS l', AS r') -> case eqAdds l' r' of
+            Refl -> Refl
+          (AZ, AZ) -> Refl
```
sheaf, line 452:
```suggestion
  ElimStruct idxL sL addL tL == ElimStruct idxR sR addR tR =
    fromMaybe False $ do
      Refl <- geq (ApEq sL) (ApEq sR)
      guard $ sL == sR
      Refl <- geq idxL idxR
      case eqAdds addL addR of
        Refl -> return $ tL == tR
```

For this you will need:

  1. A `GEq Idx` instance. See below (but it would be an orphan unless it gets added to the `debruijn` library).
  2. A `GEq Struct` instance which you can simply derive via the `ApEq` newtype I defined above.
  3. To keep `eqAdds` around (see simpler implementation below).

```hs
instance GEq Idx where
  geq i j
    | idxToInt i == idxToInt j
    = Just $ unsafeCoerce Refl
    | otherwise
    = Nothing
```

```hs
eqAdds :: Add n m pL -> Add n m pR -> pL :~: pR
eqAdds (UnsafeAdd !_) (UnsafeAdd !_) = unsafeCoerce Refl
```

hs-bindgen/src/HsBindgen/Hs/AST.hs
```diff
@@ -98,6 +100,15 @@ data Struct (n :: Nat) = Struct {
...
 
+instance GEq Struct where
+  geq :: forall n m. Struct n -> Struct m -> Maybe (n :~: m)
+  geq sL sR = Vec.withDict (structFields sL) $ Vec.withDict (structFields sR) $
+    case Nat.eqNat @n @m of
+      Just Refl
+        | sL == sR  -> Just Refl
+        | otherwise -> Nothing
+      Nothing       -> Nothing
+
```
sheaf, line 111: With the `ApEq` newtype I suggested above, this could simply be

```hs
data Struct (n :: Nat) = ...
  deriving stock (Eq, Generic, Show)
  deriving GEq via ApEq Struct
```

hs-bindgen/src/HsBindgen/C/AST/Macro.hs
```diff
@@ -122,6 +122,31 @@ data MFun arity where
...
+  geq MRelGE      MRelGE      = Just Refl
+  geq MRelEQ      MRelEQ      = Just Refl
+  geq MRelNE      MRelNE      = Just Refl
+  geq MBitwiseAnd MBitwiseAnd = Just Refl
+  geq MBitwiseXor MBitwiseXor = Just Refl
+  geq MBitwiseOr  MBitwiseOr  = Just Refl
+  geq MLogicalAnd MLogicalAnd = Just Refl
+  geq MLogicalOr  MLogicalOr  = Just Refl
+  geq _           _           = Nothing
+
```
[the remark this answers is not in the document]
TravisCardwell: Thank you *very* much for the review!

I had not seen/digested `HsBindgen.Util.TestEquality` yet.  I was just wanting `equals` yesterday but did not know how to implement it!

Your usage in the `ElimStruct` instance illustrates the value of `ApEq`, in addition to use with `deriving via`.  I added it to the `HsBindgen.Util.TestEquality` module.

hs-bindgen/src/HsBindgen/Hs/AST.hs
TravisCardwell: I implemented this in a separate commit so that it is easy to revert if necessary.

@phadej If you do not like this change, please let me know.

hs-bindgen/src/HsBindgen/Hs/AST.hs
TravisCardwell: I agree.  I did not like having to use such auxiliary functions just to bring type variables in scope.  Appropriate use of instances and auxiliary functions significantly improves the code!

I added a `geqVec` auxiliary function, since `Vec n a` does not fit `GEq`.  I previously had an `eqVec` auxiliary function that returned `Bool`, but returning a `Maybe (n :~: m)` instead is key to avoiding the need to use auxiliary functions just to bring type variables in scope.

hs-bindgen/src/HsBindgen/Hs/AST.hs
TravisCardwell: Thank you very much for the assistance!

Since `GEq` (also) checks term equality, the `guard` is not needed, correct?

```haskell
Refl <- geq (ApEq sL) (ApEq sR)
guard $ sL == sR
```

I put the orphan instance in a new `HsBindgen.Orphans` internal module, just in case we need it elsewhere in the future.

edsko: The docs are quite explicit about this (https://hackage.haskell.org/package/some-1.0.6/docs/Data-GADT-Compare.html#t:GEq):

```
geq x y = Just Refl   ⇒  x == y = True        ∀ (x :: f a) (y :: f b)
```

(Indeed, this is what makes it different from `TestEquality`.)

hs-bindgen/src/HsBindgen/Hs/AST.hs
edsko: Since this is part of Oleg's code, I assume Oleg probably had a reason for including it. Let's leave it for now, we can consider this separately (it's orthogonal to this PR).

edsko (approved):
Reviewed over meet. Feel free to merge after making the changes discussed.

hs-bindgen/src/HsBindgen/Hs/AST.hs
```diff
@@ -64,15 +64,21 @@ module HsBindgen.Hs.AST (
 
 import HsBindgen.C.AST qualified as C
 import HsBindgen.C.Tc.Macro qualified as C
+import Data.GADT.Compare (GEq(geq), defaultEq)
 import Data.Type.Equality ((:~:)(Refl))
 import Data.Type.Nat as Nat
+import Data.Vec.Lazy qualified as Vec
+import Unsafe.Coerce (unsafeCoerce)
```
phadej, line 71: wat?

phadej:
I think we shouldn't need Eq instances at all. This is all ugly and tedious code, and also error prone one.

edsko:
@phadej I think we got started on this path due to the `Eq` requirement of [ediffGolden1](https://hackage.haskell.org/package/tree-diff-0.3.3/docs/src/Data.TreeDiff.Golden.html#ediffGolden1). Can that be avoided?

phadej:
@edsko, yes, https://hackage.haskell.org/package/tree-diff-0.3.4 does exactly that.

I feel ashamed that I didn't notice that (and disappointed that nobody else either) for the seven years package existed.

hs-bindgen/src/HsBindgen/Hs/AST.hs
TravisCardwell: Reverting

hs-bindgen/src/HsBindgen/Hs/AST.hs
TravisCardwell: `Vec` is imported for `withDict`, and `unsafeCoerce` is imported for the following:

```haskell
eqAdds :: Add n m pL -> Add n m pR -> pL :~: pR
eqAdds (UnsafeAdd !_) (UnsafeAdd !_) = unsafeCoerce Refl
```

These will be removed when I remove the `Eq` instances.

TravisCardwell:
Implementing the `Eq` instances has been educational for me, at least.  I will go ahead and merge this commit and remove the `Eq` instances in a separate PR for #351.

Thank you for the update to `tree-diff`!

PR #352, opened 2025-01-08
Update tree-diff, remove Eq CHeader instance

phadej:
I didn't remove rest of Eq instances, IMO they should be. But not having `Eq CHeader` is enough to illustrate that we don't need them.


edsko:
ping @TravisCardwell

PR #357, opened 2025-01-09
Add HasFlexibleArrayMember instance generation

phadej:
AFAICT, everything else can be built as part of `-runtime` library on top, i.e. `hs-bindgen` generation part for flams is done.

### DIFF
--- a/hs-bindgen/examples/flam.h
+++ b/hs-bindgen/examples/flam.h
@@ -12,3 +12,10 @@ struct foo {
 		int y;
 	} bar[];
 };
+
+// offset and sizeof not the same
+struct diff {
+	long first;
+	char second;
+	char flam[];
+};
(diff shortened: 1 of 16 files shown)


edsko:
@phadej We need a new class other than `Storable` no? And generate instances for that..? Or is that already done and I missed it?

edsko:
Oh, no, wait, I get it. The "core" (without the flam) is `Storable`. The extended thing (newtype wrapped) is not, but as you said, we can do that in the lib. Ok, good.

edsko:
@phadej Shall I rebase or do you prefer to do it?

phadej:
> We need a new class other than Storable no? 

I don't think so. The support code for marshalling the whole struct (with flam) can be done without any additional generated code. `HasFlexibleArrayMember` and `HasFlexibleArrayLength` instances is enough; but are out of scope of this PR, which is about generation of `HasFlexibleArrayMember`.

PR #360, opened 2025-01-09
Recurse into CXCursor_UnionDecl

phadej:


TravisCardwell (approved):
Looks good to me!

If you have a header that you would like to explore with `clang-ast-dump` but find gradually updating the code tedious, please feel free to let me know which code you would like to explore, and I would be happy to help.

Since this is a trivial change, I will go ahead and merge.

PR #362, opened 2025-01-10
Print exception and break in clang_visitChildren

phadej:
Exceptions doesn't seem to fly through C-frames, so we need to handle them better.

For now, I just print and break. This makes test-suite not fail with pure exceptions.

It took me whole morning to figure out why `throwIO` weren't catched.

### DIFF
--- a/hs-bindgen-libclang/src/HsBindgen/Clang/LowLevel/Core.hs
+++ b/hs-bindgen-libclang/src/HsBindgen/Clang/LowLevel/Core.hs
@@ -188,6 +188,7 @@ module HsBindgen.Clang.LowLevel.Core (
   , nullCursor
   ) where
 
+import Control.Exception
 import Control.Monad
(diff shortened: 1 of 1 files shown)


phadej:
But we need to think of proper solution, this cannot be even called a band-aid.

edsko:
I guess the better solution would involve the Haskell function catching the exception and more explicitly returning it to the C function somehow?

phadej:
For the record https://gist.github.com/phadej/059993b3d430515fd97a3ee6033510c4 is an example: the exception is not caught, if you run that program you get

```
problem: user error (blargh)
```

i.e. uncaught exception, even if there is `try`.

> I guess the better solution would involve the Haskell function catching the exception and more explicitly returning it to the C function somehow?

I think the way to solve is to catch exception, store them in a mutable cell, and then make `clang_visitChildren` throw it after C-call returns.

PR #363, opened 2025-01-10
Make clang_visitChildren rethrow the exception thrown by visitor

phadej:
As exception don't propagate through C-call stack frames (how could they! there are no Haskell exceptions in C!), we need to propagate them manually

### DIFF
--- a/hs-bindgen-libclang/hs-bindgen-libclang.cabal
+++ b/hs-bindgen-libclang/hs-bindgen-libclang.cabal
@@ -45,6 +45,7 @@ library
       DeriveAnyClass
       DerivingStrategies
       DerivingVia
+      LambdaCase
       MagicHash
       TypeFamilies
       UnboxedTuples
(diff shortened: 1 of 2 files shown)


phadej:
With this patch `bitfield-a` branch

```
% cabal run golden -- -p bitfield
Configuration is affected by the following files:
- cabal.project
- cabal.project.local
golden
  bitfields
    treediff: FAIL
      user error (bit-fields not supported yet)
      Use -p '/bitfield/&&/treediff/' to rerun this test only.
    hs:       FAIL
      user error (bit-fields not supported yet)
      Use -p '/bitfield/&&/hs/' to rerun this test only.
    th:       FAIL
      user error (bit-fields not supported yet)
      Use -p '/bitfield/&&/th/' to rerun this test only.
    pp:       FAIL
      user error (bit-fields not supported yet)
      Use -p '/bitfield/&&/pp/' to rerun this test only.
    rust:     OK (0.60s)
```

(previously the test-suite aborted with pure exception_)

PR #369, opened 2025-01-28
Fix function argument iteration

TravisCardwell:
Problematic code:

```haskell
nargs <- liftIO $ clang_getNumArgTypes ty
args' <- forM [0 .. fromIntegral nargs - 1] $ \i -> ...
```

Function `clang_getNumArgTypes` returns an `int` while `clang_getArgType` takes an `unsigned int`, hence the need for `fromIntegral`.  This does not work when the function has no arguments, however.  When `nargs` is zero, `fromIntegral nargs - 1` underflows to the maximum value of the unsigned type.  This results in many iterations, but it generally fails quickly when attempting to get and use a non-existent argument.

This commit fixes the issue by adding an auxiliary function that works with both signed and unsigned types.  It checks for the zero case before subtraction.

A function with no arguments is added to the `simple_func.h` example.

### DIFF
--- a/hs-bindgen-libclang/clang-ast-dump/Main.hs
+++ b/hs-bindgen-libclang/clang-ast-dump/Main.hs
@@ -124,6 +124,15 @@ foldDecls opts@Options{..} cursor = do
         traceU 1 "integer value"
           =<< liftIO (clang_getEnumConstantDeclValue cursor)
         pure False -- leaf
+      Right CXCursor_FunctionDecl -> do
+        numArgs <- liftIO $ clang_getNumArgTypes cursorType
+        traceU 1 "args" numArgs
+        forM_ [0 .. numArgs - 1] $ \i -> do
+          argType <- liftIO $ clang_getArgType cursorType (fromIntegral i)
+          traceO_ 2 i =<< liftIO (clang_getTypeSpelling argType)
+        resultType <- liftIO $ clang_getResultType cursorType
+        traceU 1 "result" =<< liftIO (clang_getTypeSpelling resultType)
+        pure False
       Right CXCursor_TypedefDecl -> do
         traceU 1 "typedef name" =<< liftIO (clang_getTypedefName cursorType)
         pure True -- results in repeated information unless typedef is decl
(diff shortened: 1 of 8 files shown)


edsko (approved):
Feel free to merge after implementing the changes we discussed.

hs-bindgen-libclang/clang-ast-dump/Main.hs
```diff
@@ -124,6 +124,16 @@ foldDecls opts@Options{..} cursor = do
         traceU 1 "integer value"
           =<< liftIO (clang_getEnumConstantDeclValue cursor)
         pure False -- leaf
+      Right CXCursor_FunctionDecl -> do
+        numArgs <- liftIO $ clang_getNumArgTypes cursorType
+        traceU 1 "args" numArgs
+        when (numArgs > 0) $
+          forM_ [0 .. fromIntegral numArgs - 1] $ \i -> do
+            argType <- liftIO $ clang_getArgType cursorType i
```
phadej, line 132: just move the `fromIntegral` here. So we get `[0..-1]` above.

hs-bindgen/src/HsBindgen/C/Fold/Type.hs
```diff
@@ -301,7 +301,7 @@ processTypeDecl' relPath path unit ty = case fromSimpleEnum $ cxtKind ty of
         res' <- processTypeDeclRec relPath path unit res
 
         nargs <- liftIO $ clang_getNumArgTypes ty
-        args' <- forM [0 .. fromIntegral nargs - 1] $ \i -> do
+        args' <- forM (argIndexes (fromIntegral nargs)) $ \i -> do
             arg <- liftIO $ clang_getArgType ty i
```
phadej, line 305:
```diff
-arg <- liftIO $ clang_getArgType ty i
+arg <- liftIO $ clang_getArgType ty (fromIntegral i)
```

TravisCardwell:
Thank you very much for the feedback!

I considered just moving the `fromIntegral` but decided against it.  It performs `N` conversions instead of 1, so somebody in the future may move it back into the list comprehension and cause a regression.  My fix was an attempt to make the reasoning self-documenting, while still only performing 1 conversion.

@edsko suggested instead implementing a general solution (`enumFromToPred`).

I do not have strong feelings about it, so I will go with the smallest solution.  If anybody (future me) causes a regression, the added test will catch it, after all.

PR #373, opened 2025-01-28
Check if file exists in `withTranslationUnit`

### LINKED ISSUE #370: Report helpful error message when opening non-existent files
Currently we're getting something like

```
hs-bindgen: CallFailed "CXTranslationUnit 0x0000000000000000" CallStack (from HasCallStack):
  collectBacktrace, called at src/HsBindgen/Clang/Internal/Results.hs:41:14 in hs-bindgen-libclang-0.1.0-inplace:HsBindgen.Clang.Internal.Results
  callFailed, called at src/HsBindgen/Clang/Internal/Results.hs:68:12 in hs-bindgen-libclang-0.1.0-inplace:HsBindgen.Clang.Internal.Results
  ensureOn, called at src/HsBindgen/Clang/Internal/Results.hs:54:10 in hs-bindgen-libclang-0.1.0-inplace:HsBindgen.Clang.Internal.Results
  ensure, called at src/HsBindgen/Clang/Internal/Results.hs:74:17 in hs-bindgen-libclang-0.1.0-inplace:HsBindgen.Clang.Internal.Results
  ensureNotNull, called at src/HsBindgen/Clang/LowLevel/Core.hs:577:49 in hs-bindgen-libclang-0.1.0-inplace:HsBindgen.Clang.LowLevel.Core
  clang_parseTranslationUnit, called at src/HsBindgen/C/Parser.hs:52:15 in hs-bindgen-0.1.0-inplace:HsBindgen.C.Parser
```

which isn't exactly informative.

edsko:
Closes #370


phadej (changes requested):
This is wrong. You shouldn't ever check whether file exists before opening, but rather handle opening failure.

If `libclang` fails, we should handle that. Or have a note why we couldn't,

edsko:
I thought about this before doing this, but didn't see a way to do this better. This is "wrong" because the file might exist when we check, and no longer exist when we call libclang. (Or are there other reasons?)  But checking _after_ we call libclang if the file exists, if and when we get a NULL pointer back, is wrong for the same reason (it might exist when we call it, no longer exist after, but we get the NULL pointer for a different reason). There is also `clang_parseTranslationUnit2`, but although it returns an error code, it's still not helpful in this case. 

So what would you prefer in this case?

phadej:
> is wrong for the same reason (it might exist when we call it, no longer exist after, but we get the NULL pointer for a different reason).

It's less wrong. Then we already failed and try to figure out why.

Your current solution is trying to avoid failure, but if something fails, then we don't know what happened. It might be file pulled out, or it might be something else.

phadej:
if `clang_withTranslationUnit` returns `NULL`, we should handle that gracefully. Such result is *expected* failure.

Looking at the implementation, `CXError_ASTReadError` is probably returned (also) when the file doesn't exist.

hs-bindgen/src/HsBindgen/C/Parser.hs
```diff
@@ -74,6 +79,35 @@ withTranslationUnit relPath tracer args fp k = do
...
+clang_parseTranslationUnit' idx src args options =
+    handle callFailed $ clang_parseTranslationUnit idx src args options
+  where
+    -- If a call failed, it /might/ be because the file does exist (@libclang@
+    -- doesn't give us much info). Since this is the most likely reason, we
+    -- check for this here, throw a "does not exist" exception if the file does
+    -- not exist, and rethrow the original exception otherwise. Technically
+    -- speaking this is not correct: the file might have existed when we called
+    -- libclang, got removed /after/, and the 'CallFailed' exception was
+    -- unrelated. For now we simply accept this shortcoming.
```
phadej, line 98: Add a sentence that there are other expected but probably rarer conditions, like permissions - file exists, but is unreadable etc.

I think it's ok to not be precice then, though i'm 99% sure the `CXError_ASTReadError` is returned for all those "IO" related issues.

edsko: What about trying to open the file ourselves, and then just closing again, and relying on that for checking for existence, permissions, etc?

edsko: Meh, this results in a confusing exception. I'll use the `2` API and check for some specific cases in postprocessing. Bear with me.

phadej (approved):

PR #378, opened 2025-01-29
More careful translation of `PrimVoid`

edsko:
This also fixes the treatment of arrays of unknown size as function arguments, as the `PrimVoid` placeholder we were using there was getting in the way of a correct translation of `void`.

### DIFF
--- a/hs-bindgen/src/HsBindgen/C/AST/Type.hs
+++ b/hs-bindgen/src/HsBindgen/C/AST/Type.hs
@@ -40,6 +40,22 @@ data Type =
   | TypePointer Type
   | TypeConstArray Natural Type
   | TypeFun [Type] Type
+
+    -- | Arrays of unknown size
+    --
+    -- Arrays normally have a known size, but not always:
+    --
+    -- * Arrays of unknown size are allowed as function arguments; such arrays
+    --   are interpreted as pointers.
+    -- * Arrays of unknown size may be declared for externs; this is considered
+    --   an incomplete type.
+    -- * Structs may contain an array of undefined size as their last field,
+    --   known as a "flexible array member" (FLAM).
+    --
+    -- We treat the FLAM case separately.
+    --
+    -- See <https://en.cppreference.com/w/c/language/array#Arrays_of_unknown_size>
+  | TypeIncompleteArray Type
   deriving stock (Show, Eq, Generic)
   deriving anyclass (PrettyVal)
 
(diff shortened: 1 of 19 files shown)


edsko:
Will take a look at the compilation failure for ghc > 9.2.8.

edsko:
Turns out it was due to some fixtures that needed updating.  I'm not sure why these tests didn't fail with 9.2.8, I guess they must have been skipped?

edsko:
.. and yet more fixtures to update. Why does `golden --accept` only update some fixtures with 9.2, then some more with 9.4, and then yet more with 9.8..? :thinking:

edsko:
Oh, I see, for 9.8 we get `IO Unit` instead of `IO ()`..?

edsko:
For posterity's sake: the TH output was happening only for `>=9.4`, hence why I didn't get the golden failures (using 9.2). It was failing in 9.8 because it generates `Unit` instead of `()` due to https://github.com/ghc-proposals/ghc-proposals/pull/475 . We're now testing the TH output _only_ for ghc 9.8.

hs-bindgen/src/HsBindgen/C/AST/Type.hs
```diff
@@ -38,7 +38,22 @@ data Type =
...
+    --   are interpreted as pointers.
+    -- * Arrays of unknown size may be declared for externs; this is considered
+    --   an incomplete type.
+    -- * Structs may contain an array of undefined size as their last field,
+    --   known as a "flexible array member" (FLAM).
+    --
+    -- We treat the FLAM case separately.
+    --
+    -- See <https://en.cppreference.com/w/c/language/array#Arrays_of_unknown_size>
+  | TypeArray (Maybe Natural) Type
```
phadej, line 56: FWIW, introduce incomplete ararys in `Type` is very similar to having `void`there.

incomplete arrays cannot appear everywhere, and it somewhat depends how we'll treat them depending where we encounter them.

For that reason I'd kept `TypeConstArray` and added `TypeIncompleteArray` as a separate constructor, but that's more or less a style issue. (An argument is that we'll always want to match on the `Maybe` argument of `TypeArray` - const and incomplete arrays aren't treated anywhere the same).

edsko: Ok, I'm fine with that, will change.

Issue #307, opened 2024-11-28
Generate (optional) instances for generated types


edsko:
We should automatically derive things like `Eq`, `Ord`, etc. for the types we generate (structs, enums, typedefs, ..). However, as part of #173 we discussed the importance of using Haskell `newtype` rather than `type` to translate C `typedef`s: we should be able to provide different instances. This implies it must also be possible to _prevent_ `hs-bindgen` from generating instances for specific types (note that rust bindgen also has features for preventing the derivation of certain classes, see #21).

edsko:
The set of instances we should derive automatically will probably depend on the type (`Num` for example makes sense only for numeric types).

edsko:
As of #384 we generate `Show` and `Eq` for structs, and `Show`, `Read`, `Eq`, `Ord` and `Enum` for enums.

edsko:
Closing this; we generate the instances, and binding specifications allow to omit them.

PR #383, opened 2025-01-30
Pave the way for generating additional instances

edsko:
This takes some initial steps towards #307. Summary:

* We used to have

```haskell
DeclInstance        :: InstanceDecl -> Decl
DeclNewtypeInstance :: TypeClass -> HsName NsTypeConstr -> Decl
```

that is now

```haskell
DeclDefineInstance  :: InstanceDecl -> Decl
DeclDeriveInstance  :: Strategy -> TypeClass -> HsName NsTypeConstr -> Decl
```

* Introduce `TranslationOptions`, currently just containing a single field, and thread it through everywhere.

* Derive stock `Show` for structs, mostly to show that the infrastructure is now in place. 

We need to extend this to other type classes, as well as enums and typedefs.


hs-bindgen/src/HsBindgen/Backend/TH.hs
```diff
@@ -42,6 +43,7 @@ mkGlobal =  \case
       Applicative_seq      -> '(<*>)
       Monad_return         -> 'return
       Monad_seq            -> '(>>)
+      Show_Show            -> ''Show
```
phadej, line 46: I'd name this `Show_class` (and rename `Storable_Storable` to `Storable_class`).

hs-bindgen/src/HsBindgen/Backend/PP/Translation.hs
```diff
@@ -78,12 +78,12 @@ newtype HsModuleOpts = HsModuleOpts {
     }
   deriving stock (Show)
 
-translate :: HsModuleOpts -> C.Header -> HsModule
-translate HsModuleOpts{..} header =
+translate :: TranslationOpts -> HsModuleOpts -> C.Header -> HsModule
```
phadej, line 81: I think this function is out of place. Something like that could exist in `Lib` module, but here it unnecessary imports stuff from previous steps. I'd rather not have it in this module. Let `Transalation` modules just do the translations. Larger pipelines to be assembled elsewhere.

edsko: I'm not sure I follow. Are you saying that you'd prefer to have a separate `HsModule -> HsModule` step that adds things like instances that we want to be derived?

phadej: So, the code in `Backend` modulespace is a mess.

Everywhere else `Translation` modules contain code to translate from previous representation to the next one.

In PP code, the `Translation` module has that, but also compound functions.

In TH code, the translation is in `TH` module and compound function in `TH.Translate`.

Again, it's a mess.

I think that compound functions should only be in `Lib` which naturally imports everything. And have `Translation` modules *just do the one step translations*.

You may cleanup this in this PR, or I'll cleanup it afterwards. The current state is not good.

edsko: Ok, I'm not entirely sure what you envision here, so feel free to submit a PR with a proposed new structure.

PR #384, opened 2025-01-31
More instances

edsko:


edsko:
Just realized this is wrong, enums are not bounded as they are not "proper" Haskell enumerations.

PR #386, opened 2025-01-31
Instances for typedefs around primtypes

edsko:
In order to derive instances for typedefs for types other than newtypes, we need to do a bit more work:

* For types that we know we (just) generated, we need to know what instances we derived for those types. This would require an environment of some form. Opened #387.
* For types that came from external bindings, the definition of those bindings should record which instances are available (cc @TravisCardwell ).  Opened #388.

For now we only support to generate instances for typedefs around primitive types, which is already helpful.


edsko:
Will rebase this on top of #385.

PR #389, opened 2025-02-01
Be more specific about the target platform

edsko:
See detailed discussion in the code for rationale.


edsko:
Oops, typo (MaxOS). Will fix.

PR #391, opened 2025-02-01
Introduce `GlobalCExpr`

edsko:
This introduces

```haskell
data Global =
     .. -- all the other names
  | CExpr Target GlobalCExpr

data GlobalCExpr =
    Not_class
  | .. -- etc
```

The reason is that names from `c-expr` must be imported from a platform specific module. By adding the target to the name itself, the translation from C to our internal Haskell representations can be target platform specific, at which point further processing (such as rendering of the Haskell code) is context free.

However, it seems we might need to further improve separation of concerns in the pipeline. In particular, it seems to make sense to change a function such as

```haskell
tyConGlobal :: C.TyCon args res -> SType ctx
```

to

```haskell
tyConGlobal :: Target -> C.TyCon args res -> SType ctx
```

However, it seems less reasonable to add a `Target` argument to a function such as 

```haskell
translatePredTy :: Hs.PredType ctx -> SType ctx
```

After all, the idea was that the translation from C to Haskell would embed target specific decisions; we should not need to make further such decisions in the translation from `Hs` to `SHs`. The core of the problem seems to be that we have definitions such as this in `Hs.AST`:

```haskell
data PredType ctx
  = DictTy AClass [TauType ctx]
  | ..

data AClass where
  AClass :: C.TyCon args C.Ct -> AClass
```

Note that this is a _Haskell_ type wrapping a _C_ type. Not entirely sure how to best address this; @sheaf let's discuss this next week.

### DIFF
--- a/hs-bindgen/src/HsBindgen/Backend/PP/Names.hs
+++ b/hs-bindgen/src/HsBindgen/Backend/PP/Names.hs
@@ -13,8 +13,9 @@ module HsBindgen.Backend.PP.Names (
 
 import Data.Char qualified as Char
 
-import HsBindgen.SHs.AST
+import HsBindgen.Clang.Args (Target(..))
 import HsBindgen.Hs.AST.Type
+import HsBindgen.SHs.AST
 
 {-------------------------------------------------------------------------------
   Imports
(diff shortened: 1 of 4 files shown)


phadej:
Another option is to always use `C.Expr.BuildPlatform`, also in PP code. Eventually it will be the right module for the target, when the target code is compiled.

edsko:
Obsoleted by #395.

PR #392, opened 2025-02-01
With + 1L this doesn't work

phadej:


hs-bindgen/src/HsBindgen/SHs/AST.hs
```diff
@@ -150,9 +150,9 @@ data SExpr ctx =
   deriving stock (Show)
 
 pattern EInt :: Int -> SExpr be
-pattern EInt i <- EIntegral (fromInteger -> i) (Just HsPrimCInt)
+pattern EInt i <- EIntegral (fromInteger -> i) (Just HsPrimInt)
   where
-    EInt i = EIntegral (fromIntegral i) (Just HsPrimCInt)
```
phadej, line 155: Do we have cases with `Nothing` in `EIntegral`? We do, the enum values for example. We *could* be explicit there as well if we really want to; thus being always explicit about types of integral literals.

edsko:
We'll need a similar change to the preprocessor backend. I'll see if I can submit a PR for that.

PR #395, opened 2025-02-02
Always use C.Expr.BuildPlatform

phadej:
From the perspective of the compiler, it's always compiling code for host. Thus we can use `C.Expr.BuildPlatform` (which is really a HostPlatform - compiled code shouldn't really be aware of build platform anwyay).

This is an alternative to #391. Arguably a lot simpiler one. There are a lot of further cleanups, which can be made in `c-expr`, (e.g. renaming the `C.Expr.BuildPlatform`, potentially getting rid of non-host architecture modules, simplifying library structure, making instances non-orphans etc) but these are not strictly necessary, so not done to not obscure this change.

### DIFF
--- a/c-expr/c-expr.cabal
+++ b/c-expr/c-expr.cabal
@@ -76,12 +76,12 @@ common common
         >= 0.5   && < 0.6
 
 -- Platform independent core library
+-- Note: the C.Operator.Classes should not be publicly visible.
+-- We should only be able to import the definitions "polluted"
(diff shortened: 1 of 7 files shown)


phadej:
AFAICT this will fix #394

edsko:
Ok, that does indeed simplify things. I think what you said is a useful intuition: ghc is not a cross-compiler.

PR #396, opened 2025-02-04
Generate type annotations also in PP mode

edsko:
This is a companion to #392.


sheaf:
> Hmm we should really also add annotations for `EFloat` and `EDouble`.

Yes, #392 was missing that as well.

edsko:
> Yes, #392 was missing that as well.

Yeah. Will fix both.

PR #398, opened 2025-02-04
Generate macro newtype instances

edsko:
We generate the same instances as for `typedef`s.


edsko:
We're not completely consistent here yet; see #399.

PR #406, opened 2025-02-05
Pp th names

phadej:


TravisCardwell (approved):
Looks good!

I spotted one issue: unqualified imports may collide with mangled names.  Since changing the name mangler to do minimal name mangling by default (#331), mangled names no longer have a `C` prefix.  Aside from operators, which we do not generate, all identifiers that are imported unqualified should be reserved ([here](https://github.com/well-typed/hs-bindgen/blob/07b182bfb981666a51e7bff1dc7535e123bcf288/hs-bindgen/src/HsBindgen/Hs/AST/Name.hs#L623-L639)).

Looking over the diff, here are the current problematic unqualified imports:

* `Bounded`
* `Enum`
* `Eq`
* `FiniteBits`
* `Floating`
* `Fractional`
* `IO`
* `Int`
* `Integral`
* `Num`
* `Ord`
* `Read`
* `RealFloat`
* `RealFrac`
* `Real`
* `Show`
* `Void`

One option is to minimize use of unqualified imports in generated code.  FWIW, I am a fan of this option.

We may want to prevent translations to some of these names anyway, for "[sanity](https://github.com/well-typed/hs-bindgen/blob/07b182bfb981666a51e7bff1dc7535e123bcf288/hs-bindgen/src/HsBindgen/Hs/AST/Name.hs#L641-L676)."

Notes related to this topic but unrelated to this PR:

* [`sanityReservedTypeNames`](https://github.com/well-typed/hs-bindgen/blob/07b182bfb981666a51e7bff1dc7535e123bcf288/hs-bindgen/src/HsBindgen/Hs/AST/Name.hs#L641-L676) only has `C`-prefixed types, but perhaps we should now add many more reserved type names? @edsko
* [`hsBindgenReservedTypeNames`](https://github.com/well-typed/hs-bindgen/blob/07b182bfb981666a51e7bff1dc7535e123bcf288/hs-bindgen/src/HsBindgen/Hs/AST/Name.hs#L623-L629) currently reserves `~`, but I do not think this is necessary because the name mangler should never generate `~` even with custom rules.

If this is something that you would like me to work on in a separate PR, please let me know.

phadej:
@TravisCardwell this PR doesn't make any changes (sans `~~) to generated code, so your comment doesn't seem to be relevant to this PR.

TravisCardwell:
Indeed.  The changes in this PR look good. :+1:

I moved the unqualified imports discussion to a new issue: #407

PR #408, opened 2025-02-07
Use absolute paths in library, relative paths in text fixtures

TravisCardwell:
I added source location information to the C AST in December.  One issue that came up was use of absolute paths in test fixtures, which we cannot do because tests running on different systems have different absolute paths ([`0de1f2e9`][]).  My initial goal was to just change absolute paths to relative paths in tests, but the opaque `CHeader` type was problematic.  I implemented a fix using a type class and demonstrated how it could be used in a minimal fix for the errors we were seeing ([`ca045aed`][]).  For consistency, I ended up reverting that commit ([`477379de`][]) and implementing the change in the source location constructors themselves (which involved propagating `relPath` through many parts of the source) ([`f9799326`][]).

We now need to compare source paths, to determine which paths to import from as well as in the implementation of external bindings.  `libclang` always uses absolute paths, and we should only compare absolute paths.  We therefore need to use absolute paths in the library and translate to relative paths only in the tests.

I reverted the previous implementation.  The opaque `CHeader` still makes it difficult to implement a solution in just tests.  I considered restoring those type classes and implementing a `CHeader` instance, but I remembered hearing that dirty hacks are not so frowned upon in test the code, so I implemented a small fix that uses `unsafePerformIO` to get the current working directory for the `ToExpr SingleLoc` instance.

Is this hack acceptable, or should we go ahead and add to the `CHeader` API?

[`0de1f2e9`]: <https://github.com/well-typed/hs-bindgen/commit/0de1f2e9e008451abe2e432ecda01c1b985ba3ca>
[`ca045aed`]: <https://github.com/well-typed/hs-bindgen/commit/ca045aedc106e2139df919f425208f4cbdc98265>
[`477379de`]: <https://github.com/well-typed/hs-bindgen/commit/477379dec817f8105900c15278039f8cc9e5dfcc>
[`f9799326`]: <https://github.com/well-typed/hs-bindgen/commit/f9799326afe6e2d7d281de4a7704bfa899614354>

### DIFF
--- a/c-expr/test/CallClang.hs
+++ b/c-expr/test/CallClang.hs
@@ -321,7 +321,7 @@ clangVisitChildren args f srcContents = do
 
   unit <- clangGetTranslationUnit args srcContents
 
-  diags  <- Clang.clang_getDiagnostics Nothing unit Nothing
+  diags  <- Clang.clang_getDiagnostics unit Nothing
   let (errors, _warnings) = partition diagnosticIsSevere diags
 
 {-
--- a/hs-bindgen-libclang/hs-bindgen-libclang.cabal
+++ b/hs-bindgen-libclang/hs-bindgen-libclang.cabal
@@ -87,7 +87,6 @@ library
     , hs-bindgen-runtime
   build-depends:
       -- External dependencies
-    , filepath      >= 1.4.2.2 && < 1.6
     , mtl           >= 2.2     && < 2.4
     , pretty-show   >= 1.10    && < 1.11
     , text          >= 1.2     && < 2.2
--- a/hs-bindgen-libclang/src/HsBindgen/Clang/HighLevel/Diagnostics.hs
+++ b/hs-bindgen-libclang/src/HsBindgen/Clang/HighLevel/Diagnostics.hs
@@ -97,57 +97,53 @@ diagnosticIsError diag =
 -------------------------------------------------------------------------------}
 
 clang_getDiagnostics ::
-     Maybe FilePath -- ^ Directory to make paths relative to
-  -> CXTranslationUnit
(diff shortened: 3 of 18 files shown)


edsko (approved):
Approving but we might want to reconsider slightly, see above.

phadej:
I'll be fine with just using base names in tests.

FWIW, GHC and Cabal both "normalise" outputs, see e.g. https://github.com/haskell/cabal/blob/master/cabal-testsuite/src/Test/Cabal/OutputNormalizer.hs or https://github.com/ghc/ghc/blob/f71c2835bfacac879b294bbcd475d7acfac4adfb/testsuite/driver/testlib.py#L1345 etc.

So if we want to be more precise, we could look for current working dir in generated outputs and replace it with `/`, without needing to use `unsafePerformIO`.

TravisCardwell:
> I don't mind *terribly*, but I do wonder if there isn't an easier solution. For example, if we know what kind of paths are we expecting (that is, the *suffix*, not the *prefix*), we could just strip everything else, without having to know exactly what that is?

I doubt there is an easier solution, but it is indeed possible to mangle the paths purely.  I just pushed a commit that checks absolute paths for the `musl-include` directory.  It is not correct in general, but it should work for our needs...  Shall we go with it, at least for now?

@phadej Thank you very much for the ideas!

edsko:
Well, @phadej 's suggestion of just taking the `basename` feels like it's much simpler and while perhaps overly aggressive in theory, in practice unlikely to change the tests in such a way that we'd miss bugs?

edsko:
Anyway I'm okay with this as-is. We can readjust again if we want to.

PR #409, opened 2025-02-07
Macros: support for character and string literals

sheaf:
This commit adds support for C character literals and string literals in macros.

Some remarks:

  - The implementation assumes that the source file is encoded in UTF-8. I believe that's the only encoding that the Clang library supports anyway.
  - The implementation doesn't support wide character types, such as `char16_t`.
  - C character literals are of type `int`, not `char`. We distinguish two types of such literals, as per the C standard:
      * code unit: a specific `int` value, whose interpretation as a linguistic or symbolic character depends on the choice of a text encoding,
      * Unicode code point: a unique numeric identifier for a specific character, whose value (e.g. in a `char *` array) depends on the text encoding chosen.
  - We translate character literals to `CInt`, and use unboxed `Addr#` literals to translate string literals to `CStringLen` (allowing for the possibility of inner null bytes).

### DIFF
--- a/hs-bindgen/src/HsBindgen/Hs/Translation.hs
+++ b/hs-bindgen/src/HsBindgen/Hs/Translation.hs
@@ -24,6 +24,9 @@ import Data.Type.Nat (SNatI, induction)
 import Data.Map.Strict qualified as Map
 import Data.Text qualified as T
 import Data.Vec.Lazy qualified as Vec
+import GHC.Exts qualified as IsList (IsList(..))
+
+import C.Char qualified as C
 
 import HsBindgen.Imports
 import HsBindgen.NameHint
@@ -556,8 +559,10 @@ macroExprHsExpr = goExpr where
       C.MEmpty -> Nothing
       C.MInt i -> goInt i
       C.MFloat f -> goFloat f
+      C.MChar c -> goChar c
+      C.MString s -> goString s
       C.MVar nm args ->
-        --  TODO: removed the macro arguement used as a function check.
+        --  TODO: removed the macro argument used as a function check.
         case Map.lookup nm env of
           Just i  -> return (Hs.VarDeclVar i)
           Nothing ->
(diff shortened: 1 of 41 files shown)


phadej:
Add examples (tests)

sheaf:
> Add examples (tests)

Yes, that's the next step. This is a draft.

hs-bindgen/src/HsBindgen/SHs/AST.hs
```diff
@@ -142,6 +145,7 @@ data SExpr ctx =
   | EIntegral Integer (Maybe HsPrimType)
   | EFloat Float HsPrimType -- ^ Type annotation to distinguish Float/CFLoat
   | EDouble Double HsPrimType
+  | EString [Word8]
```
phadej, line 148: Can't we use `ByteArray`, or some better representation. `[Word8]` doesn't feel good.

hs-bindgen/src/HsBindgen/Hs/Translation.hs
```diff
@@ -568,6 +571,47 @@ macroExprHsExpr = goExpr where
     goInt (C.IntegerLiteral { integerLiteralType = mbIntTy, integerLiteralValue = i }) =
       Just $ Hs.VarDeclIntegral i (maybe HsPrimCInt (uncurry integralType) mbIntTy)
 
+    goChar :: C.CharLiteral -> Maybe (Hs.VarDeclRHS ctx)
+    goChar (C.CharLiteral { charLiteralValue = c }) =
+      ( `Hs.VarDeclIntegral` HsPrimCInt ) <$>
+        case c of
+          C.CodeUnit u
```
phadej, line 578: extract into separate functions. Preferably don't invent UTF8 parsing validation, there are plenty in the libs.

hs-bindgen/src/HsBindgen/ByteArray.hs
```diff
@@ -0,0 +1,128 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE UnboxedTuples #-}
+
+module HsBindgen.ByteArray
+  ( ByteArray(..), MutableByteArray(..) )
+  where
+
+-- #if MIN_VERSION_base(4,17,0)
```
phadej, line 9: Depened on `data-byte-array` or `primitive`.

edsko (approved):
Reviewed over meet. Feel free to merge once the changes we discussed are done.

PR #416, opened 2025-02-17
addDependentFile in Lib.templateHaskell

phadej:
This makes GHC recompile files when header changes

EDIT: perfectly we'd `addDependentFile` all the transitive includes as well.


edsko:
Have opened #422 to track the transitive headers.

Issue #381, opened 2025-01-30
Shared library mapping files


TravisCardwell:
As mentioned in [this comment][], we plan on defining mappings from C identifiers (imported from specified headers) to Haskell identifiers (in the specified packages and modules) in text files.

[this comment]: <https://github.com/well-typed/hs-bindgen/issues/294#issuecomment-2621042608>

Specifically, a mapping specifies the following:

* C name - This should be the full C spelling, specifying `struct` or `union` when required.  Examples: `int8_t`, `struct tm`
* One or more C header files - The same C type may be imported from any number of header files.  A C type with the specified C name is matched if it is made available by a specified header file, even if it is defined in a different header file that is (transitively) included by the specified header file.  The header file should be specified as used in the C source code (relative to an include directory).
* Haskell package
* Haskell module
* Haskell identifier

@edsko suggested using [YAML][].  That sounds good to me.  There are many people who *hate* YAML, and perhaps we could *also* support [JSON][] if this becomes an issue.  The [`yaml`][] package uses the [`aeson`][] data model, so it is trivial to support both.  Why do people hate YAML?  [The yaml document from hell][] is a blog post that explains some of the complications of the format, a common reason for the YAML hate.

[YAML]: <https://yaml.org/spec/1.2.2/>
[JSON]: <https://www.json.org/>
[`yaml`]: <https://hackage.haskell.org/package/yaml>
[`aeson`]: <https://hackage.haskell.org/package/aeson>
[The yaml document from hell]: <https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell>

TravisCardwell:
Do we want to specify a filename convention for the mapping files?  I am leaning toward not doing so, just requiring a `.yaml` extension.  Examples:

* `base.yaml` - mappings for types defined in `base`
* `hs-bindgen-runtime-libc.yaml` - mappings for types defined in module `HsBindgen.Runtime.LibC` of package `hs-bindgen-runtime`

Thoughts/preferences?

TravisCardwell:
@edsko and I have a call later today, and I would like to decide on the schema for mapping files.  Checking for strong preferences is probably a good idea before I implement it.

BTW, I am happy to create a [JSON Schema][] if desired.

[JSON Schema]: <https://json-schema.org/specification>

[...]

Issue #366, opened 2025-01-17
Use `LibC`


TravisCardwell:
Discussing with @edsko, we need to start using the `LibC` module in the code that we generate.

When translating C code, we must be able to recognize when a C type is one of the standard types that we support.  Such types include the following:

* Standard types defined in `base` in `Foreign.**`
* `stdint` types that map to Haskell types defined in `base` in `Data.Int` and `Data.Word`
* Standard types that we define in `HsBindgen.Runtime.LibC`

A known C type is mapped to a Haskell package, module, and identifier.  Making such a mapping implementation general would be useful, since it could be used if/when we add support for translating multiple headers/modules.

Perhaps we can identify standard types using the name and source location information provided by `libclang`.  Note that the actual source header files for many types may differ in different C library implementation, since they have different internal implementations.  For example, Musl defines many types in the `bits/alltypes.h` header file that is included by the standard header files.

@edsko has a promising idea: perhaps we can parse our `standard_headers.h` bootstrap file and keep track of where the standard types are defined in that execution (which depends on which include directories are configured).  We could then use the results to identify standard types as we translate the user code.  With this idea, the implementation details should not matter, and we do not need to worry about the same types being imported from different header files.

If we run into trouble, perhaps we can build up a call-stack-style "include stack."  I think we can do this with the information provided by `libclang` while folding, but we could parse the tokens of include directives if necessary.

Should we add checks to confirm that a parsed standard type matches the Haskell definition/instances?  We could check that size and alignment are consistent.  We could check that structure/union field names and types are consistent.

TravisCardwell:
Types may be defined differently in different C library implementations.

For example, on my platform, using my default system (`glibc`) include path:

```c
typedef signed int __int32_t;  // /usr/include/bits/types.h
typedef __int32_t int32_t;     // /usr/include/bits/stdint-intn.h
```

Using Musl:

```c
typedef signed int int32_t;    // hs-bindgen/musl-include/x86_64/bits/alltypes.h
```

We need to map the C `int32_t` to Haskell `Data.Int.Int32` for all such definitions.

------------------------------------------------------------------------------

*Details:*

Test header:

```c
#include <stdint.h>

typedef int32_t MyInt32;
```

Using my default system include path:

```
$ cabal run hs-bindgen -- dev parse --input travis.h
```

```haskell
Header
  { headerDecls =
      [ DeclTypedef
          Typedef
            { typedefName = CName { getCName = __int32_t }
[...]

Issue #333, opened 2024-12-05
How to specify foreign import headers


phadej:
For `CApiFFI` we need to include a header for declrations, e.g.

```haskell
foreign import capi "header.h foo" foo :: Int -> IO ()
```

However, with file-processing interface we don't know what's the canonical import. Is it `some-header.h` or `clang-c/Index.h` (i.e. whether to drop directories or not).

How to solve this?

Note: rust-bindgen doesn't have such problem as it doesn't have `CApiFFI` like system. If we use ordinary ForeignFunctionInterface, then the linker will find symbol, so

- if we use CApiFFI, should user specify how to import the header?
- Or should we not use `capi` (but then we need to at least care about calling conventions)

TravisCardwell:
My understanding is that users should specify header files using paths that are relative to an include directory, in which case we generate `CApiFFI` directives using the passed include file path (including any directory names).

For the `libclang` example, `pkgconf` sets include directory `/usr/include` (on my system), so `clang-c/Index.h` successfully references `/usr/include/clang-c/Index.h`.

When manually configuring, users should configure both `include-dirs` and `install-includes` in the `.cabal` file, where include files may be relative to an include directory.  Users should then specify the header file to translate as it is specified in `install-includes`.

Am I missing or misunderstanding anything?

phadej:
I'm not fan of looking up extra information (system include paths, Cabal includes) if the opposite cost is just make user specify a bit more information.

I think it will clarify when `hs-bindgen` usage, **speficially as preprocessor in `Cabal`** is fleshed out. Until then doing the simplest thing feels best.

TravisCardwell:
A quick clarification: in my understanding described above, it is up to the user to specify include directories relative to include paths (as configured by `pkgconfig` or manually in their `.cabal` file).  In our implementation, we do not have to query system include paths for Cabal configuration; we simply use the relative include paths that users specify.

phadej:
@TravisCardwell even if understand you, my point stands:

>  think it will clarify when hs-bindgen usage, speficially as preprocessor in Cabal is fleshed out. 

For TH usage nothing you said really make much sense, users would need to specify *where to find the file*, but also *how include it from C* anyway (and the latter may have a default for common cases).

It might also make sense to always require some kind of wrapper header (so it's file local to package), even if it's literally just `#include <something>`. That would require https://github.com/well-typed/hs-bindgen/issues/294 in someway, `SelectFromMainFile` won't work then.

[...]

PR #417, opened 2025-02-18
Work torward using external bindings

TravisCardwell:
I have been working toward using external bindings (#366), and there has been a *lot* of friction.  My `use-libc` branch was started more than two weeks ago, and it has accumulated a number of commits.  After multiple attempts and stashes, I finally have an implementation of include paths that I think is palatable.  There are no doubt details to be cleaned up, but this change resolves a significant amount of friction.  I am creating this draft PR now so that I can run CI tests and check for any Windows issues.  It is a significant change that would also be good to merge when ready, to minimize conflicts, and I can continue work in a new branch.

Overview:

* More types are added to `LibC`.  BTW, I have posted my idea for how to handle optional headers in #293.
* Module `HsBindgen.Runtime.Marshal` is added.  This is the module that defines the `HasStaticSize`, `Peekable`, and `Pokable` type classes.  `LibC` uses these classes already.  Once merged, we should change our generated code to use the classes as well, where appropriate, after which we can update the generated tests.
* The `HsBindgen.Runtime.LibC.Arch` modules are removed for now, as we do not have any architecture-dependent definitions yet.
* External bindings configurations for `base` and `hs-bindgen-runtime` are added.  The source is in Dhall, and we read the YAML.  See #381 for details.
* Module `HsBindgen.ExtBindings` currently implements YAML and JSON parsers.
* We have planned on using an include graph (DAG) to help match `libclang` AST nodes with external bindings.  The performance is important.  This PR includes a minimal graph implementation for doing what we planned.  (It builds up the graph as well as the transpose of the graph, and the plan was to simplify further if we only need the transposed version: we could remove the transposed part of the implementation and change updates to add transposed edges.)  As of our last call, we are now concerned that the DAG is not a viable solution.  We may remove it and implement a different solution.  I will likely work on this on next.
* To track `#include` usage, function `clang_getIncludedFile` is added.
* Both system and non-system include paths are added to `ClangArgs`.  By always using `-nostdinc` and specifying include paths, we know the include paths being used and can resolve headers specified for external bindings.
* Users may specify include path directories as absolute paths or relative paths that are relative to the current working directory.  The CLI provides `--system-include-path` and `-I`/`--include-path` options, which may be specified more than once.  The TH API takes lists of `FilePath` as arguments.
* At this time, users must specify all needed include paths.  We can add support for getting paths from a `.cabal` file section (perhaps using `pkg-config`), getting paths from the `C_INCLUDE_PATH` environment variable, and getting default system paths from `clang` in the future.  (I will soon add issues for this.)
* "Input" header files must now be specified as relative paths, relative to an include path directory, as used in C `#include` inclusions.  Note that forward slashes must be used, even on Windows.
* With this change, appropriate header paths are used in LibFFI.  (#333)


TravisCardwell:
Here are commands that I use to test the CLI:

```
$ cabal run hs-bindgen -- \
    --system-include-path 'hs-bindgen/musl-include/x86_64' \
    --include-path 'hs-bindgen/examples' \
    preprocess \
      --input 'standard_types.h' \
      --module 'StandardTypes' \
      --output 'StandardTypes.hs'
```

```
$ cabal run hs-bindgen -- \
    --system-include-path 'hs-bindgen/musl-include/x86_64' \
    --include-path 'hs-bindgen/examples' \
    dev parse \
      --input 'standard_types.h'
```

```
$ cabal run hs-bindgen -- \
    --system-include-path 'hs-bindgen/musl-include/x86_64' \
    --include-path 'hs-bindgen/bootstrap' \
    dev prelude \
      --input 'standard_headers.h'
```

```
$ cabal run hs-bindgen -- \
    --system-include-path 'hs-bindgen/musl-include/x86_64' \
    --include-path 'hs-bindgen/examples' \
    gentests \
      --input 'simple_structs.h' \
      --module 'SimpleStructs'
```

Notes:

* `preprocess` works for our tests, but I noticed that some extensions are enabled in the `.cabal` file.  The implementation needs to be updated to detect need for such extensions and emit the appropriate pragmas.
* `gentests` has not been kept in sync with code generation.  The above command fails the same was as it does in `main`.

TravisCardwell:
Some tests failed in the `cabal check` step:

```
The following errors are likely to affect your build negatively:
Error: [misplaced-c-opt] Instead of 'ghc-options: -Iexamples' use
'include-dirs: examples'
```

This is for the `test-pp` test suite, where we use Literate Haskell to load and parse the preprocessor options.  The options to `hs-bindgen literate` are specified in an `.lhs` file, but we now need to specify the appropriate include path directory, which is an option to `hs-bindgen` and cannot be specified after the `literate` command.  I worked around this by specifying the option in the `.cabal` file, resulting in the above `cabal check` error.

I changed the option to `--include-path=examples` to work around this.

The `--` breaks the highlighting in my editor, as Vim syntax highlighting is pretty primitive.

TravisCardwell:
The `struct lconv` implementation in `LibC` does not work on Windows when including from `C:/tools/ghc-9.4.8/mingw/lib/clang/14.0.6/include`.  The implementation includes the fields that were added in C99, which are unconditionally included in Musl but are apparently not included in MinGW.  I removed the type for now.

When types differ across C standards, we may want to consider defining separate versions, as discussed in https://github.com/well-typed/hs-bindgen/issues/293#issuecomment-2632518200

TravisCardwell:
The `c-expr` tests use the Musl headers for POSIX but use system default headers for Windows.  Since the header search path must now be specified, Windows tests fail.

As a temporary fix, I hard-coded the include path for the CI environment.  With this change, all tests pass!  It of course does not work for anybody using Windows with a different path, though.

edsko:
There are a few minor issues to be sorted out, but merging this now because it is a large PR which will likely clash with other work.

PR #428, opened 2025-02-20
Make (standard) include headers work in test-th&test-pp

phadej:
For ordinary "real world" usage, we most likely don't want -nostdinc. -nostdinc is useful for cross-compilation or our tests, but in simple usage it complicates things.


TravisCardwell:
Currently, we are trying to have completely control over the include search path.  The idea is to *always* use `-nostdinc` and pass all include search path directories as options.  To include default search path directories, we need to determine them in `hs-bindgen` and pass them as options.  Not using `-nostdinc` means that `hs-bindgen` does not know of all the include search path directories and cannot resolve headers (for external bindings).

We are currently looking into using Clang to resolve headers, however.  If we can get that to work, then we can avoid the awkward include search path initialization.  #433

phadej:
> Currently, we are trying to have complete control over the include search path. 

Good goal, but short term this is a blocker for me to do any further work.

phadej:
Treat this as a stop gap measure if needed. In fact in #433 you do write

> Note that we discovered that Clang always processes the C_INCLUDE_PATH and CPATH environment variables. We can work around this by unsetting them after processing them ourselves, but this is not done yet.

SDKROOT is yet another environment variable you may want to add to the list.

PR #432, opened 2025-02-20
Set target in `c-expr` test on Windows

TravisCardwell:
@sheaf asked:

> does it work if you also add `Just (Clang.Target_Linux_X86_64, Clang.TargetEnvDefault)` to the Windows path in the c-expr test-suite?

This PR includes just that change.  A PR is required to test it in CI.

### DIFF
--- a/c-expr/test/Main.hs
+++ b/c-expr/test/Main.hs
@@ -113,7 +113,12 @@ main = do
           { Clang.clangCStandard = Just Clang.C23 }
       clangArgs =
         case platformOS hostPlatform of
-          Windows -> clangArgs0
+          Windows ->
+            clangArgs0
+              { Clang.clangTarget =
+                  Just (Clang.Target_Windows_X86_64, Clang.TargetEnvDefault)
+              , Clang.clangStdInc = True
+              }
           Posix ->
             clangArgs0
               { Clang.clangTarget =


TravisCardwell:
@sheaf Many binary operator tests fail due to type conversion/promotion differences.

Please feel free to use this branch.

sheaf:
Thanks for doing this. The failures make sense; `c-expr` contains the following logic:

```hs
intLikeTypeSizeInBits :: Platform -> IntLikeType -> Word
intLikeTypeSizeInBits plat i =
  case platformWordWidth plat of
    WordWidth32 ->
      case i of
        Short    {} -> 16
        Int      {} -> 32
        Long     {} -> 32
        LongLong {} -> 64
        PtrDiff     -> 32
    WordWidth64 ->
      case i of
        Short    {} -> 16
        Int      {} -> 32
        Long     {} ->
          case platformOS plat of
            Windows -> 32
            Posix   -> 64
        LongLong {} -> 64
        PtrDiff     -> 64
```

This is correct; if the target platform is Windows (regardless of whether one is using msvc or gcc/clang) then `long` is indeed supposed to be 32 bits.

So I think I asked the wrong question, I apologise. I should have asked whether it is possible to explicitly specify the target triple as well as the include dir and get the tests to pass.

TravisCardwell:
> I should have asked whether it is possible to explicitly specify the target triple as well as the include dir and get the tests to pass.

With this target set, we are targeting Linux even when the host platform is Windows.  Perhaps the test checks that we can target Linux from Windows despite the difference in `long` width?

The Windows include directory was specified.  I just moved it into the CI configuration since the absolute path was likely specific to the CI environment.

I just rebased this branch with the current `main` ([`4fd0da29`](https://github.com/well-typed/hs-bindgen/tree/4fd0da296fb30c2c8cc64f8d19811266a17ddefa)).  The code explicitly sets `Clang.clangStdInc = True`, and the temporary CI configuration is removed.

Note that after #454 is merged, standard include directories will be used by default.  This configuration will no longer be needed, so the POSIX case sets `Clang.clangStdInc = False` (and configures the Musl headers).

I just noticed the warning and see that the POSIX case should default to using the system include directories if the `hs-bindgen` package is not found.  Perhaps this should never happen, but I went ahead and changed the POSIX configuration to match.

```haskell
Clang.clangStdInc = isNothing mbHsBindgenDir
```

I made this change to #454 as well.

sheaf:
> > I should have asked whether it is possible to explicitly specify the target triple as well as the include dir and get the tests to pass.
> 
> With this target set, we are targeting Linux even when the host platform is Windows. Perhaps the test checks that we can target Linux from Windows despite the difference in `long` width?

Sorry, I meant in this case specifiying that the target is Windows as well, not Linux in this case. Just to have it explicit as opposed to it being inferred by Clang.

PR #442, opened 2025-02-23
Add throw_TODO and basic hs-bindgen exception machinery

phadej:
Example:

Given `foo.h`

```c
#define FOO 12.3Lh
```

we will now error with

```
% cabal run hs-bindgen -- -I. preprocess  -i foo.h
hs-bindgen known issue: https://github.com/well-typed/hs-bindgen/issues/349
tcMacro: long double not supported
  throw_TODO, called at src/HsBindgen/C/Tc/Macro.hs:1266:21 in hs-bindgen-0.1.0-inplace:HsBindgen.C.Tc.Macro
```

---

Another HsBindgenException would be `PanicException` for unexpected, should not happen cases, which we cannot have issue numbers for. I'll add that in follow up PR.

---

Regarding this particular case of long double not supported, arguably macro checker should handle that, and then we could fail later in the translation phase. At the moment we don't, there we have

```
floatingType :: C.PrimFloatType -> HsPrimType
floatingType = \case
  C.PrimFloat      -> HsPrimCFloat
  C.PrimDouble     -> HsPrimCDouble
  C.PrimLongDouble -> HsPrimCDouble -- wrong (see #247)
```

---

This is another step towards resolving #340.

### DIFF
--- a/hs-bindgen/src/HsBindgen/Errors.hs
+++ b/hs-bindgen/src/HsBindgen/Errors.hs
@@ -1,0 +1,53 @@
+module HsBindgen.Errors (
+    HsBindgenException (..),
+    TODOException (..),
+    throw_TODO,
+) where
+
+import GHC.Stack (CallStack, callStack, prettyCallStack)
+import Control.Exception (SomeException (..), Exception (..), throw)
+import Data.Typeable (cast)
+
+import HsBindgen.Imports
+
+-------------------------------------------------------------------------------
+-- HsBindgenException
+-------------------------------------------------------------------------------
+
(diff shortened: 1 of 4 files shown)


hs-bindgen/src/HsBindgen/Errors.hs
```diff
@@ -0,0 +1,53 @@
+module HsBindgen.Errors (
+    HsBindgenException (..),
+    TODOException (..),
+    throw_TODO,
```
phadej, line 4: it's `throw_TODO`, as there could be `throwIO_TODO`, which I haven't yet added as not needed.

Issue #418, opened 2025-02-18
Default C include search paths


TravisCardwell:
We would like to provide an option to use Clang's default C include search paths.

This was discussed in https://github.com/well-typed/hs-bindgen/issues/366#issuecomment-2655630980

TravisCardwell:
Unfortunately, I have not been able to find a good way for us to query the default include search paths.  The "best" way that I have found is to call `clang -E - -v </dev/null` and parse the header search path directories from the output.  I have a prototype (using `typed-process`) that works on my (POSIX) system, tested using LLVM versions 14, 15, and 18.  If there is not a better way, shall I test in CI to see how it fares on Windows?

TravisCardwell:
Context: We hope to resolve headers using Clang, after which we will not have to worry about controlling the include search path.  After pulling the latest changes, however, I cannot build.

Using current `main`, `test-th` fails to compile on my system.  It uses the default include search path, but is unable to find `stddef.h`.

```
test-th/Test01.hs:1:1: error: [GHC-87897]
    Exception when trying to run compile-time code:
      TranslationUnitCErrors ["/usr/include/stdlib.h:32:10: fatal error: 'stddef.h' file not found"]
    Code: (getPackageRoot
             >>=
               \ dir
                 -> templateHaskell Nothing [dir </> "examples"] "test-th-01.h")
```

Checking my Clang default include search path:

```
$ clang -E - -v </dev/null
...
#include "..." search starts here:
#include <...> search starts here:
 /usr/lib/clang/19/include
 /usr/local/include
 /usr/include
End of search list.
...
```

Header `stddef.h` is implementation/compiler-specific, so it should be in `/usr/lib/clang/19/include`.  It does.

```
$ test -f /usr/lib/clang/19/include/stddef.h && echo found
found
```

I do not know why we get this error.  Perhaps the `libclang` default include search path is different from Clang and does not include the compiler headers?

TravisCardwell:
It seems that LLVM 19 does not work, while LLVM 15 and LLVM 14 work fine when I explicitly set the include search path using `C_INCLUDE_PATH`.  Using the Musl headers also works.

The following are my test details.

----

I tried adding `/usr/lib/clang/19/include` using `C_INCLUDE_PATH`.

```
$ env C_INCLUDE_PATH="/usr/lib/clang/19/include" \
    cabal build all 2>&1 | tee build.log
```

That fails spectacularly.  Here is the first of *many* errors:

```
[...]

PR #450, opened 2025-02-27
Resolve headers using `libclang`

TravisCardwell:
**Paths**

This PR completely changes how we deal with header paths and C include paths.

Type `CHeaderIncludePath` represents a C header path as specified in an include directive.  It is unresolved, relative to a directory in a C include search path.  Forward slashes must be used, even on Windows.  It is a sum type with two constructors:

* `CHeaderSystemIncludePath` corresponds to `#include <PATH>` syntax
* `CHeaderQuoteIncludePath` corresponds to `#include "PATH"` syntax

I decided to go ahead and use the "quote" terminology to avoid confusion.

As for syntax, I went ahead with my proposal to use a `system:`.  This can mostly be ignored, but if somebody has a header file with the same name in both the system include path and the quote include path, they can use the prefix (like `system:stdio.h`) so that the system include path is prioritized.

Type `CIncludePathDir` represents a C include path directory.  I wrote documentation about the unfortunately overloaded "path" terminology.  This type is just a `newtype` wrapper around `FilePath`, as include path directories may be absolute or relative to the current working directory.  Include path directories are passed to `libclang` via `ClangArgs` as-is: absolute or relative.  This means that `libclang` source locations may be absolute or relative even after header resolution.

Since resolved headers may be absolute or relative, there is no special type for that.  We already have a type: `SourcePath`.  This type is moved to the `Paths` module so that it can be used throughout the codebase.  (It was defined under a `HighLevel` module, which limited where it could be used due to the import graph.)

**In-Memory Headers**

Support for parsing in-memory headers using `CXUnsavedFile` is added.  The `parseTranslationUnit` functions are changed to support this.  A high-level wrapper function is added (under `HighLevel`) to scope string allocation.

Note that the `c-expr` tests are updated to use this, so temporary files are no longer used.

**Resource Management**

We have not been freeing `libclang` resources, probably because we were only parsing a single translation unit.  This PR adds bindings for `clang_disposeTranslationUnit` and `clang_disposeIndex`.  High-level wrapper functions that use `bracket` to free resources is added.  The codebase is updated to do this, except for `clang-tutorial`.

The `c-expr` tests needed to be refactored a bit so that the `libclang` functions are called within scope.

**Header Resolution**

All of the above was done in order to implement header resolution.  Header resolution is done by parsing an in-memory header that just includes the header that needs to be resolved.  The fold breaks after the inclusion directive has been processed.

Note that the module organization is a bit unfortunate.  Module `HsBindgen.Clang.Paths` defines types that are used everywhere, including `Args`, `LowLevel`, and `HighLevel` modules.  Header resolution requires all of these, so it must be defined in a separate module: `HsBindgen.Clang.Paths.Resolve`.

Since `libclang` is used to resolve headers (using the exact same arguments that we use to parse headers), we can be confident that the resolution is consistent.  As I wrote about in https://github.com/well-typed/hs-bindgen/issues/418#issuecomment-2679723289, the Clang default include paths seem to differ from those of `libclang`.  Even if we could use the C++ API to resolve headers in a less hacky way, there would likely be inconsistencies!

When starting work on this, I was worried that performance might be an issue.  Thankfully, header resolution seems to be fast even using this strategy!


TravisCardwell:
Running `clang-ast-dump` with an example header:

```
$ cabal run clang-ast-dump -- \
    -I hs-bindgen/examples \
    simple_structs.h
```

Running `clang-ast-dump` with a system header, using the Musl include directory set via `C_INCLUDE_PATH` in my worktree environment to get around the current `stddef.h` issue (https://github.com/well-typed/hs-bindgen/issues/418#issuecomment-2679723289):

```
$ cabal run clang-ast-dump stdio.h
```

Inputs default to `CHeaderQuoteIncludePath`.  One can prioritize the system include search path by using a `system:` prefix:

```
$ cabal run clang-ast-dump system:stdio.h
```

Running `clang-ast-dump` with an explicit system C include path:

```
$ cabal run clang-ast-dump -- \
    --system-include-path /usr/include \
    system:stdio.h
```

Running `preprocess`:

```
$ cabal run hs-bindgen -- preprocess \
    -I hs-bindgen/examples \
    -i simple_func.h
```

Running `dev parse`:

```
$ cabal run hs-bindgen -- dev parse \
    -I hs-bindgen/examples \
    -i simple_structs.h
```

Running `dev prelude`:

```
$ cabal run hs-bindgen -- dev prelude \
    -I hs-bindgen/bootstrap \
    -i standard_headers.h
```

Running `gentests`:

```
$ cabal run hs-bindgen -- gentests \
    -I hs-bindgen/examples \
    -i simple_structs.h
```

Note that `clang-tutorial` still just takes a regular file path as input, as it does not deal with include path directories:

```
$ cabal run clang-tutorial hs-bindgen/examples/simple_structs.h
```

TravisCardwell:
Over lunch, I recalled that I did not remove the `temporary` dependency in `c-expr.cabal`, as there was no warning.  Investigating, I see that the union of dependencies of all sections of the Cabal file are declared in the `common` section.  Using `-Wunused-packages` therefore results in many warnings since each section only uses a subset of that union.

For now, I just removed the `temporary` dependency.  I left off `-Wunused-packages` since there are so many warnings.

*If* we would like to refactor `c-expr.cabal` to be consistent with the Cabal files of the other packages in the project, feel free to let me know, and I would be happy to do it.

edsko:
Note: `-I` adds to the quoted include path, not the system include path, so defaulting to quoted paths makes sense.

edsko:
Yes, let's fix this so that we can use `-Wno-unused-packages` in `c-expr`.

PR #451, opened 2025-02-27
Ban explicit `fail`

### LINKED ISSUE #340: Fail gracefully

phadej:
Resolves #340

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -65,11 +65,10 @@
 #
 # We could use -functions as above to forbid these, but the hlint message will be a bit confusing.
 #
-# TODO: https://github.com/well-typed/hs-bindgen/issues/340
 - error: {lhs: "error", rhs: "panicPure"}
 - error: {lhs: "undefined", rhs: "panicPure"}
-#- error: {lhs: "fail", rhs: "panicIO"}
-#- ignore: {name: "Use panicIO", within: "HsBindgen.C.Reparse.Literal" }
+- error: {lhs: "fail", rhs: "panicIO"}
+- ignore: {name: "Use panicIO", within: "HsBindgen.TH" } # fail is the only way to abort in Q
 
 # Define some custom infix operators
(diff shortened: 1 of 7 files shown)


phadej:
This is not 100% solution of #340, as there is`fail`s in `hs-bindgen-libclang` (in `resolveHeader`), which perfectly shouldn't report exception conditions via exceptions.

TravisCardwell:
> I'm not sure, maybe there should yet another library (`hs-bindgen-base`) with general utilities shared among all the packages. `Imports`, error handling, etc.

I have wondered about this before, when working on `hs-bindgen-runtime` or `hs-bindgen-libclang` and wishing I could use types defined in `hs-bindgen`.  The current external bindings design, where configuration is moved into YAML files, removes most of the tension.

Here is the current package dependency graph:

```mermaid
graph TD

  BG["hs-bindgen"]
  CE["c-expr"]
  LC["hs-bindgen-libclang"]
  RT["hs-bindgen-runtime"]
  TL["hs-bindgen-testlib"]

  LC --> RT

  BG --> CE
  BG --> LC
  BG --> RT
```

Note that `hs-bindgen-testlib` will depend on `hs-bindgen-runtime` once the tests are updated to support `Peekable`, `Pokable`, etc.

Perhaps common stuff could *could* be put into `hs-bindgen-runtime`, but using a separate package (`hs-bindgen-base`) would make it clear what is *runtime*.

PR #454, opened 2025-02-27
Use standard include directories by default

TravisCardwell:
`ClangArgs` `clangStdInc` now defaults to `True`.

For the Template Haskell API, we currently have two similar functions.  Function `templateHaskell` is defined in `Lib`, while `generateBindingsFor` is defined in `TH` and is unused.  IMHO `templateHaskell` is not a good name.  I propose that we use the following names:

* `genBindings` - full API
* `genBindings'` - easy API for simple cases, implemented using `genBindings`

I am not a fan of the `Lib` module, but I propose that we implement these functions in `Lib` and remove the `TH` module, given the current design.  Here is what I implemented:

```haskell
-- | Generate bindings for the given C header
--
-- TODO: add TranslationOpts argument
genBindings ::
     FilePath -- ^ Input header, as written in C @#include@
  -> ClangArgs
  -> TH.Q [TH.Dec]
```

```haskell
-- | Generate bindings for the given C header
--
-- This function uses default Clang arguments but allows you to add directories
-- to the include search path.  Use 'genBindings' when more configuration is
-- required.
genBindings' ::
     [FilePath] -- ^ Quote include search path directories
  -> FilePath   -- ^ Input header, as written in C @#include@
  -> TH.Q [TH.Dec]
genBindings' quoteIncPathDirs fp = genBindings fp args
  where ...
```

Note that the `ClangArgs` argument is purposefully last so that users can write calls like this:

```haskell
genBindings "acme.h" defaultClangArgs
  { ...
  }
```

With this design, the `test-th` call looks like this:

```haskell
$(getPackageRoot >>= \dir -> genBindings' [dir </> "examples"] "test-th-01.h")
```

Option `--no-stdinc` is added to the CLI.  Example usage:

```
$ cabal run hs-bindgen -- dev parse \
    --no-stdinc \
    -I hs-bindgen/examples \
    -i simple_structs.h
```


TravisCardwell:
I am happy to rebase after #448, #449, and #451 have been merged.

edsko:
Discussed this PR. Merging as-is, though we'll revisit some of these changes very soon. 

Just for claritiy's sake: `HsBindgen.Lib` is intended to be the main entry point for users of hs-bindgen-as-a-library (hence the name). It should not really define any logic; it should just re-export a carefully selected set of functions, and wrap some times so that we don't leak implementation details (such as our C AST).

PR #455, opened 2025-02-28
Add transitively included headers as dependencies (TH)

TravisCardwell:
This is easy because we have the include graph.  The only annoying part is that we need to pass the final state to the `Q` monad.  I implemented this by inlining and modifying `parseCHeader`.


TravisCardwell:
I am happy to rebase after #448, #449, and #451 have been merged.

PR #456, opened 2025-02-28
Refactor `c-expr` dependencies

TravisCardwell:
This refactors the `c-expr.cabal` file as requested in https://github.com/well-typed/hs-bindgen/pull/450#issuecomment-2687128121.

I moved dependencies from the `common` section to each `library` and `test-suite`.  As in the other packages in the project, I split them into "internal dependencies," "inherited dependencies," and "external dependencies." Note that I only used "inherited dependencies" for dependencies with versions specified in the same file.  Bounds for packages `filepath` and `text` are set by internal dependency `hs-bindgen-libclang`, so the bounds for these "external dependencies" are not necessary, but perhaps it would be dangerous to leave them off if the dependencies for `hs-bindgen-libclang` were to change.

I did not check/change any of the bounds.

I added the `-Wunused-packages` GHC option to alert of unused packages as the code changes.

I reformatted the sections to match the style of the other packages in the project.  I did not touch the top settings.

### DIFF
--- a/c-expr/c-expr.cabal
+++ b/c-expr/c-expr.cabal
@@ -37,107 +37,100 @@ description:
 
 common common
   ghc-options:
-    -Wall
-    -Wno-unticked-promoted-constructors
+      -Wall
+      -Wunused-packages
+      -Wno-unticked-promoted-constructors
   default-extensions:
-    DataKinds
-    DeriveGeneric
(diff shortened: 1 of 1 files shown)


edsko:
Yes, my understanding of "inherited" is also  strictly local to the cabal file.

PR #457, opened 2025-02-28
Simplify name mangler

edsko:
I was trying to understand the existing name mangling code, and there were three cases that didn't make sense to me: name mangling of module names, name mangling of type classes, and name mangling of type variables: none of these are related to `CName`s, even though their contexts want `CName`s as input. Indeed, it turns out these three functions are not used. This PR removes them. I'm not sure if that's the right thing to do, happy to close this PR unmerged again if we decide we do want these functions in some form; this is mostly to confirm that indeed these functions are currently unused and as a basis for discussion.

ping @TravisCardwell

### DIFF
--- a/hs-bindgen/src/HsBindgen/Hs/AST/Name.hs
+++ b/hs-bindgen/src/HsBindgen/Hs/AST/Name.hs
@@ -6,10 +6,7 @@ module HsBindgen.Hs.AST.Name (
   , SingNamespace(..)
   , HsName(..)
     -- * Contexts
-  , ModuleNameContext(..)
-  , TypeClassContext(..)
   , TypeConstrContext(..)
-  , TypeVarContext(..)
   , ConstrContext(..)
   , VarContext(..)
     -- * NameMangler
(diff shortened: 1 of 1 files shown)


TravisCardwell:
Sounds good to me. :+1:

Regarding type classes and type variables, I do not know of how we might need to generate them from C code.  I just wrote implementations for each namespace in case they are needed.  Removing them sounds good.

Regarding module names, perhaps we might want to generate module names based on the C header names?  For example, header `mupdf/pdf/document.h` could generate module `Mupdf.Pdf.Document` by default.  It would be really nice (in the future) if I could just point hs-bindgen at a *directory* of headers and have it create a module for each.  Perhaps that is thinking too far ahead, though.  At any rate, `CName` is not an appropriate input.

edsko:
Ok, great. Then I will merge this for now. For Haskell modules, we currently expect the user to specify this explicitly; I agree we might want to revisit that, but feels orthogonal to name mangling per se.

Issue #427, opened 2025-02-20
Support pointers to macro-defined types


edsko:
Example:

```c
#define uint64_t int

struct foo {
  uint64_t* a;
}
```

and similarly in function signatures (#399).

edsko:
It's not entirely clear to me actually why `reparseTypeUse` has a parser for primtypes in there. Redundant? 🤔

Issue #399, opened 2025-02-04
Macros used in function signatures not recognized as such


edsko:
Given

```c
#define YEAR  int
#define MONTH int
#define DAY   int

typedef struct date {
    YEAR  year;
    MONTH month;
    DAY   day;
} date;

YEAR getYear(date* d);
```

we generate

```haskell
data Date = Date
  { date_year :: YEAR
  , date_month :: MONTH
  , date_day :: DAY
  }

newtype YEAR = YEAR
  { unYEAR :: FC.CInt
  }

foreign import capi safe "no_external_headers.h getYear" getYear :: (F.Ptr Date) -> IO FC.CInt
```

Note the result type of `getYear`.

phadej:
My gut feeling is that this will be extremely hard to fix, without reparsing the function declarations.

edsko:
Indeed, we'd have to reparse signatures to fix this. I see three paths forward (see also https://github.com/well-typed/hs-bindgen/issues/401):

1. We give up parsing type macros. This means that something like `YEAR` would not be visible at all in the generated macros; I think this would be a pity.
2. We extend the (re)parser to cover all cases we're interested in. This is potentially difficult.
3. We generate type _aliases_ instead of newtypes for type macros; this  means that where we succeed to reparse, we can use the type as specified in the C header, but if we fail, the generated code will still be correct.

(3) does feel like a reasonable middle ground to me; we could also consider giving the user a choice between `type` and `newtype` (see also https://github.com/well-typed/hs-bindgen/issues/308).

edsko:
This does mean that the reparser _will_ need to be extended to cover simple function signatures at least (this ticket).

Issue #323, opened 2024-12-02
Macro reparse doesn't use eof


phadej:
Or so it looks like

```c
#define M3 int[3]
```

is accepted and treated as `#define M3 int`.

PR #458, opened 2025-02-28
Re-parse field declarations and function declarations

sheaf:
This PR reworks the reparser to (mostly) follow the C23 specification
for the grammar of declarations & declarators. This is then used
to reparse function declarations containing macros, and struct field
declarations containing macros.

Tickets:
  - #323 (use `eof` when reparsing)
  - #427 (reparsing pointers in field declarations)
  - #399 (reparsing in function declarations)


sheaf:
My next task will be to figure out which part of the function declaration will need to be reparsed. Do I need to reparse the whole thing, just the result type, just some arguments, etc?

hs-bindgen/src/HsBindgen/C/Fold/Type.hs
```diff
@@ -268,13 +276,15 @@ processTypeDecl' path unit ty = case fromSimpleEnum $ cxtKind ty of
 
     Right CXType_Pointer -> do
         pointee <- liftIO $ clang_getPointeeType ty
-        pointee' <- processTypeDeclRec (DeclPathPtr path) unit pointee
+        pointee' <- processTypeDeclRec (DeclPathPtr path) unit declCursor pointee
```
sheaf, line 279: Investigate whether this can be changed to `Nothing` instead of `declCursor`.

sheaf: Done.

PR #459, opened 2025-02-28
Minor cleanup of the name mangler code

edsko:
This avoids needing to pass unused arguments, and makes the definition of the name mangler a bit more readable.


edsko:
@TravisCardwell let me know if you disagree with this change :)

PR #464, opened 2025-03-02
Never use error in IO context

phadej:
I wish something like stan (which can do type-based linting) were actually usable. These mistakes would be so simple to prevent, in theory, but there are no tools `</rant>`


TravisCardwell (approved):
:+1:

The function should probably be renamed to `failWithContext`.

PR #465, opened 2025-03-02
Add `ResolveHeaderException`

TravisCardwell:
Function `resolveHeader` now returns an `Either` type instead of using `fail`.

Function `resolveHeader'` just throws the exception.  It just exists to decrease boilerplate.  There is less reason for this since it no longer formats an error message, though, so we can remove it if desired.


hs-bindgen-libclang/src/HsBindgen/Clang/Paths/Resolve.hs
```diff
@@ -21,17 +23,30 @@ import HsBindgen.Runtime.Enum.Simple
 
 --------------------------------------------------------------------------------
 
+-- | Failed to resolve a header
+data ResolveHeaderException =
+    ResolveHeaderNotFound CHeaderIncludePath
+  | ResolveHeaderParseTranslationUnitError (SimpleEnum CXErrorCode)
+  deriving stock (Show)
+  deriving anyclass (Exception)
```
phadej, line 31: This doesn't solve the problem, this exception is not `HsBindgenException`, so will be reported as unexpected one.

Also always define `displayException`.

phadej:
A solution is to move `HsBindgen/Clang/Paths/Resolve.hs` to `hs-bindgen` library proper. As far as I can tell, there's nothing libclang specific, no new FFI bindings. Only stuff building on top.

TravisCardwell:
Thank you very much!

I went ahead and moved `Resolve` to `hs-bindgen`.

I changed the API so that `resolveHeader'` is the internal version that returns `Either ResolveHeaderException`.  It is used in my WIP external bindings branch, where getting a `ResolveHeaderException` directly is more convenient than casting and dealing with impossible `Nothing` case.

Function `resolveHeader` throws an `HsBindgenException`, and it is re-exported from `Lib` for use in `Main`, tests, etc.

`clang-ast-dump` resolves headers, so it needed to be moved out of `hs-bindgen-libclang`.  Putting two executables in `hs-bindgen` makes `cabal run` annoying to use, so I ended up putting it in a separate package.

I implemented `displayException` and tested it:

```
$ cabal run hs-bindgen -- dev parse --input simple_structs.h
header not found: simple_structs.h
```

A common convention is to use an `error: ` prefix, but we should probably implement that in `exceptionHandler` if we decide to do that.

We *might* want to consider adding some more information to that specific (`--input` not found) error message, informing/reminding users that it is a C header path that is relative to a directory in the include path.  If we want to do this, perhaps we could implement a `newtype` wrapper with a different `displayException` implementation and use it in a `resolveInputHeader` function.

hs-bindgen-libclang/src/HsBindgen/Clang/Paths/Resolve.hs
TravisCardwell: Thank you!

TravisCardwell:
I am currently working on the errors for external bindings, and I realized that we might want to split `ResolveHeaderException`.

```haskell
data ResolveHeaderException =
    ResolveHeaderNotFound CHeaderIncludePath
  | ResolveHeaderParseTranslationUnitError (SimpleEnum CXErrorCode)
```

`ResolveHeaderNotFound` is user-facing and should be an `HsBindgenException`.  `ResolveHeaderParseTranslationUnitError` is probably not something that the user can fix, though, so perhaps a `PanicException` is better.

edsko:
Rebased.

PR #467, opened 2025-03-05
Use external bindings

TravisCardwell:
This PR adds support for external bindings to translation.  Currently, the following kinds of C types may use external bindings:

* `typedef`s
* `struct`s
* `enum`s

When we add support for `union`s, it will be easy to include support for external bindings.

A `--external-bindings` option is added to the CLI.  Users may specify any number of external bindings configuration files (YAML or JSON).

TH API function `genBindings` now has the following signature.  Note that the simplified API (`genBindings'`) is not changed; one has to use `genBindings` to use external bindings.

```haskell
genBindings ::
     FilePath -- ^ Input header, as written in C @#include@
  -> ExtBindings
  -> ClangArgs
  -> TH.Q [TH.Dec]
```

External bindings tests are added to `test-pp` and `test-th`.  This PR does not include any changes to the golden tests.


TravisCardwell:
Only the Windows tests passed! :laughing:

Tests fail because the external bindings configuration files cannot be found.  I thought this might be yet another relative path issue, but the logged paths look fine.  Perhaps it is building packages completely separately, so that the source for `hs-bindgen-runtime` is not available when building `hs-bingen`.

I wonder if linked files are copied or if the link is stored...  That did not work.

Another idea is to move the bindings configuration to `hs-bindgen`.  Since the Windows tests are passing, indicating that the POSIX directory separators in the Literate Haskell file are not an issue for the tests, I went ahead and moved them to a `hs-bindgen/bindings` subdirectory.

While running the tests, I realized that I need to add the (YAML) files as `extra-source-files` now that they are referenced by tests!

I needed to add `template-haskell` as a dependency.  `runIO` is used to run `loadExtBindings`.  We should extend the `TH` module to export things needed when using the TH backend, and we could provide a version of `loadExtBindings` that does not (externally) need `runIO`.  I am not doing this yet, though, because I do not know if we have decided what to do about code organization.

The external bindings configuration files are now found.

TravisCardwell:
Now macOS tests are failing because header `uchar.h` cannot be found.  It was added to the standard library in C11.  It is *not* optional, so it is interesting that it is not available on macOS.

According to the following, software that needs to support macOS should just avoid using the header.

<https://stackoverflow.com/questions/24608832/uchar-h-file-not-found-on-os-x-10-9>

I removed support for `uchar.h` from `LibC`, as our current policy is to only support standard headers that are available on all supported platforms.

edsko:
Very exciting, this is a big step forwards. Some cleanup to do, but we can do that in separate PRs. Some issues to follow.

PR #469, opened 2025-03-05
Don't allow empty names

edsko:
This translation is intended for use from C identifiers, which cannot be empty, so there is no point providing cases for this.


edsko:
I realize now that there is an edge case here: if the C name starts with an underscore, and is then followed only by other underscores or digits (rather than letters), we might end up with an empty name. Will address this in a separate PR (I think the right answer here is to insist on an override in this case; _if_ this happens, it may well be that there are multiple such names, `_1`, `_2`, etc. and so mapping to `X` is likely to go wrong.)

edsko:
Fixed this edge case in #482.

PR #472, opened 2025-03-05
Very minor changes to the name mangler

edsko:
@TravisCardwell Is this comment still accurate? I'm not sure where this is implemented:

```haskell
-- * Record fields are prefixed with the type name if the data type has a single
--   constructor or the constructor name otherwise, joined using @camelCase@,
--   dropping invalid characters.
```


TravisCardwell:
```
-- * Record fields are prefixed with the type name if the data type has a single
--   constructor or the constructor name otherwise, joined using @camelCase@,
--   dropping invalid characters.
```

That is an obsolete comment.  In a previous implementation, the `ConstrContext` included the necessary context, but we removed it because we do not anticipate generating types with more than one constructor.  I missed removing that obsoleted documentation. :bow:

Issue #16, opened 2024-08-01
Construct low-level `Storable` Haskell types for all `union`s


edsko:
We need to consider what we want to do with `union`s. It is tempting to try and construct a Haskell ADT for them, but of course C unions are (in general) untagged. Perhaps the best approach is to construct an opaque type for a C union, with getters and setters.

Since this is the low-level API, the goal here is not to do anything clever; it might be possible to have (limited) [support for tagged unions in the high-level API](https://github.com/well-typed/hs-bindgen/issues/18).

edsko:
#481 provides a concrete proposal for implementing unions. 

### Running example

For our running example, consider

```c
typedef struct Dim2 {
    int x;
    int y;
} Dim2_t;

typedef struct Dim3 {
    int x;
    int y;
    int z;
} Dim3_t;

typedef union DimPayload {
    Dim2_t dim2;
    Dim3_t dim3;
} DimPayload_t;

typedef struct Dim {
    int tag;
    DimPayload_t payload;
} Dim_t;
```

### Translation

1. A union is represented by a `ByteArray` of the (statically known) size of the union. 

```haskell
newtype DimPayload = DimPayload {
      getDimPayload :: ByteArray
    }

instance Storable DimPayload where
  sizeOf    _ = #size      DimPayload_t
  alignment _ = #alignment DimPayload_t

  peek = \ptr -> DimPayload <$> peekByteArray ptr (#size DimPayload_t)
  poke = \ptr -> pokeByteArray ptr . getDimPayload
```

(using `hsc2hs` here just for illustrative purposes). This relies on these auxiliary functions:

```haskell
peekByteArray :: Ptr a -> Int -> IO ByteArray
peekByteArray src n = do
    pinnedCopy <- newPinnedByteArray n
    withMutableByteArrayContents pinnedCopy $ \dest ->
      copyBytes dest (castPtr src) n
    freezeByteArray pinnedCopy 0 n

pokeByteArray :: Ptr a -> ByteArray -> IO ()
pokeByteArray dest bytes = do
    pinnedCopy <- thawPinned bytes
    withMutableByteArrayContents pinnedCopy $ \src ->
      copyBytes dest (castPtr src) n
  where
    n = sizeofByteArray bytes

thawPinned :: ByteArray -> IO (MutableByteArray RealWorld)
thawPinned src = do
    dest <- newPinnedByteArray n
    copyByteArray dest 0 src 0 n
    return dest
  where
[...]

PR #483, opened 2025-03-06
Add bytearray utilities (for unions)

phadej:
See #16

### DIFF
--- a/hs-bindgen-runtime/src/HsBindgen/Runtime/ByteArray.hs
+++ b/hs-bindgen-runtime/src/HsBindgen/Runtime/ByteArray.hs
@@ -1,0 +1,90 @@
+-- | Utilities for dealing with 'ByteArray' and 'Storable'
+--
+-- The additional copying we have to do here is a bit annoying, but in the end
+-- an FFI implementation based on 'Storable' is never going to be /extremely/
+-- fast, as we are effectively (de)serializing. A few additional @memcpy@
+-- operations are therefore not going to be a huge difference.
+--
+-- We /could/ choose to use pinned bytearrays. This would avoid /some/ copying,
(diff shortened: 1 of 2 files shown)


hs-bindgen-runtime/src/HsBindgen/Runtime/ByteArray.hs
```diff
@@ -0,0 +1,90 @@
...
+
+peekFromByteArray :: Storable a => ByteArray -> a
+peekFromByteArray bytes = unsafePerformIO $ do
+    pinnedCopy <- thawPinned bytes
+    withMutableByteArrayContents pinnedCopy $ \ptr ->
+      peek (castPtr ptr)
+
+pokeToByteArray :: Storable a => Int -> a -> ByteArray
+pokeToByteArray n x = unsafePerformIO $ do
+    pinnedCopy <- newPinnedByteArray n
```
phadej, line 75: is `n` different from `sizeOf (undefined :: a)`?

phadej: I guess it can, but it should be always larger.

edsko: I don't feel terribly strongly, feel free to change it if you prefer.

phadej: > The reason I didn't make this

The reason is it's usage in `setUnionPayload`, where `union`  (i.e. a) can be larger than `payload`. I think someone should write a comment, as even the author is confused...

edsko: Oh, hah, you're right. Yes. Comment would be useful :)

PR #486, opened 2025-03-07
Add userland-capi PoC

phadej:


edsko:
@phadej This looks good, thank you for the detailed write-up, much appreciated! Let's go ahead and use this throughout, rather than using CApiFFI in some places and this in others for more internal consistency, unless you can think of a reason why we would want to use CApiFFI where we can?

Can we set this up in such a way that it works in both TH and in hte preprocessor mode, where for the preprocessor we spit out a Haskell module and associated C file?

PR #488, opened 2025-03-07
Add hs-bindgen-testlib to cabal.project.ci

phadej:


phadej:
tests, fail, I think this is something for @TravisCardwell to look at?

TravisCardwell:
It seems to fail here:

```
test-hs-bindgen-testlib
  HsBindgen.TestLib.Preturb
    CInt
      Preturb0XSameSemanticsX
        random:              Running 1 test suites...
```

The `Preturb` instance for `CInt`:

```haskell
instance Preturb FC.CInt where
  preturb size n = n + fromIntegral size
```

`(==~)` for `CInt` is just `(==)`.

This test is `preturb 0 x ==~ x`: adding zero to an integer results in the same integer.  This is unlikely to fail, so I suspect that the issue is unrelated.

I re-ran the failed CI job, to see if the issue is deterministic.  It failed here:

```
test-hs-bindgen-testlib
  HsBindgen.TestLib.Preturb
    CUChar
      NotPreturb1XSameSemanticsX
        max:                 Running 1 test suites...
```

In the previous CI job, that `CUChar` unit test [passed](https://github.com/well-typed/hs-bindgen/actions/runs/13723089963/job/38382938001?pr=488#step:17:298) (before the `CInt` test failed).

Unfortunately, the Windows [development VMs](https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/) are not available, making Windows issues particularly difficult to debug.

TravisCardwell:
I now have a Windows environment to test with.  It is bare-metal, so there is no rollback: any side effects will persist. :frowning_face:

Attempts to build are failing with a `llvm-config` not found error.  Checking [`install-llvm-action`](https://github.com/KyleMayes/install-llvm-action), it looks like I am installing LLVM in the same way, and the [`Debug3` output](https://github.com/well-typed/hs-bindgen/actions/runs/13723089963/job/38455664996?pr=488#step:8:1) indicates that `llvm-config` is not in the CI environment either.

It looks like the CI environment is using GHC installed via [Chocolatey](https://chocolatey.org/), while my environment uses [GHCup](https://www.haskell.org/ghcup/).  Perhaps `llvm-config` is used because I am in an [MSYS2](https://www.msys2.org/) shell, and it is not used in CI because the software is built and tested in a different shell?

I see that MSYS2 has [`mingw-w64-llvm-16`](https://packages.msys2.org/base/mingw-w64-llvm-16), available, but we need to use version 14 on Windows if I remember correctly.  I tried it out, and it provides `llvm-config`.  After successfully building the `hs-bindgen` library, it fails when linking the `golden` test.

```
ghc-9.4.8.exe:  | C:\Users\n00b\wt\hs-bindgen\dist-newstyle\build\x86_64-windows\ghc-9.4.8\hs-bindgen-libclang-0.1.0\build\libHShs-bindgen-libclang-0.1.0-inplace.a: unknown symbol `__imp_clang_getCursorKindSpelling'
ghc-9.4.8.exe:  | C:\Users\n00b\wt\hs-bindgen\dist-newstyle\build\x86_64-windows\ghc-9.4.8\hs-bindgen-libclang-0.1.0\build\libHShs-bindgen-libclang-0.1.0-inplace.a: unknown symbol `ghczuwrapperZC71ZChszmbindgenzmlibclangzm0zi1zi0zminplaceZCHsBindgenziClangziLowLevelziFFIZCwrapzugetCursorType'
ghc-9.4.8.exe:  | C:\Users\n00b\wt\hs-bindgen\dist-newstyle\build\x86_64-windows\ghc-9.4.8\hs-bindgen-0.1.0\build\libHShs-bindgen-0.1.0-inplace.a: unknown symbol `hszmbindgenzmlibclangzm0zi1zi0zminplace_HsBindgenziClangziLowLevelziFFI_wrapzugetFileName_info'
ghc-9.4.8.exe:  | C:\Users\n00b\wt\hs-bindgen\dist-newstyle\build\x86_64-windows\ghc-9.4.8\hs-bindgen-0.1.0\build\libHShs-bindgen-0.1.0-inplace.a: unknown symbol `hszmbindgenzm0zi1zi0zminplace_HsBindgenziCziPredicate_SelectFromMainFile_closure'
ghc-9.4.8.exe: ^^ Could not load 'hszmbindgenzm0zi1zi0zminplace_HsBindgenziTH_genBindingszq_closure', dependency unresolved. See top entry above.
```

</details>

I removed the GHCup installation and installed GHC using Chocolatey.  `hs-bindgen-libclang` fails to build, with the following error:

```
Error: cabal.exe: The package has a './configure' script. If you are on
Windows, This requires a Unix compatibility toolchain such as MinGW+MSYS or
Cygwin. If you are not on Windows, ensure that an 'sh' command is discoverable
in your path.
```

The CI runner includes [a lot of software](https://github.com/actions/runner-images/blob/win22/20250303.1/images/windows/Windows2022-Readme.md)...  I am not sure how to prepare a Windows environment for `hs-bindgen`.

phadej:
You should stayed with `ghcup`. Look closely at CI script, it sets

```
      - name: Debug
        shell: bash
        run: echo "$LLVM_PATH"
```

which is used if `llvm-config` is not found:

```
AC_PATH_PROG([LLVM_CONFIG],[llvm-config],[])
if test -z "$LLVM_CONFIG"; then
  if test -z "$LLVM_PATH"; then
    AC_MSG_ERROR([could not find llvm-config])
  else
    AC_MSG_NOTICE([could not find llvm-config; using LLVM_PATH=$LLVM_PATH])
  fi
fi
```

TravisCardwell:
The issue was the `preturbDouble` implementation, which assumed that `long` is at least as wide as `double`, which is not the case on Windows.  We need to use `long long`.  I went ahead and changed the property test preturbation size to type `CLLong`, matching the intention of the original implementation.

The test log output is misleading because tests are run in parallel and logs are interleaved.  This might we worth fixing, but at least I now know this and can more quickly diagnose future issues.

I was able to get my Windows environment to work by using the Clang that is in the Mingw environment for the GHC installation installed by GHCup.  I had to specify the C include path.  Here are my settings:

```bash
export PATH="/c/ghcup/ghc/9.4.8/mingw/bin:${PATH}"
export C_INCLUDE_PATH='C:\ghcup\ghc\9.4.8\mingw\x86_64-w64-mingw32\include;C:\ghcup\ghc\9.4.8\mingw\lib\clang\14.0.6\include;C:\ghcup\ghc\9.4.8\mingw\include'
```

Note that `LLVM_PATH` is *not* set, as this installation includes `llvm-config`.

----

**Test Log**

I removed the packages installed by Chocolatey and re-installed via GHCup.

I then had trouble with paths.  Perhaps the space in `C:\Program Files` was causing issues, so I re-installed LLVM in `C:\LLVM`.  I was also unsure about when to use Windows-style paths and when to use MSYS2-style paths.  After a bit of trial and error, I *think* that only `PATH` should use MSYS2-style paths, and all other environment variables should use Windows-style paths.

Standard headers were not found.  According to `clang -E - -v`, the default include search path is just `C:\LLVM\lib\clang\14.0.6\include`, which does not include headers for the full standard library.

Investigating, I found that GHCup installs [Mingw](https://www.mingw-w64.org/) with Clang.  Perhaps we need to use Clang 14.0.6 to match that version...  Setting `C_INCLUDE_PATH` allowed further progress, but then I got translation unit C errors for standard headers.

```
error: expected ';' after top level declarator
```

I wonder if we could use the Clang that GHC uses, instead of installing LLVM separately.  I noticed that it even includes `llvm-config`, and the default include search path looks promising, except that one directory path uses backslashes while the others use forward slashes.  It failed to find the standard headers (even though the standard headers are in the directory path specified using backslashes).  Perhaps this is a case of the `clang` executable having more directories in the default include search path than what we get when using `libclang`.  (I really wish we could get the default include search path from `libclang` for debugging purposes.)

I got the following error:

```
test-pp\test-pp.hs:1:1: error: parse error on input ‘../’
  |
1 | ../test-th/test-th.hs
  | ^^^
```

This file is a symbolic link, and it was saved as a file with the relative path.  I got around this issue (locally) by copying the file.

It built successfully!  The `hs-bindgen-testlib` test failed:

```
...
    CBool
      HsGenSeqNSameSemanticsCGenSeqN
        zero:                OK
        max:                 OK

divide by zero

 Attempting to reconstruct a stack trace...

   Frame        Code address
 * 0xf7bfdd270  0x7ffecfc42b96 C:\WINDOWS\System32\ucrtbase.dll+0x22b96
 * 0xf7bfdd2c0  0x7ff7137464c2 C:\Users\n00b\wt\hs-bindgen\dist-newstyle\build\x86_64-windows\ghc-9.4.8\hs-bindgen-testlib-0.1.0\t\test-hs-bindgen-testlib\build\test-hs-bindgen-testlib\test-hs-bindgen-testlib.exe+0xf64c2
 * 0xf7bfdd2f0  0x7ff71369c9a9 C:\Users\n00b\wt\hs-bindgen\dist-newstyle\build\x86_64-windows\ghc-9.4.8\hs-bindgen-testlib-0.1.0\t\test-hs-bindgen-testlib\build\test-hs-bindgen-testlib\test-hs-bindgen-testlib.exe+0x4c9a9
 * 0xf7bfdd2f8  0x7ff713684949 C:\Users\n00b\wt\hs-bindgen\dist-newstyle\build\x86_64-windows\ghc-9.4.8\hs-bindgen-testlib-0.1.0\t\test-hs-bindgen-testlib\build\test-hs-bindgen-testlib\test-hs-bindgen-testlib.exe+0x34949
 * 0xf7bfdd300  0x7ef4519f5051

Test suite test-hs-bindgen-testlib: FAIL
```

Running multiple times, `hs-bindgen-testlib` tests always fail, but different tests fail each time.

Output from multiple tests is interleaved, so I set `TASTY_NUM_THREADS=1`.  There are still test failures with this change, and now the `divide by zero` error message is easily visible.  I found that it was in the CI logs ([here](https://github.com/well-typed/hs-bindgen/actions/runs/13723089963/job/38455664996?pr=488#step:17:115)) but I had missed it because of the interleaved output.

Even though I am now testing with a single thread, I think that it is likely that the divide by zero is unrelated to the tests.  Perhaps the divide by zero is happening separately while the tests are passing.

Using `--test-option '--pattern=...'` to run a small subset of tests, the tests always pass.  I wrote a script to repeatedly test a subset that I have seen listed last before the error, until failure, and it indeed never failed through many repetitions.

I wrote a separate test with the same dependencies that just delays while counting.  It ran without failing.  I then moved the delay code into the `GenSeq.Test` module and updated `tests` to delay between each test group.  All tests passed every try, even with the delay reduced to 0 seconds!  I moved the delay to an unused function that was also exported, and it still passed all tests.  I restored the source, and it still passed all tests.  Perhaps `GenSeq` is fine...

I moved the delay code to the `Preturb.Test` module and updated `tests` to delay between each test group.  All tests passed *until* `CDouble`!  I bet it is caused by the issue that I worked on yesterday in #494!  I cherry-picked the commits from the `fix-preturb-double` branch, and sure enough: all tests pass!

TravisCardwell (approved):
This should resolve the issue.

phadej:
Symbolic links and windows: these work if you enable them, and you should; this is one guide, hopefully it works https://portal.perforce.com/s/article/3472

TravisCardwell:
Thank you!

I just learned that "Home" editions do not officially support this.  While it is possible to install "Group Policy Editor" to enable "Local Security Policy" and get the functionality, this is not something that I would like to do on the system that I am using.  I will stick to copying for now, as use of the system is a temporary solution anyway.  I will need to find another solution soon, likely a VM where such hacks will be fine if not unnecessary.

This makes me wonder, however, if we might want to avoid use of symbolic links so that users do not run into the same issue.  Perhaps anybody who does Haskell development on Windows will already have this enabled, though?

PR #492, opened 2025-03-08
Add union example; doesn't work, but doesn't fail either

phadej:
For now pretend that all unions are just `float` :P


edsko:
A reasonable choice really :)

Issue #480, opened 2025-03-05
Generate external bindings during translation


TravisCardwell:
We should support generation of external bindings configuration for a module during translation.  This is only relevant to the preprocessor mode.  We can add an option to the CLI `preprocessor` command to enable this as well as specify the output file path.

TravisCardwell:
External bindings configuration should be generated from the `Hs` declarations.  We currently do not expose this stage of the pipeline, however.  We can generate them from the `C.Header` and (re-) translate internally, as is currently done with test generation.  Alternatively, we could add a new *opaque* `HsDecls` wrapper and expose the stage.

The current pipeline can be seen in https://github.com/well-typed/hs-bindgen/pull/498#issuecomment-2716185572.

If we add the `HsDecls` wrapper, we will have the following:

```mermaid
graph TD

  Clang-- parseCHeader -->C
  C-- genHsDecls -->Hs
  Hs-- genSHsDecls --> SHs

  subgraph Preprocessor
    PPIO["`source
    (IO)`"]
    PPPure["`source
    (pure)`"]
    HsModule-- genPP -->PPIO
    HsModule-- genPPString -->PPPure
  end

  SHs-- genModule -->HsModule

  subgraph Template Haskell
    DecsQ
    Exts[Set TH.Extension]
  end

  SHs-- genTH --> DecsQ
  SHs-- genExtensions --> Exts

  Dump["`dump
  (IO)`"]
  C-- dumpCHeader -->Dump

  Hs-- genExtBindings -->ExtBindings

  Tests["`tests
  (IO)`"]
  Hs-- genTests -->Tests
```

Any opinions?  My vote is to go ahead and add `HsDecls`.

TravisCardwell:
Our external bindings design includes the Haskell package name.  It is not actually used anywhere, though we may use it in the future, and it is probably useful information for users.  We do not yet have this information (when using the `preprocessor` command via the CLI), however.

The current plan is to add a new `--gen-ext-bindings` option that takes the path to write the external bindings configuration to.  How should we specify the package name?  We could use a separate `--package` option.  It would not be used unless `--gen-ext-bindings` is specified, in which case it would be required.  This is a bit annoying, but it would work.  Are there any better ideas?

Example usage:

```
$ hs-bindgen preprocess \
    --external-bindings .../path/to/base.yaml \
    --external-bindings .../path/to/hs-bindgen-runtime.yaml \
    --include-path cbits \
    --input foo.h \
    --package acme \
[...]

PR #498, opened 2025-03-12
Add `Pipeline` module, update public API

TravisCardwell:
This PR is a first step toward improving our public API.

A new `HsBindgen.Pipeline` module is added.  This module serves two purposes:

* It defines high-level options types and defaults.  Type `Opts` contains all options that apply to both the preprocessor and TH APIs.  Type `PPOpts` contains additional options for the preprocessor API.  Note that we do not have any options specific to the TH API.
* It defines pipeline components (such as `genHsDecls`, `genTH`, etc.) and high-level API functions that are implemented using these pipeline components.  The same components are used for both the preprocessor and TH APIs, and these functions are implemented in the same module, which should help ensure that the behavior remains the same.  All of these functions are implemented using the high-level options types.

Note that `Opts` includes the name mangler as well as tracers.  With this PR, users are now able to use a different name mangler.  (Previously, the default name mangler was hard-coded in multiple places.)

Module `HsBindgen.Lib` provides an API for users who want to use `hs-bindgen` as a library.  This module should expose everything needed for users to implement preprocessing.  It should *not* export things just for our use in tests.  The module re-exports types and functions that are needed, including options types.  An exception is the `NameMangler` module, which exports a DSL; that module is exposed, and users should import from it directly when creating a custom name mangler.

Module `HsBindgen.Lib` defines opaque type `CHeader`, used because `C.Header` is not part of the public API.  It is needed because users may parse a header and then use the result to generate Haskell code (preprocessor) as well as generate tests for that code.  Functions are defined that wrap and unwrap this type, accordingly.  Note that this module should probably not define anything else; all "logic" should be implemented in non-exposed modules.

We have not decided if we should include development/debugging functionality in `HsBindgen.Lib`.  This PR removes the `dev bootstrap` command.  The function used by the `dev parse` command is still in the module, though, renamed to `dumpCHeader`.

The public API is simplified.  One notable change is that `withTranslationUnit` is no longer exposed.  Function `getTargetTriple` is re-implemented to use `CXUnsavedFile`, as the target triple depends on the Clang arguments but not the source header.

Module `HsBindgen.TH` provides the Template Haskell API.  Like `HsBindgen.Lib`, it re-exports types and functions that are needed, including options types.  A goal if this module is to reduce the amount of code that users have to write.  Note that we may want to add a function that uses `base` (and `hs-bingen-runtime`?) external bindings automatically, which would significantly reduce the code needed for that common case.  This is not done yet.

One way to reduce the amount of user code is to provide `IO` functions in the `Q` monad.  For example, the module exports a version of `loadExtBindings` in the `Q` monad.  In some cases, this allows users to avoid adding `template-haskell` as a direct dependency of their package.  The module also re-exports `getPackageRoot` from `template-haskell` or `th-compat`, allowing users to avoid the conditional dependency and `CPP` usage in their projects.

One Travis-ism that I would like to point out in case others find it unpalatable is that I strictly use qualified imports for everything (except `Prelude`) in the `HsBindgen.Lib` and `HsBindgen.TH` modules.  I like to do this for such re-exporting modules because it makes everything very clear (IMHO), but I can change it of other find it distasteful.

We need more types and functions to implement tests, as evidenced by the "exposed for the sake of tests" comment in the `.cabal` file.  This PR does not change this, and even more modules are exposed.  We should address this in our next step toward improving our public API; I will create an issue.

### DIFF
--- a/clang-ast-dump/Main.hs
+++ b/clang-ast-dump/Main.hs
@@ -436,11 +436,13 @@ main = clangAstDump . uncurry applyAll =<< OA.execParser pinfo
       ]
 
     fileArgument :: OA.Parser CHeaderIncludePath
-    fileArgument = OA.argument (OA.eitherReader parseCHeaderIncludePath) $
-      mconcat
-        [ OA.metavar "FILE"
-        , OA.help "C (header) file to parse"
-        ]
+    fileArgument =
+      OA.argument
+        (OA.eitherReader $ first displayException . parseCHeaderIncludePath)
+        $ mconcat [
+              OA.metavar "FILE"
+            , OA.help "C (header) file to parse"
+            ]
 
     mkFlag :: String -> String -> OA.Parser Bool
     mkFlag flag doc = OA.switch $ OA.long flag <> OA.help doc
--- a/hs-bindgen-libclang/src/HsBindgen/Clang/Paths.hs
+++ b/hs-bindgen-libclang/src/HsBindgen/Clang/Paths.hs
@@ -7,6 +7,7 @@ module HsBindgen.Clang.Paths (
     -- * C header paths
   , CHeaderIncludePath(..)
   , getCHeaderIncludePath
+  , ParseCHeaderIncludePathException(..)
   , parseCHeaderIncludePath
   , renderCHeaderIncludePath
 
@@ -14,6 +15,7 @@ module HsBindgen.Clang.Paths (
   , CIncludePathDir(..)
   ) where
 
+import Control.Exception (Exception(displayException))
 import Data.Text (Text)
 import Data.Text qualified as Text
 import Data.List qualified as List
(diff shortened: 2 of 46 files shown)


TravisCardwell:
I am working on #480, and a design decision relates to the pipeline.  I am creating a diagram and figured that posting the current state here would also be useful.  Here is the current pipeline:

```mermaid
graph TD

  Clang-- parseCHeader -->C
  C-- genHsDecls -->Hs
  Hs-- genSHsDecls --> SHs

  subgraph Preprocessor
    PPIO["`source
    (IO)`"]
    PPPure["`source
    (pure)`"]
    HsModule-- genPP -->PPIO
    HsModule-- genPPString -->PPPure
  end

  SHs-- genModule -->HsModule

  subgraph Template Haskell
    DecsQ
    Exts[Set TH.Extension]
  end

  SHs-- genTH --> DecsQ
  SHs-- genExtensions --> Exts

  Tests["`tests
  (IO)`"]
  C-- genTests -->Tests

  Dump["`dump
  (IO)`"]
  C-- dumpCHeader -->Dump
```

Note that tests can (*should*) be generated from the `Hs` declarations.  They are currently generated from the `C.Header` only because of the limitations of the `Lib` API.  I will discuss this in #480.

TravisCardwell:
Rebased

PR #503, opened 2025-03-13
Add NameMangler.Easy

phadej:
I find name mangler interface quite complicated; this is my attempt to simplify it.


TravisCardwell (approved):
I actually like the explicitness of defining contexts, but the easy API usage indeed looks less complicated.  I am fine with using it!

edsko:
After discussing this with @TravisCardwell , merging this as-is, and will try to unify the two modules in a later PR.

PR #506, opened 2025-03-14
Give enums a `DeclPath`

edsko:
Builds on #505.


phadej:
Do we have an exampel of

```c
struct foo {
  enum { FOO, BAR} x;
};
```

i.e. where `DeclPath` branches are exercised? I think we don't, but with change we should.

edsko:
> Do we have an exampel of
> 
> ```c
> struct foo {
>   enum { FOO, BAR} x;
> };
> ```
> 
> i.e. where `DeclPath` branches are exercised? I think we don't, but with change we should.

We do now https://github.com/well-typed/hs-bindgen/pull/541 .

PR #507, opened 2025-03-14
Use `DeclPath` for opaque structs and enums

edsko:
Builds on #505 and #506.

### DIFF
--- a/hs-bindgen/src/HsBindgen/C/AST/Type.hs
+++ b/hs-bindgen/src/HsBindgen/C/AST/Type.hs
@@ -179,7 +179,7 @@ data StructField = StructField {
 --
 -- > struct foo;
 data OpaqueStruct = OpaqueStruct {
-      opaqueStructTag       :: CName
+      opaqueStructDeclPath  :: DeclPath
     , opaqueStructSourceLoc :: SingleLoc
     }
   deriving stock (Show, Eq, Generic)
@@ -230,7 +230,7 @@ data EnumValue = EnumValue {
 --
 -- > enum foo;
 data OpaqueEnum = OpaqueEnum {
-      opaqueEnumTag       :: CName
+      opaqueEnumDeclPath  :: DeclPath
     , opaqueEnumSourceLoc :: SingleLoc
     }
   deriving stock (Show, Eq, Generic)
--- a/hs-bindgen/src/HsBindgen/C/AST/Type.hs
+++ b/hs-bindgen/src/HsBindgen/C/AST/Type.hs
@@ -40,7 +40,7 @@ data Type =
     TypePrim PrimType
   | TypeStruct DeclPath
   | TypeUnion DeclPath
-  | TypeEnum CName
+  | TypeEnum DeclPath
   | TypeTypedef CName
   | TypePointer Type
   | TypeConstArray Natural Type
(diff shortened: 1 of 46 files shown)


hs-bindgen/src/HsBindgen/C/Fold/Type.hs
```diff
@@ -312,17 +299,16 @@ processTypeDecl' path extBindings unit declCursor ty = case fromSimpleEnum $ cxt
         if anon
         then do
             -- anonymous declaration, nothing to do
-
-            -- TODO: check with struct foo { struct { ... } field; };
+            -- TODO: This is wrong, they can be nested.
```
phadej, line 302: what's wrong?

hs-bindgen/src/HsBindgen/C/AST/Type.hs
```diff
@@ -178,7 +179,7 @@ data StructField = StructField {
 --
 -- > struct foo;
 data OpaqueStruct = OpaqueStruct {
-      opaqueStructTag       :: CName
+      opaqueStructDeclPath  :: DeclPath
```
phadej, line 182: This is questionable change.

As far as I can think of, opaque struct can only be defined at the top level, so `DeclPath` should always be of top-level form. Do we really want to relax that constraint. Please add a comment.

phadej:
```c
struct foo {
        struct opaque x;
};
```

is not accepted by GCC nor clang.

edsko: Hmm, that's fair. I am trying to get to the point that we don't need the `ToTypeConstrContext` class anymore, because we always have a `DeclPath` available. But perhaps we should just construct one on the fly for these opaque types?

hs-bindgen/src/HsBindgen/C/Fold/Type.hs
edsko: Perhaps I misunderstood what was happening here, but I thought examples like this is "what's wrong":

```c
struct exA {
        enum enumA {
                VALA_1,
                VALA_2
        } fieldA1;
};     

struct exB {
        enum {
                VALB_1,
                VALB_2
        } fieldB1;
};      
```

hs-bindgen/src/HsBindgen/C/AST/Type.hs
phadej: `ToTypeConstrContext` is indeed not elegant. But if you look at the instances:

```haskell
instance ToTypeConstrContext DeclPath where
    toCtx = NM.StructTypeConstrContext

instance ToTypeConstrContext NM.CName where
    toCtx = NM.TypeConstrContext
 ```
 
 they are not "just construct one on the fly", and in fact, `DeclPath` instance is small lie, as there isn't anything struct in union or enum, we just decided to treat the names of these uniformly (I think as that we do `toTyconName` uniformly, not `fromStructName`/`fromEnumName`/`fromUnionName` - the context API though imply the latter)
 
 So TL;DR, I think the simplification is probably better done from the mangler name. It's mangler which motivates changes in otherwise "very fine AST", so maybe it's the mangler which is off.

edsko: Yes, I think I agree. I will close this PR.

PR #509, opened 2025-03-17
Generate external bindings, skip missing headers

TravisCardwell:
The pipeline is refactored so that we only expose `HsDecls` in the public API.  Not exposing `CHeader` results in a much cleaner API.

Generation of external bindings is implemented for `struct`s, opaque `struct`s, `union`s, `enum`s, opaque `enum`s, and `typedef` aliases.  Support for `union` aliases was added, as it had not been implemented yet.

In the CLI, new `preprocess` options `--package NAME` and `--gen-external-bindings PATH` allow users to specify the path of the external bindings configuration to generate.

This PR also includes a commit that refactors the code so that missing headers during external bindings header resolution is not an error.  Warnings are displayed instead.  Support for `uchar.h` is restored now that it will not cause failures on macOS.


TravisCardwell:
(Rebased)

PR #510, opened 2025-03-18
Generate getters and setters for unions

phadej:


edsko:
I'll take care of the rebasing.

edsko:
@phadej When I re-ran the golden tests, I _think_ an additional set of getters/setters were created (compared to the PR as it stood), for `DimPayloadB`. I don't fully understand why they weren't there initially (maybe they were and I just messed up somewhere, entirely possible). Either way, it looks correct to me, so merging as-is.

PR #511, opened 2025-03-19
Delete `DeclConstr`

edsko:
We only need it for Haskell-to-C translation, and in that case, we actually have sufficient information externally.

More improvements to `DeclPath` to come.


edsko:
Since this information was available from context, not having it duplicated avoids having to worry about an enum name being given to a struct.

PR #514, opened 2025-03-19
Redesign `DeclPath`

edsko:
This makes a very clear distinction between the type that is being declared, and the context in which it is declared.

It's not quite perfect yet, ideally we'd be able to insist on a non-empty context for anonymous type declarations, but I think this new design is at least a lot easier to understand, and will hopefully reduce friction working with these `DeclPath`s.

With this new definition, perhaps it would make sense to rename `DeclPath` to something else?

### DIFF
--- a/hs-bindgen/src/HsBindgen/C/Fold/Type.hs
+++ b/hs-bindgen/src/HsBindgen/C/Fold/Type.hs
@@ -44,22 +44,22 @@ processTypeDecl extBindings unit declCursor ty = do
     -- dtraceIO "processTypeDecl" ty
     s <- get
     case OMap.lookup ty (typeDeclarations s) of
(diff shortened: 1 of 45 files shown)


hs-bindgen/src/HsBindgen/C/Fold/Type.hs
```diff
@@ -579,10 +592,11 @@ partitionFields = go id where
 mkStructField ::
      ExtBindings
   -> CXTranslationUnit
-  -> DeclPath
+  -> Maybe CName         -- ^ Name of the struct (unless anonymous)
```
phadej, line 595: This and the next field should really be (fieldName -> DeclPathCtxt); then this change would make sense.

edsko: Addressed in https://github.com/well-typed/hs-bindgen/pull/523 .

Issue #75, opened 2024-08-06
Generalize to C multiple headers and/or multiple Haskell modules


edsko:
As a starting point, we will simply generate a single Haskell module from a single C header. It seems that for Rust `bindgen` this suffices: the standard workflow for multiple headers [involves creating a `wrapper.h`](https://rust-lang.github.io/rust-bindgen/tutorial-2.html). Nonetheless, we might want to generalize to multiple C headers, and (for the preprocessor only #9 , not for TH mode #11) multiple Haskell modules.

phadej:
For Cabal preprocessors it's a bad idea. The whole system works as one input file to one module (hsc2hs, alex and happy work so). That said, Cabal preprocessor setup is just awful (cannot configure new ones declaratively AFAIK), so I'd use literate haskell hack (i.e. use [`-pgmL`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/phases.html#ghc-flag-pgmL-cmd)  flag for GHC), but that is even more strongly one to one mapping.

edsko:
Keeping the mapping from C headers to Haskell modules 1:1 seems a reasonable design constraint. No problem with that.

phadej:
> and not require the user to write a single header file that #includes all of those headers

That's however a very easy way to (mostly) solve https://github.com/well-typed/hs-bindgen/issues/71 too.

edsko:
The mapping @TravisCardwell and I discussed in https://github.com/well-typed/hs-bindgen/issues/381#issuecomment-2623798141 for using types defined in other libraries could provide an answer here also.

TravisCardwell:
The mapping mentioned in the previous comment is now called *external bindings*.  The initial version of this feature is now implemented.  It is currently feasible for a user to generate multiple Haskell modules corresponding to multiple C headers.  Users can generate modules one at a time, specifying external bindings configuration for any modules (and other dependencies) used by the module currently being generated.  Users must do this in an appropriate order.

Consider a project with seven header files that should be translated to seven Haskell modules.  All headers use the standard library, and the include graph is as follows:

```mermaid
graph BT;
  B --> A
  C --> A
  E --> D
  F --> B
  G --> B
  G --> C
  G --> E
  G --> D
```

One possible generation order is as follows:

| Generate | External Bindings |
| -------- | ----------------- |
| A        | base              |
| D        | base              |
| B        | base, A           |
| C        | base, A           |
| E        | base, D           |
| F        | base, B           |
[...]

PR #520, opened 2025-03-23
Resolve input headers in the parser

TravisCardwell:
Input headers must be specified relative to an include search path directory.  Previously, we resolved the specified header and passed the `SourcePath` to the parsing function.  Due to the way that we are currently resolving headers, this meant that `libclang` was actually parsing the input header twice.  With this change, the relative path is passed to the parsing function, and the parsing function uses a `CXUnsavedFile` to import it.  Input headers are now only parsed once.

Using a `CXUnsavedFile`, it is possible to include more than one input header.  This enables translating multiple C headers to one Haskell module.  This PR includes the necessary changes to the implementation but does not yet change the public API.  We need to decide if/how we should expose this feature to users (#75).  Note that it is now possible to support translating multiple C headers to one Haskell module (this change) as well as to multiple modules (using external bindings).

This change requires changing how we manage input "header not found" errors.  Previously, they were generally caught during header resolution, and a pretty error message was output.  A `TranslationUnitCannotOpen` constructor of `TranslationUnitException` was implemented, but it was generally not used because most error cases were caught during resolution.  With this PR, a missing input header is reported as a diagnostic.

There are three types of errors:

* Header not found
* Error(s) in a C file
* Unknown error

Which of these are user errors (`HsBindgenException`), and which are internal errors that users should report as bugs?  Note that the exception was previously just thrown using `throwIO`.  For now, I made all of these errors user errors.

We can easily differentiate input headers using the source location in the diagnostic.  The current implementation does this so that input "header not found" errors do not leak internal `CXUnsavedFile` details.  Non-input "header not found" errors are considered "error(s) in a C file" so that they are formatted with source location information.

Function `resolveHeader` is removed from the public API.  All header resolution is internal.

Some related dead code is removed.


edsko:
> Which of these are user errors (HsBindgenException), and which are internal errors that users should report as bugs? Note that the exception was previously just thrown using throwIO. For now, I made all of these errors user errors.

"Header not found" and errors in C headers seem clear-cut examples of user errors, not bugs in bindgen. "Unknown error" is less clear, I guess these should be reported, if only because we should make them known errors so that we can blame the user again? :)

Issue #499, opened 2025-03-12
Separate public and internal APIs


TravisCardwell:
PR #498 refactors the public `HsBindgen.Lib` and `HsBindgen.TH` APIs, but we still expose many modules for our tests.  We should make the public and internal APIs very clear.  Options include:

* Use an internal library (`hs-bindgen:internal`)
    * Tests can use the internal API, while users cannot.
    * The internal and public libraries always have the same version since they are in the same package.
    * A separate source directory should be used.  Prefix `HsBindgen.Internal` should be used for all `hs-bindgen:internal` modules *except for* those that are re-exported by the public library.
* Use one or more `Internal` modules
    * We could add an `HsBindgen.Internal` module that re-exports only what we need for tests.
    * Alternatively, we could use prefix `HsBindgen.Internal` for all modules that are not part of the public API.
* Use an `hs-bindgen-internal` package à la [Internal convention is a mistake](https://nikita-volkov.github.io/internal-convention-is-a-mistake/))
    * This is the Pythonic "we're all consenting adults here" option.
    * Versions for `hs-bindgen-internal` and `hs-bindgen` may differ.

Related decisions:

* How strictly do we want to make `hs-bindgen-libclang` internal?  If use use an `hs-bindgen:internal` library, might we also want to move `hs-bindgen-libclang` to an internal `hs-bindgen:libclang` library?
* Would it be worth it to strictly separate public API tests from tests that use internal APIs?  I think it would be a good idea.

edsko:
Having never worked with internal libraries, I'm unsure if there are any caveats I am not thinking of, but using an internal library does sound like the cleanest solution. Let's make sure that that works with stack also, though, just to be sure.

Then let's also

* Make `hs-bindgen-bootstrap`  an executable within `hs-bindgen`
* Make `ls-bindgen-libclang` an internal library

We _could_ make `hs-bindgen-runtime` a _public_ internal library, but working support for public internal libraries requires very recent versions of `cabal` so let's not do that just yet; and then rename `hs-bindgen-testlib` to `hs-bindgen-test-runtime` for consistency.

Eventually `ansi-diff` and `c-expr` can probably move to separate repos.

phadej:
Few points:
- `hs-bindgen-bootstrap` doesnt need to be part of `hs-bindgen` package, and in fact it couldn't. Because it's used for *bootstrapping*
[...]

PR #355, opened 2025-01-09
Rename to `hs-bindgen-runtime`

TravisCardwell:
All modules now have prefix `HsBindgen.Runtime`.  Module `HsBindgen.Runtime.Patterns` continues to re-export internal submodules.

PR #521, opened 2025-03-25
Separate public and internal APIs

TravisCardwell:
This PR implements the changes discussed in #499.  Overview:

* The `hs-bindgen-libclang` dependency on `hs-bindgen-runtime` is removed.  The modules used (noted below) are moved into `hs-bindgen-libclang`.  With this change, the `hs-bindgen-runtime` dependency is also removed from `c-expr` and `clang-ast-dump`.
    * `HsBindgen.Runtime.Backtrace` → `HsBindgen.Clang.Backtrace`
    * `HsBindgen.Runtime.Enum.Bitfield` → `HsBindgen.Clang.Enum.Bitfield`
    * `HsBindgen.Runtime.Enum.Simple` → `HsBindgen.Clang.Enum.Simple`

* Package `hs-bindgen-libclang` is renamed to `clang`.  It is not specific to `hs-bindgen` and can be moved into a separate repository in the future.

* Package `hs-bindgen-testlib` is renamed to `hs-bindgen-test-runtime`.

* Executable `hs-bindgen` is renamed to `hs-bindgen-cli`.

* Development commands are moved out of the `hs-binggen-cli` executable to a new `hs-bindgen-dev` executable.  It is only built when the `dev` flag is set.

* Executable `clang-ast-dump` is moved into `hs-bindgen`.  It is only built when the `dev` flag is set.  Note that it is *not* part of `hs-bindgen-dev` because it does not have the same global options.

* The `hs-bindgen` implementation is moved into an internal library.
    * This PR implements the minimal change for moving the `hs-bindgen` implementation into an internal library.  Currently, all modules are exposed, except for those that are documented in the Cabal file as being re-exported.  We should probably clean up the API *after* this change has been merged and everybody has already rebased to resolve any merge conflicts in other branches.
    * The public library implements the `Lib` and `TH` APIs as well as re-exports the following modules:
        * `HsBindgen.Errors`
        * `HsBindgen.Hs.NameMangler`
        * `HsBindgen.Hs.NameMangler.DSL`
    * If we do not want to expose all of the `Errors` module, we should refactor that.
    * Note that I followed the suggestion to *not* use a unique module name prefix (`HsBindgen.Internal`) for the internal library.  If issues arise, I am happy to refactor the library to use one.
    * Executable `hs-bindgen-cli`, test `test-pp`, and test `test-th` do *not* depend on the internal library.  I think that we should maintain this, as these should only use/test the public API.
    * Development executable `hs-bindgen-dev`, development executable `clang-ast-dump`, and test `golden` depend on the internal library, by design.

* Executable `clang-bootstrap` is moved into `clang`, so that it is only built when the `dev` flag is set.  There is a tight coupling, so it is nice to version them together.  This also reduces the number of packages that will need to be maintained on Hackage.

    * Example workflow:

        ```
        $ cabal build clang -f dev
        $ cabal run clang-bootstrap -f dev
        $ cabal build all -f dev
        $ cabal test all
        ```

    * Just `cabal-bootstrap` can be built and run, but note that `cabal clean` must then be run before further development:

        ```
        $ cabal build clang:clang-bootstrap -f dev
        $ cabal run clang-bootstrap -f dev
        $ cabal clean
        ```

* CI is configured to set the `dev` flag for both `clang` and `hs-bindgen`, so that building of development tools is tested in CI.

This PR changes a lot of paths, making rebasing a pain.  Apologies in advance! :bow:

This PR is being created as a draft because it must be rebased before merging.  PR #520 should be merged before this one, and the corresponding squashed commit in this branch will be dropped.  Any other merges will need to be integrated via edits of the other commits.


phadej:
https://github.com/well-typed/hs-bindgen/pull/355#pullrequestreview-2538938993

TravisCardwell:
(Rebased)

The `diff`s for this PR are probably not easy to review.  It might be easier to just look at the [`refactor-packages-and-modules`](https://github.com/well-typed/hs-bindgen/tree/refactor-packages-and-modules) branch.

edsko:
Merging this as-is since it has a large surface area. We can adjust smaller parts in separate PRs if needed.

Issue #493, opened 2025-03-08
Add tests for ExtBindings


phadej:
For now the related code paths seem to be dead code, i.e. if removed nothing breaks.

edsko:
@TravisCardwell I'm closing this as we have golden tests for the external bindings, though we might eventually want to improve how we test this more thoroughly (but that is a larger discussion on how we want to test bindgen in general, not just external bindings).

PR #524, opened 2025-03-26
Add external bindings tests

TravisCardwell:
This PR adds unit tests for the *internal* API.  In general the objective of such tests is to confirm the behavior both now an in the future, making it easier/safer to maintain the implementation.

We had only one unit test (for `getTargetTriple`) that was tacked onto the golden tests.  This PR refactors the test suite to put unit tests and golden tests in different test groups.  (The source is renamed from `golden.hs` to `Main.hs`, and the test suite is renamed to `tests`.)  Unit tests are put into `Test` modules; for example tests for `HsBindgen.ExtBindings` are put into a `HsBindgen.ExtBindings.Test` module.

I plan on discussing this with @edsko on Thursday.  If this is not a convention that we would like to use in this project, I am happy to change it.

Regarding external bindings, I made a number of fixes.

Additions:

* External bindings for `newtype` wrappers representing `#define` macros of types are now generated.

Bug fixes:

* Ignoring missing headers when resolving headers for external bindings was no implemented correctly.  This issue is fixed.
* Generated external bindings for opaque `struct`s and `enum`s did not use the necessary prefix.  This issue is fixed.

Minor changes:

* When resolving the headers for external bindings, the API now returns a set of of errors instead of a list.  They were already a set internally, so this is not a significant change.  It just makes it easier to compare the errors (in tests).
* Similarly, conflicting names when merging external bindings is now stored in a set instead of a list.
* Some `Eq` instances are added to error types.

### DIFF
--- a/hs-bindgen/hs-bindgen.cabal
+++ b/hs-bindgen/hs-bindgen.cabal
@@ -259,8 +259,12 @@ test-suite test-internal
   hs-source-dirs: test/internal
   other-modules:
       Test.HsBindgen.C.Parser
+      Test.HsBindgen.ExtBindings
       Test.Internal.Misc
+      Test.Internal.QuickCheck
+      Test.Internal.QuickCheck.Orphans
       Test.Internal.Rust
+      Test.Internal.Tasty
       Test.Internal.TastyGolden
       Test.Internal.TH
       Test.Internal.TreeDiff.Orphans
@@ -274,23 +278,28 @@ test-suite test-internal
   build-depends:
       -- Inherited dependencies
     , bytestring
+    , containers
     , debruijn
     , mtl
     , template-haskell
+    , text
     , vec
   build-depends:
       -- External dependencies
     , ansi-diff
     , deepseq
-    , directory     ^>=1.3.6.2
-    , filepath      ^>=1.4.2.2 || ^>=1.5.2.0
+    , directory        ^>=1.3.6.2
+    , filepath         ^>=1.4.2.2 || ^>=1.5.2.0
     , process
-    , syb           ^>=0.7.2.4
-    , tasty         ^>=1.5
-    , tasty-hunit   ^>=0.10.2
+    , QuickCheck       ^>=2.15.0.1
+    , syb              ^>=0.7.2.4
+    , tasty            ^>=1.5
+    , tasty-hunit      ^>=0.10.2
+    , tasty-quickcheck ^>=0.11.1
     , temporary
-    , tree-diff     ^>=0.3.4
-    , utf8-string   ^>=1.0.2
+    , text-icu         ^>=0.8
+    , tree-diff        ^>=0.3.4
+    , utf8-string      ^>=1.0.2
 
 test-suite test-th
   type:           exitcode-stdio-1.0
--- a/hs-bindgen/hs-bindgen.cabal
+++ b/hs-bindgen/hs-bindgen.cabal
@@ -297,9 +297,11 @@ test-suite test-internal
     , tasty-hunit      ^>=0.10.2
     , tasty-quickcheck ^>=0.11.1
     , temporary
-    , text-icu         ^>=0.8
     , tree-diff        ^>=0.3.4
     , utf8-string      ^>=1.0.2
+  if os(linux)
+    build-depends:
+      , text-icu ^>=0.8
 
 test-suite test-th
   type:           exitcode-stdio-1.0
--- a/.github/workflows/simple.yml
+++ b/.github/workflows/simple.yml
@@ -69,6 +69,15 @@ jobs:
         if: ${{ startsWith(matrix.os, 'macos') }}
         run: echo "LLVM_CONFIG=$LLVM_PATH/bin/llvm-config" >> "$GITHUB_ENV"
 
+      - name: ICU PKG_CONFIG_PATH (macos)
+        shell: bash
+        if: ${{ startsWith(matrix.os, 'macos') }}
+        run: |
+          export PKG_CONFIG_PATH=$(brew --prefix)/opt/icu4c/lib/pkgconfig
+          echo "PKG_CONFIG_PATH=${PKG_CONFIG_PATH}" >> "$GITHUB_ENV"
+          echo "${PKG_CONFIG_PATH}"
+          ls -l "${PKG_CONFIG_PATH}"
+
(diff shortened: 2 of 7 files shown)


TravisCardwell:
One thing to point out is that I use the following:

```haskell
newtype HiddenException = HiddenException CallStack
  deriving Show

instance Exception HiddenException where
  displayException (HiddenException cs) = unlines
    [ "A test accessed data that was explicitly hidden."
    , prettyCallStack cs
    ]

hidden :: HasCallStack => a
hidden = throw $ HiddenException callStack
```

For example, function `newtypeUnion` creates a `newtype` declaration (named `Foo`) for a C `union` with the specified `DeclPath` and aliases, and anything that should not be used for external bindings is marked `hidden`:

```haskell
newtypeUnion :: C.DeclPath -> [C.CName] -> Hs.Decl
newtypeUnion declPath aliases = Hs.DeclNewtype Hs.Newtype
    { Hs.newtypeName   = "Foo"
    , Hs.newtypeConstr = hidden
    , Hs.newtypeField  = hidden
    , Hs.newtypeOrigin = Hs.NewtypeOriginUnion C.Union
        { C.unionDeclPath  = declPath
        , C.unionAliases   = aliases
        , C.unionSizeof    = hidden
        , C.unionAlignment = hidden
        , C.unionFields    = hidden
        , C.unionSourceLoc = hidden
        }
    }
```

Some benefits to this:

* It makes it clear what is relevant to the tests.
* We do not "waste" time implementing parts of test data that we do not need.
* The burden of maintaining the test code is reduced.

TravisCardwell:
Discussing with @edsko, we decided to use `Test` prefix for the module names.  This will allow us to move the test library code out of the top-level.

We also think that it will be worthwhile to separate tests for the public and internal APIs.  The tests for the public API will *not* depend on `hs-bindgen:internal` and may only use the public API.  Some tests can serve as examples to users.

We will also add property tests, so perhaps a top-level `unit` group is not appropriate.  Tests should be organized by component, under which we may have both property and unit tests.  Note that `golden` tests are organized by example, but I will not change that at this time.

I would like to make these changes *before* merging.

hs-bindgen/hs-bindgen.cabal
```diff
@@ -250,12 +251,23 @@ executable clang-ast-dump
       -- External dependencies
     , optparse-applicative >= 0.18 && < 0.19
 
-test-suite golden
+test-suite tests
```
phadej, line 254: `tests` is a terrible name, partly because it's a magic keyword in `cabal-install` selector syntax.

Use something more specific, `hs-bindgen-tests`.

hs-bindgen/hs-bindgen.cabal
```diff
@@ -250,12 +251,23 @@ executable clang-ast-dump
...
   import:         lang
   type:           exitcode-stdio-1.0
-  main-is:        golden.hs
+  main-is:        Main.hs
```
phadej, line 257: My principle is to name `main-is` files reflecting componetn name, `hs-bindgen-tests.hs` (or previously `golden.hs`).

Because in big projects ,like this one, you'll end up with multiple `Main.hs` files, and it gets very confusing very fast.

phadej:
What I meant by #493, is not some internal-unit tests, but a full fledged end-to-end tests, more like `test-th` and `test-pp`. So this PR doesn't resolve #493 in my opinion. (The name of this PR implies so).

TravisCardwell:
IMHO, unit tests provide a good way to associate assertions with implementations, especially combinations of possibilities and edge cases, so that future development such as maintenance and refactoring can be done with less risk of breaking things.  I understand that property tests and end-to-end tests are preferred for this project, however, so I will course correct.

`Test02` (of `test-th` and `test-pp`) uses external bindings.  It only covers a happy path.  The error cases are more difficult to test in end-to-end tests because they result in build failures.

The `golden` tests use `.extbindings.yaml` fixtures for generated external bindings.

So far, I wrote some property tests that roundtrip to/from configuration files.

TravisCardwell:
In the `Arbitrary` instance for `CName`, I used [`text-icu`](https://hackage.haskell.org/package/text-icu) to generate names based on `XidStart` and `XidContinue` character properties (which is how C identifier syntax is specified).  This dependency failed to install on both Windows and macOS.

The `text-icu` repo uses GitHub Actions to test on Windows and macOS: [`cabal-mac-win.yml`](https://github.com/haskell/text-icu/blob/master/.github/workflows/cabal-mac-win.yml).  Apparently ICU is already installed on macOS, and setting `PKG_CONFIG_PATH` indeed fixes the issue for us as well.  On Windows, `C:\msys64\mingw64\bin` is added to `GITHUB_PATH`, and `C:\msys64\usr\bin` is temporarily added to the `PATH` to install both `pkg-config` and ICU using `pacman`.  Two paths were used because of an issue, but that issue has since been resolved.  Putting these in the same step fixes the issue for us.  Listing `C:\msys64\mingw64\bin` did not produce any output, but installation fails if it is not added to `GITHUB_PATH`; strange.  I wish I could look at the filesystem and better understand how Windows is configured.  I restored the `GITHUB_PATH` setting and all tests pass in CI.

This may make testing more difficult in development Windows environments, however.  I am booting my Windows test environment and will try it.

If using `text-icu` is problematic, it is possible to remove the dependency and simply not generate as complete a range of random `CName`s.  I can hard-code some ranges of Unicode code points so that we can test using a subset of the valid Unicode characters.

edsko:
Some end to end testing of external bindings is already happening `Test02`, which depends on external bindings for `base` and `hs-bindgen-runtime`. I've agreed with @TravisCardwell that we'll have a dedicated test also for "non-standard" external bindings, with the following ingredients:

* test/cbits/header1.h 
* test/cbits/header2.h
* external bindings for header1 (genrated or handwritten, either way)
* generate bindings for header2, using external bindings for ehader1
* haskell module referenced by external bindings for header1
* haskell test module that uses stuff from header2
* secondary goal: have this be a "reference" for the kinds of things we can  have external bindings for (therefore use public API only)

I think we can close #493 once such a thing has been added?

TravisCardwell:
This task has been pushed down the stack and is now dated.  We have generalized external bindings to binding specifications, and now we are doing a *major* refactoring.  I am therefore closing this PR and will create a new one when the new implementation is ready.

PR #539, opened 2025-04-03
Remove ToHs type-class

phadej:


phadej:
This would been a first commit of #538, there C-to-HS translation is modified so `ToHs` would lose its last instances anyway

edsko (approved):
Perhaps it would be good if we could change the "known to break"  fixtures to not include line numbers?

phadej:
> Perhaps it would be good if we could change the "known to break" fixtures to not include line numbers?

Yeah, probably a good idea to do some more mangling (than just windows path slash) to reduce noise. I'll do that next.

PR #540, opened 2025-04-03
Clarify control flow

edsko:
It was a little difficult to see which cases we were distinguishing exactly, and we had some repeated code paths also. This clarifies this control flow, and uses the same code paths in multiple places, clarifying where there are inconsistencies (e.g., external bindings for unions).

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/C/Fold/Type.hs
+++ b/hs-bindgen/src-internal/HsBindgen/C/Fold/Type.hs
@@ -169,205 +169,111 @@ processTypeDecl' ctxt extBindings unit declCursor ty = case fromSimpleEnum $ cxt
 
     -- structs and unions
     Right CXType_Record -> do
-        decl <- liftIO (clang_getTypeDeclaration ty)
-        ki <- liftIO $ fromSimpleEnum <$> clang_getCursorKind decl
+        decl <- liftIO $ clang_getTypeDeclaration ty
+        ki   <- liftIO $ fromSimpleEnum <$> clang_getCursorKind decl
+        sloc <- liftIO $ HighLevel.clang_getExpansionLocation
+                     =<< clang_getCursorLocation decl
+
         case ki of
             Right CXCursor_StructDecl -> do
-                name <- liftIO (clang_getTypeSpelling ty)
-                anon <- liftIO (clang_Cursor_isAnonymous decl)
-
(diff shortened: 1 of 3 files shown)


hs-bindgen/src-internal/HsBindgen/C/Fold/Type.hs
```diff
@@ -169,205 +169,111 @@ processTypeDecl' ctxt extBindings unit declCursor ty = case fromSimpleEnum $ cxt
...
-                                    { unionDeclPath  = declPath
-                                    , unionAliases   = []
-                                    , unionSizeof    = fromIntegral sizeof
-                                    , unionAlignment = fromIntegral alignment
-                                    , unionFields    = fields
-                                    , unionSourceLoc = sloc
-                                    }
+                    case flavour of
+                      TypeDeclExternal _extId ->
+                        panicIO "external bindings for unions not implemented #537"
```
phadej, line 224: Should this be `throwIO_TODO`? Or is `TypeDeclExternal` case here unreachable? Why we cannot do `addAlias ty $ TypeExtBinding extId` here?

hs-bindgen/src-internal/HsBindgen/C/Fold/Type.hs
```diff
@@ -575,6 +483,84 @@ updateDeclAddAliases aliases = \case
...
+-- arbitrary type.
+--
+-- TODO: Should we warn when we see these?
+data AnonTopLevel = AnonTopLevel Type
+
+-- | Classify type declaration
+classifyTypeDecl ::
+     DeclPathCtxt
+  -> ExtBindings
+  -> (CXType, CXCursor, Either CInt CXCursorKind, SingleLoc)
```
phadej, line 512: why a tuple?

hs-bindgen/src-internal/HsBindgen/C/Fold/Type.hs
edsko: Ah, yes, it should be `throwIO_TODO`.

hs-bindgen/src-internal/HsBindgen/C/Fold/Type.hs
edsko: I use this a way to signal that these are all closely related.

PR #545, opened 2025-04-04
Fix double escape bug in name mangler

### LINKED ISSUE #535: Name mangler double-escape bug
While researching this morning, I thought of a case that we might want to be sure to test: use of combining diacritical marks.  These generally have property `XID_Continue` but do not satisfy `isAlphaNum`.

Example:

```c
typedef int adiós;

adiós main() {
  return 0;
}
```

This example is hand-coded to use a combining acute accent (`U+0301`), *not* the canonical `ó` character (`U+00F3`).  It is therefore *not* NFC (Normalization Form Canonical Composition) normalized.

```
$ hexdump -C adios.c
00000000  74 79 70 65 64 65 66 20  69 6e 74 20 61 64 69 6f  |typedef int adio|
00000010  cc 81 73 3b 0a 0a 61 64  69 6f cc 81 73 20 6d 61  |..s;..adio..s ma|
00000020  69 6e 28 29 20 7b 0a 20  20 72 65 74 75 72 6e 20  |in() {.  return |
00000030  30 3b 0a 7d 0a                                    |0;.}.|
00000035
```

With default options, Clang compiles without warning, while GCC warns that it is not NFC-normalized.

```
$ clang -o adios adios.c
$ gcc -o adios adios.c
adios.c:1:13: warning: `adio\U00000301s' is not in NFC [-Wnormalized=]
    1 | typedef int adio<U+0301>s;
      |             ^~~~~~~~~~~~~
adios.c:3:1: warning: `adio\U00000301s' is not in NFC [-Wnormalized=]
    3 | adio<U+0301>s main() {
      | ^~~~~~~~~~~~~
```

Running `hs-bindgen-cli` on a header with just the `typedef`, there is a bug!  It looks like it correctly escapes the type constructor name but then escapes the single quote in the type constructor name to create the data constructor name, which is incorrect.

`$ cabal run hs-bindgen-cli -- preprocess --input adios.h`

```haskell
...
newtype Adio'0301s = Adio'00270301s
  { un_Adio'00270301s :: FC.CInt
  }
...
```

[`adios.h.gz`](https://github.com/user-attachments/files/19581392/adios.h.gz)

Note that we *could* normalize ([`nfc`](https://hackage.haskell.org/package/text-icu-0.8.0.5/docs/Data-Text-ICU-Normalize2.html#v:nfc)) C names before translation, but this would require `text-icu`.

edsko:
Closes #535


TravisCardwell (approved):
Looks good!  That was an easy fix!

Issue #542, opened 2025-04-04
Incorrect `Enum` instance for generated `enum` types


TravisCardwell:
We represent C `enum`s using a `newtype` wrapper around a C integral type (such as `CUInt`) and define a pattern for each constant of the enumeration.  We are generating Haskell `Enum` instances as follows ([example](https://github.com/well-typed/hs-bindgen/blob/33bb1894fd7eb45bd516ad4d700923b1764debef/hs-bindgen/fixtures/enums.pp.hs#L42)):

```haskell
deriving newtype instance Enum First
```

This is not correct because it derives the instance over the full domain of the underlying integral type.

```
λ: take 4 (enumFrom FIRST1)
[First {un_First = 0},First {un_First = 1},First {un_First = 2},First {un_First = 3}]
it :: [First]
```

Defining a correct `Enum` instance for such types is not trivial because C enumeration constant values may not be sequential.  Perhaps it would be worthwhile to have two implementations: one for enumerations with sequential values, because it is a common case that we can make performant, as well as one for enumerations with non-sequential values.

Multiple C enumeration constants may have the same value, but I do not think that this is a problem since our representation is a wrapper around the values themselves.

TravisCardwell:
I am running into issues with code generation.  The `SHs` AST defines a `Global` sum type that contains all global references, including constructors and variables (as well as type constructors and classes).  Code generation for `EGlobal` and `EInfix` both use `TH.varE`, however, which does not work for constructors.  For example, when trying to generate a `Maybe` expression, both `EGlobal Maybe_Nothing` and `EApp (EGlobal Maybe_Just) _` both result in "illegal variable name" errors when splicing.

What is the best way to resolve this issue?  One option that does not complicate the AST is to add a function `mkGlobalExpr :: Quote q => Global -> q TH.Exp` that uses the correct `TH.Exp` constructor for each global reference, using the existing `mkGlobal` function to resolve names.

PR #552, opened 2025-04-08
Fix `Enum` instances for generated `enum` types

TravisCardwell:
The incorrect `Enum` instance for generated `enum` types issue (#542) sniped me for some reason.  I wanted to test a few things and kept pulling threads until I had a full prototype implementation.  I went ahead and created an implementation while it was still paged in.

Two type classes are implemented:

* `SequentialCEnum` is used when the enumeration values are sequential values of the underlying integral type.  Instances for `Bounded` and `Enum` are derived via `newtype` wrapper `SeqCEnum`.
* `GeneralCEnum` is used in other cases.  The implementation uses a list of values, so performance is generally linear.  (I wonder if the Haskell ecosystem has a sorted tree that provides ordered enumeration from a value found in logarithmic time?)  Instances for `Bounded` and `Enum` are derived via `newtype` wrapper `GenCEnum`.

Since type classes/families are used, I just "tested" some specific cases using unit tests.  In addition, I added a non-sequential enumeration to the `enums.h` example and added some `enum` tests to `test_01` (of `test-pp` and `test-th`).


TravisCardwell:
Sorry for the noise.  This is the *second* time I forgot that `unsnoc` was only added to `base` relatively recently.  For some reason it feels like something that should have been there for a long time.

hs-bindgen-runtime/src/HsBindgen/Runtime/CEnum/General.hs
```diff
@@ -0,0 +1,305 @@
...
+    values = (if i < j then id else reverse) $
+      generalCEnumValues (Proxy :: Proxy a)
+
+{-------------------------------------------------------------------------------
+  Instance Deriving
+-------------------------------------------------------------------------------}
+
+-- | Type used to derive 'Bounded' and 'Enum' instances when a type has a
+-- 'GeneralCEnum' instance
+newtype GenCEnum a = GenCEnum a
```
edsko, line 274: Let's rename this to `AsCEnum`, and explicitly mention "deriving-via" in the comments.

hs-bindgen-runtime/src/HsBindgen/Runtime/CEnum/General.hs
```diff
@@ -0,0 +1,305 @@
+{-# LANGUAGE TypeFamilies #-}
+
+module HsBindgen.Runtime.CEnum.General (
```
edsko, line 3: Could consider changing naming convention for qualified imports, would make this easier perhaps. No strong preference.

hs-bindgen-runtime/src/HsBindgen/Runtime/CEnum/General.hs
```diff
@@ -0,0 +1,305 @@
...
+    -- * Type Class
+    GeneralCEnum(..)
+    -- * Bounded API
+  , generalCEnumMinBound
+  , generalCEnumMaxBound
+    -- * Enum API
+  , generalCEnumSucc
```
edsko, line 10: Could hide these, would provide for more implementation hiding, giving us the ability to change internals whilst keeping all functions still mutually consistent.

edsko (commented):
After discussing this with @TravisCardwell , we decided to go a slightly different route here. It is somewhat inconsistent that we define a set of pattern synonyms for enums and explicitly _not_ declare that set to be complete, because C enums do not restrict the underlying range, and yet have implementations of `Enum` and `Bounded` which pretend that that range _is_ complete. 

Instead we will define a new `CEnum` class, which makes some different design choices than `Enum` does, and not derive `Enum` or `Bounded` at _all_ for C enum types. This gives a very clear signal that the semantics of these functions requires a bit of thought.

TravisCardwell:
I realized that `Map` also provides a `split` API, so a single `Map` can specify both valid values and associated names.  With that change, the two variants would be quite similar, so I went ahead and redesigned so that there is a single `CEnum` type class that handles both sequential and non-sequential values.  The optimization for sequential values is still implemented, and it now supports empty enumerations.

```haskell
class Integral (CEnumZ a) => CEnum a where
  type CEnumZ a
  wrap :: CEnumZ a -> a
  unwrap :: a -> CEnumZ a
  declaredValueMap :: proxy a -> Map (CEnumZ a) (NonEmpty String)
  sequentialValueBounds :: proxy a -> Maybe (CEnumZ a, CEnumZ a)
```

Functions are now named for qualified import.  The whole API is implemented in one module.  In addition to the above, the following functions are provided:

```haskell
isDeclared :: CEnum a => a -> Bool
mkDeclared :: CEnum a => CEnumZ a -> Maybe a
showCEnum  :: CEnum a => String -> a -> String
getNames   :: CEnum a => a -> [String]
```

Function `showCEnum` takes the name of the constructor.  It is used in the definitions of generated `Show` instances.

```haskell
instance Show StatusCode where
  show = CEnum.showCEnum "StatusCode"

show StatusOK         == "StatusOK"
show (StatusCode 418) == "StatusCode 418"
```

Type `AsCEnum` defines instances for `Bounded` and `Enum`, to be used with `DerivingVia`.  These instances are *not* derived by default.  The functions that implement each method are *not* exported.

This implementation no longer generates instances using `DerivingVia`, but I kept support for it because we will use it for `Peekable` and related instances.

This implementation generates code that include constructors referenced using `Global`.  The TH backend used `TH.varE` for all globals, which does not work.  There are a number of ways to resolve this.  I implemented the least intrusive: function `mkGlobalExpr`.

This PR should be rebased/merged after #576.

edsko (approved):
We discussed some cosmetic changes:

```haskell
class Integral (CEnumZ a) => CEnum a where
  type CEnumZ a
  fromCEnumZ     :: CEnumZ a -> a
  toCEnumZ       :: a -> CEnumZ a
  declaredValues :: proxy a -> Map (CEnumZ a) (NonEmpty String)

class CENUm a => SequentialCEnum where  
  rangeIsSequential :: (a, a)
```

and a few others. Feel free to merge once you're happy that these are done.

TravisCardwell:
I realized that one downside of this design is that functions `isDeclared` and `mkDeclared` do not take advantage of `SequentialCEnum` optimization (providing `O(1)` instead of `O(log n)` time).  We could provide *separate* functions if needed, but perhaps this is not necessary.

I made one small change.  `rangeIsSequential` was the name we came up with when it was part of `CEnum`.  I decided to specify the bounds separately, which is easier to work with.

```haskell
class CEnum a => SequentialCEnum a where
  minDeclaredValue :: a
  maxDeclaredValue :: a
```

PR #565, opened 2025-04-09
Deal with awkward names

### LINKED ISSUE #562: Unicode characters are escaped in foreign imports
For example,  given

```c
void hi你好(void);
void adiós(void);
```

(where the first example is and the second example is not in [NFC](#535)), we generate

```haskell
foreign import capi safe "example.h hi\20320\22909" hi你好 :: IO ()
foreign import capi safe "example.h adio\769s" adio'0301s :: IO ()
```

I think this is only an issue for functions, not for datatypes.

### LINKED ISSUE #563: Name mangler generates invalid names if `toUpper` is "broken"
Consider

```c
typedef int 东西;
```

The first character here is considered `isAlphaNum` but (interestingly and quite reasonably) _neither_ `isLower` _nor_ `isUpper`. However, this this is a type name, we do try to call `toUpper` on this to generate the Haskell name, which fails (it returns the character unchanged), resulting in invalid Haskell.

edsko:
Closes #562
Closes #563


hs-bindgen/src-internal/HsBindgen/Hs/NameMangler/DSL/FixCandidate.hs
```diff
@@ -172,47 +173,117 @@ singNameRuleSet _ =
   Applying name rule sets
 -------------------------------------------------------------------------------}
 
+data CannotApplyRuleset ns where
+  CannotApplyRuleset ::
+       -- Since there are no characters which /are/ 'isUpper' but for which
+       -- 'toLower' fails (returns the character unchanged), this can only fail
+       -- for 'NameRuleSetOther'.
```
TravisCardwell, line 180: This is not true.  Counterexample: [U+03D2 Greek Upsilon with Hook Symbol](https://codepoints.net/U+03D2)

```
λ: import qualified Data.Char as Char
λ: let c = Char.chr 0x03d2
λ: Char.isUpper c
True
it :: Bool
λ: c == Char.toLower c
True
it :: Bool
```

hs-bindgen/src-internal/HsBindgen/Hs/NameMangler/DSL/FixCandidate.hs
```diff
@@ -172,47 +173,117 @@ singNameRuleSet _ =
...
+--    > toUpper '_' == '_'
+--    > toUpper '事' == '事'
 --
--- * If we are generating a 'NameRuleSetVar' name, the underscore is fine
---   as-is; any letters can be made uppercase.
--- * If we are generating a 'NameRuleSetOther', and the identifier starts
---   with an underscore, we strip off everything that's not a letter and
---   call @onInvalidFirst@ on this prefix and the remaining suffix.
+--    When this happens we strip off all unuseable characters, and then call the
+--    supplied function to deal with this edge case.
```
TravisCardwell, line 222: I am sorry to nitpick about wording, but perhaps the description for step (3) is confusing.  Suggestion:

> When this happens, we call the supplied function with the unusable prefix and the remainder of the identifier.

hs-bindgen/src-internal/HsBindgen/Hs/NameMangler/DSL/FixCandidate.hs
edsko: Hah! Delightful :D

PR #567, opened 2025-04-10
Improved union experiment

edsko:


edsko:
We decided against this alternative proposal for dealing with unions. The nail of the coffin is that not all unions should be considered _either or_, but are instead _both_. For example, in

```c
union {
  struct {
    float x;
    float y;
    float z;
  }
  float[3] arr;
};
```

these are two different views on the _same_ data, that are both valid at at all times. That use case is not covered by this alternative proposal. 

"Actual" tagged unions are part of the high-level API instead.

PR #573, opened 2025-04-11
Less reserved names

phadej:


phadej:
E.g.

```
$ ghci
GHCi, version 9.6.6: https://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/phadej/.ghci
ghci> safe x = x
ghci>
```

nothing wrong here.

hs-bindgen/src-internal/HsBindgen/Hs/NameMangler/DSL/ReservedNames.hs
```diff
@@ -90,34 +90,37 @@ haskellKeywords =
...
+    -- , "safe"
     , "static"
-    , "stdcall"
+    -- , "stdcall"
     , "stock"
-    , "unsafe"
-    , "using"
-    , "via"
+    -- , "unsafe"
+    -- , "using"
```
TravisCardwell, line 122: Name `using` results in a parse error when extension `TransformListComp` is enabled.

TravisCardwell:
Great!

These keywords are added by extensions, so they are only problematic when the relevant extension is enabled.  If we want to test them, perhaps it would be better to use the TH interface and enable the relevant extensions?

Here is a test module that enables extensions that use these keywords: [Experiment.hs.gz](https://github.com/user-attachments/files/19727437/Experiment.hs.gz).  I do not know what extension uses `label`.

Here is a test module that enables many extensions (including deprecated extensions, tested with GHC 9.12.2): [ExperimentAll.hs.gz](https://github.com/user-attachments/files/19727438/ExperimentAll.hs.gz).

Names that result in failure:

* `by` (`TransformListComp`)
* `forall`
* `mdo` (`RecursiveDo`)
* `pattern` (`PatternSynonyms`)
* `proc` (`Arrows`)
* `rec` (`RecursiveDo`)
* `static` (`StaticPointers`)
* `using` (`TransformListComp`)

Names that do not result in failure:

* `anyclass`
* `capi`
* `cases`
* `ccall`
* `dynamic`
* `export`
* `family`
* `group`
* `interruptible`
* `javascript`
* `label`
* `prim`
* `role`
* `safe`
* `stdcall`
* `stock`
* `unsafe`
* `via`

edsko (changes requested):
Names that _never_ result in problems (@TravisCardwell 's second list in https://github.com/well-typed/hs-bindgen/pull/573#issuecomment-2800142257) are fine to remove from the list of reserved names, but I think the names in his _first_ list, which cause problems if particular extensions are enabled, we should continue to treat as reserved and escape when they are used. It would be confusing I think if the bindings we generate (especially in TH mode) suddenly change because the user enabled an extension.

edsko:
I'll make the change.

edsko (approved):

PR #586, opened 2025-04-18
Change ext bindings `.yaml` ordering

edsko:
This now matches the Haskell convention `package.module.identifier`.

### DIFF
--- a/hs-bindgen/fixtures/adios.extbindings.yaml
+++ b/hs-bindgen/fixtures/adios.extbindings.yaml
@@ -1,11 +1,11 @@
 types:
-- cname: adiós
-  headers: adios.h
+- headers: adios.h
+  cname: adiós
+  package: example
+  module: Example
   identifier: Adio'0301s
-  module: Example
+- headers: adios.h
+  cname: 数字
   package: example
-- cname: 数字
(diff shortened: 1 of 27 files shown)


edsko:
Abstracted out the calls to `Yaml.encode`, and renamed `Config` to `ConfigFile` to avoid confusion with the _yaml_ config, with corresponding functions `fromConfigFile`/`toConfigFile` (instead of ` mkUnresolvedExtBindings` / `encodeUnresolvedExtBindings`, the latter of which I found confusingly named, as it does not do any encoding as such).

TravisCardwell:
I just tried to change the key ordering in the `base` and `hs-bindgen-runtime` bindings, and I discovered that Dhall re-orders the keys when generating YAML.  ([Issue](https://github.com/dhall-lang/dhall-haskell/issues/1187))

What shall we do?  Ideas include:

* Not worry about it
* Remove the Dhall versions and maintain the YAML by hand
* Replace the Dhall with Haskell scripts that are easy to maintain and generate the desired key ordering

edsko:
I think if the input is Dhall, then the yaml output is essentially just a build artifact, and so the ordering does not really matter?

PR #590, opened 2025-04-18
test-pp.hs: don't use a symlink

sheaf:
This avoids usage of symlinks in `test-pp.hs`. The workaround in the cabal file is perhaps a bit unsavoury but it seems to work.

### DIFF
--- a/hs-bindgen/hs-bindgen.cabal
+++ b/hs-bindgen/hs-bindgen.cabal
@@ -321,7 +321,7 @@ test-suite test-th
 
 test-suite test-pp
   type:           exitcode-stdio-1.0
-  main-is:        test-pp.hs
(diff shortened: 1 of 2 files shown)


hs-bindgen/hs-bindgen.cabal
```diff
@@ -321,7 +321,7 @@ test-suite test-th
 
 test-suite test-pp
   type:           exitcode-stdio-1.0
-  main-is:        test-pp.hs
+  main-is:        ../th/test-th.hs
```
phadej, line 324: > The workaround in the cabal file is perhaps a bit unsavoury but it seems to work.

Add a comment or I'll reintroduce the symlink once I forget why there is a HACK.

PR #594, opened 2025-04-18
Preserve source order for `Enum` names

edsko:
The first name declared in the enum is now the first name in the list. This is not always the right choice (in some cases this will result in a suboptimal `Show` instance) but it's probably the less surprising choice.


edsko:
cc @TravisCardwell

PR #605, opened 2025-04-22
Rename to avoid confusion with `Storable`

TravisCardwell:
`HasStaticSize` is renamed to `StaticSize`, and the methods are renamed to `staticSizeOf` and `staticAlignment`.

`Peekable` is renamed to `ReadRaw`, and the method is renamed to `readRaw`.  Function `peekByteOff` is renamed to `readRawByteOff`, function `peekElemOff` is renamed to `readRawElemOff`, and function `maybePeek` is renamed to `maybeReadRaw`.

`Pokable` is renamed to `WriteRaw`, and the method is renamed to `writeRaw`.  Function `pokeByteOff` is renamed to `writeRawByteOff`, and function `pokeElemOff` is renamed to `writeRawElemOff`.

The instances for the `base` types are *already* defined in the module. I had forgotten that.

~The instances that we define in `LibC` require `peekByteOff` and `pokeByteOff` shims in order to use `hsc2hs`.~

Note that `ReadRawWithCtxt` is *not* implemented yet.

### DIFF
--- a/hs-bindgen-runtime/src/HsBindgen/Runtime/LibC.hsc
+++ b/hs-bindgen-runtime/src/HsBindgen/Runtime/LibC.hsc
@@ -50,7 +50,7 @@ import Data.Ix (Ix)
 import Data.Word (Word16, Word32)
 import Foreign.C.Types qualified as C
 import Foreign.Ptr (Ptr)
-import Foreign.Storable (Storable)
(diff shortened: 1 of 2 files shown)


TravisCardwell:
Fixed `Storable` import, rebased

PR #610, opened 2025-04-24
Bugs

edsko:
This example exposes _two_ bugs:

* `UINT8_T` for some reason is defined as _signed_ rather than unsigned
* Our `BOOLEAN_T` wrapper uses `unsigned char` instead of `UINT8_T`


edsko:
Fix for the first bug: https://github.com/well-typed/hs-bindgen/pull/611 . Thanks @sheaf !

sheaf:
Both bugs are fixed as of e8a7592. The test case from this MR is in `macro_types.h` so we can close this.

Issue #568, opened 2025-04-10
Sort generated definitions by C source location


edsko:
To improve the understandability/readability of the generated code, it would be helpful if the order matched the order in the C headers.

phadej:
Aren't they modulo some reordering in case of forward declarations (and there are various options for "original" order).

Issue #558, opened 2025-04-09
Keep track of available instances


edsko:
Given

```c
struct person {
  char* name;

  union {
    struct student {
      char* university;
      int year;
    } student;

    struct employee {
      char* company;
      struct person* supervisor;
      int salary;
    } employee;
  } occupation;
};
```

we currently generate

```haskell
data Person = Person { .. }
  deriving stock (Show, Eq)
newtype Person_occupation = Person_occupation ByteArray
```

but we do not derive `Show` and `Eq` for `Person_occupation`, so this fails to compile.

This ticket is the generalization of #484.

PR #612, opened 2025-04-25
Track available instances

TravisCardwell:
By tracking which types have which instances, we are able to check which instances are available defining or deriving new instances.  For example, a `data` or `newtype` can only derive an `Eq` instance if all fields have `Eq` instances.

Currently a declaration is only generated when all required instances are available.  In a future PR, we will instead generate declarations that have constraints for any missing instances.  For example, a user may configure `hs-bindgen` to *not* generate a `Show` instance for some type `A` because they want to implement a custom instance.  Some type `B` that uses `A` can still have a generated `instance Show A => Show B`, however, so that the user can implement the `Show A` instance but not need to implement the `Show B` instance.

High-level notes:

* External bindings now include instances.  To support this, relevant `Hs` AST nodes now include a set of instances.

* Getter and setter functions for a union field are now only defined if the field type has a `Storable` instance.

* Optional instances for `newtype` wrappers translated from `typedef`s may now be generated with any wrapped type, not just primitives.

* For macros, we only generate a `newtype` for primitive types.  Experimenting with a non-primitive type (`#define I int64_t`) results in a parse error, and code is generated without a `newtype`.  I did not change this behavior.

* Type classes `StaticSize`, `ReadRaw`, and `WriteRaw` are added because the `base` and `hs-bindgen-runtime` external bindings use them.  Generation of instances during translation is *not* added yet, however.

Implementation notes:

* Aside from external bindings changes, this PR is almost entirely implemented within `HsBindgen.Hs.Translation`.

* The signature for top-level function `generateDeclarations` is unchanged, and the `State` monad is used to track instances internally.  The type of the state is as follows.

    ```haskell
    type InstanceMap = Map (HsName NsTypeConstr) (Set HsTypeClass)
    ```

    I initially used the `RWS` monad but decided against it because it complicated scoping.

* The implementation strictly separates code that does not depend on state, code that depends on state but cannot modify it, and code that may modify the state.  Example:

    ```haskell
    -- | Generate declarations for given C struct
    structDecs :: forall n m.
         (SNatI n, State.MonadState InstanceMap m)
      => TranslationOpts
      -> NameMangler
      -> C.Struct
      -> Vec n C.StructField
      -> m [Hs.Decl]
    structDecs opts nm struct fields = do
        (insts, decls) <- aux <$> State.get
        State.modify' $ Map.insert structName insts
        return decls
      where
        declPath :: C.DeclPath
        declPath = C.structDeclPath struct

        ...

        -- everything in aux is state-dependent
        aux :: InstanceMap -> (Set HsTypeClass, [Hs.Decl])
        aux instanceMap = (insts,) $
            structDecl : storableDecl ++ optDecls ++ hasFlamDecl
          where
            insts :: Set HsTypeClass
            insts = getInstances instanceMap structName candidateInsts $
              Hs.fieldType <$> Vec.toList structFields

        ...
    ```

    Everything in the `where` clause up until `aux` does not depend on state.  Everything defined under `aux` depends on state (exposed as `instanceMap`) but cannot modify it.  Only the (minimal) code in the `do` block makes use of the monad to get and modify the state.  I am *not* a fan of nested `where` clauses, but I think that this is a case where it is worthwhile.

* Auxiliary function `getInstances` computes intersections of sets of type classes, looking up type references in the current state.  It needs to know the type being processed in order to handle recursive types.

    ```haskell
    getInstances ::
       InstanceMap         -- ^ Current state
    -> HsName NsTypeConstr -- ^ Name of current type
    -> Set HsTypeClass     -- ^ Candidate instances
    -> [HsType]            -- ^ Dependencies
    -> Set HsTypeClass
    ```

    If a type is not found in the instance map, the function fails with `panicPure`.  This has not happened yet, across our existing tests.

I tested the example in #558 and confirmed that the `Eq` and `Show` instances are no longer derived.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/ExtBindings.hs
+++ b/hs-bindgen/src-internal/HsBindgen/ExtBindings.hs
@@ -2,6 +2,7 @@ module HsBindgen.ExtBindings (
     -- * Types
     HsModuleName(..)
   , HsIdentifier(..)
+  , HsTypeClass(..)
   , ExtIdentifier(..)
   , UnresolvedExtBindings(..)
   , ExtBindings(..)
@@ -34,7 +35,7 @@ module HsBindgen.ExtBindings (
 import Control.Applicative
 import Control.Exception (Exception(displayException))
 import Control.Monad ((<=<))
-import Data.Aeson ((.=), (.:))
+import Data.Aeson ((.=), (.:), (.:?), (.!=))
 import Data.Aeson qualified as Aeson
 import Data.Aeson.Types qualified as Aeson
 import Data.ByteString (ByteString)
@@ -50,6 +51,7 @@ import Data.Typeable
 import Data.Yaml qualified as Yaml
 import Data.Yaml.Internal qualified
 import Data.Yaml.Pretty qualified as Yaml.Pretty
+import Text.Read (readMaybe)
 
 import Clang.Args
 import Clang.CNameSpelling
@@ -82,10 +84,51 @@ newtype HsIdentifier = HsIdentifier { getHsIdentifier :: Text }
   -- 'Show' instance valid due to 'IsString' instance
   deriving newtype (Aeson.FromJSON, Aeson.ToJSON, Eq, IsString, Ord, Show)
 
+-- | Haskell type class
+data HsTypeClass =
+    -- Haskell98 derivable classes
+    -- <https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/deriving.html>
+    Eq
+  | Ord
+  | Enum
+  | Ix
+  | Bounded
+  | Read
+  | Show
+
+    -- Classes we can only derive through newtype deriving
+  | Bits
+  | FiniteBits
+  | Floating
+  | Fractional
+  | Integral
+  | Num
+  | Real
+  | RealFloat
+  | RealFrac
+
+    -- Classes we can only generate when all components have instances
+  | StaticSize
+  | ReadRaw
+  | WriteRaw
+  | Storable
+  deriving stock (Eq, Generic, Ord, Read, Show)
+
+instance Aeson.FromJSON HsTypeClass where
+  parseJSON = Aeson.withText "HsTypeClass" $ \t ->
+    let s = Text.unpack t
+    in  case readMaybe s of
+          Just htc -> return htc
+          Nothing  -> Aeson.parseFail $ "unknown type class: " ++ s
+
+instance Aeson.ToJSON HsTypeClass where
+  toJSON = Aeson.String . Text.pack . show
+
 -- | External identifier
 data ExtIdentifier = ExtIdentifier {
       extIdentifierModule     :: HsModuleName
     , extIdentifierIdentifier :: HsIdentifier
+    , extIdentifierInstances  :: Set HsTypeClass
     }
   deriving stock (Eq, Generic, Ord, Show)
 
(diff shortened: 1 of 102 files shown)


hs-bindgen/src-internal/HsBindgen/Errors.hs
```diff
@@ -85,3 +87,11 @@ panicPure msg = throw (PanicException callStack msg)
...
 panicIO :: (HasCallStack, MonadIO m) => String -> m a
 panicIO msg = liftIO (throwIO (PanicException callStack msg))
+
+-------------------------------------------------------------------------------
+-- MonadFail
+-------------------------------------------------------------------------------
+
+-- | Fail Aeson parsing
+aesonFail :: String -> Aeson.Parser a
+aesonFail = fail
```
phadej, line 97: Remove. Use https://hackage.haskell.org/package/aeson-2.2.3.0/docs/Data-Aeson-Types.html#v:parseFail

hs-bindgen/src-internal/HsBindgen/Hs/Translation.hs
```diff
@@ -133,20 +139,165 @@ generateDeclarations opts nm (C.Header decs) =
...
-  C.DeclTypedef d      -> typedefDecs opts nm d
-  C.DeclMacro m        -> macroDecs opts nm m
-  C.DeclFunction f     -> functionDecs opts typedefs nm f
+generateDecs ::
+     State.MonadState InstanceMap m
+  => TranslationOpts
+  -> NameMangler
+  -> Map C.CName C.Type
+  -> C.Decl
+  -> m [Hs.Decl]
```
phadej, line 291: Does this state approach work for mutually recursive definitions (using forward declarations?)

TravisCardwell: This would be a problem if mutually recursive definitions were used without pointers.  I do not know of any valid code that would test this, and none of our examples do.  (The `recursive_struct.h` example uses a pointer.)  My attempts result in "field has incomplete type" (Clang) errors.

If it is possible, this is the kind of thing that I expect would trigger the `panicPure` mentioned in the top description.  Translation from `C` to `Hs` is done using a single traversal of the `C` declarations, so the state only has data for types that have already been processed and therefore would not work with mutually-recursive declarations that require forward declarations (yet do not use pointers).  This limitation will be resolved when I implement instance constraints (soon).

phadej: If it works even with `typedef`s "hiding" the pointer, then it's ok.

phadej:
```c
typedef struct bar bbar;

struct foo;

struct bar {
  struct foo *f;
};

struct foo {
  struct bar b;
};
```

results into 

```haskell
data Foo = Foo
  { foo_b :: Bar
  }

data Bar = Bar
  { bar_f :: F.Ptr Foo
  }
```

without typedef, `data Bar` would be first.

The first order won't work with single-pass state approach AFAICT. So either we should make the approach order-invariant (resilient to order changes) or make C construction more precise, so structs come in "C definition" order, even if there are typedefs or forward-declarations which shuffle it.

The important bit here, that this is very subtle, so a good note inside the code should be written, so future refactoring don't break invariant if we go for the latter approach.

TravisCardwell:
Thank you very much for the example and guidance!

I realized that the following also results in `Foo` before `Bar`.

```c
struct bar;

struct foo;

struct bar {
  struct foo *f;
};

struct foo {
  struct bar b;
};
```

Regarding ordering the C declarations, I have thought about this in the context of #568, about sorting declarations to improve the readability of the generated code.  We could sort declarations by source path (topologically ordered), source line, and source column.  This would put `Bar` before `Foo` in this case, but it would also put nested declarations later, which would not work with the current implementation.

Instance constraints would resolve the issue.  I think it would be acceptable with the current order of declarations, because constraints that are technically not necessary would only be added in (relatively rare) cases where forward declarations or `typedef`s preturb the declaration order.  I do not think it would be acceptable with declarations ordered as described above, however.

I therefore think that we should indeed implement order-invariant translation, so that it works regardless of how we resolve #568.  It must resolve instances regardless of the order of declarations, but the translation must maintain the order so that generated code reflects the order of the C source.  I have an idea for how to do this and will give it a try.  It depends on some new code also needed for #568, so I am implementing that one first (#620).

edsko:
The ordering problem will be fixed by #620, specifically https://github.com/well-typed/hs-bindgen/pull/620#discussion_r2068160178 .

hs-bindgen/src-internal/HsBindgen/ExtBindings.hs
```diff
@@ -82,10 +84,51 @@ newtype HsIdentifier = HsIdentifier { getHsIdentifier :: Text }
...
+  | Real
+  | RealFloat
+  | RealFrac
+
+    -- Classes we can only generate when all components have instances
+  | StaticSize
+  | ReadRaw
+  | WriteRaw
+  | Storable
+  deriving stock (Eq, Generic, Ord, Read, Show)
```
edsko, line 115: This doesn't feel like the right module for this to be defined, but we can reconsider this in a separate PR.

TravisCardwell: I agree.  In the (WIP) binding specs branch, the following types would best be defined separately, in module(s) that do not depend on any other `hs-bindgen` modules:

* `ExtType`
* `HsRef`
* `HsModuleName`
* `HsIdentifier`
* `HsTypeClass`
* `DeriveStrategy`

When faced with a similar issue before, we ended up creating `Clang.Paths` and `Clang.CNameSpelling` modules for similarly shared types.

It would be convenient to do this after the binding specs work has been merged (or as part of it).

PR #620, opened 2025-04-30
Sort by C source location

TravisCardwell:
1. Source path, topological order of the include graph
2. Line number
3. Column number

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/C/Parser.hs
+++ b/hs-bindgen/src-internal/HsBindgen/C/Parser.hs
@@ -14,6 +14,7 @@ module HsBindgen.C.Parser (
 import Control.Exception
 import Data.List qualified as List
 import Data.List.Compat ((!?))
+import Data.Map.Strict qualified as Map
 import Data.Maybe qualified as Maybe
 import Data.Text qualified as Text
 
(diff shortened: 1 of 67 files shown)


hs-bindgen/src-internal/HsBindgen/C/Parser.hs
```diff
@@ -159,3 +160,39 @@ getTargetTriple args =
...
     opts :: BitfieldEnum CXTranslationUnit_Flags
     opts = bitfieldEnum []
+
+{-------------------------------------------------------------------------------
+  Auxiliary functions
+-------------------------------------------------------------------------------}
+
+-- | Sort declarations by source location
+--
+-- 1. Source path, in specified order
```
edsko, line 170: Should this be something other than topological? (Unsure.)

TravisCardwell: We discussed and decided to sort by dependency primarily and source location (as implemented in the initial version of this PR) secondarily.  The order of generated code should be easily understandable, and the current implementation of instance tracking (#612) should work as-is.  I will implement this and update this PR.

edsko: Discussed this over meet. We will opt for a slightly different ordering, which will be "good enough" for an intuitive (roughly matching the C header) ordering of the Haskell definitions but provides for a more convenient ordering internally, avoiding further ordering later.

TravisCardwell:
(Last commit message)

The C declarations are generally sorted by source location, but dependencies of `struct`, `union`, and `typedef` types are now put before the relevant declaration.

Caveats:

* This only applies to types.  Function and variable declarations do not affect dependency ordering.
* Dependencies behind pointers (including function pointers) are ignored.  This is necessary to avoid cycles in the dependency graph.

Additional notes:

* This is almost entirely implemented in `HsBindgen.C.Parser` function `sortDecls` (and functions).  It is implemented step-by-step in the `where` clause, with documentation.  Context-free auxiliary functions are implemented separately to emphasize that they are context-free.
* The `DynGraph` insertion order is significant, as it affects how the spanning forest of the graph is constructed.  Comments are updated to reflect this.
* An `Ord` instance is added for `C.Type` because it is used as a `Map` key type.

TravisCardwell:
PR #626 should be merged before this PR.  This PR will then be ready after a rebase.

PR #626, opened 2025-05-01
Include `DeclPath` in `OpaqueStruct` / `OpaqueEnum`

TravisCardwell:
Opaque types `OpaqueStruct` and `OpaqueEnum` only included the `CName` because there was no need for a full `DeclPath`.  When taking dependencies into account when sorting (#620), however, we need the full `DeclPath` for dependency type resolution.

### DIFF
--- a/hs-bindgen/fixtures/manual_examples.hs
+++ b/hs-bindgen/fixtures/manual_examples.hs
@@ -2515,8 +2515,17 @@
       emptyDataOrigin =
       EmptyDataOriginOpaqueStruct
         OpaqueStruct {
-          opaqueStructTag = CName
-            "person",
+          opaqueStructDeclPath =
+          DeclPathName
+            (CName "person")
+            (DeclPathCtxtPtr
+              (DeclPathCtxtField
+                (Just (CName "employee"))
+                (CName "supervisor")
+                (DeclPathCtxtField
+                  (Just (CName "occupation"))
+                  (CName "employee")
+                  DeclPathCtxtTop))),
           opaqueStructAliases = [],
           opaqueStructSourceLoc =
           "manual_examples.h:77:12"}},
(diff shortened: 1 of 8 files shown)


TravisCardwell:
Example from `manual_examples.h` ([code](https://github.com/well-typed/hs-bindgen/blob/1fede8055e95243ecd835eef5f2da6bf43ed3adc/hs-bindgen/examples/manual_examples.h#L77)):

```c
typedef union occupation {
  struct student {
    char* university;
    int year;
  } student;

  struct employee {
    char* company;
    struct person* supervisor;
    int salary;
  } employee;
} occupation;
```

In this example, `struct person` is opaque; it is not defined anywhere.  The `DeclPath` in the type for that field has context ([code](https://github.com/well-typed/hs-bindgen/blob/1fede8055e95243ecd835eef5f2da6bf43ed3adc/hs-bindgen/fixtures/manual_examples.tree-diff.txt#L446-L463)):

```haskell
StructField {
  fieldName = CName "supervisor",
  fieldOffset = 64,
  fieldWidth = Nothing,
  fieldType = TypePointer
    (TypeStruct
      (DeclPathName
        (CName "person")
        (DeclPathCtxtPtr
          (DeclPathCtxtField
            (Just (CName "employee"))
            (CName "supervisor")
            (DeclPathCtxtField
              (Just (CName "occupation"))
              (CName "employee")
              DeclPathCtxtTop))))),
            fieldSourceLoc =
            "manual_examples.h:77:20"},
```

The opaque structure declaration does not include context, however ([code](https://github.com/well-typed/hs-bindgen/blob/1fede8055e95243ecd835eef5f2da6bf43ed3adc/hs-bindgen/fixtures/manual_examples.tree-diff.txt#L416-L422)):

```haskell
OpaqueStruct {
  opaqueStructTag = CName
    "person",
  opaqueStructAliases = [],
  opaqueStructSourceLoc =
  "manual_examples.h:77:12"},
```

When traversing the dependency graph, the type of the field (including the context) does not match the type of the opaque structure (lacking the context).

edsko:
We discussed this, and concluded this is is a fix "in the wrong direction": rather than adding more information to the `DeclPath`, we should instead add _less_; replacement PR in #627.

PR #628, opened 2025-05-02
Nix development environment

dschrempf:
Build and tests pass in the Nix development shell. If you prefer to not have Nix-related code in the repo, I can amend the `.gitignore` file and use the files locally.

I also highlighted the version of `rust-bindgen` used by tests.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,68 @@
+{
+  description = "hs-bindgen development environment";
+
+  inputs.flake-utils.url = "github:numtide/flake-utils";
+
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+
+  outputs =
+    {
+      self,
+      flake-utils,
+      nixpkgs,
+    }:
+    flake-utils.lib.eachDefaultSystem (
+      system:
+      let
+        pkgs = import nixpkgs {
+          inherit system;
+          overlays = [
+            (final: prev: {
+              # See ./test/internal/Test/Internal/Rust.hs::rustBindgenVersion.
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -0,0 +1,68 @@
...
+            hpkgs.ghc
+            hpkgs.haskell-language-server
+
+            # Other.
+            lpkgs.clang
+            lpkgs.llvm
+            pkgs.rust-bindgen
+          ];
+          shellHook = ''
+            # When running `clang` during GHC compilation with Template Haskell,
```
phadej, line 56: Technically we don't run `clang`. 

And https://github.com/well-typed/hs-bindgen/issues/366#issuecomment-2655630980 comment is related. I haven't followed it, so cannot say whether a proper issue should be created out of it. @TravisCardwell can probably elaborate.

dschrempf: Do you mean, when we "access the `libclang` interface" instead of "run `clang`"?

dschrempf: (And I guess that's why the wrapper does not work, it only wraps the executable, and probably not the API)?

TravisCardwell: Yes.  We use the `libclang` API.  We do not run the `clang` executable.  I understand that you have it working now, and I indeed see `libclang` as an input in the [current flake](https://github.com/dschrempf/hs-bindgen-flake/blob/main/flake.nix).

In the past, we considered calling the `clang` executable to query the include search path, but we decided to *not* do this.  Users can specify directories to add, which we do via Clang arguments, but we do not know the (full) include search path from within `hs-bindgen`.

PR #630, opened 2025-05-02
Provide location hint, even if not precise

edsko:
We don't always have a `CXCursor` available, which means we don't always have a location. There _is_ a way to get a `CXCursor` from a `CXType` (`clang_getTypeDeclaration`) but it (often? sometimes?) results in a "zero" location. In this PR we therefore accept that we don't always have a precise location, but when we don't, we give a location in terms of another location ("this is the underlying type of the type at location such and such").


edsko:
Waiting for @FinleyMcIlwaine to give this a go, to see if it actually helps us pinpoint down the problem he was seeing.

PR #632, opened 2025-05-02
Make `clang` bindings polymorphic over `m`

edsko:
All of the `clang` functions worked in the concrete `IO` monad, but we never actually _use_ it in `IO`, resulting in a lot of unnecessary visual noise.

### DIFF
--- a/clang/clang.cabal
+++ b/clang/clang.cabal
@@ -97,10 +97,11 @@ library
       Clang.HighLevel.Wrappers
   build-depends:
       -- External dependencies
-    , filepath      >= 1.4     && < 1.6
-    , mtl           >= 2.2     && < 2.4
-    , text          >= 1.2     && < 2.2
-    , unliftio-core ^>=0.2.1.0
+    , exceptions    >= 0.10  && < 0.11
+    , filepath      >= 1.4   && < 1.6
+    , mtl           >= 2.2   && < 2.4
+    , text          >= 1.2   && < 2.2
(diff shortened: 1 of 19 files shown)


sheaf:
My opinion is that `foo :: ... -> IO a` is much better than `foo :: MonadIO m => ... -> m a`.

My main argument is that now all functions take an abstract `MonadIO` dictionary, so the entire codebase needs to be specialised at the `IO` type to recover the specialised functions that don't take any typeclass dictionaries.

clang/clang.cabal
```diff
@@ -97,10 +97,11 @@ library
...
       -- External dependencies
-    , filepath      >= 1.4     && < 1.6
-    , mtl           >= 2.2     && < 2.4
-    , text          >= 1.2     && < 2.2
-    , unliftio-core ^>=0.2.1.0
+    , exceptions    >= 0.10  && < 0.11
+    , filepath      >= 1.4   && < 1.6
+    , mtl           >= 2.2   && < 2.4
+    , text          >= 1.2   && < 2.2
+    , unliftio-core >= 0.2.1 && < 0.3
```
phadej, line 104: ??

phadej:
I agree with @sheaf. :-1:

edsko:
Virtually _every. single. call_. into our `clang` bindings is of the form `liftIO $ clang_foo`

image

and so forth. This PR just moves that into the `clang` bindings itself, so that at the use sites, the code becomes more readable.

Issue #613, opened 2025-04-25
`CEnum` `Read` instances


TravisCardwell:
We already generate `Show` instances.  Perhaps we should generate `Read` instances as well?  It would be nice if it could read *all* of the patterns as well as `Constructor NUM` strings, including declared and undeclared values.

TravisCardwell:
I prefer to avoid both `Show` and `Read` in the vast majority of cases, only using `Show` for debugging and almost never using `Read`.  Enumerations are an exception, however, when the textual representation of a value is the Haskell constructor name.  In this case, `Show` and `Read` provide zero-maintenance encoding and decoding.  For example, they may be used in Aeson `ToJSON` and `FromJSON` instances.

PR #634, opened 2025-05-05
readPrec for Read instances of `CEnum a`

dschrempf:
See issue https://github.com/well-typed/hs-bindgen/issues/613; this implements the first part of the solution excluding code generation.

- Add a `readPrecUndeclared` method to the `CEnum` type class

- Add `readPrecWrappedUndeclared` wrapper function which allows easy definition of `readPrecUndeclared`

- Define `readPrecCEnum` in terms of `readPrecUndeclared`; `readPrecCEnum` allow easy definition of `Read` instances

- Add unit and property tests. The parsers either fails completely (i.e., does not consume input), or succeeds.

- Define and use proper `Show` instances of test cases.


dschrempf:
Superseded by #636.

PR #635, opened 2025-05-05
Use upstream install-llvm-action version

phadej:
Now as it has llvm-20 support


.github/workflows/simple.yml
```diff
@@ -50,7 +50,7 @@ jobs:
 
       - name: Install LLVM and Clang
         # Use our fork for time being to get latest LLVMs
-        uses: phadej/install-llvm-action@v3
+        uses: KyleMayes/install-llvm-action@v2
```
dschrempf, line 53: I guess we should also remove the comment.

PR #636, opened 2025-05-05
CEnum: Read instance with code generation

### LINKED ISSUE #613: `CEnum` `Read` instances

dschrempf:
Closes #613.

CEnum:
- Add a `readPrecUndeclared` method to the `CEnum` type class
- Add `readPrecWrappedUndeclared` wrapper function which allows easy definition of `readPrecUndeclared`
- Define `readPrecCEnum` in terms of `readPrecUndeclared`; `readPrecCEnum` allow easy definition of `Read` instances
- Add unit and property tests. The parsers either fails completely (i.e., does not consume input), or succeeds.
- Define and use proper `Show` instances of test cases.

Code generation:
- Generate relevant code in `hs-bindgen`.
- `Text.Read` now maps to `Text.Read` (although TH resolves it to `GHC.Read`). Only the `Read` type class resolves to the Prelude.
- Accept golden tests (please review!).
- I can not guarantee that the generated code works, but I checked it manually (the `Text.Read` imports seem to be OK).

This PR does not yet add Read-related stuff to the manual.

### DIFF
--- a/hs-bindgen-runtime/src/HsBindgen/Runtime/CEnum.hs
+++ b/hs-bindgen-runtime/src/HsBindgen/Runtime/CEnum.hs
@@ -16,6 +16,8 @@ module HsBindgen.Runtime.CEnum (
   , showCEnum
   , showsCEnum
   , showsWrappedUndeclared
+  , readPrecCEnum
+  , readPrecWrappedUndeclared
   , seqIsDeclared
   , seqMkDeclared
     -- * Deriving via support
@@ -34,7 +36,12 @@ import Data.List.NonEmpty qualified as NonEmpty
 import Data.Map.Strict (Map)
 import Data.Map.Strict qualified as Map
 import Data.Proxy (Proxy(Proxy))
-import GHC.Show (appPrec1, showSpace)
+import GHC.Show (appPrec1, showSpace, appPrec)
+import Text.Read (ReadPrec, (+++))
+import Text.Read.Lex (Lexeme (..), expect)
+import Text.Read qualified as Read
+import Text.ParserCombinators.ReadP qualified as ReadP
+import Text.ParserCombinators.ReadPrec qualified as ReadPrec
 
 {-------------------------------------------------------------------------------
   Type classes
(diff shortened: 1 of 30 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/PP/Names.hs
```diff
@@ -101,6 +102,22 @@ importQ name = ResolvedName
   where
     s = TH.nameBase name
 
+-- TODO!
+-- - TH maps the module 'Text.Read' to 'GHC.Read';
+-- - 'moduleOf' maps 'GHC.Read' to 'Prelude';
+-- - 'readPrec', ... are not in 'Prelude'.
+--
+-- For now we just ignore 'moduleOf', but that's not a permanent solution.
+importQTemp :: TH.Name -> ResolvedName
```
dschrempf, line 111: Please note! Is there a better way to solve the "GHC.Read" problem?

dschrempf:
(I pushed a tiny fix which moves the parsing of parentheses into the wrapper function).

dschrempf:
I have rebased the PR after /Instance tracking/ was merged. I have added the Hs.Read instance to the tracked instances (because it is not anymore stock derived). Please check!

hs-bindgen-runtime/src/HsBindgen/Runtime/CEnum.hs
```diff
@@ -151,44 +164,51 @@ class CEnum a => SequentialCEnum a where
...
 {-------------------------------------------------------------------------------
   Instance support
 -------------------------------------------------------------------------------}
 
--- | Declared values and associated names (opaque)
-newtype DeclaredValues a = DeclaredValues
-    { unDeclaredValues :: Map (CEnumZ a) (NonEmpty String)
+-- | Declared values (opaque)
+data DeclaredValues a = DeclaredValues
+    { declaredValuesList :: ![String]
```
edsko, line 175: A bang on a list is a red flag to me (it forces only the head of the list, which is almost certainly not what we want). To avoid delaying this PR further I'll just fix this myself, by removing this field altogether and adding

```haskell
declaredValuesList :: DeclaredValues a -> [String]
declaredValuesList = Map.keys . declaredValueToIntegral
```

edsko:
Nice work, good to have you on board :)

PR #641, opened 2025-05-08
Read instance: Manual section and some other girl scout changes

dschrempf:
- I also updated the generated code.
- Also add `readEitherCEnum` for consistency with `showCEnum`.


manual/LowLevel/Enums.md
```diff
@@ -247,7 +250,74 @@ showCursorKind :: CXCursorKind -> String
...
+`Read` instances inherit the `Read`-related properties of the underlying
+integral type. In particular, we parse negative integers, even when the integral
+type is unsigned:
+
+```haskell
+read "Index (-1)" :: Index -- Evaluates to 'Index 4294967295'
+```
+
+We can ensure the value is within a given bound by providing a
+`readEitherIndexWith` function like so:
```
edsko, line 307: Nice example :)

PR #642, opened 2025-05-09
Clang extra args: `withExtraClangArgs`  (#640)

### LINKED ISSUE #640: Support `BINDGEN_EXTRA_CLANG_ARGS` and `BINDGEN_EXTRA_CLANG_ARGS_<TARGET>`
`hs-bindgen-cli` supports setting `libclang` arguments using command line arguments. In particular, we can set
- `--system-include-path DIR`,
- `--include-path DIR`,
- `--clang-option` (other options directly passed to `libclang`).

However, we still do not support configuration of `hs-bindgen(-cli)` via environment variables. Configuration via environment variables is relevant for at least three reasons:

- Software repository maintainers may want to configure the behavior of `hs-bindgen-cli`. This applies to NixOS, but maybe also to other systems?
- From the `rust-bindgen` README:
  > End-users should set these environment variables to modify bindgen's behavior without modifying the source code of direct consumers of bindgen.
- We want to configure `hs-bindgen` without changing the source code when running `hs-bindgen` via Template Haskell (`genBindings`).

For reference: [Rust Bindgen supports two environment variables](https://github.com/rust-lang/rust-bindgen?tab=readme-ov-file#environment-variables):
- `BINDGEN_EXTRA_CLANG_ARGS`, and
- `BINDGEN_EXTRA_CLANG_ARGS_<TARGET>`

<details>

Edsko and I had a discussion about where we could put the machinery. We see two options:
- In our `clang` library, we could read the variables in
  - `withTranslationUnit`, and
  - `withTranslationUnit2`.
- However, as the name `BINDGEN_EXTRA_CLANG_ARGS` suggests, we think we should read the variable in `hs-bindgen`. There, we have more entry points to our `clang` interface:
  - `clangAstDump`,
  - `parseCHeaders`,
  - `getTargetTriple`, and
  - `resolveHeader'`.
 
</details>

dschrempf:
Implement and use `withExtraClangArgs` honoring `BINDGEN_EXTRA_CLANG_ARGS` and `BINDGEN_EXTRA_CLANG_ARGS_<TARGET>`.

The environment variables are documented in the optparse-applicative help footer, and the "Invocation" section in the manual.

This was a bit of a rabbit hole. Since we use `libclang` directly, we need to parse the extra arguments and split them honoring shell-escape syntax. (However, `unescapeArgs` in `base` can do that for us).

There are still some open questions:
- [x] Logging. (Right now when reading `BINDGEN_EXTRA_CLANG_ARGS`, we print to `stdout` which is suboptimal, especially when `hs-bindgen` is used as a library). However, I was not sure if I should re-use one of the tracers, or if we still setup a proper logger. -> Seprate issue/PR (#647).
- [x] ~~I tested the parser well, but there may be more edge cases I have missed. (I was surprised I coulnd't find a library function on Hackage doing this for us; do you know one?)~~ We use `unescapeArgs` from `base`.
- [x] Should we add an integration-like test, checking if `BINDGEN_EXTRA_CLANG_ARGS` is acutally used? (It does work here, but in the long run this is desirable). This may also be especially relevant when using other targets. -> Done.
- [x] We document the behavior in the manual, and in the footer of the client(s).
- [x] Rebase on #651.
- [x] Use tracer in `withExtraClangArgs`.

Closes #640

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /.dir-locals.el
 /.direnv/
+/.stylish-haskell.yaml
 
 .envrc
 .vscode/
--- a/hs-bindgen/test/internal/test-internal.hs
+++ b/hs-bindgen/test/internal/test-internal.hs
@@ -3,6 +3,7 @@ module Main (main) where
 
 import Control.DeepSeq (force)
 import Control.Exception (displayException, evaluate, try)
+import Control.Tracer (Tracer)
 import Data.ByteString.UTF8 qualified as UTF8
 import Data.List qualified as List
 import Data.TreeDiff.Golden (ediffGolden1)
@@ -19,6 +20,7 @@ import HsBindgen.Lib
 import HsBindgen.Pipeline qualified as Pipeline
 
 import Test.HsBindgen.C.Parser qualified
+import Test.HsBindgen.Clang.Args qualified
 import Test.Internal.Misc
 import Test.Internal.Rust
 import Test.Internal.TastyGolden (goldenTestSteps)
@@ -35,15 +37,17 @@ import Test.Internal.TH
 main :: IO ()
 main = do
     packageRoot <- findPackageDirectory "hs-bindgen"
-    defaultMain . withRustBindgen $ tests packageRoot
+    withTracerStdOut defaultTracerConf $ \tracer ->
+      defaultMain . withRustBindgen $ tests tracer packageRoot
 
 {-------------------------------------------------------------------------------
   Tests
 -------------------------------------------------------------------------------}
 
-tests :: FilePath -> IO FilePath -> TestTree
-tests packageRoot rustBindgen = testGroup "test-internal" [
-      Test.HsBindgen.C.Parser.tests args
+tests :: Tracer IO (TraceWithCallStack Trace) -> FilePath -> IO FilePath -> TestTree
+tests tracer packageRoot rustBindgen = testGroup "test-internal" [
+      Test.HsBindgen.C.Parser.tests tracer args
+    , Test.HsBindgen.Clang.Args.tests tracer
     , testGroup "examples" [
           golden "simple_structs"
         , golden "recursive_struct"
@@ -170,9 +174,9 @@ tests packageRoot rustBindgen = testGroup "test-internal" [
     mkOpts :: (String -> IO ()) -> Pipeline.Opts
     mkOpts report =
       let tracerConf = defaultTracerConf { tVerbosity = Verbosity Warning }
-          tracer = mkTracer EnableAnsiColor tracerConf report
+          tracerGolden     = mkTracer EnableAnsiColor tracerConf report
       in  opts {
-              Pipeline.optsTracer = tracer
+              Pipeline.optsTracer = tracerGolden
             }
 
     ppOpts :: Pipeline.PPOpts
--- a/hs-bindgen/app/HsBindgen/App/Cli.hs
+++ b/hs-bindgen/app/HsBindgen/App/Cli.hs
@@ -26,6 +26,7 @@ getCli = customExecParser p opts
     opts = info (parseCli <**> helper) $
       mconcat [
           header "hs-bindgen - generate Haskell bindings from C headers"
+        , footerDoc (Just $ environmentVariablesFooter p)
         ]
 
 {-------------------------------------------------------------------------------
--- a/hs-bindgen/app/HsBindgen/App/Common.hs
+++ b/hs-bindgen/app/HsBindgen/App/Common.hs
@@ -7,6 +7,7 @@ module HsBindgen.App.Common (
   , parseInput
     -- * Auxiliary hs-bindgen functions
   , loadExtBindings'
+  , environmentVariablesFooter
     -- * Auxiliary optparse-applicative functions
   , cmd
   , cmd'
@@ -14,12 +15,17 @@ module HsBindgen.App.Common (
 
 import Control.Exception (Exception (displayException))
 import Control.Tracer (Tracer)
-import Data.Bifunctor (first)
+import Data.Bifunctor (Bifunctor (bimap), first)
 import Data.Char qualified as Char
 import Data.List qualified as List
-import GHC.Stack (callStack)
+import Data.Text (Text)
+import Data.Text qualified as Text
+import GHC.Stack (HasCallStack, callStack)
 import Options.Applicative
 import Options.Applicative.Extra (helperWith)
+import Options.Applicative.Help (Doc, align, extractChunk, pretty, tabulate,
+                                 vcat)
+import Prettyprinter.Util (reflow)
 
 import HsBindgen.Lib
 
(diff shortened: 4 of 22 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 /.dir-locals.el
 /.direnv/
+/.stylish-haskell.yaml
```
dschrempf, line 3: I am using a configuration that only handles imports. Is this OK with you?

hs-bindgen/src-internal/HsBindgen/C/Wrapper.hs
```diff
@@ -0,0 +1,83 @@
...
+parseStringArguments :: String -> Either ParseError [String]
+parseStringArguments str = parse arguments "" str
+
+getExtraClangArgs :: MonadIO m => m [String]
+getExtraClangArgs = do
+  extraClangArgsStr <- liftIO $ lookupEnv extraClangArgsEnvName
+  case extraClangArgsStr of
+    Nothing -> pure []
+    Just str -> case parseStringArguments str of
+      -- TODO: Do we have proper logging? Use Tracer?
```
dschrempf, line 36: See main PR comment.

dschrempf:
After our discussion @edsko, I did some research with respect to target-specific extra args. In `rust-bindgen`, they use `BINDGEN_EXTRA_CLANG_ARGS_<TARGET>` only when cross compiling to a specific target (which may be the same architecture as the current machine). When cross compiling, they first try to read the target-specific environment variable, and then `BINDGEN_EXTRA_CLANG_ARGS`.

I have implemented the same behavior. I guess it is good to mimic the behavior of `rust-bindgen`, but let me know @edsko if you disagree!

This behavior makes sense in terms of the environment variable template name (i.e., `BINDGEN_EXTRA_CLANG_ARGS_<TARGET>`). The term `TARGET` implies a cross-compilation environment. However, it doesn't make much sense when looking at the actual environment variable name, e.g., `BINDGEN_EXTRA_CLANG_ARGS_x86_64-pc-linux`, because here we only see the architecture, and not the term `target`, which does not imply cross-compilation...

Relevant function in the `rust-bindgen` codebase: https://github.com/rust-lang/rust-bindgen/blob/97ab9152b5edb1fda1ced9bc1604f5e4dc9cfaa9/bindgen/lib.rs#L1230

dschrempf:
BTW, they use `shlex::split` to do the shell-escaped splitting of arguments.

hs-bindgen/src-internal/HsBindgen/C/Environment.hs
```diff
@@ -0,0 +1,75 @@
+-- | Handle `libclang`-specific environment variables.
+
+module HsBindgen.C.Environment (
```
phadej, line 3: `HsBindgen.C` is a wrong namespace.

This is not C-*AST* related.

hs-bindgen/test/internal/test-internal.hs
```diff
@@ -2,22 +2,23 @@
 module Main (main) where
 
 import Control.DeepSeq (force)
-import Control.Exception (try, displayException, evaluate)
+import Control.Exception (displayException, evaluate, try)
```
phadej, line 5: is it your editor adding those changes?

Please don't. These are  (mostly) noise: they are not related to the change at hand.

If you really must, please put those into separate commits.

hs-bindgen/src-internal/HsBindgen/C/Environment.hs
dschrempf: Can you suggest a better namespace?

hs-bindgen/test/internal/test-internal.hs
dschrempf: Yes, I have configured my editor to auto-format imports which saves a lot of time. I configured the formatting so that it it honors the current way we format our imports. It does sort htings alphabetically though, and I think this is a good thing.

If you insist, I can put changes like these into separate commits.

hs-bindgen/src-internal/HsBindgen/C/Environment.hs
dschrempf: `HsBindgen.Environment`? Would that work?

phadej: This is clang specific, so i'd expect seeing `Clang` (or rather `ClangArgs`) in the module name.

phadej: .. so maybe something like `HsBindgen.Clang.Args` as an extension of `Clang.Args`.

But to be honest, the only `hs-bindgen` specific bit is the environment variable name (and maybe tracer type in the future). Both specificities could be abstracted away.

hs-bindgen/test/internal/test-internal.hs
dschrempf: Actually, I thought about this a bit more, and it is not too much of a hazzle for me to create separate commits if I know this from the beginning. In this PR, the changes are interleaved into the different commits, I only see one easy way out: Squash all commits in this PR into one, and extract the commits ordering existing imports. Would that be OK for you, @phadej ?

hs-bindgen/src-internal/HsBindgen/C/Environment.hs
dschrempf: Ok, thanks for the suggestion. I will move it to `HsBindgen.Clang.Args`. The only dawback I see is that potentially other supported env vars in the future do not have a space there. But we can extend and move the module when the need arises.

hs-bindgen/test/internal/test-internal.hs
dschrempf: I extracted the relevant changes into a separate commit. My opinion though is that such changes should be seen as (girls scout) improvements, and not as changes with merge-conflict potential. I hope the status quo is OK with you!

phadej: TBH, I'm usually more annoyed with such "formatting" changes. If people have editors configured differently, those changes go back and forth, so I'm basically against them unless they are enforced.

We could use `stylish-haskell` and **enforce** its usage in CI, then there won't be style pendulums (and FWIW there are, people format module export lists, and even in this project they have been formatted forward and back).

This is annoying issue.

dschrempf: I understand your point. What about enforcing them then? I have a working `stylish-haskell` setup that **only sorts and formats imports**, and does this in an unintrusive way. In particular, it keeps import groups when they are separated by a newline. `stylish-haskell` comes with HLS, so we can assume everybody has it installed. I could add the check to CI and commit my configuration. What do you think? @edsko @sheaf @TravisCardwell @phadej

sheaf: I had a bad experience with formatters with the Cabal project. The main issue I had was that the formatter would increase the size of diffs because one small change could lead the formatter to decide a whole block of code needed different indentation. This then caused major headaches when rebasing. I don't think this is a concern in this case if it's just about imports/exports.

I also would want to make sure that formatting failures don't block or cancel other CI jobs, as it is a very frustrating workflow when one wants CI results.

edsko (approved):
Feel free to merge once the tracing PR is merged.

PR #652, opened 2025-05-17
Replace frontend

edsko:
This refactors the C folding and processing into separate passes, significantly simplfying the code. This is WIP.

TODO:

- [x] Introduce use-def graph and def-use graph
- [x] Introduce separate pass for parsing
- [x] Introduce separate pass for sorting (uses the use-def graph)
- [x] Introduce separate pass for renaming (uses the def-use graph)
- [x] Support full AST 
    - [x] Structs
    - [x] Forward and opaque struct declarations
    - [x] Typedefs
    - [x] Macros
    - [x] Unions
    - [x] Enums
    - [x] Opaque enums
    - [x] Function declarations
- [x] Parse macros: in the `Parsed` pass, macros are simply a list of tokens, after parsing, split into `MacroType` and `MacroVal` (macro parsing must occur _after_ sorting).
- [x] Support reparsing in the AST: list of tokens _if_ macro expansion has happened in `Parsed`, after `Reparsing` this should be `()`
- [x] Add pass to incorporate binding specifications
- [x] Add pass to for macro reparsing
- [x] Add pass for generation of Haskell names (maybe)
- [x] Add pass for selection
    - [x] Anything with an empty source location should always be omitted (I think?) -- these are clang internal definitions
- [x] Probably more that I haven't thought of yet


edsko:
This is obviously a huge PR, and is the work of three people over multiple weeks, with PRs into this branch.

The `.hs` and `tree-diff.txt` output fixtures are basically unuseable in this PR, as the internal representation has changed. Changes to `.pp.hs` however should be taken seriously. Fortunately there are not too many; I've gone through all of the ones that _have_ changed, and have classified the changes:

* Two tests have been modified to avoid `stdint`. I've opened a new ticket about this (https://github.com/well-typed/hs-bindgen/issues/700).

* A type such `__sFILE` is called `C__sFILE` in Haskell, becausd we cannot turn that underscore into an uppercase letter. We were previously then calling the fields of this type `c__SFILE__<foo>`; we now call then `__sFile__<foo>` (without the `C` prefix), because the underscore is unproblematic for the field name. I think both designs are excusable, this more just fits better with the new way that name mangling works.

* For anonymous structs inside of structs, we previously added a `_Deref` suffix (if the field was a pointer to the anonymous struct). We no longer add this suffix because (unlike with typedefs) there can be no name clashes.

* The order of the declarations is sometimes different: when `A` depends on `B`, then A now always comes first. This sometimes results in confusing diffs on GitHub; for these cases I've reordered the declarations manually and did diff against that to see if anything else changes.

* We now more correctly track when we squash typedefs, so that we keep the information around for generating correct C, so that we for example generate

  ```c
  signed int testmodule_index_triple (struct triple *arg1, enum index arg2) {
    return index_triple(arg1, arg2);
  }
  ```

  instead of

  ```c
  signed int testmodule_index_triple (triple *arg1, index arg2) {
    return index_triple(arg1, arg2);
  }
  ```

  (https://github.com/well-typed/hs-bindgen/issues/667)

Issue #648, opened 2025-05-14
Generate instances with constraints


TravisCardwell:
Currently, an instance for a `data` or `newtype` type is only derived/generated if there are instances for all of the fields.  This is not very user-friendly, however, in cases where users provide instances.  Since user-defined instances are not available when `hs-bindgen` is run, it does not derive/generate any instances that require the user-defined instances, forcing users to do so themselves.

By generating code with constraints, we can derive/generate instances in such cases.  User-defined instances can satisfy the constraints, saving users from having to write code that we can generate.

This was discussed in the context of binding specifications [here](https://github.com/well-typed/hs-bindgen/issues/608#issuecomment-2826591126), with an example.

The following is my current plan.  It shows all of the supported type classes for each C kind.  For each type class, the supported strategies are listed.  The `hs-bindgen` strategy indicates that we generate the instance (not get GHC to derive it).  A strategy annotated with `(default constraints)` means that the instance is derived/generated, and constraints are used for any field type that is not known to have an instance.  Constraints for field types that are known to have instances are *not* added.  A strategy annotated with `(default strict)` means that the instance is only derived/generated when all fields are known to have instances.  In this case, instances with constraints are *not* derived/generated by default.  Since we generate a `newtype` around an integral type for `Enum`s, instances for those types will never have constraints.  A type class where none of the strategies are marked as default is only derived/generated when explicitly requested via binding specifications.  When an instance that is explicitly requested via binding specifications, it is always derived/generated with constraints when necessary.

* `Struct`
    * `Eq`: `stock` (default constraints)
    * `Ord`: `stock`
    * `Show`: `stock` (default constraints)
    * `Storable`: `hs-bindgen` (default constraints)

* `Union`
    * `Storable`: `hs-bindgen` (default)

* `Enum`
    * `Bounded`: `hs-bindgen`, `newtype`
    * `Enum`: `hs-bindgen`, `newtype`
    * `Eq`: `stock` (default)
    * `Ord`: `stock` (default)
    * `Read`: `hs-bindgen` (default), `newtype`, `stock`
    * `Show`: `hs-bindgen` (default), `newtype`, `stock`
    * `Storable`: `hs-bindgen` (default)

* `Typedef`
    * `Bits`: `newtype` (default strict)
    * `Bounded`: `newtype` (default strict)
    * `Enum`: `newtype` (default strict)
    * `Eq`: `stock` (default constraints)
    * `FiniteBits`: `newtype` (default strict)
[...]

Issue #608, opened 2025-04-23
Generalize "external bindings" to "binding specifications"


edsko:
External bindings currently map C identifiers to Haskell identifiers:

```yaml
types:
- headers: simple_structs.h
  cname: S2_t
  module: Example
  identifier: S2_t
```

This will need to be extended with additional information:

- We need to track which instances are available (#558)
- We might need to to know which types are newtypes around other types for the userland CAPI work (and indeed, which types are _aliases_, #308)
- Macro types (#423)
- ...

At the moment we use these only when referring to _existing_ bindings. The main idea of this ticket is to generalize this so that it's also usable when generating _new_ bindings; hence changing the name to binding _specification_. 

We consider the consequences of this generalization for each piece of information separately.

We should have the following consistency property: if we are _generating_ bindings according to a certain specification, then that same specification should be a valid specification _of_ those generated bindings (that is, it should be usable in reference mode). It might be an underspecification; since `hs-bindgen` also _produces_ a binding spec for bindings that it generates, this produced spec should be a strict "superset" of the input specification.

### Haskell module/identifier

```yaml
types:
- headers: simple_structs.h
  cname: S2_t
  haskell:
    module: Example
    identifier: S2_t
```

**Reference mode**: the C identifier is defined in module `Example` and has name `S2_t`.  
**Generation mode**: please call the type you're going to generate `S2_t` and place it in module `Example`.

This would then subsume #564 (support for name overrides) and (part of) #75 (generating multiple Haskell modules from a single C header).

_Absent value_: We should be careful about what it means for sections to be _absent_. For the `haskell` section: In reference mode it currently means "there is no Haskell type for this _yet_, please generate one". In generation mode we probably don't want it to mean "do not generate a Haskell type", but rather "just do the default". It would perhaps be good to have an _explicit_ syntax for "no value", maybe something like

```yaml
types:
- headers: simple_structs.h
  cname: S2_t
  haskell: nil
```

### Instances

```yaml
types:
- headers: simple_structs.h
  cname: S2_t
  instances: [Eq, Ord, Show, Storable]
```

**Reference mode**: The Haskell type corresponding to the C identifier has the specified instances.
**Generation mode**: Please generate instances for the specified classes. 

This therefore subsumes #307 (configure which instances to generate).

[...]

PR #656, opened 2025-05-19
Binding specifications (part 1)

TravisCardwell:
This PR generalizes "external bindings" to "binding specifications" (#608).  Binding specifications are used for both configuration of code generation and external binding specification.  Note that the terminology "external binding" is still used.  For example, `hs-bindgen-cli` now accepts zero or one binding specifications for configuration via the `--binding-specs` option, as well as any number of external binding specifications via the `--external-bindings` option (unchanged).

Internally, `TypeSpec` is the general specification for a type, while `ExtType` is used for external bindings.  Configuration of the module and identifier in a `TypeSpec` is optional, while the module and identifier are required in an `ExtType`.  Also, a `TypeSpec` includes instance specifications, while an `ExtType` contains the set of available instances.

Type `BindingSpecs` is the public type, which is opaque in the public API.  It wraps internal type `IBindingSpecs` in order to hide a type parameter that specifies the header representation.  This is a different design from that of the original external bindings implementation.  Since it is used for configuration, the type will likely grow over time, and it is therefore worthwhile to define it in a single place and parameterize the two possible representations (unresolved and resolved headers).

The JSON/YAML representation is implemented via *internal* auxiliary data types, using an `A` prefix by convention.

The `omit` syntax wraps what is omitted, where the object must have a single item with key `omit`.

```haskell
data AOmittable a = ARequire a | AOmit a

instance Aeson.FromJSON a => Aeson.FromJSON (AOmittable a) where
  parseJSON = \case
    Aeson.Object o | KM.size o == 1 && KM.member "omit" o ->
      AOmit <$> o .: "omit"
    v -> ARequire <$> Aeson.parseJSON v
```

Parsing the wrapped value is required in order to know what is being omitted.  In the exported API, type `Omittable` is isomorphic to `Maybe`.

Type name customization is implemented, aside from types created from macros.

This PR does not include tests yet.  The plan is to add tests as part of a second PR, which needs to be done after instance constraints are implemented (#648).


TravisCardwell:
Here is a summary of the changes to the `C` phase, to help with #652:

* `HsBindgen.C.AST.Type`
    * `TypeExtBinding ExtIdentifier Type` is changed to `TypeExtBinding ExtType Type`; as `ExtIdentifier` is removed.  This specification is done because it is no longer just a reference to an identifier.  `ExtType` includes information specific to types: instance information.
    * Type specification fields of type `Maybe TypeSpec` are added to the AST.  This is the information for the type from the binding specifications, used to customize how we translate the type.

* `HsBindgen.C.Fold.DeclState`
    * Constructor `TypeDeclOmitted CNameSpelling` is added to `TypeDecl`.  It allows us to keep track of which types were explicitly omitted in binding specifications.

* `HsBindgen.C.Fold.Decl`
    * We used to have an `ExtBindings` argument that was threaded through the fold, but now we have two `IBindingSpecs SourcePath` arguments: the "input" binding specifications and the external binding specifications.

* `HsBindgen.C.Fold.Type`
    * The "input" binding specifications and the external binding specifications are threaded through the fold.
    * The `TypeDeclOmitted` cases are handled.  An exception is thrown if an omitted type is used.
    * Function `lookupExtBinding` is replaced with function `lookupExtTypeAndTypeSpec`, which looks up a type in both binding specifications.
    * Function `addOmitted` registers a type as omitted in the state.

I am of course happy to help with the transition.

edsko (changes requested):
We decided to migrate this work to the refactoring branch.

TravisCardwell:
I will implement binding specifications in a new branch based off of the refactoring.  I am therefore closing this PR and will create a new one when the new implementation is ready.

PR #658, opened 2025-05-19
Poke with FLAM

### LINKED ISSUE #490: Implement `pokeWithFLAM`
This is still missing, as it wasn't entirely clear what to do if the provided vector does not match the size as specified by `flexibleArrayMemberLength`. Let's just make this a precondition, and add an assertion that they must match.

dschrempf:
Closes #490 

- [x] `assert` does not work well with Tastt HUnit tests, so we use `throwIO`.
- [x] We use `AssertionFailed`. Let me know if you prefer to use a separate exception type

### DIFF
--- a/hs-bindgen-runtime/src/HsBindgen/Runtime/FlexibleArrayMember.hs
+++ b/hs-bindgen-runtime/src/HsBindgen/Runtime/FlexibleArrayMember.hs
@@ -4,8 +4,12 @@ module HsBindgen.Runtime.FlexibleArrayMember (
     HasFlexibleArrayLength (..),
     WithFlexibleArrayMember (..),
     peekWithFLAM,
+    FLAMLengthMismatch (..),
+    pokeWithFLAM,
 ) where
 
+import Control.Exception (Exception, throwIO)
+import Control.Monad (unless)
 import Data.Vector.Storable qualified as VS
 import Data.Vector.Storable.Mutable qualified as VSM
 import Foreign
@@ -23,7 +27,7 @@ data WithFlexibleArrayMember element struct = WithFlexibleArrayMember
     }
   deriving Show
 
--- | Peek structure together with contents of flexible array member.
+-- | Peek structure with flexible array member.
 peekWithFLAM :: forall struct element. (Storable struct, Storable element, HasFlexibleArrayLength element struct)
     => Ptr struct -> IO (WithFlexibleArrayMember element struct)
 peekWithFLAM ptr = do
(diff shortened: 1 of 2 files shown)


hs-bindgen-runtime/src/HsBindgen/Runtime/FlexibleArrayMember.hs
```diff
@@ -35,3 +38,26 @@ peekWithFLAM ptr = do
...
+-- | Poke structure with flexible array member.
+pokeWithFLAM
+  :: forall struct elem.
+     (Storable struct, Storable elem, HasFlexibleArrayLength elem struct)
+  => Ptr struct -> WithFlexibleArrayMember elem struct -> IO ()
+pokeWithFLAM ptrToStruct (WithFlexibleArrayMember struct' vector')  = do
+  struct <- peek ptrToStruct
+  let !lenFLAM = flexibleArrayMemberLength struct
+      !lenVector' = VS.length vector'
+  -- TODO: 'assert' does not work with Tasty?
```
phadej, line 51: `assert` is disabled in optimised builds. See https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/assert.html#assertions

> GHC ignores assertions when optimisation is turned on with the [-O](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/using-optimisation.html#ghc-flag-O) flag. That is, expressions of the form assert pred e will be rewritten to e. You can also disable assertions using the [-fignore-asserts](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/using-optimisation.html#ghc-flag-fignore-asserts) option. The option [-fno-ignore-asserts](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/using-optimisation.html#ghc-flag-fignore-asserts) allows enabling assertions even when optimisation is turned on.

TL;DR, if it's  a functional requirement that exception is thrown, `assert` cannot be used.

IMHO `assert` in GHC is not right design, and many libraries indeed do checks in own way, without relying on `assert`. (E.g. `vector`s bounds checks).

dschrempf: Thank you! This means, we use `throwIO`. If you think it is advisable to create a new Exception Type, let me know! Otherwise, we can leave it like it is, and use `AssertionFailed`.

edsko (approved):
Feel free to merge once the exception is changed.

PR #663, opened 2025-05-21
Tracers: Custom log levels and tracking log levels

### LINKED ISSUE #662: Improvements to tracing infrastructure
In a meeting with Edsko, the following tasks came up which we decided to tackle after merging PR #651 , and PR #642 .

### No warnings on internal tests

We want to reduce warnings when running internal tests. There are two options:

- We set the verbosity level to `Error`. Like so, we do not observe new warnings althouth these warnings may be of interest.
- We implement machinery that allows suppression of specific trace messages. Then, we observe new warnings popping up in tests.

### Tracing should track the maximum log level

We want to know, if an error or warnings have ocurred, inform the user, and act accordingly.

- The `withTracer` functions should store and return the maximum log level that was traced.

- Possibly replace existing panics with error traces.

dschrempf:
Closes #662.

### Support tracers with custom log levels
- Use the machinery to suppress some known traces in tests.
- The tests are now trace-free (but we do still show quite some other putStrLn log messages).

### Track and return maximum log level of traces

### DIFF
--- a/hs-bindgen/app/HsBindgen/App/Common.hs
+++ b/hs-bindgen/app/HsBindgen/App/Common.hs
@@ -39,7 +39,7 @@ data GlobalOpts = GlobalOpts {
     , globalOptsClangArgs   :: ClangArgs
     , globalOptsExtBindings :: [FilePath]
     }
-  deriving (Show)
+  deriving stock (Show)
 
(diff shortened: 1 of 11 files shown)


dschrempf:
I decided to combine this with Part 2, and convert it to draft.

PR #672, opened 2025-05-23
Manual section about structures

dschrempf:
We postponed some stuff:
- Implementation of FLAM so we can _use it_ in `RunManual.hs`.
- Opaque structures.


edsko:
Looks good. I'll merge as-is for now, we can always readjust later if and when we need to.

PR #674, opened 2025-05-26
Fold with selections

dschrempf:
Remarks

- Provide predicate for main files (plural).

- We use `Text` for storing `SourcePath`, but have to compare using
  `equalFilePath` which converts to `FilePath`. -> See https://github.com/well-typed/hs-bindgen/issues/678.

- Add `Predicate` and `SourcePath`s of the "main files" to the parse environment.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/C/Predicate.hs
+++ b/hs-bindgen/src-internal/HsBindgen/C/Predicate.hs
@@ -8,12 +8,14 @@ module HsBindgen.C.Predicate (
     Predicate(..)
   , Regex -- opaque
     -- * Execution (this is internal API)
+  , SkipReason (..)
   , match
   ) where
 
 import Control.Monad.Except (ExceptT, runExceptT, throwError)
+import Data.Text qualified as Text
 import Text.Regex.PCRE qualified as PCRE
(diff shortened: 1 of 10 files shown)


hs-bindgen/src-internal/HsBindgen/C/Predicate.hs
```diff
@@ -36,11 +37,8 @@ data Predicate =
     -- Used to define the 'Semigroup' instance.
   | SelectIfBoth Predicate Predicate
 
-    -- | Check if the definition is from the \"main\" file
-    --
-    -- Corresponds directly to @clang_Location_isFromMainFile@ in @libclang@;
-    -- <https://clang.llvm.org/doxygen/group__CINDEX__LOCATIONS.html#gacb4ca7b858d66f0205797ae84cc4e8f2>.
-  | SelectFromMainFile
+    -- | Include definitions in main files (and not in included files)
+  | SelectFromMainFiles
```
dschrempf, line 42: I decided against adding `Set SourceFile` here. The reason is just that the option parsind would be more difficult, because we already need to know the main files then (or we create an intermediate option datatype). @edsko

PR #679, opened 2025-05-27
Fold with `enum`s

dschrempf:


dschrempf:
Merging after meeting with Edsko.

PR #680, opened 2025-05-27
Detect ANSI color support in golden tests

### LINKED ISSUE #675: hs-bindgen should not use ANSI colors when output doesn't support them.
```
cabal run test-internal -- -p attri | cat
```

I still see colors. That is wrong

dschrempf:
Fixes #675.

This was actually tricky because we need to have access to Tasty options (`UseColor`), and also to `stdout` (to query ANSI color features).

### DIFF
--- a/hs-bindgen/hs-bindgen.cabal
+++ b/hs-bindgen/hs-bindgen.cabal
@@ -297,6 +297,7 @@ test-suite test-internal
     , deepseq
   build-depends:
       -- Inherited dependencies
+    , ansi-terminal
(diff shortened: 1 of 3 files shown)


dschrempf:
Also change the way how we degrade traces in tests. Now, we check for diagnostic category names instead of IDs.

dschrempf:
Travis confirmed that it is working on his side, I would merge this, if you agree @edsko

edsko (approved):
Go for it :)

PR #699, opened 2025-06-06
Add source map to support external bindings for non-selected types

TravisCardwell:


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Monad.hs
```diff
@@ -160,6 +186,31 @@ checkHasMacroExpansion extent = do
       , any (\e -> fromMaybe False (rangeContainsLoc range e)) expansions
       ]
 
+recordSource :: CXCursor -> M ()
+recordSource curr = do
+    mNamespace <- dispatch curr $ \case
+      CXCursor_InclusionDirective -> return Nothing
```
edsko: Let's add a catch-all, with a comment: 

We intentionally do selection as part of parsing, rather than a separate step: if the user does not select certain declarations (perhaps because they live deep in the bowels of some system libraries), we also do not need to parse them. Since `recordSource` is called on _all_ declarations, selected or not, we must ensure that we don't error out on such unsupported cases.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/RenameAnon.hs
```diff
@@ -283,3 +286,15 @@ nameForAnon = \case
         nameForAnon useOfAnon <> "_" <> field
       UsedByAnon (UsedInFunction _valOrRef) _useOfAnon ->
         panicPure $ "nameForAnon: unexpected anonymous function argument or return type"
+
+{-------------------------------------------------------------------------------
+  Source map
+-------------------------------------------------------------------------------}
+
+renameSourceMap :: DefUseGraph -> SourceMap Parse -> SourceMap RenameAnon
```
edsko, line 294: Add instance to `CoercePass` for `SourceMap` instead.

edsko, line 294: Going forward: 

* Introduce `OmittedDecls` (essentially the current `SourceMap` type) as a "writer" part of `ParseMonad`. Only record declarations we omitted; it does not need a pass index (implied: `RenameAnon`; we cannot have top-level anonymous declarations).
* In `ResolveBindingSpecs`, when resolving external bindings, traverse both the actual set of declarations, and the omitted declarations (or alternatively, build a local `SourceMap` of all declarations, omitted or not).

But we can do this in a separate commit, after merging into `main`.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ResolveBindingSpec/IsPass.hs
```diff
@@ -24,6 +24,7 @@ import HsBindgen.Language.C
 type ResolveBindingSpec :: Pass
 data ResolveBindingSpec a deriving anyclass (ValidPass)
 
+-- We don't need the @SourceMap@ anymore.
```
edsko: Suggestion: "We do not need the `SourceMap` anymore, so we drop it from the annotation on ... "

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/RenameAnon.hs
TravisCardwell: PR #704

PR #704, opened 2025-06-09
Refactor to use `OmittedDecls` and resolve binding specification in two passes

TravisCardwell:
Previously, type `SourceMap p` mapped from `QualId p` to `SourcePath` for all declarations, including those that are not selected by the predicate.  This is refactored to type `OmittedDecls`, which is essentially the same but is only used for declarations that are not selected by the predicate.  The motivation is to reduce duplication.

The key of the `OmittedDecls` map is essentially `QualId RenameAnon`, since we decided to only deal with top-level types, which cannot be anonymous.  To avoid referencing the `RenameAnon` pass from within the `Parse` pass, the equivalent type `(CName, Namespace)` is used instead.  Since `OmittedDecls` is not parameterized, the annotations do not need to be renamed.

The `OmittedDecls` values are accumulated in the `Parse` pass.  This is essentially a `Writer`, but the accumulator is included in the state for simplicity and performance.

The `ResolveBindingSpec` pass checks selected and omitted declarations separately.

Resolving binding specification in two passes was something that was lost during the translation to the `ResolveBindingSpec` pass.  A reference to a declaration may come before the declaration itself, via a pointer.  With this change, information about external bindings and omitted types of declarations is accumulated during a first pass, and then the types within each declaration is resolved in a second pass.


hs-bindgen/src-internal/HsBindgen/Frontend/OmittedDecls.hs
```diff
@@ -0,0 +1,37 @@
...
+
+import Clang.Paths (SourcePath)
+import HsBindgen.Imports
+import HsBindgen.Language.C.Name (CName, Namespace)
+
+{-------------------------------------------------------------------------------
+  Type
+-------------------------------------------------------------------------------}
+
+newtype OmittedDecls = OmittedDecls {
```
edsko: Let's rename this to `NonSelectedDecls`.

edsko: We should add a comment explaining that these are the declarations that were not selected during parsing.

Add further comment: It is important to keep track of these non-selected declarations because they can be given external bindings. We do _not_ support external bindings for _anonymous_ non-selected declarations; _if_ you want to provide an external binding for some local type, for example

```c
struct rect {
  struct { int x; int y; } bottomleft;
  struct { int x; int y; } topright;
}
```

then you need to make sure that you _traverse_ `rect`, so that the `RenameAnon` pass can do its work.

edsko (approved):
Nice. Feel free to merge once the changes we discussed are done.

PR #706, opened 2025-06-10
Rename graphs UseDef to UseDecl and DefUse to DeclUse

dschrempf:


dschrempf:
@edsko I think this should be ready now.

Issue #707, opened 2025-06-10
Do not emit warning "unexpected unsupported member declaration list in struct specifier"


dschrempf:
When parsing the following C header file:

```c
#define MY_TYPE int

typedef struct {
  int x;
  MY_TYPE y;
} bar;
```

we first try to reparse the entire struct, fail wilt the following warning,

```
[Warning] [HsBindgen] "./warn.h" (line 6, column 3):
  unexpected unsupported member declaration list in struct specifier
  typedef struct { int x ; MY_TYPE y ; } bar
```

Then, we reparse the fields, and succeed.

dschrempf:
I was analyzing this ticket, and I am not exactly sure what is going on.

It seems like we label the typedef and the fields as "needs reparse".

Then, we reparse the `typedef`, and fail when we encounter the struct. Then, we go on reparsing the fields and succeed.

In essence, we should not label the `typedef` to need a reparse, but only the individual fields. Is that so?

Edit: Relevant function is `HsBindgen.Frontend.Pass.HandleMacros.processDecl`, and is a bit hard to debug, because of lazyness.

dschrempf:
Digging a bit further, in `HsBindgen.Frontend.Pass.Parse.Decl.Monad.getReparseInfo`, we set `ReparseNeeded` for all declarations that "have a macro  extension". This is true for the `struct`, as well as the fields of the `struct`.

<details><summary> General thoughts about the "check for reparse" algorithm</summary>

Further, in the same module, in `checkHasMacroExpansion`, we check if the extent of the current declaration contains any of the macro expansions.

This makes me think, maybe we should just traverse the macro expansions, and look for the "smallest" declaration completely containing the expansion, and only reparse those? Maybe this is also faster?

Also, I am pretty sure we can improve the `O(n)` algorithm (scan through all macro expansions) by using a data structure with some type of range order (but I am not an expert in this regard). (Maybe keep an eye on this when working on #517).
</details>

Assuming we do not change the algorithm for now:
- for a given declaration with `extent`, check if the `extent` contains macro expansions
- if it contains macro expansions, check for each macro expansion:
  - is there an inner declaration also containing the macro expansion?
    - if yes, do not reparse this declaration (the inner declaration, or, recursively, an even more inner declaration, will be reparsed)
    - if no, reparse this declaration

Problem: what if we have two macro expansions, one of the is also covered by an inner declaration, but the other is not?

dschrempf:
[...]

PR #708, opened 2025-06-10
Reduce traces in tests, improve tracer infrastructure, handle max log level

### LINKED ISSUE #710: Do not generate output after errors
The new tracer infrastructure keeps track of the maximum log level used, allowing us to detect if any errors were reported (we try not to throw exceptions, whenever possible, so that we can report as many user errors at once). We should check for this maximum log level, and prevent the generation of any output (`exitFailure` or throw exception at that point) if it there were any errors.

dschrempf:
- Set default log level to Warning
- Correctly detect ANSI colors in TH tests
- Improve tracer infrastructure in tests (no need to parse options twice)
- Handle max log level (closes #710)
- Add unit test for exception on error

### The following warnings remain

1. unexpected Unexpected primitive type "void"
2. unexpected unsupported enumerator list in enum specifier
3. unexpected unsupported member declaration list in struct specifier
4. Failed to typecheck macro

I think 2. and 3. have a ticket: https://github.com/well-typed/hs-bindgen/issues/707

I don't know if we want to degrade or fix 1. and 4.

### DIFF
--- a/hs-bindgen/app/hs-bindgen-cli.hs
+++ b/hs-bindgen/app/hs-bindgen-cli.hs
@@ -20,8 +20,7 @@ import HsBindgen.Lib
 main :: IO ()
 main = handle exceptionHandler $ do
     cli@Cli{..} <- getCli
-    withTracerStdOut (globalOptsTracerConf cliGlobalOpts) DefaultLogLevel $ \tracer ->
-      execMode cli tracer cliMode
+    execMode cli cliMode
 
 data LiterateFileException = LiterateFileException FilePath String
   deriving Show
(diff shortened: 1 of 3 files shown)


dschrempf:
I thhink this can be merged, we have to tackle the warnings separately anyways.

dschrempf:
I have added the "withResource" function, as agreed @edsko . I decided to pass on the ANSI color setting instead of the actual tracer, because Tasty only provides us with an IO action to get the resource anyways.

This means we can use the `withXYZ` functions, and don't need to export the lower-level `mkTracer`.

dschrempf:
We decided to include #710 into this PR.

dschrempf:
Actually, I think it is a good idea to add a test case checking that we do not generate output. (Probably using file output, which can be monitored in an easier way than `stdout`).

edsko (approved):

Issue #703, opened 2025-06-07
TH mode should be made more convenient (and full featured)


edsko:
It would ideally be as simple as

```hs
hashInclude "header.h"
```

and nothing else.

dschrempf:
Most likely also related to https://github.com/well-typed/hs-bindgen/issues/728#issuecomment-2974684626

Issue #419, opened 2025-02-18
Cabal C include search paths


TravisCardwell:
We would like to provide an option to get C include search path information from a section of a `.cabal` file.  The file should of course be parsed using the Cabal library.

* Any directories specified in [`include-dirs`][] should be resolved and added to the non-system include search path.
* For packages specified in [`pkgconfig-depends`][], we can call `pkg-config` and add to the include search path according to the result.

[`include-dirs`]: <https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#pkg-field-include-dirs>
[`pkgconfig-depends`]: <https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#pkg-field-pkgconfig-depends>

This was discussed in https://github.com/well-typed/hs-bindgen/issues/366#issuecomment-2655630980

edsko:
Note that we should not try to parse the cabal file, but we should run cabal, or use information from a previous cabal run. Cabal's runtime flags, as well as files like `cabal.project` and `cabal.project.local`, all have influence on this; moreover, `cabal` will run `pkg-config` as well. The information in the `.cabal` file itself is insufficient.

PR #718, opened 2025-06-11
Simple TH `hashInclude "simple.h"`

dschrempf:
This implements part 1 of #703 (simple interface).

It does not include part 2: Access to all client options.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Pipeline.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Pipeline.hs
@@ -3,9 +3,7 @@
 module HsBindgen.Pipeline (
     -- * Options
     Opts(..)
-  , defaultOpts
   , PPOpts(..)
-  , defaultPPOpts
 
     -- * Translation pipeline components
   , parseCHeader
@@ -90,12 +88,11 @@ data Opts = Opts {
     , optsTracer      :: Tracer IO (TraceWithCallStack Trace)
     }
 
--- | Default 'Opts'
-defaultOpts :: Opts
-defaultOpts = Opts {
-      optsClangArgs   = defaultClangArgs
+instance Default Opts where
+  def = Opts {
+      optsClangArgs   = def
     , optsExtBindings = BindingSpec.empty
-    , optsTranslation = Hs.defaultTranslationOpts
+    , optsTranslation = def
     , optsPredicate   = SelectFromMainFiles
     , optsTracer      = nullTracer
     }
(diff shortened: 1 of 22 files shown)


hs-bindgen/src-internal/HsBindgen/Pipeline.hs
```diff
@@ -174,18 +177,52 @@ preprocessIO ppOpts fp = genPP ppOpts fp . genModule ppOpts . genSHsDecls
   Template Haskell API
 -------------------------------------------------------------------------------}
 
+-- | Generate bindings for the given C header (simple)
+--
+-- Use default Clang arguments and the directory of the source file as include
+-- search path.
```
TravisCardwell: I am not sure if adding the directory of the Haskell source file to the include search path is very useful, as projects generally keep C headers and source separate from Haskell source.  This would encourage mixing the two, which might not be a good idea.

How about we do not add anything to the include search path by default?  Users can do so via environment variables after all.

In the future, we can (also) get C include search path directories by processing the `.cabal` file.  (#419)

dschrempf: The idea was to provide a simple wrapper that does not need any configuration. That's why I added the path of the source file to the include path.

We also export the other function which allows specification of include paths directly.

~~Of course, we could check C_INCLUDE_PATH and/or parse the `cabal` file.~~ <- Clang does this for us.

I think we just have to make a decision. I liked the possibility to include C header files relative to the source file without further ado, but I understand, that it may not be very useful in larger projects. @edsko?

dschrempf: We have discussed this a meeting. The outcome is:

We provide two functions:
  1. A simple interface that does not temper with include dirs.
  2. An interface that allows addition of include dirs relative to the package root.

dschrempf:
We could add `Data.Default` instances to our other options too.

EDIT: I was rethinking this, and I think we should either use Data.Default consistently, or not use it at all.

dschrempf:
We decided to provide Data.Default instances for all user-facing option types.

dschrempf:
@edsko I think this is done now. It also touches some c-expr files, formatting some imports. I took care that the nice separation of Sam is not lost.

hs-bindgen/src-internal/HsBindgen/Pipeline.hs
```diff
@@ -174,18 +182,81 @@ preprocessIO ppOpts fp = genPP ppOpts fp . genModule ppOpts . genSHsDecls
...
+  deriving stock (Eq, Show)
+
+toFilePath :: FilePath -> QuoteIncludeDir -> FilePath
+toFilePath root (PackageRoot     x) = root </> x
+toFilePath _    (QuoteIncludeDir x) = x
+
+toFilePaths :: [QuoteIncludeDir] -> TH.Q [FilePath]
+toFilePaths xs = do
+  root <- getPackageRoot
+  pure $ map (toFilePath root) xs
```
TravisCardwell: Perhaps these should go in the `where` clause of the single function that uses them, just to make it clear that they are not part of the public API.  The names do not include context, which brings unnecessary attention to them.  :shrug:

PR #729, opened 2025-06-13
Avoid panic on missing declaration in name mangler

### LINKED ISSUE #723: Missing selected declarations result in name mangler panic
Consider `a.h`

```c
struct point;
```

and `b.h`

```c
#include "a.h"

void f(struct point* p);
```

If we only select bindings from `b.h`, we get a panic: 

```
$ cabal run -- hs-bindgen-cli preprocess -i b.h 
Name mangler bug: no name for QualId "point" NamespaceStruct in fromList [(QualId "f" NamespaceFunction,"f")]
```

(If we pass `--select-all`, it works just fine.) 

It's fine that we report an error here, but the error should be more informative.

dschrempf:
Closes #723


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames.hs
```diff
@@ -59,9 +61,25 @@ mangleNames unit =
...
+  | MissingDeclaration !(C.QualId HandleTypedefs)
   deriving stock (Show, Eq)
 
+instance PrettyTrace MangleError where
+  prettyTrace (CouldNotMangle name) =
+    "Could not mangle C name: " <> unpack name
+  -- NB: We did not register any declaration with the given ID. This is most
+  -- likely because the user did not select the declaration. If the declaration
+  -- was completely missing, Clang would have complained already.
+  prettyTrace (MissingDeclaration qualId) =
```
edsko, line 71: Let's have this comment where we generate the message, rather than where we pretty-print it.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames.hs
```diff
@@ -59,9 +61,25 @@ mangleNames unit =
 type NameMap = Map (C.QualId HandleTypedefs) HsIdentifier
 
 data MangleError =
-    CouldNotMangle Text
+    CouldNotMangle !Text
+  | MissingDeclaration !(C.QualId HandleTypedefs)
   deriving stock (Show, Eq)
```
edsko, line 66: What's the purpose of the bangs?

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Internal.hs
```diff
@@ -246,6 +247,9 @@ deriving instance Show (Id p) => Show (QualId p)
 deriving instance Eq   (Id p) => Eq   (QualId p)
 deriving instance Ord  (Id p) => Ord  (QualId p)
 
+instance (PrettyTrace (Id p)) => PrettyTrace (QualId p) where
```
edsko, line 250: I wonder if we should rename `PrettyTrace`, since `QualId` is not a trace message :thinking: Not necessary to do as part of this PR though.

hs-bindgen/src-internal/HsBindgen/Language/C/Name.hs
```diff
@@ -25,3 +31,12 @@ data Namespace =
...
   deriving stock (Show, Eq, Ord, Generic)
 
+instance PrettyTrace Namespace where
+  prettyTrace = \case
+    NamespaceTypedef  -> "typedef"
+    NamespaceStruct   -> "struct"
+    NamespaceUnion    -> "union"
+    NamespaceEnum     -> "enum"
+    -- I made these two up.
+    NamespaceMacro    -> "C macro"
```
edsko: I guess the "C" part can be omitted here?

edsko: Also, @TravisCardwell might have some thoughts here, as we were working towards not having this `Namespace` at all. Not sure what we should do.

edsko (approved):
Looks good, with some minor changes, but let's give @TravisCardwell a chance to chime in regarding the `Namespace`.

hs-bindgen/src-internal/HsBindgen/Language/C/Name.hs
edsko: I'm also okay with merging this (modulo the changes), and we can revisit the `Namespace` next week.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames.hs
dschrempf: I was expecting you to ask that :-). It is a mixture of
- the linter telling me to put those bangs
- me programming a lot of number crunching simulations in Haskell and having serious performance problems with lazy data structures.

That's why I have the habit to make constructor fields strict.

I will remove those.

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Internal.hs
dschrempf: Yes, I was also thinking the same. On the other hand, the `PrettyTrace` could also indicate how we should print it "in a trace message", and then it would be fine.

hs-bindgen/src-internal/HsBindgen/Language/C/Name.hs
dschrempf: I have removed the 'C' part, and we can always change the exact output later, if Travis would like to change it.

PR #730, opened 2025-06-13
Deal with redeclarations

edsko:
Redeclaring a macro with the _same_ definition does not result in any warnings or errors (and we test that we get the appropriate output for this).

Redeclaring a macro with a _different_ value results in a warning (and no binding), and we test that we get this warning.


edsko:
@dschrempf One thing we could think about is checking that for the "succeeding" fixtures we don't get any warnings at all (it would be good if I checked this for `redeclaration_identical.h` but our current infrastructure does not make that very easy.)

dschrempf:
Well, actually, I think we do generate some warnings, and degrade them to debug messages. So I think a good way to do this is upgrade all traces to errors, and then selectively downgrade known traces to debug (if that makes sense).

edsko:
Yes, selectively move warnings to either errors or info would be good.

PR #743, opened 2025-06-14
Check for varargs functions

edsko:
If we see a declaration of a varargs function, we skip over it (with a warning); if we use a _use_ of a varargs function, we error out.


edsko:
Going to submit a PR that changes how we parse types first.

PR #745, opened 2025-06-14
Introduce `ParseTypeException`

edsko:
See code for discussion. TL;DR: This allows us to skip generating bindings for declarations which contain unsupported types (and produce a helpful error message).


edsko:
This PR is wrong; we should not be using exceptions while folding. Will fix.

edsko:
Fix in https://github.com/well-typed/hs-bindgen/pull/773 .

PR #746, opened 2025-06-14
Warn about long double in parser

edsko:
Unsupported features should be detected in the frontend; panics in the backend should be limited to bugs in hs-bindgen.


edsko:
As of #773 we can now skip over any declarations that use `long double`.

Issue #705, opened 2025-06-09
Maintain C spelling and aliases


TravisCardwell:
For a given `Hs` declaration, we need to know the corresponding C spelling (if any) as well as all "aliases."  This is something that was in the previous implementation but was lost in the refactoring.

An *alias* is a C spelling for an `Hs` declaration that is not the primary one.  With default settings, we only get an alias if a (`struct`, `union`, or `enum`) tag exactly matches a `typedef` name for the type.  For example, `typedef struct foo {...} foo;` results in a primary C spelling of `struct foo` and an alias `foo`.  Once implemented, a binding specification will allow users to configure other aliases, so a given `Hs` declaration may have any number of aliases.  For example, `typedef struct foo {...} foo_t` normally translates to two `Hs` declarations, but a user may configure it such that `foo_t` is instead an alias for the `struct foo` declaration.

Knowing the C spelling is required for generating a binding specification, generating C code for FFI as well as tests, and documentation.  Knowing all aliases is required in order to load a binding specification correctly.

TravisCardwell:
**Minimal Example**

Here is a `struct` declaration with just a tag:

```c
struct foo {
  int x;
  int y;
};
```

<details>
<summary>
<tt>cabal run hs-bindgen-dev parse -- -i ex_tag_only.h 2>/dev/null | friendly</tt>
</summary>

```haskell
TranslationUnit {
    unitDecls = [
      Decl {
        declInfo = DeclInfo {
          declLoc = "./ex_tag_only.h:1:8"
        , declId = NamePair {nameC = "foo", nameHsIdent = "Foo"}
      }
      , declKind = DeclStruct (
          Struct {
            structNames = RecordNames {recordConstr = "Foo"}
          , structSizeof = 8
          , structAlignment = 4
          , structFields = [
              StructField {
                structFieldLoc = "./ex_tag_only.h:2:7"
              , structFieldName = NamePair {nameC = "x", nameHsIdent = "foo_x"}
              , structFieldType = TypePrim (PrimIntegral PrimInt Signed)
              , structFieldOffset = 0
              , structFieldWidth = Nothing
            }
            ,StructField {
                structFieldLoc = "./ex_tag_only.h:3:7"
              , structFieldName = NamePair {nameC = "y", nameHsIdent = "foo_y"}
              , structFieldType = TypePrim (PrimIntegral PrimInt Signed)
              , structFieldOffset = 32
              , structFieldWidth = Nothing
            }
          ]
[...]

PR #748, opened 2025-06-17
Maintain C spelling and aliases (part 1)

TravisCardwell:
This is a first PR for #705.

The goal is to maintain enough information to be able to reconstruct C spellings, as well as know when there is no C spelling for something.  Before refactoring the project, we used `DeclPath` to do this.  The refactored design removed that type, and this PR changes some things so that we have the necessary information.

We have been using a C `Namespace` type defined as follows:

```haskell
data Namespace =
    NamespaceTypedef
  | NamespaceStruct
  | NamespaceUnion
  | NamespaceEnum
  | NamespaceMacro
  | NamespaceFunction
```

This differs from C namespaces.  C has the following namespaces:

* Label namespace, which we are not concerned with
* Ordinary namespace, including variable, function, and `typedef` names
* Tag namespace, which is a *single* namespace for `struct`, `union`, and `enum` tags
* Field namespaces, separate per `struct` and `union`

We decided to change the type to reflect (the) C namespaces (that we use), and I initially refactored it to the following.  The motivation was to use the type when the tag kind is determined by some context, such as a declaration kind or type constructor.  The type is not needed, however, so it is removed.

```haskell
data Namespace =
    NamespaceOrdinary
  | NamespaceTag
```

We need a type that distinguishes tag kinds when there is no such context.  For example, `QualId` is used as keys in maps and vertices in graphs, so it must use such a type.  The following type is used.  Note that it closely matches C spelling prefixes.

```haskell
data NameKind =
    NameKindOrdinary
  | NameKindStruct
  | NameKindUnion
  | NameKindEnum
```

For declarations, a `DeclKind` constructor determines the `NameKind`.  For some kinds of declarations, we need origin information to determine the spelling (if one exists).  The following type is used.

```haskell
data NameOrigin =
    NameOriginInSource
  | NameOriginGenerated
  | NameOriginRenamedFrom CName
```

For example, as indicated in https://github.com/well-typed/hs-bindgen/issues/705#issuecomment-2960778933, a `Decl` with kind `DeclStruct` implicitly has an `Id` of kind `NameKindStruct`, and we need a `NameOrigin` to know where the name comes from.  With `Id` `foo`:

* If `NameOriginInSource`, then `struct foo` is valid.
* If `NameOriginGenerated`, then `struct foo` is *not* valid.  We must look at the list of aliases to determine if there is a `typedef` name for the type.  If there are no aliases, then the type is not representable in C.
* If `NameOriginRenamedFrom "bar"`, then `struct foo` is *not* valid, but `struct bar` is valid.

We need to determine spellings for types as well.  A key distinction is that types may represent squashed `typedef` declarations.  An anonymous `struct` with a name determined by a `typedef` is represented by a `TypeTypedef` that references a `TypeStruct`.  This distinguishes the `NameKind`, but we still need the `NameOrigin` (for `struct`, `union`, and `enum` types).  For example, the name of an anonymous `struct` may be generated from a `struct` and field name, and we must track that there is no valid C name for the type.  The origin also tracks renaming, such as in the `typedef struct foo {...} *foo;` case.


TravisCardwell:
Tests on macOS pass with LLVM 14 but fail with later versions!  Investigating.

TravisCardwell:
Debugging macOS failures is difficult/frustrating; I wish it was legal to run macOS VMs on non-Apple hardware.  I was just able to reproduce the failure locally (in Linux) using LLVM 20, however!  This is an issue with newer versions of LLVM, not macOS.  Now I am able to debug.

Minimal example:

```c
typedef struct {
  int a;
} foo;
```

`getDeclId` should return a `DeclAnon` for the anonymous `struct`, but it is not doing so!  The call to `clang_getCursorSpelling` returns `foo` and the call to `clang_Cursor_isAnonymous` returns `False`.

I am out of time today, but I will continue in the morning.  If anybody has any hints, please share!

TravisCardwell:
Investigating names:

* `clang_getCursorDisplayName` also returns `foo`.
* `llvm-project` issue [#58429](https://github.com/llvm/llvm-project/issues/58429) indicates that empty names are no longer reported for anonymous declarations.  `rust-bindgen` issue: [#2312](https://github.com/rust-lang/rust-bindgen/issues/2312)
* `clang_getTypeSpelling` returns the empty string.  We raise an exception with `ensure (not . Text.null)`.

Investigating [`clang_Cursor_isAnonymous`](https://clang.llvm.org/doxygen/group__CINDEX__TYPES.html#ga6e0d2674d126fd43816ce3a80b592373):

* `clang_Cursor_isAnonymousRecordDecl` also returns `False`.
* I confirmed that `clang_Cursor_isAnonymous` returns `False` even with earlier versions of LLVM.  It never did what we expected; it just worked because of the empty name check.
* I found a Stack Overflow question [libclang ,clang_Cursor_isAnonymous what is an anonymous declaration?](https://stackoverflow.com/questions/33582339/libclang-clang-cursor-isanonymous-what-is-an-anonymous-declaration) that states that "anonymous structures or unions are c++ features."  That is counter-intuitive, and I see no reference to it in official documentation.  The corrected code passes `-x c++` arguments to Clang.
* I tried passing that Clang argument (`cabal run hs-bindgen-dev -- parse --clang-option '-xc++' example.h`), but this is not something that we designed for, and I get an error ("unknown error parsing C headers: `simpleEnum CXError_ASTReadError`").

TravisCardwell:
`clang_Cursor_isAnonymous` [source](https://github.com/llvm/llvm-project/blob/a9811340b75baae8e06fb9ab83015a90d61510ee/clang/tools/libclang/CXType.cpp#L1360-L1372):

```c++
unsigned clang_Cursor_isAnonymous(CXCursor C){
  if (!clang_isDeclaration(C.kind))
    return 0;
  const Decl *D = cxcursor::getCursorDecl(C);
  if (const NamespaceDecl *ND = dyn_cast_or_null<NamespaceDecl>(D)) {
    return ND->isAnonymousNamespace();
  } else if (const TagDecl *TD = dyn_cast_or_null<TagDecl>(D)) {
    return TD->getTypedefNameForAnonDecl() == nullptr &&
           TD->getIdentifier() == nullptr;
  }

  return 0;
}
```

It does not return `True` when there is a `typedef`.  This will not work for us, regardless of the dubious `-x c++` answer on Stack Overflow.

TravisCardwell:
> `clang_getTypeSpelling` returns the empty string. We raise an exception with `ensure (not . Text.null)`.

I now find cases where the `typedef` name is returned.

> I found a Stack Overflow question [libclang ,clang_Cursor_isAnonymous what is an anonymous declaration?](https://stackoverflow.com/questions/33582339/libclang-clang-cursor-isanonymous-what-is-an-anonymous-declaration) that states that "anonymous structures or unions are c++ features." That is counter-intuitive, and I see no reference to it in official documentation.  The corrected code passes `-x c++` arguments to Clang.

I found the source in the mailing list:

1. [Initial question](https://lists.llvm.org/pipermail/cfe-dev/2015-November/045949.html)
2. [Reproduction](https://lists.llvm.org/pipermail/cfe-dev/2015-November/046046.html)
3. [Reply](https://lists.llvm.org/pipermail/cfe-dev/2015-November/046107.html) stating that anonymous declarations are a C11 feature
4. [Clang args solution](https://lists.llvm.org/pipermail/cfe-dev/2015-November/046116.html)

Note that we use C17 by default.  Configuring C11 results in no change.  This is expected given the above implementation.

I found the [extension diagnostics](https://github.com/llvm/llvm-project/blob/a9811340b75baae8e06fb9ab83015a90d61510ee/clang/lib/Sema/SemaDecl.cpp#L5559-L5565).

TravisCardwell:
Investigating [`clang_Cursor_isAnonymousRecordDecl`](https://clang.llvm.org/doxygen/group__CINDEX__TYPES.html#ga59aaf3b8329a35e400ee3735229a8cb6), I found the [commit](https://github.com/llvm/llvm-project/commit/50be573ed207baa477e209cf9e8b247ddd402bc5) that added it.

I found the following comment for function `isAnonymousStructOrUnion`:

```c++
/// Whether this is an anonymous struct or union. To be an anonymous
/// struct or union, it must have been declared without a name and
/// there must be no objects of this type declared, e.g.,
/// @code
///   union { int i; float f; };
/// @endcode
/// is an anonymous union but neither of the following are:
/// @code
///  union X { int i; float f; };
///  union { int i; float f; } obj;
/// @endcode
```

In my minimal example above, I think that the `typedef` around the anonymous `struct` counts as an object, which is why the anonymous `struct` is *not* considered anonymous by Clang.

We have different ideas of what "anonymous" means.

TravisCardwell:
The best way I could think of to fix the issue is to check the underlying tokens.  Function `isNotTagged` returns `True` if the cursor is for a `struct`, `union`, or `enum` declaration and the underlying tokens starts with that keyword and a left bracket.

This passes all tests across the supported versions of LLVM, but I am of course open to a better solution if anybody has any ideas.

dschrempf:
Impressive analysis, thank you!

Why is it actually, that we think the struct in your example is anonymous? We can refer to it using `foo`, can't we?

I would think that anonymous means that we have no means of referring to it, i.e., neither a name nor a type synonym.

TravisCardwell:
Indeed, we think of a `struct` *declaration* as "anonymous" when there is no tag, even though the *type* may not be anonymous due to a `typedef`.  The distinction is important because we treat `struct` and `typedef` declarations as distinct types in Haskell, using `newtype`.  We "squash" the `typedef` when the `typedef` name is the same as the tag name, but even this will be configurable.

The issue that I am working on is primarily about maintaining C spellings for types.  Given an `Hs.Decl`, we have to be able to generate C code.  Given an `Hs.Decl` that represents a C `struct`, we must know if the type is representable in C.  If it is, we need to know the spelling to use: either `struct foo` or `foo`.  This PR implements the information required to distinguish all of these cases, for both `Decl` and `Type`.

TravisCardwell:
One of our goals is to be able to reconstruct the original `DeclId`:

```haskell
data DeclId =
    DeclNamed CName
  | DeclAnon AnonId

newtype AnonId = AnonId SingleLoc
```

Given a `Decl` with name `cname` and source `sloc`, we can inspect the `NameOrigin`:

* If `NameOriginInSource`, then the `DeclId` is `DeclNamed cname`.
* If `NameOriginGenerated`, then the `DeclId` is `DeclAnon (AnonId sloc)`.
* If `NameOriginRenamedFrom fromCName`, then the `DeclId` is `DeclNamed fromCName`.

edsko (approved):
Good stuff. Feel free to merge once the `AnonId` is added to `GeneratedFrom` and we use the different way of detecting anonymous declarations we discussed.

TravisCardwell:
The change to `GeneratedFrom` is done.

Unfortunately, we really cannot detect anonymous declarations via names.  Clang generates names (that include parentheses) only when `clang_Cursor_isAnonymous` returns `True`.

@edsko suggested that we implement the token hack more strictly, checking the name of tags against what `libclang` reports.  I will implement this in a separate PR.

PR #753, opened 2025-06-18
Ignore type attributes

edsko:


edsko:
`rust-bindgen` [crashes on this example](https://github.com/rust-lang/rust-bindgen/blob/v0.70.1/bindgen/codegen/mod.rs#L2220). I took a look at the code we generate and I _think_ it's correct, which would suggest this is a limitation of `rust-bindgen` as opposed to some problematic situation that we are failing to detect, though I am not _completely_ sure. The [gcc manual says](https://gcc.gnu.org/onlinedocs/gcc-4.1.0/gcc/Type-Attributes.html) that `transparent` unions just affect how functions are called, nothing else.

PR #759, opened 2025-06-18
Test-specific trace predicates

### LINKED ISSUE #752: Tests still emit some unexpected warnings
After https://github.com/well-typed/hs-bindgen/pull/751, I still see the following warnings:

```
[Warning] [HsBindgen] Encountered unsupported type while parsing f at "/home/dominik/work/hs-bindgen/hs-bindgen/examples/golden/varargs.h:4:6": Unsupported variadic (varargs) function.. No bindings generated for f.
```

```
        clang diag: examples/golden/unnamed-struct.h:1:1: warning: declaration does not declare anything [-Wmissing-declarations]
```
(coming from `rust-bindgen`).

```
        [Warning] [HsBindgen] Failed to typecheck macro:
          Unbound variable: '_Pragma'
```

We should fix or degrade those and then activate "strict" mode in that we set all other traces to log level Error.

dschrempf:
Closes #752.

Tests fail if an unexpected trace is emitted, or if an expected trace is not emitted.

By default, we ignore Info and Debug traces.

TODOs:
- [x] Test the `withTracePredicate` framework.


hs-bindgen/test/internal/test-internal.hs
```diff
@@ -43,166 +48,177 @@ main :: IO ()
...
         , expectTrace
             "redeclaration_different"
-            "expected trace Redeclaration"
-            (\case
-                TraceFrontend (FrontendSort (SortErrorDeclIndex (Redeclaration {}))) -> True
-                _otherwise -> False
+            (customTracePredicate $ \case
+              TraceFrontend (FrontendSort (SortErrorDeclIndex (Redeclaration {})))
+                -> Just Expected
+              TraceDiagnostic x |
```
dschrempf: @edsko I had to add this expectation.

(But maybe it is good if we look through all exceptions anyways).

edsko: Yeah, I guess that's reasonable enough. I wonder if we're going to check the clang diagnostic that we're getting here, perhaps we should lok for `redefined` in the diagnostic text rather than check the category..? But perhaps that will not work across all clang versions?

dschrempf: Yes, I have found the `diagnosticCategoryText` to be the most stable, but we can try to be more specific in these cases. Should I check for a more detailed predicate?

edsko: I think it would be good if we could check for `redefined` (`isInfixOf`). Not just because it makes the test more precise, but also because it makes the test more _readable_ / easier to audit (why are we allowing this warning?).

dschrempf:
Note, I also moved `type_attributes` to `golden-norust`.

dschrempf:
The only warning I see when running `cabal run -- test-internal` is now:

```
         clang diag: examples/golden/unnamed-struct.h:1:1: warning: declaration does not declare anything [-Wmissing-declarations]
```

dschrempf:
We now check that `rust-bindgen` actually returns a non-zero exit code on all `norust` examples.

PR #760, opened 2025-06-18
Update binding specification terminology

TravisCardwell:


dschrempf:
Actually, we should probably also change `--external-bindings` to `--external-binding-spec`? (or plural, whatever works).

edsko:
Suggested naming:

```
--generate-binding-spec (-g)
--descriptive-binding-spec ("external") (-d)
--prescriptive-binding-spec ("input" / "config") (-p)
```

Difference:

* _describes_ existing bindings
* _prescribes_ how to generate new bindings

TravisCardwell:
(Commit message)

Update binding specification terminology

There is a single *binding specification* type, which can be represented using YAML or JSON formats.

The single type serves two purposes:

* A *prescriptive binding specification* is used to configure how bindings are generated.
* An *external binding specification* is used to specify existing bindings that should be used, /external/ from the module being generated.

A *generated binding specification* may be used for either/both of these two purposes.

Regarding plurality, a single binding specification (singular) describes multiple bindings (plural).  We generally do not use "binding specifications" (plural) unless we are specifically talking about multiple/separate binding specifications (multiple `BindingSpec` values or multiple YAML/JSON files).

This commit includes the following:

* Documentation and variable names are changed to match the new conventions.  As a convention, a variable for a prescriptive binding specification is now named `pSpec` (was `confSpec`).
* Type name `StdlibBindingSpecs` confusing, as it sounds like the binding specification itself.  I tentatively renamed it to `StdlibBindingSpecConf` to make it clear that it is a configuration type.
* A pending change is fixing the public API to export an opaque wrapper around `ResolvedBindingSpec`.  This is *not* fixed yet because it is pending discussion (#777).  To fix terminology, a `type` alias is added, with a `TODO`.  Because of this change, `test-internal.hs` is refactored to not use the `Lib` library, which is something that we discussed wanting to do anyway.  FWIW, I would be happy to use qualified imports more often...

PR #761, opened 2025-06-19
Default external bindings

TravisCardwell:
~Since we are avoiding use of Template Haskell in `hs-bindgen` because it greatly complicates cross compilation,~ I added a new `HsBindgen.BindingSpec.Stdlib` module that defines both the `base` and `hs-bindgen-runtime` bindings.  New `hs-bingen-dev` commands output the YAML for these bindings:

* `hs-bindgen-dev binding-spec base`
* `hs-bindgen-dev binding-spec runtime`

The binding specification files in `hs-bindgen/bindings` are now created using these commands.  Dhall is no longer used.

The standard external bindings are now used by default.  Note that binding specifications are resolved *before* parsing and translating headers.  `Pipeline.Opts` contains resolved external bindings as `optsExtBindings :: ResolvedBindingSpec`.  We cannot simply include standard external bindings as a default because bindings must be resolved, which is environment-specific.  To use standard external bindings by default, you must resolve them and configure `Pipeline.Opts` with the result.

This works fine in the `Lib` API.  It maintains the status quo of the `TH` API, and `hashInclude` of course does this.  This is implemented in the internal tests using a Tasty resource.  Changing tests to use the standard external bindings does not work, however: the Template Haskell tests fail because there are no imports for the external bindings used.

CLI option `--no-stdlib` can be used to disable the standard external bindings.

I have not yet added a CI job that confirms that the binding specification files in `hs-bindgen/bindings` are current.  I will add a commit to this PR that does this, once I confirm that all tests pass.

### DIFF
--- a/hs-bindgen-runtime/src/HsBindgen/Runtime/Prelude.hs
+++ b/hs-bindgen-runtime/src/HsBindgen/Runtime/Prelude.hs
@@ -1,4 +1,8 @@
 -- | Types used by the C standard library external binding specification
+--
+-- The binding specification for the types in this module is defined in
+-- @HsBindgen.BindingSpec.Stdlib@ in the @hs-bindgen:internal@ library, in the
+-- same order.
 module HsBindgen.Runtime.Prelude (
     -- * Primitive types
     -- $PrimitiveTypes
--- a/hs-bindgen/src-internal/HsBindgen/BindingSpec/Stdlib.hs
+++ b/hs-bindgen/src-internal/HsBindgen/BindingSpec/Stdlib.hs
@@ -3,6 +3,9 @@
 -- Intended for qualified import.
 --
 -- > import HsBindgen.BindingSpec.Stdlib qualified as Stdlib
+--
+-- The types for these bindings are defined in @HsBindgen.Runtime.Prelude@ in
+-- the @hs-bindgen-runtime@ library, in the same order.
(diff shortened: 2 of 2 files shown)


TravisCardwell:
Thinking about it over lunch, another option is to *remove* the `hs-bindgen/bindings` binding specification files.  Since they are used by default, perhaps their value is in reference, but YAML is a terrible format for that.  Users would still be able to inspect the YAML using the `hs-bindgen-dev` command.  Users can also use the command to save to a file that they can edit if the built-in version is causing problems.

Is keeping the `binding-spec` commands in `hs-bindgen-dev` appropriate, or should we go ahead and make them part of the public API and put them into `hs-bindgen-cli`?  Note that if they are part of the public API, we will also need to expose them in `HsBindgen.Lib`.

If we decide to keep the files and create a CI job to confirm that the files are current, we need to decide how to implement the job.  It requires building the project, so it is not cheap.  Perhaps we can have a single job that builds the project and then runs *all* such checks that require the CLI.  This is currently just checking the `base` and `hs-bindgen-runtime` binding specifications, but there might be more such checks in the future (for generated documentation).

edsko (approved):
Nice. Feel free to merge once

* Remove spec files from repo (just have them in Haskell)
* Move command to generate external files to the public executable (rather than dev)

PR #770, opened 2025-06-20
Use C origin and alias information when generating binding specification

TravisCardwell:
* Spellings `struct foo`, `union foo`, and `enum foo` are only used when they are valid spellings in C.
* Spelling `foo` is used for ordinary name kinds (`typedef` and `macro` names), as well as for generated names.


TravisCardwell:
This PR should unblock @dschrempf on the `vector` compilation.

It does not include the `BindingSpec` refactoring.  I will implement that in a separate PR, which should be the final one for #705.  Restoration of test generation will be done separately.

PR #771, opened 2025-06-20
Refactor the `.Resolve`

edsko:
This teases apart the control flow in the `Fold` from the decision making process, making subsequent changes to the `Fold` infrastructure easier.

While I was here also tidied up a code a bit in a way that I find a bit more readable. Hope that's okay.

@TravisCardwell The only "real" difference is that this code now _always_ uses `Continue` rather than `Break` whenever a particular cursor doesn't allow us to resolve the target, but I think that's actually correct. Bu  perhaps you were intentionally making a subtle distinction when we should `Break` early, but if so, I don't understand it :)

### DIFF
--- a/hs-bindgen/hs-bindgen.cabal
+++ b/hs-bindgen/hs-bindgen.cabal
@@ -190,6 +190,7 @@ library internal
     , template-haskell   >= 2.18      && < 2.24
     , text               >= 1.2       && < 2.2
     , time               >= 1.11      && < 1.15
+    , transformers       >= 0.5       && < 0.7
     , unliftio-core      ^>=0.2.1.0
     , vec                >= 0.5       && < 0.6
     , witherable         >= 0.4.2     && < 0.6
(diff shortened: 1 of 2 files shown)


TravisCardwell:
Looks good!

One thing that I noticed is that putting all conditionals at the bottom of `tryResolve` means that it now does more work that needed.  When we discussed this previously, we wanted to make resolution as fast as possible because we resolve all headers in descriptive binding specifications.  The extra work is probably just extra copying of memory, though, so it is not significant.

We developed a way to resolve any number of headers using a single translation unit (for the input headers).  Perhaps we should change `Resolve` to do something similar, as this *is* significant.  For example, resolving `libc` headers within a single translation unit would make Clang only parse internal headers once.

edsko:
> We developed a way to resolve any number of headers using a single translation unit (for the input headers). Perhaps we should change Resolve to do something similar, as this is significant. For example, resolving libc headers within a single translation unit would make Clang only parse internal headers once.

Yes, I think this is a good idea; like you said, it could make a big difference if some internal or system headers are reused. That said, this whole thing feels clunky, having to parse all of this only to then re-parse it all again... perhaps we should take another look at if we can't do that more sensibly...?

PR #775, opened 2025-06-23
Program slicing

dschrempf:


hs-bindgen/fixtures/program_slicing.pp.hs
```diff
@@ -0,0 +1,38 @@
...
+{-# LANGUAGE StandaloneDeriving #-}
+
+module Example where
+
+import qualified Foreign as F
+import qualified HsBindgen.Runtime.Prelude
+import Prelude ((<*>), (>>), Eq, Int, Show, pure)
+
+data Foo = Foo
+  { foo_sixty_four :: HsBindgen.Runtime.Prelude.Word64
```
dschrempf, line 47: Somehow, it prefers the default descriptive binding specs, which was not the case a few days ago. (I guess this is because of changes Travis has made in the meantime).

dschrempf: I guess this is about `NonSelectedDecls`, which I have not yet adapted.

dschrempf: Fixed by disabling descriptive binding specifications in this test. I also adjusted the nonSelectedDecls, but I am unsure whether this is actually required.

dschrempf: Also, I noticed that we still call those _external_ bindings specifications instead of _descriptive_. Is this on somebody's radar?

TravisCardwell: Yes.  There are a number of binding specification changes that refactor the same code and therefore should be done sequentially.  That is what I am currently working on, and the change of terminology is in the queue.

* [x] Add `C.NameKind`, `C.NameOrigin`, and aliases (#748)
* [x] Use C origin and alias info when generating a binding specification (#770)
* [ ] Refactor `BindingSpec` map, factoring out `CSpelling` (**current**)
* [ ] Refactor `ResolvedBindingSpec` to include *both* `CHeaderIncludePath` and `SourcePath`
* [ ] Refactor binding specification terminology ("prescriptive" and "descriptive")

edsko (approved):
Nice, great that this is so easy :)

PR #776, opened 2025-06-24
Factor out textual `C.Spelling`

TravisCardwell:
Binding specifications have used a textual `C.Spelling` representation as map keys.  Now that we have `C.NameKind`, this can be refactored.

The parameterized `QualId p` type is removed, replaced by its two distinct uses:

* Type `Id p ~ DeclId => QualId p` is used in the beginning passes of the frontend.  New type `QualDeclId` replaces this usage.

* Type `Id p ~ CName => QualId p` is equivalent to `(CName, C.NameKind)`, which is a parsed `C.Spelling`.  New type `C.QualName` replaces this usage.

This change improves consistency and IMHO makes the code easier to understand.

Multiple parts of the code base now use `C.QualName` as map keys.  The textual representation is largely factored out; it is parsed when binding specification files are loaded.

Parsing this added a new type of error, which made multiple types of errors possible in many functions.  That was possible in one place before, and I used a specific exception type then.  That is no longer sufficient with this change, and that is something that has been on my list to improve anyway.  I added a `MultiException` type that makes it easier to accumulate multiple errors.

Generated binding specifications are ordered differently because we the type of map keys is changed.

`ResolvedBindingSpec` is refactored to store *both* the unresolved `CHeaderIncludePath` and the resolved `SourcePath`.  We discussed refactoring so that only `ResolvedBindingSpec` is part of the public API, which requires always resolving the `stdlib` binding specification.  I have *not* implemented this yet, as it requires more changes.  I will create an issue for this, as I think we should do this after the terminology refactoring.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/IsPass.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/IsPass.hs
@@ -100,7 +100,7 @@ getDeclId curr = do
 -- | Check if the cursor is for a non-tagged declaration
 --
 -- This function returns 'True' if the cursor is for a @struct@, @union@, or
--- @enum@ declaration and the underlying tokens starts with that keyword and a
+-- @enum@ declaration and the underlying tokens start with that keyword and a
 -- left bracket.
 isNotTagged :: MonadIO m => CXCursor -> m Bool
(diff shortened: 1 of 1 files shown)


dschrempf:
I think this affects https://github.com/well-typed/hs-bindgen/pull/775.

Issue #777, opened 2025-06-24
Change how missing headers in binding specifications are handled


TravisCardwell:
Binding specifications were originally designed for specifying external bindings (descriptive binding specifications).  The initial implementation threw an exception if a header could not be resolved.  We decided to instead ignore such issues since not all environments provide the same headers.  In particular, macOS does not provide standard header `uchar.h`, which we want to include in the `stdlib` bindings.  (Note that this could alternatively be addressed by #554.)

This behavior needs to change.  For prescriptive binding specifications, a missing header should be considered an error.

TravisCardwell:
We discussed the idea of including information about how headers are resolved in YAML output.  YAML supports comments, but the library that we use (`yaml`) does not, as it is built on top of `aeson`.  Perhaps we could output the information as part of the data, which could be ignored when loading the file.

I am not (yet) a fan of the idea.  Users can check how headers are resolved using the `hs-bindgen-cli resolve` command (#774).  What if we trace all header resolution, with level `info` or perhaps `debug`?

TravisCardwell:
We would like to only expose an opaque `BindingSpec` wrapper around the internal `ResolvedBindingSpec` type.  We discussed always resolving the `stdlib` binding specification and exposing it (in `Lib`) as `stdlibBindingSpec :: IO BindingSpec`.  If all headers for a type are missing, then that type is removed, but this is probably not what we want for this use case.  The `hs-bindgen-cli binding-spec stdlib` command should output the full binding specification, IMHO, not just what is available in the current environment.

edsko:
> What if we trace all header resolution, with level info or perhaps debug?

We discussed this and agreed it would be useful.

edsko:
@TravisCardwell can this close now? At least this part

> This behavior needs to change. For prescriptive binding specifications, a missing header should be considered an error.

is now done (as of #792).

TravisCardwell:
Treating missing headers as warnings vs. errors is done.  Tracing header resolution is done, and doing this instead of adding it to the output YAML file is good (IMHO).

[...]

Issue #647, opened 2025-05-13
Unify `Tracer`


dschrempf:
At the moment, we have two domain-specific tracers (please correct me if wrong):
- a tracer for diagnostic messages (`Diagnostic`), and 
- a tracer for skipped declarations (`Skipped`).

Ideally, we also support general-purpose tracing/logging such as
- Picking up environment variables,
- Output of current stage in the pipeline,
- ...You name it!

Futher, we should use a single tracer that either writes to `stdout`, or `stderr`, or log file (or a combination of those), and specialize the tracer using `contramap`.

For example,
```haskell
-- | Parse a C header
parseCHeader :: Opts -> CHeaderIncludePath -> IO ([SourcePath], C.Header)
parseCHeader Opts{..} headerIncludePath =
    C.parseCHeaders
      (contramap show optsDiagTracer)
      (contramap prettyLogMsg optsSkipTracer)
      ...
```
uses two tracers `optsDiagTracer`, and `optsSkipTracer`, but should probably only use one, which is specialized using the `contramap`s.

phadej:
FWIW, I think we use the whole idea of "Contravariant Tracer" wrong. We currently have an ordinary logger setup (say, like `fast-logger`) but we try to pretend its something fancier.

If we change the `optsDiagTracer  :: Tracer IO String` and `optsSkipTracer` to `optsTracer :: Tracer IO ActualTraceADT`, then I'm :+1:. In this case something like `show` or `prettyLogMsg` (`prettyActualTraceADT`) is provided at the very edge of the API (by executable, or TH wrapper using `report` e.g.)

phadej:
Another issue, I think that `Level` in

```haskell
newtype Tracer m a = Wrap {
      unwrap :: Contra.Tracer m (Level, a)
    }
```

is wrong. in contravariant tracer approach it's not the log-call-site who decides the severity of a message, it's the provider of the tracer. The "severity" is intrinsic property of `a`.

edsko:
I'm fine with moving the severity logic to the tracer.

dschrempf:
> Another issue, I think that `Level` in
> 
> newtype Tracer m a = Wrap {
>       unwrap :: Contra.Tracer m (Level, a)
>     }
> is wrong. in contravariant tracer approach it's not the log-call-site who decides the severity of a message, it's the provider of the tracer. The "severity" is intrinsic property of `a`, as interpreted by whoever constructs the tracer.

[...]

PR #787, opened 2025-06-25
Refactor `BindingSpec` and `Resolve` traces/errors

TravisCardwell:
(From the commit messages)

**`BindingSpec`**

A new `TraceBindingSpec` is used.

The `Exception` types are renamed to indicate that they are traces.  (Example: `ReadBindingSpecTrace`)  They are currently all errors, but we may add other types of traces in the future.

The (upstream) YAML error types do not have `Eq` instances, but we now need `Eq` instances for the `Trace` type.  I resolved this by rendering those errors and just storing `String`s.

The functions are now even more lenient than before.  For example, an empty binding specification is returned when unable to load a file at all.

Type `MultiException` is removed, as we are now managing errors via traces.

This change means that we need tracers in more places.  Notable, the tracer is in `Opts` but not `PPOpts`.  We have discussed making `Opts` opaque and instead using new `Config` types for configuration.  Perhaps `Opts` and `PPOpts` will be combined, resolving this issue.  For now, I simply pass both where `PPOpts` was previously sufficient.

**`Resolve`**

This change is motivated by two things:

* We would like to trace header resolution, both successes and failures.
* Failure to resolve is an error in some cases and a warning in others.  (See #777)

Type `ResolveHeaderException` is removed, and the API now returns a `Maybe`, where `Nothing` indicates failure.  The caller can decide how to handle failure.

Type `ResolveHeaderTrace` is added.  Successful resolution is traced with level `info` and failure is traced with level `warning`.

----

With the way that we configure traces, we do indeed need to expose the trace types in the public API, so `Lib` and `TH` need to be updated/organized.

### DIFF
--- a/hs-bindgen/app/hs-bindgen-cli.hs
+++ b/hs-bindgen/app/hs-bindgen-cli.hs
@@ -33,33 +33,33 @@ instance Exception LiterateFileException where
 
 execMode :: Cli -> IO ()
 execMode Cli{cliGlobalOpts=GlobalOpts{..}, ..} = case cliMode of
-    ModePreprocess{..} -> genDecls >>= outputDecls
-      where
-        genDecls = withTracer $ \tracer -> do
-          extBindings <- loadExtBindings tracer
-                           globalOptsClangArgs
-                           globalOptsStdlibSpecs
-                           globalOptsExtBindings
-          let opts = cmdOpts {
-                  optsExtBindings = extBindings
-                , optsTranslation = preprocessTranslationOpts
-                , optsTracer      = tracer
-                }
-          -- to avoid potential issues it would be great to include unitid in module unique
-          -- but AFAIK there is no way to get one for preprocessor
-          -- https://github.com/well-typed/hs-bindgen/issues/502
-          let mu :: ModuleUnique
-              mu = ModuleUnique $ filter isLetter $ hsModuleOptsName $ preprocessModuleOpts
-          translateCHeaders mu opts preprocessInputs
-        outputDecls decls = do
-          let ppOpts = (def :: PPOpts) {
-                ppOptsModule = preprocessModuleOpts
-              , ppOptsRender = preprocessRenderOpts
-              }
-          preprocessIO ppOpts preprocessOutput decls
-          case preprocessGenExtBindings of
-            Nothing   -> return ()
-            Just path -> genExtBindings ppOpts preprocessInputs path decls
+    ModePreprocess{..} -> withTracer $ \tracer -> do
+      extBindings <-
+        loadExtBindings
+          tracer
+          globalOptsClangArgs
+          globalOptsStdlibSpecs
+          globalOptsExtBindings
+      let opts = cmdOpts {
+              optsExtBindings = extBindings
+            , optsTranslation = preprocessTranslationOpts
+            , optsTracer      = tracer
+            }
+          ppOpts = (def :: PPOpts) {
+              ppOptsModule = preprocessModuleOpts
+            , ppOptsRender = preprocessRenderOpts
+            }
+      -- to avoid potential issues it would be great to include unitid in module
+      -- unique but AFAIK there is no way to get one for preprocessor
+      -- https://github.com/well-typed/hs-bindgen/issues/502
+      let mu :: ModuleUnique
+          mu = ModuleUnique $
+            filter isLetter (hsModuleOptsName preprocessModuleOpts)
+      hsDecls <- translateCHeaders mu opts preprocessInputs
+      preprocessIO ppOpts preprocessOutput hsDecls
+      case preprocessGenExtBindings of
+        Nothing   -> return ()
+        Just path -> genExtBindings opts ppOpts preprocessInputs path hsDecls
 
     ModeGenTests{..} -> do
       extBindings <- withTracer $ \tracer ->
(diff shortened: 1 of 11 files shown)


hs-bindgen/src-internal/HsBindgen/Pipeline.hs
```diff
@@ -321,6 +325,7 @@ loadExtBindings tracer args stdlibSpecs =
     BindingSpec.load
       (useTrace TraceExtraClangArgs tracer)
       (useTrace TraceResolveHeader tracer)
+      (useTrace TraceBindingSpec tracer)
```
dschrempf: The provision of multiple tracers is unfortunate. What we really need is something like: I need to trace A, B, and C. Pease provide me with an object that can do that.

I guess that would mean to implement type classes like `CanTraceExtraClangArgs`, which isn't much nicer either.

The alternative was to use structured, but weaker typed traces. I.e., one trace type, but with multiple fields to provide details (Option 3 in https://github.com/well-typed/hs-bindgen/issues/647#issuecomment-2879905010).

dschrempf (approved):

hs-bindgen/src-internal/HsBindgen/Pipeline.hs
TravisCardwell: Another alternative is to just pass `Tracer IO (TraceWithCallStack Tracer)` tracers.  All `traceWithCallStack` calls would need to specify the full path of constructors, which does not work in general but would probably work in our case.

dschrempf: This would work, but then we need to be very careful to avoid circular module imports.

edsko: Sorry for being slow here, but why not just introduce a sum type here just like we do everywhere else?

dschrempf: I think we need some of those traces in multiple locations. For example, the extra clang args info messages are used all over. Of course, we can add them to the tree of traces multiple times (but then writing trace predicates like we did also becomes more problematic).

TravisCardwell:
I added documentation *to the function* noting that header resolution failure is traced as a warning because it is an error in some cases but not in others.

hs-bindgen/src-internal/HsBindgen/Pipeline.hs
edsko: Perhaps the root problem here is not the tracers at all, but rather the fact that we can have these errors in multiple contexts. It would be cleaner if we could do header resolution once and for all, for example (and more efficient).

But if that's not feasible, then I think having the sum type is the least bad option for now. It is at least consistent, and perhaps the fact that the trace emphasizes that the same error can happen in multiple places isn't actually the worst idea.

hs-bindgen/app/hs-bindgen-cli.hs
```diff
@@ -84,11 +84,14 @@ execMode Cli{cliGlobalOpts=GlobalOpts{..}, ..} = case cliMode of
...
-              Right path -> isSuccess <$ putStrLn path
-              Left  err  -> False     <$ putStrLn (displayException err)
+        let tracerClangArgs = useTrace TraceExtraClangArgs tracer
+            tracerResolve   = useTrace TraceResolveHeader  tracer
+            args            = optsClangArgs cmdOpts
+            step isSuccess header =
+              resolveHeader tracerClangArgs tracerResolve args header >>= \case
+                Just path -> isSuccess <$ putStrLn path
+                Nothing ->
+                  False <$ putStrLn ("header not found: " ++ show header)
```
dschrempf: This could actually be an `Error` trace, which would allow you to skip the error handling part. Not saying that it has to be.

edsko:
@TravisCardwell This got a bit entangled in my cleanup of the tracers, and dealing with clang errors and warnings. I will port some of the changes you've made here to my branch; I'll report back.

hs-bindgen/app/hs-bindgen-cli.hs
TravisCardwell: Header resolution is an error in some cases but not others, so using an `Error` trace is problematic.

TravisCardwell:
This seems to be the error that is causing failure to build `c-expr` in all macOS jobs:

```
<no location info>: error:
    dlopen(/Users/runner/.cabal/store/ghc-9.4.8/lib/libHSspltmx-0.1.3-7e4c0808-ghc9.4.8.dylib, 0x0005): symbol not found in flat namespace '_kSecRandomDefault'
```

[`splitmix-0.1.3`](https://hackage.haskell.org/package/splitmix-0.1.3) was released very recently, and I confirmed in the log that it is used.

hs-bindgen/src-internal/HsBindgen/Pipeline.hs
TravisCardwell: This is being resolved in #792.

hs-bindgen/app/hs-bindgen-cli.hs
TravisCardwell: With the changes in #792, we are making the default `Error`, which will be nice for cases like the above.  The level is overridden in cases where it should instead be `Warning`.

hs-bindgen/src-internal/HsBindgen/Pipeline.hs
edsko: The conclusion is that @dschrempf 's "traces in multiple locations" is actually a feature, not a bug; see https://github.com/well-typed/hs-bindgen/pull/792#issuecomment-3007575269 for details.

PR #789, opened 2025-06-25
Cleanup tracing infrastructure

edsko:
* Consistently import from `HsBindgen.Util.Tracer`
* Move `.Util.Trace` to `.TraceMsg`
* Rename the main `Trace` type to `TraceMsg`
* Make `Tracer` an opaque wrapper, so that we hide the callstack
* Rename `PrettyTrace` to `PrettyForTrace` (to emphasize that this does not just apply to whole messages)

Also ensured consistent naming of the message types. Renamed:

* `ExtraClangArgsLog` -> `ExtraClangArgsMsg`
* `FrontendTrace` -> `FrontendMsg`
* `ParseTrace` -> `ParseMsg`
* `SortError` -> `SortMsg`
* `SliceError` -> `SliceMsg`
* `MacroError` -> `HandleMacrosMsg`
* `BindingSpecError` -> `ResolveBindingSpecsMsg`
* `MangleError` -> `MangleNamesMsg`

`ResolveHeaderException` should probably change also, but doing this in a future PR, as going to change how we interact with `clang` a bit.


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Slice.hs
```diff
@@ -24,7 +24,7 @@ import HsBindgen.Frontend.Pass.Sort.IsPass (DeclMeta (declNonSelected, declUsage
 import HsBindgen.Language.C.Name qualified as C
 import HsBindgen.Util.Tracer (HasDefaultLogLevel (getDefaultLogLevel),
```
dschrempf: Here we could also reduce the imports like you did in other modules.

hs-bindgen/src-internal/HsBindgen/TraceMsg.hs
```diff
@@ -1,14 +1,16 @@
-module HsBindgen.Util.Trace (
-    Trace (..)
+-- | Top-level sum-type that captures all trace messages we might emit
+--
+-- Intended for unqualified import.
+module HsBindgen.TraceMsg (
+    TraceMsg(..)
```
dschrempf, line 5: We may want to reexport all sub-types here (although this may not even be possible because we may need to traverse multiple levels).

We have to make sure though that the sub-types are exported from `Lib`, and `TH`. Travis was mentioning that.

hs-bindgen/src-internal/HsBindgen/Util/Tracer.hs
```diff
@@ -52,6 +54,62 @@ import System.IO (Handle, IOMode (AppendMode), hPutStrLn, stdout, withFile)
...
+--
+-- This is internal API.
+simpleWithCallStack :: Applicative m => (MsgWithCallStack a -> m ()) -> Tracer m a
+simpleWithCallStack =
+      Wrap
+    . ContraTracer.Tracer
+    . ContraTracer.emit
+
+-- | See 'ContraTracer.squelchUnless'
+squelchUnless :: Monad m => (a -> Bool) -> Tracer m a -> Tracer m a
```
dschrempf, line 97: Do we even need this function?

edsko: Ah, it's used internally only. Ok, will remove the export. Well spotted.

hs-bindgen/src-internal/HsBindgen/TraceMsg.hs
edsko: Opened https://github.com/well-typed/hs-bindgen/issues/790 to track this.

PR #792, opened 2025-06-25
Introduce `HsBindgen.Clang`

edsko:
This is now the main way that we call `clang` from `hs-bindgen`.

Importantly, this also removes the treatment of clang errors as exceptions, and now uses the standard "error" trace messages.


edsko:
@TravisCardwell This includes some small parts of #787: `resolve` now uses a tracer rather than reporting an exception. It was more useful to do that as part of this PR as it fit the new structure much better.

I apologize for the many conflicts this has with your PR; hopefully they should not be too difficult to resolve, and hopefully you'll agree that this new structure is better.

hs-bindgen/test/internal/Test/HsBindgen/Clang/Args.hs
```diff
@@ -73,4 +73,4 @@ getExtraClangArgsTests = testGroup "getExtraClangArgs" [
   where assertExtraClangArgs :: [(EnvVar, Content)] -> Maybe Target -> [String] -> IO ()
         assertExtraClangArgs xs mtarget x = do
           withTracePredicate defaultTracePredicate $ \tracer ->
-            assertWithEnv xs (getExtraClangArgs (contramap TraceExtraClangArgs tracer) mtarget) x
+            assertWithEnv xs (getExtraClangArgs tracer mtarget) x
```
edsko, line 76: @dschrempf I am not 100% sure if this change is correct, I was a bit confused by this code. But the tests pass, so I guess it's okay..?

TravisCardwell (approved):
Looks good.  Perhaps we can implement an option for `withClang` that allows it to run without displaying/checking diagnostics, for use by `resolveHeader`?  The `Frontend` could continue to check diagnostics.

TravisCardwell:
~We should probably `cherry-pick` commit [`bf269e1d`](https://github.com/well-typed/hs-bindgen/pull/787/commits/bf269e1d180966b9082281f7ac48d6b6f0e4f3d1) (from `tcard/bindings-spec-errors`).  ([details](https://github.com/well-typed/hs-bindgen/pull/787#issuecomment-3006453916))~

This commit was merged in #796.

hs-bindgen/test/internal/Test/HsBindgen/Clang/Args.hs
dschrempf: It should be ok. The `assertWithEnv` function sets the `BINDGEN_EXTRA_CLANG_ARGS` environment variable (and it's cross-compilation derivatives), runs the action, and unsets the variables again.

edsko:
> Looks good. Perhaps we can implement an option for withClang that allows it to run without displaying/checking diagnostics, for use by resolveHeader? The Frontend could continue to check diagnostics. 

We now do this differently. The trace message records contexts (where needed), which allows us to do things like this:

```hs
instance HasDefaultLogLevel ResolveBindingSpecMsg where
  getDefaultLogLevel = \case
      ResolveExternalBindingSpecHeader _x ->
        -- Any errors that happen while resolving /external/ headers are 'Info'
        -- only: the only consequence is that those headers will then not match
        -- against anything (and we might generate separate warnings/errors
        -- for that anyway while resolving the binding specification).
        Info
```

This logic was previously just a comment, now it is explicit.

PR #794, opened 2025-06-25
Resolve #279. Cleanup DVar treatment.

### LINKED ISSUE #279: Split DVar into two declarations
- the variable binding
- and the type annotation

Then we can return to having one-to-one translations.

phadej:
Treat DVar as "a binding (with type)".
I opened #279 because I didn't like that it was the only statement, which required two TH declarations when translated. It's not true anymore, as pattern synonyms also require two TH declarations (similarly, type and "term").

I think it's fine to have it this way, and by *always* requring the type we avoid conditional treatment.
If we would ever need top-level bindings ("vars") without type annotations (which I think we shouldn't need), then DVar could be split into type-signature and declaration.


edsko:
Ok, sounds good. I will close #279 for now then.

PR #802, opened 2025-06-26
Common module for exports shared between Lib and TH

dschrempf:
This is an attempt to shed some light on the exports from `Lib` and `TH` by introducing a (hidden) `Common` module that contains all declarations exported by both, `Lib`, and `TH`.

Direct reexport of `Common` is, sadly, not possible because we loose the headings in the Haddock.

However, like this we can see in the export lists of `Lib`, and `TH`, which exports are specific to the module, and which exports are common to both modules.

It's not perfect, and may be a controversial change. Let me know what you think.

### DIFF
--- a/hs-bindgen/clang-ast-dump/Main.hs
+++ b/hs-bindgen/clang-ast-dump/Main.hs
@@ -87,7 +87,7 @@ clangAstDump opts@Options{..} = do
   where
 
     tracerConf :: TracerConf
-    tracerConf = defaultTracerConf {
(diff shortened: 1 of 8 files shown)


dschrempf:
We still need to fix tracer reexports (only export what is really necessary, and do so from Common).

edsko (approved):
Nice. As discussed, I think you ended up with the best possible solutions out of only suboptimal solutions. Feel free to merge once the `Tracer` stuff is minimized a bit.

PR #816, opened 2025-06-27
Simple zero copy demo

edsko:


dschrempf:
I just had a look at this. Do I understand correctly:

The advantage of this approach is that we can directly refer to inner elements of nested structures/unions, without the need to first extract them level-by-level. This is because we provide composable, optics-like getters and setters (peekField/pokeField).

I like the approach. The generated code looks much cleaner also, but I guess this is partly because it was written by you :-).

bolt12:
I also took a look at this, looks nice, but I also wonder about if there are any cons of making a heavier use of type classes for larger code bases. I think it looks nicer and more correct in this way, but I think I find the currently generated code, more straightforward to read and a more direct translation of the C code. If the idea is to have a high level layer on top of the low-level layer anyway, is it worth the hassle of how things are generated ?

dschrempf:
I think in this case it is also about speed. Copying fewer values is faster. But I totally see your point!

bolt12:
Oh that's right, I completely overlooked that! I think that's enough of a reason to do this then!

edsko:
Obsoleted by #1247.

PR #820, opened 2025-06-27
`withTracer` functions return Maybe

### LINKED ISSUE #791: Idea: Make exception handling of tracers more explicit by providing an `onFailure` argument
At the moment, we throw an exception when we encounter an error trace. We do this before writing any output.

That is, the output and the exit code of `hs-bindgen` are determined in a rather implicit way.

We discarded that `withTracer...` returns an `Either Error Result` because it is rather annoying to perform exception handing at every use site.

Instead, we could add an `onFailure` argument to `withTracer...`. Then, it is explicit to the user what is going to happen on errors.

dschrempf:
Closes #791

### DIFF
--- a/hs-bindgen/app/HsBindgen/App/Common.hs
+++ b/hs-bindgen/app/HsBindgen/App/Common.hs
@@ -294,10 +294,18 @@ parseInputs = some . argument (eitherReader parseHeader) $ mconcat [
   Auxiliary hs-bindgen functions
 -------------------------------------------------------------------------------}
 
+-- | Run an action with a tracer.
+--
+-- Return 'Nothing' if errors happened.
 withTracer :: GlobalOpts -> (Tracer IO TraceMsg -> IO b) -> IO (Maybe b)
 withTracer GlobalOpts{..} =
     withTracerStdOut globalOptsTracerConf DefaultLogLevel
 
+-- | Extract the result or exit gracefully with an error message.
+--
+-- Helper function to be used in conjunction with 'withTracer'. We carefully
+-- separate running actions from error handling; before we continue to process
+-- the result.
 fromMaybeWithFatalError :: MonadIO m => Maybe b -> m b
 fromMaybeWithFatalError k = maybe fatalError pure k
 
--- a/hs-bindgen/src-internal/HsBindgen/Util/Tracer.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Util/Tracer.hs
@@ -227,7 +227,7 @@ withTracerCustom
 withTracerCustom ansiColor tracerConf customLogLevel report action =
   nothingOnError (withTracerCustom' ansiColor tracerConf customLogLevel report action)
 
--- | Report that errors happened while generating bindings and exit with failure.
+-- | Report that errors happened and exit with failure.
 fatalError :: MonadIO m => m a
 fatalError = liftIO $ do
   putStrLn "An error happened (see above)"


dschrempf:
I have amended this PR so it contains all points we have discussed:
- The `withTracer` functions return a `Maybe` type
- There is a function `fatalError` which can be used to handle the `Nothing`
  branch (Lib-mode only)

You may want to merge, @edsko , but we can also discuss on Tuesday.

I noticed that `hs-bindgen-cli/execMode/ModePreprocess` uses two tracers; one for getting the declarations, and another for generating the binding specifications. In particular, `genBindingSpec` uses a tracer and directly writes to files. We may want to change this in the future.

hs-bindgen/app/hs-bindgen-cli.hs
```diff
@@ -53,7 +53,8 @@ execMode Cli{cliGlobalOpts=GlobalOpts{..}, ..} = case cliMode of
               , optsTracer         = tracer
               }
         translateCHeaders mu opts preprocessInputs
-      withTracer $ \tracer -> do
+      hsDecls <- maybe fatalError pure maybeHsDecls
+      maybeSuccess <- withTracer $ \tracer -> do
```
dschrempf: This is suboptimal. Here, we create output, but creation of the prescriptive binding specifications may still fail.

hs-bindgen/src-internal/HsBindgen/Util/Tracer.hs
```diff
@@ -219,46 +207,38 @@ withTracerStdOut ::
...
   -> CustomLogLevel a
   -> (String -> m ())
   -> (Tracer m a -> m b)
-  -> m b
+  -> m (Maybe b)
 withTracerCustom ansiColor tracerConf customLogLevel report action =
-  exceptionOnError (withTracerCustom' ansiColor tracerConf customLogLevel report action)
+  nothingOnError (withTracerCustom' ansiColor tracerConf customLogLevel report action)
+
+-- | Report that errors happened while generating bindings and exit with failure.
```
dschrempf: adjust comment

hs-bindgen/app/HsBindgen/App/Common.hs
```diff
@@ -282,10 +284,13 @@ parseInputs = some . argument (eitherReader parseHeader) $ mconcat [
   Auxiliary hs-bindgen functions
 -------------------------------------------------------------------------------}
 
-withTracer :: GlobalOpts -> (Tracer IO TraceMsg -> IO b) -> IO b
+withTracer :: GlobalOpts -> (Tracer IO TraceMsg -> IO b) -> IO (Maybe b)
 withTracer GlobalOpts{..} =
     withTracerStdOut globalOptsTracerConf DefaultLogLevel
 
+fromMaybeWithFatalError :: MonadIO m => Maybe b -> m b
```
dschrempf, line 309: comment

hs-bindgen/app/HsBindgen/App/Common.hs
```diff
@@ -282,10 +284,13 @@ parseInputs = some . argument (eitherReader parseHeader) $ mconcat [
   Auxiliary hs-bindgen functions
 -------------------------------------------------------------------------------}
 
-withTracer :: GlobalOpts -> (Tracer IO TraceMsg -> IO b) -> IO b
+withTracer :: GlobalOpts -> (Tracer IO TraceMsg -> IO b) -> IO (Maybe b)
```
dschrempf, line 300: comment (scope matters)

hs-bindgen/app/hs-bindgen-cli.hs
```diff
@@ -39,15 +39,15 @@ execCli Cli{..} = case cliCmd of
...
     preprocessIO ppOpts preprocessOutput hsDecls
     case preprocessGenBindingSpec of
-      Nothing   -> return ()
-      Just path -> withTracer globalOpts $ \tracer ->
-        genBindingSpec tracer ppOpts preprocessInputs path hsDecls
-
+      Nothing   -> pure ()
+      Just path ->  fromMaybeWithFatalError =<<
+        withTracer globalOpts (\tracer ->
+          genBindingSpec tracer ppOpts preprocessInputs path hsDecls)
```
dschrempf, line 48: We are going to separate binding spec generation and output generation @TravisCardwell so that we can use one tracer, and do not generate any output if there are errors.

dschrempf: We are going to open a ticket.

TravisCardwell: I do not follow.  The current code does the following:

1. Resolve binding specs and translate, then exit on error
2. Run `PP` backend, which cannot error (aside from IO which we do not handle)
3. When binding spec generation requested, do so and exit on error

What is the desired behavior?  Perhaps you mean to write "combine" instead of "separate?"  If we generate the binding spec before running the `PP` backend, then we would not generate any output if there are errors.

dschrempf: Sorry, we should have been more specific but were a bit pressed on time.

The idea is to only produce output when there is no error. That is, we want to generate the bindings and the binding specs, then exit if there was an error, and only if no error has happened, we go on to writing the bindings and the binding specs.

hs-bindgen/app/hs-bindgen-cli.hs
dschrempf: We discuss this in a separate conversation below.

hs-bindgen/app/hs-bindgen-cli.hs
TravisCardwell: No problem!

Generating the binding spec before running the `PP` backend would resolve the issue, correct?

I could refactor the code to return `ByteString` results that can be output only when we know that there are no errors.  Simply changing the order would let us avoid this, though.

Currently, the only possible error when writing a binding specification is if the user specifies a filename with an unknown extension (aside from `IO` errors of course).  Another option is to refactor it so that the filename is parsed separately.  We could even parse the filename within the `Frontend`/`Hs` call to `withTrace` so that any error is always output.  Alternatively, we could parse binding spec filenames as part of the CLI parser, the logic being that such errors are usage errors.

edsko: Let's go with the latter option. Observations: generating a binding specification for bindings _we_ generate should _never_ result in errors; if there are unsupported features, that _could_ result in errors, we should detect those already in the frontend. 

The validation of the filename is such an example of something that could be validated in the frontend, so the second option here seems the best. That said, not convinced we need to validate this at all; if the user wants to call their file `bindingspec.yabadabadoo`, do we really care..?

Either way, we can address this in a separate PR.

dschrempf: This will be handled in a separate issue #854

dschrempf:
@edsko I am going to merge this if it is green. I think we resolved all issues, and outsourced separation of the `genBindingSpecs` function.

Issue #377, opened 2025-01-29
Support extern declaration of array of unknown size


edsko:
Example:

```c
extern int x[]; // the type of x is "array of unknown bound of int"
```

See https://en.cppreference.com/w/c/language/array#Arrays_of_unknown_size

phadej:
#42 is a prerequisite.

edsko:
Indeed, that's why this uses the fancy new GitHub feature of "subissues" 😬

edsko:
The [`bindings-DSL` manual](https://github.com/rethab/bindings-dsl/wiki/GlobalVariablesAndConstants) lists some examples also:

```c
extern int numbers[];
extern char name[];
extern char *another_name;
```

it also explains the difference between the array and the pointer:

> The difference between an array and a pointer is that arrays are always read-only pointers.

We need to think about how we want to reflect this on the Haskell side. 

(For functon arguments the distinction is less critical, because it primarily imposes a constraint on the _implementation_ of the function, which is in C, not Haskell.)

edsko:
> The difference between an array and a pointer is that arrays are always read-only pointers.

Some tests suggests that this is simply not true; given `api.h`

```c
extern int arr[];
```

and `impl.c`

```c
int arr[5] = {1, 2, 3, 4, 5};
```

The following code is accepted by the C compiler (even with `-Wall`), and runs just fine:

```c
#include <stdio.h>

#include "api.h"

int main() {
  arr[0] = 1234;
  for(int i = 0; i < 5; i++) {
    printf("%d\n", arr[i]);       
  }
}
```

The `gcc` manual says  in [16.5 Incomplete Array Types
](https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Incomplete-Array-Types.html)

> An array is equivalent, for most purposes, to a pointer to its zeroth element.  (..) These declarations are examples of incomplete array types, types that are not fully specified. The incompleteness makes no difference for accessing elements of the array, but it matters for some other things. For instance, sizeof is not allowed on an incomplete type.

with no mention of "read only", nor does the section on [Arrays of unknown size](https://en.cppreference.com/w/c/language/array.html#Arrays_of_unknown_size) on `cppreference.com` say anything about it. I think we should be able to just use a `Ptr`.

One thing we should verify that we do correctly is examples with _multiple_ dimensions, such as this one (example from the `gcc` manual):

```
struct chesspiece *funnyboard[][8];
```

Only the first dimension can be unknown.

[...]

PR #834, opened 2025-06-28
Add support for global variables

### LINKED ISSUE #42: Generate bindings for global variables
When a C header defines a global variable:

```c
extern int x;
```

we should generate bindings for `x`, perhaps `setX :: CInt -> IO ()` and `getX :: IO CInt`.

Generating bindings for global _constants_ is tracked in a [separate ticket](https://github.com/well-typed/hs-bindgen/issues/41).

### LINKED ISSUE #684: Handle type declarations nested in function declarations
The following declaration declares a function signature together with a named structure
```c
void declared_inline_named(struct inline_named { int x_5; } * arg);
```

The named structure is declared inline and unavailable outside of the function signature definition.

By default, `libclang` issues a warning when parsing such a declaration. For example,

``` text
[Warning] [Libclang ] Diagnostic {diagnosticFormatted = "./signature.h:27:35: warning: declaration of 'struct inline_named' will not be visible outside of this function [-Wvisibility]", diagnosticSeverity = simpleEnum CXDiagnostic_Warning, diagnosticLocation = "./signature.h:27:35", diagnosticSpelling = "declaration of 'struct inline_named' will not be visible outside of this function", diagnosticOption = Just "-Wno-visibility", diagnosticDisabledBy = Just "-Wvisibility", diagnosticCategory = 2, diagnosticCategoryText = "Semantic Issue", diagnosticRanges = [], diagnosticFixIts = [], diagnosticChildren = [Diagnostic {diagnosticFormatted = "hs-bindgen-root.h:1:10: note: in file included from hs-bindgen-root.h:1:", diagnosticSeverity = simpleEnum CXDiagnostic_Note, diagnosticLocation = "hs-bindgen-root.h:1:10", diagnosticSpelling = "in file included from hs-bindgen-root.h:1:", diagnosticOption = Nothing, diagnosticDisabledBy = Nothing, diagnosticCategory = 0, diagnosticCategoryText = "", diagnosticRanges = [], diagnosticFixIts = [], diagnosticChildren = []}]}
```

Futher, `libclang` does not return the declaration affected by the warning.

### Problem

This results in a declaration missing in our usage-definition graph and a runtime error during parsing:

``` text
"Unknown key: QualId (DeclNamed \"inline_named\") NamespaceStruct"
```

### Discarded solution

We also decided against recursing into parsing function parameters, because doing so leads to duplicate declarations in all other cases. For example, parsing

``` c
void declared_inline_anon(struct { int x_6; } * arg);
```

leads to three declarations
- 2x `DeclAnon (AnonId "./signature.h:29:27")`, and
- `DeclNamed "declared_inline_anon"`, which in turn leads to

``` text
"duplicate declaration for QualId (DeclAnon (AnonId \"./signature.h:29:27\")) NamespaceStruct"
```

### Possible solution

Detect this the `libclang` warning and panic with an informative message. (However, we can not rely on `libclang` always issuing the warning because users may disable warnings using custom arguments).

edsko:
This also adds plumbing for constants in the front-end, but does not yet deal with them in the backend.

As I was working on this, I also realized that we resolved #684 in a slightly suboptimal way: in principle we _can_ support _named_ declarations inside function signatures, and WG14-N3037 suggests that perhaps we should. We therefore leave it up to the C compiler to warn for this case or not; either way, we can generate code. One benefit of fixing this is that the parser for global variables and the parser for function signatures is now very very similar.

Closes #42


edsko:
Perhaps this isn't quite right yet; https://github.com/analogdevicesinc/no-OS/blob/855c4b3c34f2297865e448661ba4fcc0931bf430/drivers/rf-transceiver/talise/firmware/talise_stream_binary.h#L322-L325 is an example of a global variable that is _not_ extern..? :thinking:

edsko:
Still to do:

- [x] Make sure all examples from #42 work (all examples work except the ones requiring arrays of unknown size)
- [x] In particular, make sure non-extern globals work (and is documented carefully)
- [ ] Support arrays of unknown size (#377); not strictly necessary as part of this PR, but I think easy to do.

Edit: turns out arrays of unknown size _do_ require a bit of thought; see https://github.com/well-typed/hs-bindgen/issues/377#issuecomment-3048803269 .

PR #836, opened 2025-06-30
Build manual on Ubuntu, MacOs and Windows

bolt12:


edsko:
@bolt12 This looks good to me, thank for your all the clear comments, much appreciated! I'm going to merge this as-is, because I want to make some modifications to the manual, which would conflict with this PR. I haven't looked at the details of the CI integration and the Windows/OSX options; @TravisCardwell if you could review those, and then perhaps we can fix them after the fact?

Issue #807, opened 2025-06-26
Improve evaluation of `--select-by-element-name`


edsko:
Currently we evaluate this against the `clang_getCursorSpelling`, but we should ideally evaluate it against the "C spelling"; that is, `struct foo` rather than `foo`. This is backwards compatible (the regexp `foo` would still match `struct foo` also) but would allow users to be more precise. 

We should do #806 first.

edsko:
This should use the `QualName` introduced by @TravisCardwell , which I think should have a function to construct the "C spelling" string already?

TravisCardwell:
> This should use the `QualName` introduced by [@TravisCardwell](https://github.com/TravisCardwell) , which I think should have a function to construct the "C spelling" string already?

`C.qualNameText :: QualName -> Text` renders, and `C.parseQualName :: Text -> Maybe QualName` parses.

PR #837, opened 2025-06-30
Pure selection predicates

### LINKED ISSUE #806: Make evaluation of selection predicate pure
This is necessary for #784.

dschrempf:
Closes #806


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -27,22 +29,26 @@ import HsBindgen.Language.C qualified as C
...
 foldDecl :: HasCallStack => Fold ParseDecl [C.Decl Parse]
-foldDecl = foldWithHandler handleTypeException $ \curr ->
-    evalPredicate curr >>= \case
+foldDecl = foldWithHandler handleTypeException $ \curr -> do
+    info <- getDeclInfo curr
+    mNameKind <- dispatch curr $ return . C.toNameKindFromCXCursorKind
+    let mQualName = QualName <$> (C.isNamedDecl $ C.declId info) <*> mNameKind
+    evalPredicate (C.declLoc info) mQualName >>= \case
       Right () ->
+        -- TODO: This gets the CXCursorKind again.
```
dschrempf: Note, we extract the cursor kind twice.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -27,22 +29,26 @@ import HsBindgen.Language.C qualified as C
...
         dispatchFold curr $ \case
           CXCursor_AlignedAttr        -> attribute
-          CXCursor_EnumDecl           -> enumDecl
-          CXCursor_FunctionDecl       -> functionDecl
+          CXCursor_EnumDecl           -> enumDecl info
```
dschrempf, line 43: I tried to pass on the declaration info. Some functions are also used recursively, and so I didn't pass the top level `info` object, but let the function extract it's own `info` object. This may be a little slower than it could be, but is safer.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/Monad.hs
```diff
@@ -190,6 +180,12 @@ recordNonSelectedDecl curr = do
         -- use descriptive binding specification with anonymous declarations, we
         -- __must__ select these declarations.
         DeclAnon{} -> return ()
+      -- We intentionally do selection as part of parsing, rather than a
```
dschrempf, line 184: I moved the note here, since I extracted the conversion function from `CXCursorKind` to `NameKind`.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type/DeclId.hs
```diff
@@ -0,0 +1,83 @@
+module HsBindgen.Frontend.Pass.Parse.Type.DeclId (
```
dschrempf, line 1: Before implementing #807 , I used `DeclId` in the predicate `match` function. To avoid circular module dependencies, I created a separate module for `DeclId`. Then, I moved on to using `QualName`, and this separate module is not strictly necessary anymore. However, it can stand on its own, so I left it for now.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
edsko: I'm not worried about that. It's just a field read on the C(++) structure. If it yields more readable code, then go for it.

dschrempf: I changed it to a NOTE.

hs-bindgen/src-internal/HsBindgen/Language/C/Name.hs
```diff
@@ -82,11 +84,20 @@ data NameKind =
     --
     -- An enum name is written with an @enum@ prefix.
   | NameKindEnum
-  deriving stock (Show, Eq, Ord, Generic)
+  deriving stock (Show, Eq, Ord, Bounded, Enum, Generic)
 
 instance PrettyForTrace NameKind where
   prettyForTrace = showToCtxDoc
 
+toNameKindFromCXCursorKind :: CXCursorKind -> Maybe NameKind
```
dschrempf, line 93: Needs to be changed. We also want to select functions.

hs-bindgen/src-internal/HsBindgen/C/Predicate.hs
```diff
@@ -50,7 +56,7 @@ data Predicate =
 
     -- | Match element against regex
   | SelectByElementName Regex
-  deriving (Show)
+  deriving (Show, Eq)
 
 instance Semigroup Predicate where
```
dschrempf: Remove semigroup instance, use conjunction for negative predicates, disjunction for positive predicates (in the client).

hs-bindgen/src-internal/HsBindgen/Language/C/Name.hs
dschrempf: @edsko We stopped discussing this yesterday. We need to convert from `CXCursorKind` to `NameKind`. We do this already when recording non-selected declarations `recordNonSelectedDecl`, and now also for all declarations beofre evaluating the selection predicate.

Do you want to use two different functions for this conversion, or should we just add another clause from `CXCursorKind` to `NameKind` covering function declarations?

edsko: I think having a single function with more clauses makes sense. My  only slight hesitation is whether this is a sign that the separation of concerns isn't quite right.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
dschrempf: Fixed, we now separate parse and skip actions.

hs-bindgen/src-internal/HsBindgen/Language/C/Name.hs
dschrempf: We extended this function and added an explicit panic if we have not selected the 'Just' or 'Nothing' case for a particular 'CXCursorKind'.

hs-bindgen/src-internal/HsBindgen/C/Predicate.hs
dschrempf: If we provide the merge function (with dis- and conjunctions) in the client, it is harder (but perfectly possible) to construct selection predicates when using the library. Is that on our radar? @edsko

edsko (approved):
Nice, will be good to get this in :) Feel to merge once the final changes are done.

PR #840, opened 2025-07-01
Refactor `optparse-applicative` usage

TravisCardwell:
The primary change is the separation of commands (implemented using sum types) and options (implemented using product types).  This allows us to easily dispatch commands to separate functions that implement them.

Note that we previously used (internal) terminology *mode*, which I refactored to *command*.  This is consistent with `optparse-applicative` terminology and works better with hierarchical commands, which I assume we will have more of as we develop more CLI utilities.  I had already implemented `binding-spec` as a hierarchical command since we plan on providing various utilities for working with binding specifications.

The `gentests` command had some issues, as a result of refactoring code that is not actively developed/tested/used.  It used a tracer for loading binding specifications but not for generating tests (where the default `nullTracer` was used), and the `ModuleUnique` was `TODO`.  This PR resolves both of those issues.

This PR also fixes the `HEADER...` syntax, used to indicate that one or more headers may be specified.  `optparse-applicative` does not get this right when just using `some`.

### DIFF
--- a/hs-bindgen/app/HsBindgen/App/Cli.hs
+++ b/hs-bindgen/app/HsBindgen/App/Cli.hs
@@ -1,14 +1,14 @@
 {-# LANGUAGE ApplicativeDo #-}
 module HsBindgen.App.Cli (
     Cli(..)
-  , CliMode(..)
-  , PreprocessMode(..)
-  , GenTestsMode(..)
-  , LiterateMode(..)
-  , BindingSpecMode(..)
-  , ResolveMode(..)
+  , CliCmd(..)
+  , PreprocessOpts(..)
+  , GenTestsOpts(..)
+  , LiterateOpts(..)
+  , BindingSpecCmd(..)
+  , ResolveOpts(..)
   , getCli
-  , pureParseModePreprocess
+  , pureParseCmdPreprocess
   ) where
 
 import Data.Default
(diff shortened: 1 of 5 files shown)


hs-bindgen/app/HsBindgen/App/Cli.hs
```diff
@@ -37,110 +41,146 @@ getCli = customExecParser p opts
...
-pureParseModePreprocess =
-      getParseResult
-    . execParserPure (prefs subparserInline) (info parseCli mempty)
-    . ("preprocess" :)
-
-{-------------------------------------------------------------------------------
-  Mode selection
--------------------------------------------------------------------------------}
+data CliCmd =
+    -- | The main command: preprocess C headers to Haskell modules
```
dschrempf: We may want to remove the description here since it is repeated below. (Albeit this one is developer-facing, while the other one is user-facing).

dschrempf (approved):
Thanks, this looks very nice!

It will conflict somewhat with #820 , which adds explicit error handling to `withTracer` functions.

hs-bindgen/app/HsBindgen/App/Cli.hs
TravisCardwell: It also conflicts with #841.  I think we should merge both of those first, and I can rebase.

edsko (approved):
Nice, good to see this cleaned up.

PR #853, opened 2025-07-02
Introduce `ExtBinding` type family

edsko:
This is instantiated to `Void` prior to `ResolveBindingSpec`, and to a new record type `ResolvedExtBinding` thereafter, making it clear when we have resolved external bindings and when we have not.

This also does some code cleanup made possible by having just one argument, and removes a bit of code duplication.


edsko:
cc @TravisCardwell

PR #856, opened 2025-07-02
Fix detection of anonymous structs

### LINKED ISSUE #795: Detect "anonymous" declarations
We need to determine *valid* C spellings.  For this purpose, we consider a declaration to be "anonymous" if there is no valid C spelling for it.  For example, a `struct` declaration is "anonymous" if no tag is specified.

* In `typedef struct foo { ... } foo;` the `struct` declaration is *not* anonymous because we can write `struct foo`.  We "collapse" the `typedef` by default since the `typedef` name is the same as the tag, creating "alias" `foo`, but note that users will be able to configure different behavior.

* In `typedef struct { ... } foo;` the `struct` declaration is anonymous even though there is an "alias" `foo`.  We must know that `foo` is a valid spelling while `struct foo` is invalid.

Clang has a different definition of "anonymous" that does not match our needs.  For example, in `typedef struct { ... } foo;` the `struct` declaration is not anonymous because the *type* may be referenced via the `typedef` name `foo`.

We did not have problems with older versions of Clang because the spelling of the declaration was an empty string when there was no tag.  Newer versions of Clang no longer have this behavior.  Testing confirms that our usage of `clang_Cursor_isAnonymous` never worked for our needs even in older versions; the spelling check was what made it work for us.  We need a new solution now that we can no longer rely on the spelling check.

In #748, I implemented a hack that checks the tokens for a keyword (such as `struct`) followed by an opening brace.  This does *not* work when there are attributes.

It turns out that Edsko worked on this problem before, and the code still exists in [`Clang.HighLevel.UserProvided`](https://github.com/well-typed/hs-bindgen/blob/20a40e85d052e7cfe74d84fc6e03255e5efb3583/clang/src/Clang/HighLevel/UserProvided.hs).  Note that in the following tests I am only concerned with the `UserProvided` constructor, to distinguish between anonymous and named cases.  It only fails when a macro is used to determine the tag name.

### `tagged.h`

```c
struct foo {
  char c;
  int  i;
};
```

Clang 15 and 20:

* `clang_getCursorSpelling`: `"foo"`
* `clang_Cursor_isAnonymous`: `False`
* `UserProvided` name: `UserProvided "foo"` ✓
    * Name range: line 1, columns 8-11 (`foo`)

Relevant tokens:

* `CXToken_Keyword "struct"`
* `CXToken_Identifier "foo"`
* `CXToken_Punctuation "{"`

### `tagged_macro.h`

```c
#define STRUCT(name) name##_s

struct STRUCT(foo) {
  char c;
  int  i;
};
```

Clang 15 and 20:

* `clang_getCursorSpelling`: `"foo_s"`
* `clang_Cursor_isAnonymous`: `False`
* `UserProvided` name: `LibclangProvided "foo_s"` ✗
    * Name range: line 3, columns 8-19 (`STRUCT(foo)`)

Relevant tokens:

* `CXToken_Keyword "struct"`
* `CXToken_Identifier "STRUCT"`
* `CXToken_Punctuation "("`
* `CXToken_Identifier "foo"`
* `CXToken_Punctuation ")"`
* `CXToken_Punctuation "{"`

### `tagged_attr.h`

```c
struct __attribute__((__packed__)) foo {
   char c;
   int  i;
};
```

Clang 15 and 20:

* `clang_getCursorSpelling`: `"foo"`
* `clang_Cursor_isAnonymous`: `False`
* `UserProvided` name: `UserProvided "foo"` ✓
    * Name range: line 1, columns 36-39 (`foo`)

Relevant tokens:

* `CXToken_Keyword "struct"`
* `CXToken_Keyword "__attribute__"`
* `CXToken_Punctuation "("`
* `CXToken_Punctuation "("`
* `CXToken_Identifier "__packed__"`
* `CXToken_Punctuation ")"`
* `CXToken_Punctuation ")"`
* `CXToken_Identifier "foo"`
* `CXToken_Punctuation "{"`

### `tagged_attr_macro.h`

```c
#define PACKED __attribute__((__packed__))

struct PACKED foo {
   char c;
   int  i;
};
```

Clang 15 and 20:

* `clang_getCursorSpelling`: `"foo"`
* `clang_Cursor_isAnonymous`: `False`
* `UserProvided` name: `UserProvided "foo"` ✓
    * Name range: line 3, columns 15-18 (`foo`)

Relevant tokens:

* `CXToken_Keyword "struct"`
* `CXToken_Identifier "PACKED"`
* `CXToken_Identifier "foo"`
* `CXToken_Punctuation "{"`

### `anon.h`

```c
typedef struct {
  char c;
  int  i;
} foo;
```

Clang 15:

* `clang_getCursorSpelling`: `""`
* `clang_Cursor_isAnonymous`: `False`
* `UserProvided` name: `LibclangProvided ""` ✓
    * Name range: line 1, columns 9-15 (`struct`)

Clang 20:

* `clang_getCursorSpelling`: `"foo"`
* `clang_Cursor_isAnonymous`: `False`
* `UserProvided` name: `LibclangProvided "foo"` ✓
    * Name range: line 1, columns 9-15 (`struct`)

Relevant tokens:

* `CXToken_Keyword "struct"`
* `CXToken_Punctuation "{"`

### `anon_attr.h`

```c
typedef struct __attribute__((__packed__)) {
  char c;
  int  i;
} foo;
```

Clang 15:

* `clang_getCursorSpelling`: `""`
* `clang_Cursor_isAnonymous`: `False`
* `UserProvided` name: `LibclangProvided ""` ✓
    * Name range: line 1, columns 9-15 (`struct`)

Clang 20:

* `clang_getCursorSpelling`: `"foo"`
* `clang_Cursor_isAnonymous`: `False`
* `UserProvided` name: `LibclangProvided "foo"` ✓
    * Name range: line 1, columns 9-15 (`struct`)

Relevant tokens:

* `CXToken_Keyword "struct"`
* `CXToken_Keyword "__attribute__"`
* `CXToken_Punctuation "("`
* `CXToken_Punctuation "("`
* `CXToken_Identifier "__packed__"`
* `CXToken_Punctuation ")"`
* `CXToken_Punctuation ")"`
* `CXToken_Punctuation "{"`

### `anon_attr_macro.h`

```c
#define PACKED __attribute__((__packed__))

typedef struct PACKED {
  char c;
  int  i;
} foo;
```

Clang 15:

* `clang_getCursorSpelling`: `""`
* `clang_Cursor_isAnonymous`: `False`
* `UserProvided` name: `LibclangProvided ""` ✓
    * Name range: line 3, columns 9-15 (`struct`)

Clang 20:

* `clang_getCursorSpelling`: `"foo"`
* `clang_Cursor_isAnonymous`: `False`
* `UserProvided` name: `LibclangProvided "foo"` ✓
    * Name range: line 3, columns 9-15 (`struct`)

Relevant tokens:

* `CXToken_Keyword "struct"`
* `CXToken_Identifier "PACKED"`
* `CXToken_Punctuation "{"`

edsko:
Fix #795

This fix only works for `clang >= 19.1.0`, due to a bug in earlier versions dealing with locations (https://github.com/llvm/llvm-project/pull/72400). See `Clang.HighLevel.UserProvided.clang_getCursorSpelling` for a detailed discussion of the fix.

In order to properly detect the `clang` version (and be able to distinguish between clang 19.1 and before), I needed to change how we get the `clang` version number. Since we're now parsing something that is explicitly described as "should not be parsed", I've added some tests and some plausibility checks (unfortunately, `clang` says that the version string should not be parsed but does not provide an feasible alternative).


edsko:
If I am right, this PR should fail only for llvm-20 (currently CI is broken; https://status.canonical.com/).

edsko:
cc @TravisCardwell

PR #861, opened 2025-07-03
Parse binding spec paths separately

TravisCardwell:
This enables checking for (unknown extension) errors before any output is written.

A new `BindingSpec.Path` type represents the parsed paths, which are parsed by a `BindingSpec.parsePath` function.  Function `writeFile` no longer needs a tracer, as it does not throw any errors (aside from IO). Function `readFile` is changed for consistency, and parsing is done within `BindingSpec.load`.  The functions for writing specific formats are no longer exported; they are now defined within the `BindingSpec.read` and `BindingSpec.write` functions.

All trace message types need to re-exported via the public API, probably using a module that exports them all.  They `BindingSpec` module is designed for qualified import, but the trace message types should have names for non-qualified import.  I refactored them so that they all have a `BindingSpec` prefix and `Msg` suffix, where `BindingSpecMsg` is the sum type and the other functions are specialized.

An opaque `BindingSpecPath` wrapper is added to `Lib`.  Function `parseBindingSpecPath` wraps the type as well as specifies the passed tracer, which is still `Tracer IO TraceMsg` like in the previous implementation.  Function `genBindingSpec` no longer needs a tracer and unwraps the type.


TravisCardwell:
As discussed, the only remaining reference to "unresolved" outside of `HsBindgen.BindingSpec*` is in the internal implementation of the opaque `BindingSpec` type and related functions used for writing JSON/YAML binding specifications.

PR #867, opened 2025-07-04
Passes have associated configuration and trace message types

dschrempf:


hs-bindgen/src-internal/HsBindgen/Frontend/Pass.hs
```diff
@@ -69,5 +71,20 @@ class IsPass (p :: Pass) where
...
 data NoAnn = NoAnn
   deriving stock (Show, Eq)
+
+data NoConfig = NoConfig
+  deriving stock (Show, Eq)
+
+data NoMsg = NoMsg
+  deriving stock (Show, Eq)
+
+-- TODO: Cyclic dependency.
```
dschrempf: I wanted to define a result type for passes, but this leads to cyclic module imports. Do we want a result type?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames.hs
```diff
@@ -35,7 +31,7 @@ import Text.SimplePrettyPrint (hcat, textToCtxDoc, (><))
 
 mangleNames ::
      C.TranslationUnit HandleTypedefs
-  -> (C.TranslationUnit NameMangler, [MangleNamesMsg])
+  -> (C.TranslationUnit MangleNames, [Msg MangleNames])
```
dschrempf, line 34: I also renamed `NameMangler` -> `MangleNames` so that module name and pass match.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass.hs
edsko: No, I don't think that's necessary, it won't provide that much benefit I think.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass.hs
```diff
@@ -69,5 +71,20 @@ class IsPass (p :: Pass) where
   -- are showable, for example).
   type Ann (ix :: Symbol) p :: Star
 
+  -- | Configuration required to run the pass
+  type Config p :: Star
+
+  -- | Trace messages possibly emitted by the pass
+  type Msg p :: Star
```
edsko: Let's make this a `data` family rather than a `type` family. Passes that provide no messages are very rare, and data families generally work better with type inference.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass.hs
```diff
@@ -69,5 +71,20 @@ class IsPass (p :: Pass) where
   -- are showable, for example).
   type Ann (ix :: Symbol) p :: Star
 
+  -- | Configuration required to run the pass
+  type Config p :: Star
```
edsko, line 75: Let's give this a default (`NoConfig`).

hs-bindgen/src-internal/HsBindgen/Frontend.hs
```diff
@@ -110,15 +116,16 @@ processTranslationUnit
 
     let (afterSort, msgsSort) =
           sortDecls afterParse
+        sliceConfig = SliceConfig configProgramSlicing configPredicate
```
edsko: Let's not define this here, it breaks the flow, and it suggests that the config may depend on the results of the previous passes. Let's move it to the `where` clause.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/HandleTypedefs/IsPass.hs
```diff
@@ -31,6 +31,8 @@ instance IsPass HandleTypedefs where
   type MacroBody  HandleTypedefs = C.CheckedMacro HandleTypedefs
   type ExtBinding HandleTypedefs = ResolvedExtBinding
   type Ann ix     HandleTypedefs = AnnHandleTypedefs ix
+  type Config     HandleTypedefs = NoConfig
```
edsko: If we add the default, this (and similar) lines will become unnecessary.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames.hs
edsko: Yes, good. That was an omission on my part when I cleaned up the passes previously.

edsko (approved):
Nice, loving the consistency! A few minor things to fix; feel free to merge after that.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass.hs
dschrempf: Done. I created an associated data family.

Issue #847, opened 2025-07-01
We should test with `clang` and `gcc` backends


edsko:
The "official" way to run `hs-bindgen` is with `clang` throughout; this means that [ghc should also be configured to use the llvm backend](https://minoki.github.io/posts/2025-01-06-ghc-llvm-backend.html). If ghc is used with the default backend (gcc), _most_ things will work, but in some cases there might be divergence. It would be very good to test this (not just generating bindings, but actually executing them). Ideally we'd end up with a list of features for which `gcc` is fine, and a list of features that require caution.

phadej:
>  It would be very good to test this (not just generating bindings, but actually executing them)

`test-th` (and `test-pp`) do run stuff.

phadej:
Also GHC's LLVM backend has nothing to do with compiling C code. 

GHC will use its backed in C compiler (unless told otherwise which is generally not a good idea)

```
% ghc --info
 [("Project name","The Glorious Glasgow Haskell Compilation System")
 ,("GCC extra via C opts","")
 ,("C compiler command","gcc")
```

regardless whether you use `-fllvm` or not (i.e. LLVM or NCG code generation backends).

There is no "GCC" backend.

edsko:
I see, fair enough. So what option should we pass to use `clang` instead?

/cc @bolt12

edsko:
Or does that question not even make sense? When does `ghc` actually call a C compiler..? 🤔  (Other than with CAPI perhaps?). Perhaps it's _cabal_ we should be configuring instead..? 🤔

phadej:
> Or does that question not even make sense? When does ghc actually call a C compiler..? 🤔 (Other than with CAPI perhaps?).

Yes.

> Perhaps it's cabal we should be configuring instead..?

Well, `Cabal` shouldn't call C compiler directly either (it uses GHC as C compiler most of the time). There is no "C compiler" option to `cabal-install`. there is `--with-gcc` which kind of does that, but its semantics are very unclear. (Also it's not plumbed internally everywhere atm).

TL;DR, changing C compiler is *a bad idea*. If you really want that, you should make your own GHC distribution so it's `"C compiler command","clang")`

bolt12:
according to: https://downloads.haskell.org/ghc/latest/docs/users_guide/phases.html

[...]

PR #869, opened 2025-07-04
Added section about how to setup the Environment in each Platform

bolt12:


edsko:
@bolt12 Thank you for this, I think it's a great start! There some things I think we should address before merging this; especially we need to clear up the confusion about what `ghc` is using the `C` compiler for, if anything; the section on "configuring ghc with a different C compiler" should also be tightened up a bit, I think; it currently reflects a bit too much our own uncertainty :)

I also wonder if @TravisCardwell has any feedback here, I'll request his review.

edsko:
@bolt12 By the way, how come this only runs 5 checks...? Do we have some clever infrastructure in place now that avoids running the bindgen tests themselves if we're just changing the manual?

edsko:
Link for rendered version: https://github.com/well-typed/hs-bindgen/blob/bolt12/added-section-about-setting-up-env/manual/LowLevel/Invocation.md#setting-up-an-environment-that-is-able-to-build-and-run-hs-bindgen

TravisCardwell:
We should probably also document how to use `hs-bindgen` without shared libraries.  I think that it is a common case for users to want to compile in their own project C code, so that the only shared libraries used are those of dependencies.  Perhaps this should be done in a separate PR, however, which adds such an example.

Perhaps it would be nice to be able to build the manual with shared libraries or without them.  The `.cabal` files differ, but we could use a flag.  In my quick test, I simply created symbolic links to the `c` directory, but we could refactor the manual so that C source is stored within the relevant Haskell package directory.

<details><summary><code>ldd run-manual</code></summary>

```
  linux-vdso.so.1 (0x00007f535151d000)
  libm.so.6 => /usr/lib/libm.so.6 (0x00007f53513eb000)
  libgmp.so.10 => /usr/lib/libgmp.so.10 (0x00007f5351345000)
  libc.so.6 => /usr/lib/libc.so.6 (0x00007f5351155000)
  /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f535151f000)
```

</details>

If we refactor the manual so that C source is stored within the relevant Haskell package directory, the `extra-include-dirs` and `extra-lib-dirs` configuration in `cabal.project.local` files would also be come unnecessary.  A drawback is that more paths would need to be added to `LD_LIBRARY_PATH` (Linux), `DYLD_LIBRARY_PATH` (macOS), and `PATH` (Windows) to execute the resulting executable as we currently do it.  Alternatively, we could collect the executable and shared libraries into a build directory.  We would not need to set `PATH` at all on Windows.  I am not sure about macOS.  We would still need `LD_LIBRARY_PATH` in Linux:

```
$ env LD_LIBRARY_PATH=build build/run-manual
```

manual/LowLevel/Invocation.md
```diff
@@ -57,3 +57,213 @@ with command-line arguments. For instance, they can set:
...
+`-pgmc` and `-pgml` ([see reference here](https://downloads.haskell.org/ghc/latest/docs/users_guide/phases.html#replacing-the-program-for-one-or-more-phases))
+which allow one to tell GHC which C compiler to use, however after some
+experimentation proved that just setting these flags was not enough. For more
+details see the discussions at issues
+[#846](https://github.com/well-typed/hs-bindgen/issues/846)
+[#847](https://github.com/well-typed/hs-bindgen/issues/847).
+
+All in all, using Clang for both your C code and the code generated by `hs-bindgen`
+would ensure consistency and correctness. But we currently don't thoroughly
+test with `clang`.
```
TravisCardwell, line 102: I think that we need to fully understand this before we can write good documentation for it.  (#847)

manual/LowLevel/Invocation.md
```diff
@@ -57,3 +57,213 @@ with command-line arguments. For instance, they can set:
...
+      <path-to-hs-bindgen>/manual/c
+  extra-lib-dirs:
+      <path-to-hs-bindgen>/manual/c
+
+package hs-vector
+  extra-include-dirs:
+      <path-to-hs-bindgen>/manual/c
+  extra-lib-dirs:
+      <path-to-hs-bindgen>/manual/c
+```
```
TravisCardwell, line 136: Perhaps this is an artifact of the specific way we are organizing our manual?  Projects generally should not need this because such things are specified in `.cabal` files.  Perhaps we should reorganize our manual source.

manual/LowLevel/Invocation.md
```diff
@@ -57,3 +57,213 @@ with command-line arguments. For instance, they can set:
...
+
+        **NOTE**: Without `--no-stdinc`, there are likely multiple directories
+        in the C include path that provide the same headers.  If different,
+        then which header is used is order-dependent, which is not good.
+
+      * In [CI](../../.github/workflows/build-manual-ubuntu.yml) we do a bit
+        of parsing to get the results automatically.
+
+      * Note that setting this environment variable this is only needed in
+      Linux.
```
TravisCardwell: This is what we used to do, but implemented in Haskell, and we decided to instead use the default include directories for the current environment as determined by `libclang`.  Include directories should generally be set with `C_INCLUDE_PATH`, which *add* to those default directories.

Generally recommending using `BINDGEN_EXTRA_CLANG_ARGS` to set include paths does not seem correct.

Using `BINDGEN_EXTRA_CLANG_ARGS` with `-nostdinc` makes sense when we need to completely *replace* the default include directories.  Doing this may be a sign that something is wrong, though, and should be investigated further.

Note that `--no-stdinc` is an `hs-bindgen-*` option, while `-nostdinc` is a `clang` option.

Saying that it is only needed in Linux is misleading.  One use is to set preprocessor definitions, which is valid on all platforms.  We are not doing so in the Windows build only because we disable Unicode by default.

A number of details are still unclear for me in the Windows manual CI jobs.  We should reconsider this section once we fully understand all of the CI builds.

manual/LowLevel/Invocation.md
```diff
@@ -57,3 +57,213 @@ with command-line arguments. For instance, they can set:
...
+
+  * `LLVM_PATH`, `LLVM_CONFIG`: `hs-bindgen` may need to know where to find your LLVM installation.
+
+    ```bash
+    export LLVM_PATH=/usr/lib/llvm-14
+    export LLVM_CONFIG=$LLVM_PATH/bin/llvm-config
+    ```
```
TravisCardwell: This is only needed when you want to use a version of LLVM/Clang that is not in your current `PATH`.  (I never use these in Linux; I just set my `PATH`.)

manual/LowLevel/Invocation.md
```diff
@@ -57,3 +57,213 @@ with command-line arguments. For instance, they can set:
...
+  extra-lib-dirs:
+      <path-to-hs-bindgen>/manual/c
+```
+
+#### Setting Environment Variables
+
+Properly setting environment variables is crucial for `hs-bindgen` to find the necessary tools and libraries in Linux.
+
+  * `LD_LIBRARY_PATH`: To ensure that the C libraries you build can be found
+                       at runtime, you need to add their location to this variable.
```
TravisCardwell: I have never seen this indentation style before for Markdown.  I suggest that we stick with standard indentation and avoid lines longer than 80 columns when possible:

```markdown
* `LD_LIBRARY_PATH`: To ensure that the C libraries you build can be found at
  runtime, you need to add their location to this variable.
```

manual/LowLevel/Invocation.md
```diff
@@ -57,3 +57,213 @@ with command-line arguments. For instance, they can set:
...
+    ```bash
+    export LLVM_PATH=/usr/lib/llvm-14
+    export LLVM_CONFIG=$LLVM_PATH/bin/llvm-config
+    ```
+
+  * `SUPPORTS_UNICODE`: This is only specific to building the manual. The
+                        manual has some Unicode Haskell definitions and depending on the platform
+                        these are supported or not. Linux supports them so if you want to build and
+                        run the manual using these Unicode definitions you should also set this flag
+                        and add `-DSUPPORTS_UNICODE` to `BINDGEN_EXTRA_CLANG_ARGS`
```
TravisCardwell: Perhaps we should leave this out, or perhaps move it to a manual README.

The current `SUPPORTS_UNICODE` implementation is not straightforward.  We set it for GHC, GHC's C compiler, and `libclang`.  To build the shared libraries (using `gcc`), we compute a value separately, first setting it if running Linux and then unsetting it if `LLVM_BACKEND` is set.

`manual/c/Makefile` still uses `gcc`.  As I mentioned before, I am not sure if this is a good idea.  It would be cleaner to use the C compiler that is used elsewhere (by GHC).

manual/LowLevel/Invocation.md
```diff
@@ -57,3 +57,213 @@ with command-line arguments. For instance, they can set:
...
+
+  * Dynamic-link library loading order: If your application fails silently
+                                        or with an `ExitFailure` and a cryptic error code like `(-1073741515)`, it
+                                        is very likely a DLL loading issue. Adding the directory containing your C
+                                        libraries' DLLs to the system `PATH` is the solution.
+
+  * Resolving issues with underlying type mismatches (`FC.CInt` vs. `FC.CUInt`): You might encounter Haskell type errors where, for example, a
+                                                                                 C `int` is being interpreted as a `CUInt` instead of a `CInt`. This is
+                                                                                 often due to how different compilers and platforms define basic types.
+                                                                                 Carefully check your C and Haskell type definitions to ensure they match.
```
TravisCardwell: Perhaps it is worth pointing out that bindings generated by `hs-bindgen` are *not portable*.  Somebody who wants to create a portable API must do so at a higher level, using CPP to create an abstraction layer over the low-level, platform-specific bindings.  Indeed, I see that the `CUInt` issue is resolved in this way.

This is particularly important because people could get frustrated with `hs-bindgen` if users upload platform-specific APIs to Hackage without making it clear that they are for the specific platform that they were generated for and they can be successfully used on other platforms with incorrect values.

TravisCardwell (approved):
This is a good start!

I would like to include a lot more detail, but perhaps that can be done in future changes.  I traversed a lot of logs and did some more local tests today, and I still have a number of open questions.  I need to experiment in CI to better understand what is happening across all of our tested environments.

manual/LowLevel/Invocation.md
TravisCardwell: Given that `hs-bindgen` only supports generating bindings for a subset of targets people use Hackage with, perhaps *all* generated bindings uploaded to Hackage should include appropriate gates?  Minimal example:

```
if !(os(linux) && arch(x86_64))
  buildable: false
```

manual/LowLevel/Invocation.md
bolt12: Yeah, I do link to #847 so people can at least look into that discussion to get more details

manual/LowLevel/Invocation.md
bolt12: Right, I'll add a paragraph making that clear, but in my defense I do say "one might need that" and what flags are needed, so they can set that in their .cabal files and not cabal.project.local files.

manual/LowLevel/Invocation.md
bolt12: Windows assembler on CI also gave me errors regarding Unicode, but it might be yet another quirk of those GH CI Windows machines.

But yes, I agree, I have rephrased things a little bit to make it clear this is not supposed to do and might be only a CI quirk.

manual/LowLevel/Invocation.md
bolt12: Added a clarification phrase.

manual/LowLevel/Invocation.md
bolt12: > manual/c/Makefile still uses gcc. As I mentioned before, I am not sure if this is a good idea. It would be cleaner to use the C compiler that is used elsewhere (by GHC).

What do you mean with this?

manual/LowLevel/Invocation.md
bolt12: Added!

bolt12:
I have opened an issue to document how to run the manual without shared libraries #873 . I won't merge this PR yet because I want an extra round of feedback.

I agree with most of the things that @TravisCardwell mentions like the `SUPPORTS_UNICODE` flag and the paragraph about `BINDGEN_EXTRA_CLANG_ARGS`, etc... but I wanted to leave it documented somewhere my struggles and findings regarding #836 so that it is documented in the manual somewhere. With this being said I am fine with having them removed too

manual/LowLevel/Invocation.md
TravisCardwell: Indeed.  We can update this section when we are ready.

manual/LowLevel/Invocation.md
TravisCardwell: I learned that the point of the manual is to demonstrate how to use shared libraries of external dependencies.

Looks good!

manual/LowLevel/Invocation.md
TravisCardwell: I should look more into this.  On Windows, we have (at least) three candidate C compilers:

* Clang already installed in the job runner image - This one has a default target of `x86_64-pc-windows-msvc`.
* Clang installed by GHCup as the C compiler for GHC - This one is part of MinGW and likely has a default target of `x86_64-unknown-windows-gnu` or `x86_64-unknown-mingw32`.
* GCC - I am currently not sure what GCC we are using, but I suspect that it is already installed in the job runner image.  Each installation of GCC has a fixed target, and I wonder what it is.

I figure that it would be cleanest to use the same compiler as GHC to build the DLLs.

The `-fPIC` error screenshot is from the Clang already installed in the job runner image, targeting `x86_64-pc-windows-msvc`.  That error is expected, because Windows EXEs/DLLs have a different way of handling addressing/relocation, and I do not think `-fPIC` is a valid option for them.  That GCC allows `-fPIC` indicates that it must have a non-Windows target...  Perhaps it is `x86_64-unknown-mingw32` and MinGW supports PIC EXEs/DLLs because it runs them differently, not doing standard relocation in the kernel?  Alternative, perhaps it is ignoring the option?

Perhaps none of this matters, but my Windows ignorance makes me uneasy, and I would at least like to know what we are doing.

For our tests, perhaps it is sufficient to just test `x86_64-unknown-mingw32`, but here is a bonus question that I would like to figure out if it is easy to do so:

* Is an executable or DLL built targeting `x86_64-unknown-windows-gnu` or `x86_64-unknown-mingw32` able to be used outside of a MinGW environment, such as from the Windows command prompt or PowerShell?  If not, perhaps the GHC target of `x86_64-unknown-mingw32` prevents any of the built Haskell programs from running outside of a MinGW environment.

Unfortunately, this is not something that I can test in my local environment.  Perhaps I can build in CI and find a way to put the built file in a place where I can download it...

bolt12:
I completely forgot about this PR. I'll merge it!

PR #886, opened 2025-07-08
Fix #793

### LINKED ISSUE #793: Bug in type checker: "incoherent substitution"
The following file

```c
#define SWAP32(w) (((w)>>24)&0xff) | (((w)<<8)&0xff0000) | (((w)>>8)&0xff00) | (((w)<<24)&0xff000000)
```

results in

```
Semigroup Subst: incoherent substitution
TyVar with unique 30 mapped to two different types
ty1: (a_30[tau],IntLike a_43[tau])
ty2: (a_30[tau],IntLike a_47[tau])
```

With 20a40e85d052e7cfe74d84fc6e03255e5efb3583, the `CallStack` is

```
  panicPure, called at src-internal/HsBindgen/C/Tc/Macro.hs:197:3 in hs-bindgen-0.1.0-inplace-internal:HsBindgen.C.Tc.Macro
  substClashErr, called at src-internal/HsBindgen/C/Tc/Macro.hs:169:35 in hs-bindgen-0.1.0-inplace-internal:HsBindgen.C.Tc.Macro
```

sheaf:
This is a collection of bugfixes to the macro typechecker that are necessary to fix #793.


hs-bindgen/src-internal/HsBindgen/C/Tc/Macro.hs
```diff
@@ -1882,8 +1882,9 @@ tcMacro tyEnv macroNm args body =
 
     -- Step 4: generalise.
     let
-      qtvsList = reverse $ seenTvsRevList $ getFVs noBoundVars $
-                   freeTyVarsOfTypes ( fmap ( applySubstNormalise plat ctSubst ) $ bodyTy : Vec.toList argTys )
+      qtvsList = reverse $ seenTvsRevList $ getFVs noBoundVars do
+        freeTyVarsOfTypes $ fmap ( applySubstNormalise plat ctSubst ) $ bodyTy : Vec.toList argTys
+        freeTyVarsOfTypes $ fmap ( applySubstNormalise plat ctSubst ) $ map fst simpleCts
```
sheaf: Observations:

  1. We shouldn't try to quantify over type variables that appear only in constraints; that's an ambiguous type. If we really run into a type like `AddRes alpha Int ~ Int => Int -> Int` then that is an internal error because we failed to default `alpha`.
  2. There is currently no mechanism in place that will default an unfilled metavariable `alpha` if it only appears in e.g. `AddRes alpha Int ~ Int`. But then the question arises: can we ever end up with such a constraint without an accompanying `Add alpha Int` constraint, in which defaulting **does** happen currently?

Conclusion: remove this logic, and add a panic if there are any unfilled metavariables that appear in the constraints that don't appear in the argument and result types.

hs-bindgen/src-internal/HsBindgen/C/Tc/Macro.hs
```diff
@@ -1539,33 +1539,41 @@ solveDictCt handleOverlap doDefault ctOrig cls instEnv args = do
...
+  all noRefinement $ IntSet.toList qtvs
+    where
+      noRefinement tv =
+        case IntMap.lookup tv dfltSubst of
+          Nothing -> True
+          Just ( _, dfltTy ) ->
+            case IntMap.lookup tv matchSubst of
+              Nothing -> False
+              Just ( _, matchTy ) ->
+                matchTy `eqType` dfltTy
```
sheaf, line 1645: Explanation of what's going on here, via an example (to be turned into a comment.)

Original qtvs: `{alpha}`. We should not default this type variable.
Matching substitution: `[ alpha :-> IntLike beta ]`

Case 1, defaulting substitution = `[ alpha :-> IntLike beta ]`
Case 2, defaulting substitution = `[ alpha :-> IntLike (Signed Int), beta := Signed Int ]`
Case 3a, defaulting substitution = `[ alpha :-> IntLike beta, beta := Signed Int ]` impossible because of the internal invariant that substitutions are idempotent.
Case 3b, defaulting substitution = `[ beta := Signed Int  ]` impossible because the defaulting substitution is finer than the matching substitution.

hs-bindgen/src-internal/HsBindgen/C/Tc/Macro.hs
```diff
@@ -1539,33 +1539,41 @@ solveDictCt handleOverlap doDefault ctOrig cls instEnv args = do
...
+                      debugTraceM $
+                        unlines
+                          [ "solveDictCt: matchOne SUCCESS (defaulting)"
+                          , "qtvs: " ++ show qtvs
+                          , "ct: " ++ show ct
+                          , "matchSubst: " ++ show matchSubst
+                          , "dfltSubst: " ++ show dfltSubst1
+                          ]
+                      return dfltSubst1
+                return $ Just $ Subst $ IntMap.withoutKeys finalSubst qtvs
```
sheaf: The `IntMap.withoutKeys finalSubst` is the only change in this commit other than fixing the logic in `doesNotRefine`.

This `IntMap.withoutKeys` is wrong because updating the substitution like this doesn't make sense.
The reason I wrote it was to work around a mismatch between macro types and Haskell types in relation with `IntLike`. Example: we want to quantify over `alpha` and have an unsolved `Shift alpha Int` constraint. The solver will solve it with `alpha := IntLike beta`. So now we would be inferring a type signature which quantifies over `beta :: IntLikeWidthAndSign` which is something that exists in the macro type language but not in Haskell.

Really there are two choices here:

  1. Solve `Shift alpha Int` by `alpha := IntLike beta`, but then properly translate the `IntLike beta` type to Haskell (in `translateTau`), e.g. by accumulating additional constraints that we wrap around the whole thing. So `forall b. TcMacro.IntLike b -> TcMacro.IntLike b -> TcMacro.IntLike b` turns into `forall c. Hs.IntLike c => c -> c -> c`.
  2. Don't solve `Shift alpha Int`; leave it as-is, and then quantify over `alpha` and over this constraint.

sheaf: I think it is better to avoid modifying `translateTau`, so choosing option (2). A robust fix which means the translation never sees polymorphic `IntLike` types would be to change how instance lookup works to require monomorphic `IntLike blah` types rather (e.g. it behaves like a collection of instances `IntLike (W64, Signed)`, `IntLike (W32, Unsigned)` etc).

hs-bindgen/src-internal/HsBindgen/C/Tc/Macro.hs
```diff
@@ -1551,14 +1596,33 @@ solveDictCt handleOverlap doDefault ctOrig cls instEnv args = do
...
+-- Assumes that the second substitution refines the first one, i.e. that one
+-- can arrive at the second substitution by adding more substitutions to the
+-- first.
+--
+-- Example: @qtvs = {α}@, @subst1 = {α ↦ IntLike β}@.
+--
+--  1. @subst2 = {α ↦ IntLike β}@.
+--     OK: @α@ maps to the same thing in both substitutions.
+--  2. @subst2 = {α ↦ IntLike (Int Signed), β ↦ Int Signed }@
+--     Not OK: @α@ is further subtituted.
```
sheaf:
```suggestion
--     Not OK: @α@ is further substituted.
```

hs-bindgen/src-internal/HsBindgen/C/Tc/Macro.hs
```diff
@@ -222,6 +222,45 @@ applySubst subst = goTy
...
+-- This means there is no Haskell equivalent of the type @IntLike alpha@ for
+-- an unfilled metavariable @alpha@; it really corresponds to a family of types.
+--
+-- One might wonder why the macro type system departs from Haskell in this way;
+-- the foundational reason is that it allows one to easily write families of
+-- typeclass instances which cover all int-like types (see 'classInstancesWithDefaults').
+isAtomicType :: Type ki -> Bool
+isAtomicType = \case
+  Data IntLikeTyCon args
+    | TyConAppTy {} ::: VNil <- args
```
sheaf: Add comment: there are only two well-kinded arguments here:

  - `TyConAppTy (PrimIntInfoTyCon inty) VNil` which is atomic
  - `TyVarTy tv` which is not atomic

sheaf:
Two additional observations:

  1. The only place where we could possibly introduce non-atomic `IntLike alpha` types is when solving a class constraint.
  2. We don't default in equalities, e.g. we don't know what to do in `AddRes alpha Int ~ Int`. But we should never get these on their own without an additional `Add alpha Int` constraint. We rely on the fact that equality constraints don't appear on their own but are going to be accompanied by a class constraint (for which we **do** do defaulting). NB: this was not the case before, e.g. given `Shift alpha Int`, we would solve it with `alpha := IntLike beta` and quantify over `beta` which meant we could end up with e.g. `AddRes gamma Int ~ IntLike b` with no other constraints.

PR #887, opened 2025-07-08
Enable "globals" examples in the manual

edsko:
This reverts commit a698bfa70f360ac6fd30f7f1cae1e6409a8af7bb.

This will (I think) fail on Windows, but I don't know why. Bug _somewhere_, but probably not hs-bindgen, I _think_. We should figure out what's going wrong though.

### DIFF
--- a/.github/workflows/build-manual-windows.yml
+++ b/.github/workflows/build-manual-windows.yml
@@ -140,6 +140,9 @@ jobs:
         run: |
           echo "PATH=${{ github.workspace }}\\manual\\c;$PATH" >> "$GITHUB_ENV"
 
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
+
       - name: Run the manual executable
         working-directory: ./manual/hs
         shell: bash


edsko:
@TravisCardwell if this fails again on Windows with

```
    Mingw-w64 runtime failure:
    32 bit pseudo relocation at 00007FF70A9676A7 out of range, targeting 00007FFAD8683010, yielding the value 00000003CDD1B965.
```

would you be able to reproduce this on your Windows machine?

edsko:
The hope is that this will be addressed by #898.

TravisCardwell:
Here are my (still relevant) notes from yesterday.  We discussed them, but I did not have time to post them until this morning. :bow:

----

I am able to reproduce this in my local environment.  I get the same error with different integral values.

----

This issue seems to be related to the use of DLLs.

I have been wondering about the `-fPIC` option because Windows has a different way of doing relocation.  Here is the [`-fPIC`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/phases.html#ghc-flag-fPIC) documentation for GHC:

> Generate position-independent code (code that can be put into shared libraries).  This currently works on Linux `x86` and `x86-64`.  On Windows, position-independent code is never used so the flag is a no-op on that platform.

I experimented with removing the `-fPIC` option.  When compiling the `.o` files, the same exact output is created without `-fPIC` as with `-fPIC`.  It must be a no-op for Clang (targeting `x86_64-unknown-windows-gnu`) as well.

Note that I have not tested with GCC locally because it is not installed.

Note that the DLLs differ with each build, even with the same options.

----

Searching online:

* [`MINGW-packages` #10547](https://github.com/msys2/MINGW-packages/issues/10547): `[ecppack]` Mingw-w64 runtime failure: 32 bit pseudo relocation (...) out of range
    * A comment indicates that this is related to loading DLLs, as expected.

* [`haskell.nix` #2228](https://github.com/input-output-hk/haskell.nix/issues/2228): Regression: Mingw-w64 runtime failure: 32 bit pseudo relocation
    * I tried that minimal example program.  It works without error in my environment.  Indeed a comment states that it does not happen on a native Windows GHC.
    * The quoted overlay configuration [still exists in `HEAD`](https://github.com/input-output-hk/haskell.nix/blob/1af1e0d8d8e997c35efea80c983b62732050f7ef/overlays/windows.nix#L26-L34).

* [GHC #23194](https://gitlab.haskell.org/ghc/ghc/-/issues/23194): Likely bug under Windows with mimimal reproducer

* [Introduction to PIC](https://wiki.gentoo.org/wiki/Hardened/Introduction_to_Position_Independent_Code)

TravisCardwell:
When building, I ran into the "couldn't match expected type `FC.CUInt` with actual type `FC.CInt`" error.  Investigating, I found that we use the following macro:

```c
#if defined(mingw32_HOST_OS)
```

I am running MinGW64 (the shell prepared by GHCup), so I tried changing it to `mingw64_HOST_OS` and that indeed worked.

FWIW, I think that the CI environment is also running MinGW64, because the log says "Mingw-w64 runtime failure:" (as does my error).

edsko:
Obsoleted by #927.

PR #888, opened 2025-07-08
Enable `rust-bindgen` on `globals.h`

edsko:
This will fail, because on CI the output from `rust-bindgen`, despite it being the same version, is slightly different. Not sure why. Submitting this PR so that we don't lose sight of this.


dschrempf:
Should we open an issue at `rust-bindgen` and ask about this?

edsko:
No longer relevant as of https://github.com/well-typed/hs-bindgen/pull/1295 .

PR #892, opened 2025-07-09
More sophisticated handling of typedefs

### LINKED ISSUE #566: Generate only one definition when struct/union/enum and typedef defined together
Currently we generate only one definition for

```c
typedef enum shape_tag {
    RECT,
    CIRCLE
} shape_tag;
```

but not for

```c
typedef enum shape_tag {
    RECT,
    CIRCLE
} shape_tag_t;
```

I feel like a useful rule of thumb might be: generate only a single definition if the typedef and the type are defined together like this, and generate two separate defintions otherwise. When they are defined together, it's a clear signal from the programmer that they think of them as being essentially the same.

MIght be a bit finicky though to implement with clang's representation.

### LINKED ISSUE #589: Eponymous pointer to struct results in duplicate identifiers
This definition

```c
typedef struct game_state *game_state;
```

results in invalid Haskell 

```haskell
data Game_state

newtype Game_state = Game_state
  { un_Game_state :: F.Ptr Game_state
  }
```

edsko:
See `Analysis.Typedefs` for detailed discussion.

Closes #566
Closes #589


hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/Typedefs.hs
```diff
@@ -0,0 +1,287 @@
...
+      (taggedNameKind taggedKind)
+
+{-------------------------------------------------------------------------------
+  Internal auxiliary: restore original names
+
+  This allows us to look up the type in the 'DeclUseGraph'.
+  TODO: Maybe this should live somewhere more general?
+-------------------------------------------------------------------------------}
+
+origDeclId :: CName -> C.NameOrigin -> DeclId
```
edsko, line 283: @TravisCardwell I suspect you might have a version of this function somewhere in your `PrelimDeclId` refactoring branch.

PR #899, opened 2025-07-10
Simplify evaluation of a selection predicate

edsko:
We were trying to construct a _reason_ why a selection predicate matched or failed, but I realized that this is rather difficult to do well in the presence of arbitrary boolean expressions, and indeed we were not doing it correctly (though this was rather hidden: for example, logical conjunction was implemented using monadic `>>`, effectively ignoring the reason why the first predicate matched). In this PR we simplify this, mostly in preparation for further changes.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/C/Predicate.hs
+++ b/hs-bindgen/src-internal/HsBindgen/C/Predicate.hs
@@ -7,9 +7,6 @@
 module HsBindgen.C.Predicate (
     Predicate (..)
   , Regex -- opaque
-    -- * Trace messages
-  , SkipReason (..)
-  , Match (..)
     -- * Execution (internal API)
   , IsMainFile
   , match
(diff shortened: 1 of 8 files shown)


edsko:
Simplified this further still:

* The control flow of `foldDecl` is now improved and I think much clearer
* `toNameKindFromCXCursorKind` is gone entirely: we already had this information, reconstructing it led to having to think about unnecessary cases 
* Various edge cases in `matchPredicate` are gone; in particular, it will _always_ get a `QualDeclId`

All of these changes together are paving the way for adding another case to `DeclId` for builtins.

dschrempf:
>  logical conjunction was implemented using monadic >>, effectively ignoring the reason why the first predicate matched

I think the reason we did it like so was as to reduce the length of traces.

It's great that `toNameKindFromCXCursorKind` is gone. That panic really made me feel uneasy :-). The conversion to the NameKind seems to be a bit hidden though (part of the case statement). Maybe we can still have our own function. I think you did it that way to avoid returning a `Maybe` (because we only convert when we actually need the kind). @edsko

PR #904, opened 2025-07-11
Program slicing with selection predicates

### LINKED ISSUE #784: Common predicate for selection and program slicing
We would like to **unify selection of declarations** as implemented by the `Predicate`, and selecting roots and their transitive dependencies (program slicing).

The idea is to provide one set of predicates (most likely the ones we already have), but allow them to be used with normal selection, and with program slicing.

When program slicing is enabled (`--enable-program-slicing`), the selection predicate becomes the root selection predicate. That is, all declarations should be parsed and the predicate should be applied during the `Slice` step.

For this to work, matching the predicate needs to be a pure function (i.e., not use libclang as it does now). We think this is possible by providing the location of the declaration and it's name.

Further, we would like to add **"low-level" options to distinguish between what we want to select and what we actually want to parse** (for example, to exclude declarations that we can not parse).

dschrempf:
After mergin this PR, we can use selection predicates together with program slicing.

Closes #784 , we outsourced #906 .

Please also see #910 .

TODO:
- ~~[ ] After this is merged, rebase PR #888 and reactivate `rust-bindgen` (which failes because fixtures differ between PC and CI).~~ Not required anymore after https://github.com/well-typed/hs-bindgen/pull/908


edsko (approved):
Very nice! I feel happy that we have sufficient infrastructure in place now that this is possible. Good work.

PR #913, opened 2025-07-11
Example use of blocks

edsko:
Note the use of `-pgmc`, `-optc` and `-pgma` in the `.cabal` file.

This PR is not meant to be merged as-is; we need to think about what to do with examples like this, which are in a similar spirit to the `manual` example.


.github/workflows/examples.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         # NOTE: add new example projects here. The example project has to
         # satisfy some requirements for this workflow to work correctly. See
         # ./examples/README.md (with respect to the repository root).
-        example: ['botan', 'c-minisat', 'libpcap', 'c-qrcode']
```
jorisdral, line 45: I can seen an error in CI because the example is named `blocktest` here so it expects an `examples/blocktest` directory to exist, but the directory is called `examples/blocks` instead

edsko: Yeah, just noticed that myself. Sorry :)

PR #917, opened 2025-07-16
Refactor naming/identity types

TravisCardwell:
Module `HsBindgen.Language.C.Name` now just contains standard C stuff, not stuff specific to `hs-bindgen`.

Module `HsBindgen.Language.C` is now imported qualified (as `C`).  This implements the long-pending change of renaming `CName` so that it is used qualified as `C.Name`.  Sub-modules are not used directly.

* `Name`
* `TypeNamespace` is added
* `NameKind`
* `QualName`

Module `HsBindgen.Frontend.Naming` now contains the `hs-bindgen` name/identity types.

* `AnonId`
* `PrelimDeclId` was previously `DeclId`
* `NsPrelimDeclId` is added for use in the index and graphs
* `QualPrelimDeclId` is just used for slicing
* `NameOrigin` now includes `NameOriginBuiltin`
* `DeclId` is added, combining a `C.Name` and a `NameOrigin`

Note that `Located` is moved into that module and `Decl p` helper functions are moved into `HsBindgen.Frontend.AST.Internal`, to avoid import issues.

Topics to discuss:

* `Id MangleNames` is now `(NamePair, NameOrigin)`.  Those types are kept separate because `NamePair` is used in other places (such as fields).
* The `External` AST is unchanged: the origin is still separate.
* The `showsName` function now outputs `<anon>` for anonymous types and retains the old behavior for types of other origins.


edsko (approved):
Excellent, will be great to get this in. Let's merge this ASAP to avoid merge conflicts, and deal with the minor issues we discussed afterwards.

PR #921, opened 2025-07-16
Handle `const` and `pure` function attributes specially

### LINKED ISSUE #67: Take advantage of `pure` and `const` attributes
Example of a `pure` function:

```c
int hash (char *) __attribute__ ((pure));
```

[From the `gcc` manual](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-pure-function-attribute):

> Calls to functions that have no observable effects on the state of the program other than to return a value may lend themselves to optimizations such as common subexpression elimination. Declaring such functions with the `pure` attribute allows GCC to avoid emitting some calls in repeated invocations of the function with the same argument values.
>
> The `pure` attribute prohibits a function from modifying the state of the program that is observable by means other than inspecting the function’s return value. However, functions declared with the `pure` attribute can safely read any non-volatile objects, and modify the value of objects in a way that does not affect their return value or the observable state of the program.

The `const` attribute is [similar but even stricter](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-const-function-attribute).

For such functions we should be able to automatically create a pure Haskell binding, using `unsafePerformIO` or perhaps even `unsafeDupablePerformIO`.

jorisdral:
Resolves #67 

This PR threads through function attributes `const` and `pure` from the parser all the way to code generation, where we modify foreign imports based on the function attributes.

* For `const` C functions we generate a foreign import that is a pure Haskell function.

* For `pure` C functions we generate a foreign import that is an `IO` Haskell action, but we add a comment to that action that it is "pure" not in the Haskell sense, but in the C sense. Still, a higher level API could take advantage of the "pure" property to remove `IO`.

* For C functions that are both `const` and `pure`, we ignore the `pure` and consider the C function to be `const`.


hs-bindgen/src-internal/HsBindgen/Frontend/AST/Internal.hs
```diff
@@ -181,9 +182,21 @@ data EnumConstant p = EnumConstant {
...
       functionArgs :: [Type p]
     , functionRes  :: Type p
+    , functionAttrs :: [FunctionAttr]
     , functionAnn  :: Ann "Function" p
     }
 
+-- | Function attributes specify function properties that may help the C
+-- compiler.
+--
+-- TODO: should this have a @p@ type argument as well?
```
edsko: No need, unless or until we add something like a nested type or identifier in here.

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Internal.hs
```diff
@@ -181,9 +182,21 @@ data EnumConstant p = EnumConstant {
 data Function p = Function {
       functionArgs :: [Type p]
     , functionRes  :: Type p
+    , functionAttrs :: [FunctionAttr]
```
edsko: I wonder if we should do the resolution already here; so rather than having a `[FunctionAttr]`, and then having to search through the list in later, we could do something like

```hs
data FunctionAttr = FunctionAttr {
     functionPurity :: Purity
  }

data Purity = 
    Impure
  | HaskellPure
  | CPure
```

or some-such (these names are perhaps not perfect). It feels like the parser is the best place to do this, and then we depend on a clear AST later.

hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
```diff
@@ -195,10 +196,20 @@ data Typedef = Typedef {
 
 data Function = Function {
       functionArgs :: [Type]
+    , functionAttrs :: [FunctionAttr]
     , functionRes  :: Type
     }
   deriving stock (Show, Eq, Generic)
 
+-- | Function attributes specify function properties that may help the C
```
edsko: No need to repeat the defintion, `External` can just re-export definitions from `Internal` when they are unchanged.

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Finalize.hs
```diff
@@ -215,15 +215,24 @@ instance Finalize Int.Function where
...
           functionArgs
         , functionRes
+        , functionAttrs
         , functionAnn = NoAnn
         } = function
 
+      -- TODO: for now, @Int.FunctionAttr@ and @Ext.FunctionAttr@ are the exact
+      -- same types, but defined twice. Maybe that will change, see the TODO on
+      -- @Int.FunctionAttr@. If it does not change, maybe we should define a
+      -- common @FunctionAttr@ type in a shared module?
```
edsko: It doesn't have to be a new module, `External` pulls together definitions from lots of places. Just re-exporting from `Internal` is fine.

hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden.hs
```diff
@@ -245,6 +245,17 @@ testCases = [
                Nothing
         , testRustBindgen = RustBindgenFail
         }
+    , (defaultTest "fun_attributes_conflict-17") {
+          testClangVersion = Just (< (18, 0, 0))
+        }
+    , (defaultTest "fun_attributes_conflict") {
```
edsko, line 248: I hadn't realized `testClangVersion` could be used to have different tests for different versions, clever :)

hs-bindgen/src-internal/HsBindgen/Hs/Translation.hs
```diff
@@ -1090,11 +1093,63 @@ functionDecs mu typedefs info f _spec =
...
+    hsIO :: HsType -> HsType
+    ioComment :: Maybe SHs.SDecl
+    (hsIO, ioComment) = case preferred attrs of
+        Just C.ConstAttr -> (id  , Nothing)
+        Just C.PureAttr  -> (HsIO, Just (SHs.DComment pureComment))
+        Nothing          -> (HsIO, Nothing)
+      where
+        -- In case we try to handle new attributes in the future, this case
+        -- expression throws a compiler error, which should hopefully indicate
+        -- to the reader that 'hsIO' has to be updated.
```
edsko: Now where did you pick up that habit I wonder :thinking: :grimacing:

hs-bindgen/src-internal/HsBindgen/Hs/Translation.hs
```diff
@@ -1090,11 +1093,63 @@ functionDecs mu typedefs info f _spec =
...
+          C.ConstAttr -> ()
+          C.PureAttr  -> ()
+
+        preferred :: [C.FunctionAttr] -> Maybe C.FunctionAttr
+        preferred []     = Nothing
+        preferred (a:as) = Just $ foldr prefer a as
+          where
+            prefer C.ConstAttr _ = C.ConstAttr
+            prefer _ C.ConstAttr = C.ConstAttr
+            prefer _ _           = C.PureAttr
```
edsko: So I think this should live in the parser.

hs-bindgen/src-internal/HsBindgen/Hs/Translation.hs
```diff
@@ -1042,7 +1042,9 @@ functionDecs ::
 functionDecs mu typedefs info f _spec =
     [ Hs.DeclInlineCInclude $ getCHeaderIncludePath $ C.declHeader info
     , Hs.DeclInlineC $ PC.prettyDecl (wrapperDecl innerName wrapperName res args) ""
-    , Hs.DeclForeignImport $ Hs.ForeignImportDecl
+    ] ++
+    toList (Hs.DeclSimple <$> ioComment) ++
```
edsko, line 1046: This is I think my only "real" concern with the PR as it stands: the generated header does not make it clear _which_ particular function is "C pure". We should attach it to the function definition itself. But perhaps it would make more sense to fix that after @bolt12 is done with the haddock generation, as it should slot in nicely there.

edsko (approved):
Ok, looks great! Thanks for your attention to detail and comments, much appreciated.

Some minor suggestions; we can merge this as-is and address them in in separate PR, or modify this one and then merge, your preference.

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Internal.hs
jorisdral: Removed the TODO

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Internal.hs
jorisdral: Rewrote the decision procedure, and I've put it in the `Internal` module. The documentation for what Haskell-pure and C-pure (etc.) now also lives there.

hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
jorisdral: Remove the redefinition

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Finalize.hs
jorisdral: Removed the redefinition

hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden.hs
jorisdral: There is a downside to this approach and that is that if you want to regenerate the golden files, you'll have to run the golden tests twice with different clang versions: once with <18, once with >=18

hs-bindgen/src-internal/HsBindgen/Hs/Translation.hs
jorisdral: Not from you, of course...

hs-bindgen/src-internal/HsBindgen/Hs/Translation.hs
jorisdral: Done

hs-bindgen/src-internal/HsBindgen/Hs/Translation.hs
jorisdral: I've added a TODO to add the comment directly as haddocks on the foreign import.

hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden.hs
jorisdral: Actually, now that https://github.com/well-typed/hs-bindgen/pull/921#discussion_r2222744223 is done, we don't need two versions of the same golden test anymore!

edsko: Ah, nice that it's not needed anymore. And good point re clang, hadn't realized that. Perhaps it would be useful to document this somewhere.

PR #927, opened 2025-07-21
Generate stubs for global variables

### LINKED ISSUE #898: Generate stub for globals
Given

```c
int simpleGlobal;
```

we are currently generating

```hs
foreign import capi safe "&simpleGlobal" simpleGlobal :: F.Ptr FC.CInt
```

We should instead generate a stub

```c
void * get_foo_ptr() { return &foo; }
```

At least, we should do so on Windows, to avoid problems like this

```
    Mingw-w64 runtime failure:
    32 bit pseudo relocation at 00007FF70A9676A7 out of range, targeting 00007FFAD8683010, yielding the value 00000003CDD1B965.
```

which are due to a ghc bug: https://gitlab.haskell.org/ghc/ghc/-/issues/26187

jorisdral:
Resolves #898 

After this, #887 should pass, but I've enabled the globals section of the manual and extended it in the current PR, so maybe the current PR should supersede #887?

For information about the changes, see the commit messages


hs-bindgen/fixtures/globals.pp.hs
```diff
@@ -71,37 +71,37 @@ instance F.Storable Inline_struct where
...
-foreign import capi safe "&nesInteger" nesInteger :: F.Ptr FC.CInt
+foreign import ccall safe "get_nesInteger_ptr" nesInteger :: F.Ptr FC.CInt
 
-foreign import capi safe "&nesFloating" nesFloating :: F.Ptr FC.CFloat
+foreign import ccall safe "get_nesFloating_ptr" nesFloating :: F.Ptr FC.CFloat
 
-foreign import capi safe "&nesString1" nesString1 :: F.Ptr (F.Ptr FC.CChar)
+foreign import ccall safe "get_nesString1_ptr" nesString1 :: F.Ptr (F.Ptr FC.CChar)
 
 foreign import capi safe "&nesString2" nesString2 :: F.Ptr ((HsBindgen.Runtime.ConstantArray.ConstantArray 3) FC.CChar)
```
edsko, line 82: Note to self. We're _not_ generating a wrapper for constant arrays. We could probably avoid the wrapper in more cases (probably for all global variables that are pointers); however, the only place where it's really _useful_ to not generate the wrapper is constant arrays, because it's harder to keep type information in that case. So for now we'll just treat this as the exception to the rule; we might reconsider this later.

hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
```diff
@@ -265,3 +266,9 @@ data TypedefRef =
 isVoid :: Type -> Bool
 isVoid TypeVoid = True
 isVoid _        = False
+
+-- | Arrays can be of known or unknown size.
+isArray :: Type -> Bool
+isArray TypeConstArray{}      = True
+isArray TypeIncompleteArray{} = True
```
edsko, line 276: This may or may not be correct; we don't actually support this yet. We should consider when looking at https://github.com/well-typed/hs-bindgen/issues/377 .

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Internal.hs
```diff
@@ -129,6 +129,10 @@ data DeclKind p =
   | DeclEnumOpaque
   | DeclMacro (MacroBody p)
   | DeclFunction (Function p)
+    -- TODO: contrary to what the constructor name suggests, we use 'DeclExtern'
+    -- both for global variables that are extern and that are non-extern. Should
+    -- we rename this constructor to 'DeclGlobal', and propagate similar name
+    -- changes elsewhere in the pipeline?
```
edsko: Agreed. Mind making that change?

hs-bindgen/src-internal/HsBindgen/Hs/Translation.hs
```diff
@@ -1147,20 +1148,86 @@ functionDecs mu typedefs info f _spec =
...
+    varName = T.unpack (C.getName . C.nameC . C.declId $ info)
+
+    stubType :: C.Type
+    stubType = C.TypePointer ty
+
+    -- TODO: it seems that it is currently not possible to use handy
+    -- abstractions like "HsBindgen.PrettyC" to generate the stub, because the
+    -- stub function is not a closed expression. Could we add an abstraction or
+    -- AST constructor somewhere that would prevent us from generating the
+    -- string directly? If so, where should it be added?
```
edsko: Let's add `FreeVar` to `Expr`.

hs-bindgen/src-internal/HsBindgen/Hs/Translation.hs
```diff
@@ -1147,20 +1148,86 @@ functionDecs mu typedefs info f _spec =
...
   where
-    header :: FilePath
-    header = getCHeaderIncludePath $ C.declHeader info
+    importName :: HsName 'NsVar
+    importName = C.nameHs (C.declId info)
+
+    importType :: HsType
+    importType = typ stubType
+
+    stubName :: String
```
edsko, line 1206: Let's add a TODO here that this really should go through the name mangler.

edsko (approved):
Nice work. Feel free to merge when you think it's ready, some of these minor adjustments we can address in later PRs.

hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
jorisdral: Added a TODO here that references #377.

hs-bindgen/src-internal/HsBindgen/Hs/Translation.hs
jorisdral: I've ended up going with `NamedVar` instead of `FreeVar`, because "free variable" is relative to the expression/statement you're looking at, which can be ambiguous. `NamedVar` makes it clearer that we're using a name instead of a DeBruijn index. Still, I've add an example to the documentation of `NamedVar` that it can be used to refer to variables that are *free with respect to the current function definition*, like global variables.

hs-bindgen/src-internal/HsBindgen/Hs/Translation.hs
jorisdral: Done

jorisdral: See #946

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Internal.hs
jorisdral: Made the change in the current PR

Issue #906, opened 2025-07-11
Low-level parse predicates: Skip declarations while parsing


dschrempf:
We already support selection predicates and program slicing with selection predicates. However, at the moment, when program slicing is enabled, we parse __all declarations__ and then select root declarations and their transitive dependencies.

The "parse all" approach with program slicing enabled may be problematic when we encounter declarations that we simply can not parse. Therefore, we would like to add "low-level" options to skip declarations already while parsing when program slicing is enabled.

edsko:
Leaving this at medium priority for now, because I am still worried that we'll choke on some inputs, and then this option would rescue the situation. But we might degrade this to low priority if we're feeling very confident that that cannot happen :)

dschrempf:
We now allow `type ParsePredicate = Predicate HeaderPathPredicate`.

To fix this issue #906, we could make `ParsePredicate` more fine-grained. If we only want to avoid parse failures, we could match all declarations against a regex (similar to the element regex we had before). What is your opinion?

TravisCardwell:
The reason why the `Parse` pass does not support predicates on declaration names is that we do not yet have names for anonymous declarations.  The `Select` pass supports predicates on declaration names after anonymous declarations have been named, consistent with names in binding specifications.

If we restore support for declaration name predicates to `Parse`, perhaps only named declarations can be specified?  This presents a challenge: perhaps it also needs transitive dependency filtering since anonymous dependencies of parsed declarations must also be parsed, but we do not calculate the `UseDeclGraph` until the `Sort` pass.

Alternatively, transitive dependency filtering is not strictly necessary if `Parse` predicates can specify anonymous declarations.  Specifying a header, line number, and column number would be both tedious and fragile, however, so perhaps that is not viable?

I wonder if it would be possible to instead "delay" errors until after selection, so that only selected declarations can have (fatal) errors?

BTW, this topic of predicate and slicing design is related to [a comment](https://github.com/well-typed/hs-bindgen/issues/592#issuecomment-3148830352) that I wrote yesterday about the possibility of (also) filtering after slicing, which would make it possible to generate a module that includes only the necessary dependencies for a public interface defined in one or more separate modules.  It is easy to implement, and I think that it would be a useful feature, but it would complicate the API to add another set of predicate options.

EDIT

The very next email I processed is for #905 (Delay emission of warnings about unsupported declarations until they are truly selected)!  Is this issue still necessary if we do that?

dschrempf:
[...]

Issue #866, opened 2025-07-04
Macros: We incorrectly parse string concatenation


edsko:
```c
#define STRUCT(name) name##_s
```

results in 

```
[Warning] [HsBindgen] Failed to typecheck macro:
  Unbound variable: '_s'
```

but this is actually not a variable at all, but a constant string.

edsko:
Closing as we don't parse string concatenation at all anymore since https://github.com/well-typed/hs-bindgen/pull/1191 .

PR #929, opened 2025-07-22
Refactor predicate selection

TravisCardwell:
The objective of this PR Is to refactor how we do predicate selection.

* Selection must happen after the `ResolveBindingSpec` pass because it must take into account types that have external bindings or are omitted by a prescriptive binding specification.
* Selection and slicing must happen after the `NameAnon` pass so that selection can be done on anonymous types.
* We still want to be able to select during the `Parse` pass, because of performance concerns.

The `Slice` pass is moved after `ResolveBindingSpec` and renamed to `Select`.

The `Predicate` type is refactored to separate general boolean logic from concrete predicates using a type parameter (`Predicate a`).  Concrete predicate `HeaderPathPredicate` implements predicates on header paths, and concrete predicate `DeclPredicate` implements predicates on declarations.  Type alias `type ParsePredicate = Predicate HeaderPathPredicate` is used for predicates for the `Parse` pass, which may only include header path predicates.  Type alias `type SelectPredicate = Predicate (Either HeaderPathPredicate DeclPredicate)` is used for the `Select` pass, which may include either.  There are now two separate match functions: `matchParse` and `matchSelect`.  The check for built-ins is only done in `matchParse`.

Some things are renamed for consistent terminology.

* Predicates are used to select *declarations*, not *files* or *elements*.
* We can select from *main headers*, not *main files*.
* We select by header *paths* that match a PCRE, not just *filenames*.

Previous CLI options:

```
--select-all                    Process all elements
--select-by-filename PCRE       Select files with names that match PCRE
--skip-by-filename PCRE         Skip files with names that match PCRE
--select-by-element-name PCRE   Select elements with names that match PCRE
--skip-by-element-name PCRE     Skip elements with names that match PCRE
--select-from-main-files        Only process elements from main files (default)
```

New CLI options:

```
--parse-select-all                   Select all declarations (Parse pass)
--parse-select-from-main-headers     Select declarations in main headers (Parse pass, default)
--parse-select-by-header-path PCRE   Select declarations in headers with paths that match PCRE (Parse pass)
--parse-skip-by-header-path PCRE     Skip declarations in headers with paths that match PCRE (Parse pass)

--select-all                   Select all declarations (Select pass, default)
--select-from-main-headers     Select declarations in main headers (Select pass)
--select-by-header-path PCRE   Select declarations in headers with paths that match PCRE (Select pass)
--skip-by-header-path PCRE     Skip declarations in headers with paths that match PCRE (Select pass)
--select-by-decl-name PCRE     Select declarations with names that match PCRE (Select pass)
--skip-by-decl-name PCRE       Skip declarations with names that match PCRE (Select pass)
```

Note that a prescriptive binding specification can *require* or *omit* a declaration, while a predicate can be used to *select* or *skip* a declaration.

This PR is being created as a draft because I am currently trying to understand an issue, which I will write about in a separate comment below.  In addition, the following tasks are pending:

* [x] Add subdirectory support to file predicates
* [x] Investigate separation of selection and slicing into separate passes


TravisCardwell:
For reference, here are the current associated types for each pass.

|                      | `Id`                     |
| :---                 | :---                     |
| `Parse`              | `PrelimDeclId`           |
| `Sort`               | `PrelimDeclId`           |
| `HandleMacros`       | `PrelimDeclId`           |
| `NameAnon`           | `DeclId`                 |
| `ResolveBindingSpec` | `DeclId`                 |
| `Select`             | `DeclId`                 |
| `HandleTypedefs`     | `DeclId`                 |
| `MangleNames`        | `(NamePair, NameOrigin)` |

|                      | `FieldName` |
| :---                 | :---        |
| `Parse`              | `C.Name`    |
| `Sort`               | `C.Name`    |
| `HandleMacros`       | `C.Name`    |
| `NameAnon`           | `C.Name`    |
| `ResolveBindingSpec` | `C.Name`    |
| `Select`             | `C.Name`    |
| `HandleTypedefs`     | `C.Name`    |
| `MangleNames`        | `NamePair`  |

|                      | `TypedefRef`                       |
| :---                 | :---                               |
| `Parse`              | `C.Name`                           |
| `Sort`               | `C.Name`                           |
| `HandleMacros`       | `C.Name`                           |
| `NameAnon`           | `C.Name`                           |
| `ResolveBindingSpec` | `C.Name`                           |
| `Select`             | `C.Name`                           |
| `HandleTypedefs`     | `RenamedTypedefRef HandleTypedefs` |
| `MangleNames`        | `RenamedTypedefRef MangleNames`    |

|                      | `MacroBody`                       |
| :---                 | :---                              |
| `Parse`              | `UnparsedMacro`                   |
| `Sort`               | `UnparsedMacro`                   |
| `HandleMacros`       | `CheckedMacro HandleMacros`       |
| `NameAnon`           | `CheckedMacro NameAnon`           |
| `ResolveBindingSpec` | `CheckedMacro ResolveBindingSpec` |
| `Select`             | `CheckedMacro ResolveBindingSpec` |
| `HandleTypedefs`     | `CheckedMacro HandleTypedefs`     |
| `MangleNames`        | `CheckedMacro MangleNames`        |

|                      | `ExtBinding`         |
| :---                 | :---                 |
| `Parse`              | `Void`               |
| `Sort`               | `Void`               |
| `HandleMacros`       | `Void`               |
| `NameAnon`           | `Void`               |
| `ResolveBindingSpec` | `ResolvedExtBinding` |
| `Select`             | `ResolvedExtBinding` |
| `HandleTypedefs`     | `ResolvedExtBinding` |
| `MangleNames`        | `ResolvedExtBinding` |

|                      | `Ann "TranslationUnit"` |
| :---                 | :---                    |
| `Parse`              | `NonSelectedDecls`      |
| `Sort`               | `DeclMeta`              |
| `HandleMacros`       | `DeclMeta`              |
| `NameAnon`           | `DeclMeta`              |
| `ResolveBindingSpec` | `DeclMeta`              |
| `Select`             | `DeclMeta`              |
| `HandleTypedefs`     | `DeclMeta`              |
| `MangleNames`        | `DeclMeta`              |

|                      | `Ann "Decl"`           |
| :---                 | :---                   |
| `Parse`              | `NoAnn`                |
| `Sort`               | `NoAnn`                |
| `HandleMacros`       | `NoAnn`                |
| `NameAnon`           | `NoAnn`                |
| `ResolveBindingSpec` | `BindingSpec.TypeSpec` |
| `Select`             | `BindingSpec.TypeSpec` |
| `HandleTypedefs`     | `BindingSpec.TypeSpec` |
| `MangleNames`        | `DeclSpec`             |

|                      | `Ann "Struct"` |
| :---                 | :---           |
| `Parse`              | `NoAnn`        |
| `Sort`               | `NoAnn`        |
| `HandleMacros`       | `NoAnn`        |
| `NameAnon`           | `NoAnn`        |
| `ResolveBindingSpec` | `NoAnn`        |
| `Select`             | `NoAnn`        |
| `HandleTypedefs`     | `NoAnn`        |
| `MangleNames`        | `RecordNames`  |

|                      | `Ann "StructField"` |
| :---                 | :---                |
| `Parse`              | `ReparseInfo`       |
| `Sort`               | `ReparseInfo`       |
| `HandleMacros`       | `NoAnn`             |
| `NameAnon`           | `NoAnn`             |
| `ResolveBindingSpec` | `NoAnn`             |
| `Select`             | `NoAnn`             |
| `HandleTypedefs`     | `NoAnn`             |
| `MangleNames`        | `NoAnn`             |

|                      | `Ann "Union"`  |
| :---                 | :---           |
| `Parse`              | `NoAnn`        |
| `Sort`               | `NoAnn`        |
| `HandleMacros`       | `NoAnn`        |
| `NameAnon`           | `NoAnn`        |
| `ResolveBindingSpec` | `NoAnn`        |
| `Select`             | `NoAnn`        |
| `HandleTypedefs`     | `NoAnn`        |
| `MangleNames`        | `NewtypeNames` |

|                      | `Ann "UnionField"` |
| :---                 | :---               |
| `Parse`              | `ReparseInfo`      |
| `Sort`               | `ReparseInfo`      |
| `HandleMacros`       | `NoAnn`            |
| `NameAnon`           | `NoAnn`            |
| `ResolveBindingSpec` | `NoAnn`            |
| `Select`             | `NoAnn`            |
| `HandleTypedefs`     | `NoAnn`            |
| `MangleNames`        | `NoAnn`            |

|                      | `Ann "Enum"`   |
| :---                 | :---           |
| `Parse`              | `NoAnn`        |
| `Sort`               | `NoAnn`        |
| `HandleMacros`       | `NoAnn`        |
| `NameAnon`           | `NoAnn`        |
| `ResolveBindingSpec` | `NoAnn`        |
| `Select`             | `NoAnn`        |
| `HandleTypedefs`     | `NoAnn`        |
| `MangleNames`        | `NewtypeNames` |

|                      | `Ann "Typedef"` |
| :---                 | :---            |
| `Parse`              | `ReparseInfo`   |
| `Sort`               | `ReparseInfo`   |
| `HandleMacros`       | `NoAnn`         |
| `NameAnon`           | `NoAnn`         |
| `ResolveBindingSpec` | `NoAnn`         |
| `Select`             | `NoAnn`         |
| `HandleTypedefs`     | `NoAnn`         |
| `MangleNames`        | `NewtypeNames`  |

|                      | `Ann "Function"` |
| :---                 | :---             |
| `Parse`              | `ReparseInfo`    |
| `Sort`               | `ReparseInfo`    |
| `HandleMacros`       | `NoAnn`          |
| `NameAnon`           | `NoAnn`          |
| `ResolveBindingSpec` | `NoAnn`          |
| `Select`             | `NoAnn`          |
| `HandleTypedefs`     | `NoAnn`          |
| `MangleNames`        | `NoAnn`          |

|                      | `Ann "CheckedMacroType"` |
| :---                 | :---                     |
| `Parse`              | `NoAnn`                  |
| `Sort`               | `NoAnn`                  |
| `HandleMacros`       | `NoAnn`                  |
| `NameAnon`           | `NoAnn`                  |
| `ResolveBindingSpec` | `NoAnn`                  |
| `Select`             | `NoAnn`                  |
| `HandleTypedefs`     | `NoAnn`                  |
| `MangleNames`        | `NewtypeNames`           |

Note that `MacroBody Select` is `CheckedMacro ResolveBindingSpec`, *not* `CheckedMacro Select`.  This is done intentionally because the `Select` pass implementation uses `coercePass` on the translation unit, which does not work with `CheckedMacro p` because of the type parameter.  Coercion is done in the `HandleTypedefs` pass on the macro body.

Also worthy of note is that `Ann "TranslationUnit"` preserves non-selected declarations through all of the passes.  The `Parse` pass annotates `NonSelectedDecls` while the rest of the passes include it in `DeclMeta`.  These are needed for the `ResolveBindingSpec` pass.  The `Select` pass is now after the `ResolveBindingSpec` pass, and it updates the `NonSelectedDecls` to include the types that are skipped by the selection predicate.  Should we keep this behavior, or perhaps we should instead remove the `NonSelectedDecls` from the annotation in the `ResolvedBindingSpec` pass because it is no longer needed?

TravisCardwell:
Selection tests fail because the `Select` pass is moved later in the pipeline.  We previously did selection after the `Sort` pass, so many declarations were skipped *before* the `HandleMacros` pass.  By moving selection to the `Select` pass after the `ResolveBindingSpec` pass, the `HandleMacros` pass now processes *all* declarations.  We get many (more) issues that we need to either resolve or tolerate.

For example, many macros fail to typecheck because of the string concatenation issue (#866).

TravisCardwell:
Note that the user-specified `ParsePredicate` is ignored if slicing is enabled.

```haskell
let predicateParse = case configProgramSlicing of
      EnableProgramSlicing  -> SelectAll
      DisableProgramSlicing -> configParsePredicate
```

This is done because slicing selects declarations.  For example, the `program_slicing_simple` test "checks that program slicing generates bindings for `uint32_t` if we remove it from the standard external binding specification."  If we were to pass the parse predicate as-is, the user would need to configure one of the following options:

* `SelectAll`
* `SelectIfEither (SelectIfMatch SelectFromMainHeaders) (SelectIfMatch (SelectByHeaderPath re))` where the regular expression matches the header where `uint32_t` is defined.  The above example depends on the C library implementation; it is defined in `bits/alltypes.h` in Musl.
* (Contrived predicates such as `SelectIfEither (SelectIfMatch (SelectByHeaderPath re))` where the regular expression selects declarations in all headers.)

FWIW, I think that we should change the `program_slicing_simple` test should just not use the standard external binding specification, so that bindings for both `uint64_t` and `uint32_t` are created.  Modifying an existing binding specification is not a part of the API, and I am uneasy about doing it even in internal tests.  Alternatively, we could create an (internal test) function to do it and include a comment that warns readers that it is a dirty hack just for tests.

----

Reference:

* The default `ParsePredicate` (only used when slicing is disabled) is `SelectIfMatch SelectFromMainHeaders`.
* The default `SelectPredicate` (always used) is `SelectAll`.

TravisCardwell:
I made the requested changes:

* `Predicate` is refactored to have logical constructor names (such as `PTrue`), avoiding *select* terminology since the type is used in both the `Parse` and `Select` passes.

* Command-line options are changed to reflect the desired terminology.  `Parse` options are either `parse-*` or `parse-except-*`.  `Select` options are either `select-*` or `select-except-*`.  Note that we avoid *skip* terminology in both cases.

    ```
    --parse-all                          Parse all declarations
    --parse-from-main-headers            Parse declarations in main headers (default)
    --parse-by-header-path        PCRE   Parse declarations in headers with paths that match PCRE
    --parse-except-by-header-path PCRE   Parse except declarations in headers with paths that match PCRE

    --select-all                           Select all declarations (default)
    --select-from-main-headers             Select declarations in main headers
    --select-by-header-path        PCRE    Select declarations in headers with paths that match PCRE
    --select-except-by-header-path PCRE    Select except declarations in headers with paths that match PCRE
    --select-by-decl-name          PCRE    Select declarations with names that match PCRE
    --select-except-by-decl-name   PCRE    Select except declarations with names that match PCRE
    ```

* We no longer disable the `Parse` predicate when slicing is enabled.  If a user wants slicing to include declarations from the standard library, for example, they must configure the predicate to do so.  This is done in the slicing golden tests.

* `NonSelectedDecls` is renamed to `NonParseDecls`, to reflect that it only contains information about declarations that are excluded in the `Parse` pass.  This information is used in the `ResolveBindingSpec` pass to resolve external bindings.  It is no longer used afterwards, so the code in the `Select` pass to add information about non-selected declarations is removed.

Up next: rebasing on #932 so that macro warnings are only `Info`.

TravisCardwell:
After rebasing so that macro warnings are `Info`, there are much fewer unexpected traces.  I will work on those next.

TravisCardwell:
Since the parse predicate is always executed, even when slicing is enabled, we are now able to exclude declarations while parsing.

closes #906

TravisCardwell:
(Posting my notes)

I am now getting "transitive dependency unavailable" errors.  For example, the `program_slicing_simple` test results in the following error:

> Program slicing: Transitive dependency unavailable: 'uint64_t' (ordinary)

The problem is that the set of transitive dependencies is calculated using the `UseDeclGraph`, which does not reflect external bindings.

We talked a bit about this before, and now the code is ready to resolve it.  There are a number of different ways that this can be done, generally in two categories:

* We could update (or copy+update) the `UseDeclGraph` to remove edges as external bindings are resolved.
* We could add logic to exclude external bindings as we traverse the graph to determine the set of transitive dependencies.

Current usage of the `Ann "TranslationUnit"` annotation:

|                      | `NonParsedDecls` | `DeclIndex` | `UseDeclGraph` | `DeclUseGraph` |
| :---                 | :--:             | :--:        | :--:           | :--:           |
| `Parse`              | ✔                |             |                |                |
| `Sort`               | ✔                | ✔           | ✔              | ✔              |
| `HandleMacros`       |                  |             |                |                |
| `NameAnon`           |                  | ✔           |                | ✔              |
| `ResolveBindingSpec` | ✔                |             |                |                |
| `Select`             |                  |             | ✔              |                |
| `HandleTypedefs`     |                  |             |                | ✔              |
| `MangleNames`        |                  |             |                |                |

The `UseDeclGraph` is just used by `Select` (after `Sort`).  I think that updating it in `ResolveBindingSpec` could work well.  This is probably much more performant than alternatives anyway.

BTW, `DecluseGraph` is used in the `HandleTypedefs` pass, but it should not matter that external bindings are not reflected.

There are various possible ways to update the `UseDeclGraph` in the `ResolveBindingSpec` pass.  Here are some candidates that I considered:

1. We could add the `UseDeclGraph` to the state and update it each time a type is replaced with a `ResolvedExtBinding`.  A nice property of this approach is that the performance cost is proportional to usage of external bindings.
2. That requires threading the declaration identity through the `resolve` calls, and a lookup is performed on the same identity each time the graph is updated.  In addition, a declaration that uses the same externally bound type multiple times needlessly performs edge deletion for that type multiple times.  To address these issues, we could instead accumulate a set of externally bound types in the `resolve` calls and delete edges per declaration in bulk.  The declaration is only looked up once, the use of a set means that each externally bound type is only processed once (per declaration), and filtering can be done using `IntSet.notMember`.  (Note that `O(n)` filtering is required because edges have labels.)  The performance cost is still proportional to usage of external bindings.
3. We could instead do a "bulk update" of the `UseDeclGraph` that removes all edges to any of the external bindings used.  We already keep track of the externally bound types while processing.  This avoids having to lookup *any* of the "from" vertices, but it must traverse every declaration, not just those that use external bindings.  Filtering can of course be done using `IntSet.notMember`.

I suspect that (3) is the most performant when external bindings are used extensively, but it has a higher cost when external bindings are used sparingly.  I like "paying for what you use," so perhaps (2) is a good solution.  I implemented it, and the `program_slicing_simple` tests now pass!

The `program_slicing_selection` test still has an error.  I have rolled back my configuration changes for that test to serve as a checkpoint, and I will work on it next.

TravisCardwell:
I added a predicate that matches declarations in headers in main header directories, including subdirectories.  I did so using a separate constructor, which is the cleanest design with the current options, IMHO.

Note that there are no tests for this yet.  Shall we add multi-header tests?

Such an option does not make sense for the `HeaderPathMatches` predicate because the user specifies a regular expression, not a directory.  Indeed, users can write a regular expression to match any header path within a specified directory.

Here are the updated CLI options:

```
--parse-all                           Parse all declarations
--parse-from-main-headers             Parse declarations in main headers (default)
--parse-from-main-header-dirs         Parse declarations in main header directories
--parse-by-header-path PCRE           Parse declarations in headers with paths that match PCRE
--parse-except-by-header-path PCRE    Parse except declarations in headers with paths that match PCRE

--select-all                          Select all declarations (default)
--select-from-main-headers            Select declarations in main headers
--select-from-main-header-dirs        Select declarations in main header directories
--select-by-header-path PCRE          Select declarations in headers with paths that match PCRE
--select-except-by-header-path PCRE   Select except declarations in headers with paths that match PCRE
--select-by-decl-name PCRE            Select declarations with names that match PCRE
--select-except-by-decl-name PCRE     Select except declarations with names that match PCRE

--enable-program-slicing              Enable program slicing: Select declarations using the selection
                                      predicate, and also select their transitive dependencies
```

TravisCardwell:
I found an issue while testing manually.  Investigating...

TravisCardwell:
I do not think that separation of selection and slicing into separate passes is viable.  We would need to keep around the non-selected declarations in order to re-select them, complicating the implementation and decreasing performance.

edsko:
Let's choose different defaults:

* `--parse-from-main-header-dirs`
* `--select-from-main-headers`

Issue #895, opened 2025-07-10
Module organization


edsko:
We have an internal library (simply called `internal`) and a public library (`hs-bindgen`). 

# Organization of the public library

We have two main entry points in `h-bindgen`

* `HsBindgen.Lib`, intended for "using `hs-bindgen` as a library", both for power-users who want to configure `hs-bindgen` in ways that aren't available through simple options (this will become increasingly important with the high-level API), as well as for our own clients (the CLI)
* `HsBindgen.TH`, intended for import in client modules who want to use `hs-bindgen` in TH mode.

If there is public functionality that we users _may_ need to use, but only in specific cases, we can add these as additional modules in the public library, for example

* `HsBindgen.Config`
* `HsBindgen.TraceMsg`

This could also be done for modules that are really intended for qualified import. `HsBindgen.Lib` and `HsBindgen.TH` should still be a pretty complete picture though of how `hs-bindgen` is meant to be used (the `HsBindgen.Config` fits in this view, because these configuration types have `Default` instances, and so you don't _need_ the `Config` module in order to call a `Lib` function).

Finally, we may want to expose some stuff from the internal library on a "no guarantees basis", this should live in

* `HsBindgen.Internal.*`

# Organization of the _internal_ library

Even _within_ `internal` there are subcomponents (such as binding specifications) that have "private", "protected", and "public" parts:

* "private" would refer to types needed _only_ by that subcomponent (`UnresolvedBindingSpec`)
* "protected" would refer to how the rest of `internal` would use this component (`BindingSpec`)
* "public" would be types that we re-export in the public `hs-bindgen` lib

The distinction between "protected" and "public" is the distinction between whether something lives (only) in `internal` or `hs-bindgen`.

In order to have "private" types, suggestion (using `BindingSpec` as an example):

* Introduce `HsBindgen.<...>.BindingSpec.Private` (`UnresolvedBindingSpec` would live here)
*Convention: imports from `HsBindgen.<...>.BindingSpec.Private` should _only_ exist in other modules in `HsBindgen.<...>.BindingSpec.*`, nowhere else in `internal`.

The goal would be that if something changes in the binding spec internals, the rest of the library is unaffected; in a way this is pretending that the binding specs is its own little sub-library.

PR #930, opened 2025-07-23
Refactor `BindingSpec` API

TravisCardwell:
Module `HsBindgen.BindingSpec.Internal` is renamed to `HsBindgen.BindingSpec.Private`, and the `StdLib` is moved to `HsBindgen.BindingSpec.Private.Stdlib`.  The code is refactored so that these modules are only used within `HsBindgen.BindingSpec` and `HsBindgen.BindingSpec.Gen` (which must be kept separate due to the dependency graph).

Module `HsBindgen.BindingSpec` exports a public as well as an internal API, both using the `BindingSpec` type that wraps unresolved and resolved binding specifications.  Functions are added to unwrap this opaque public type and call the private implementations.


TravisCardwell:
This PR can be considered part of #895.

PR #952, opened 2025-07-28
improve tracer and trace messages

### LINKED ISSUE #934: Tracing: TH API configuration and custom API
I made macro warning `Info` by default and added a `--macro-warnings` CLI option that restores the reparse and typecheck messages back to warnings (#884, #932).

We need to make this option available in the TH API as well.  Doing this cleanly likely requires some refactoring of the tracing API.  Note that I simply implemented a `withCliTracer` auxiliary function to work with the current API, but feel free to completely remove that when refactoring.

In addition, use of `withTracerCustom` (to configure a `CustomLogLevel`) requires determining `AnsiColor`.  We should probably either implement a function that avoids this *or* expose `getAnsiColor` so that users do not need to use `hSupportsANSIColor` directly in their code.

dschrempf:
tracer: link report/handle to ANSI color setting

trace messages: custom log level settings

th: expose "macros are warnings" functionality

closes #934


hs-bindgen/src-internal/HsBindgen/Util/Tracer.hs
```diff
@@ -205,6 +206,40 @@ instance Default TracerConfig where
...
+
+-- | The report function needs to know about the log level. For example, with
+-- Template Haskell, errors and warnings are treated differently compared to
+-- debug and info messages.
+type Report m = Level -> String -> m ()
+
+data OutputConfig m =
+    OutputHandle {
+      _outputHandle           :: Handle
+    , _outputAnsiColorSetting :: Maybe AnsiColor
```
dschrempf, line 222: Add comment about `Nothing`.

hs-bindgen/app/HsBindgen/App/Common.hs
```diff
@@ -414,23 +411,10 @@ withCliTracer ::
   -> (Tracer IO TraceMsg -> IO a)
   -> IO (Maybe a)
 withCliTracer GlobalOpts{..} action' = do
-    ansiColor <- getAnsiColor
     let customLogLevel = case macroWarnings of
           MacroLogInfo    -> DefaultLogLevel
```
dschrempf: Cleaner to assemble the list directly. (turn option(s) to list) and then always use `customLogLevelFrom`.

PR #954, opened 2025-07-29
Change CLI include options

TravisCardwell:
When initially designing the command-line options, we decided to make `-I` add to the quote include path.  Users must use `--system-include-path` to add to the system include path.

```
--no-stdinc                 Disable standard include directories
--system-include-path DIR   System include search path directory
-I,--include-path DIR       Quote include search path directory
```

I propose that we instead make `-I` add to the system include path, for consistency with Clang and GCC.  Include directories specified in `.cabal` or `cabal.project(.local)` configuration are passed to the C compiler using `-I`, so they are added to the system include path as well.

```
--no-stdinc                    Disable standard include directories
-I,--system-include-path DIR   System include search path directory
--quote-include-path DIR       Quote include search path directory
```

Note that the long options now both explicitly specify `system` or `quote`.  I think that this will be easiest for users to understand/use.

### DIFF
--- a/hs-bindgen/app/HsBindgen/App/Common.hs
+++ b/hs-bindgen/app/HsBindgen/App/Common.hs
@@ -224,15 +224,15 @@ parseGnuOption = switch $ mconcat [
 
 parseSystemIncludeDirOptions :: Parser [CIncludePathDir]
 parseSystemIncludeDirOptions = many . strOption $ mconcat [
-      long "system-include-path"
+      short 'I'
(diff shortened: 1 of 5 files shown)


TravisCardwell:
Errors may be related to a current GitHub outage.  I will re-run them once [GitHub Status](https://www.githubstatus.com/) is green.

UPDATE: All tests indeed pass once the GitHub outage was resolved.

PR #955, opened 2025-07-29
Add version command

TravisCardwell:
A `version` command is added that prints version information for both `hs-bindgen` and `libclang`.  The raw Clang version string is printed, to give users as much information as possible.  This can be helpful to confirm which version of `libclang` is being used by `hs-bindgen` in an environment with more than one version available.

Note that a command is used instead of a `--version` option because it is impure.  The `hs-bindgen` version is static, but the `libclang` version depends on the `libclang` shared library that is dynamically linked.

### DIFF
--- a/hs-bindgen/app/HsBindgen/App/Cli.hs
+++ b/hs-bindgen/app/HsBindgen/App/Cli.hs
@@ -11,25 +11,36 @@ module HsBindgen.App.Cli (
   , pureParseCmdPreprocess
   ) where
 
+import Data.List qualified as List
+import Data.Text qualified as Text
+import Data.Version (showVersion)
 import GHC.Generics (Generic)
 import Options.Applicative
 
 import Clang.Paths
+import Clang.Version (clang_getClangVersion)
 import HsBindgen.Lib
 
 import HsBindgen.App.Common
+import Paths_hs_bindgen qualified as Package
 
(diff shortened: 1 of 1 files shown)


hs-bindgen/app/HsBindgen/App/Cli.hs
```diff
@@ -81,6 +82,9 @@ parseCliCmd = subparser $ mconcat [
     , cmd "resolve" (CliCmdResolve <$> parseResolveOpts) $ mconcat [
           progDesc "Resolve C headers to source paths, for debugging"
         ]
+    , cmd "version" (pure CliCmdVersion) $ mconcat [
```
edsko: One option you could try is

```hs
parseCliCmd = asum [
    flag' CliCmdVersion $ mconcat [ long "version" ]
  , subparser $ mconcat [ .. ] 
  ]
```

edsko (approved):
Let's try to see if we can make `--version` work, but either way feel free to merge.

hs-bindgen/app/HsBindgen/App/Cli.hs
TravisCardwell: This indeed works, but we decided to simply get the Clang version before executing the parser, so that we do not complicate the parser.

PR #956, opened 2025-07-29
Improvements to Template Haskell API

### LINKED ISSUE #728: Avoid need for list of headers as argument to `hashInclude`
As of #727, `hashIncludes` takes a _list_ of headers to process. Intuitively, this should not be necessary; after all, in C we can simply do

```c
#include "a.h"
#include "b.h"
```

It would be nice if we could translate this more directly to

```hs
hashInclude "a.h" def
hashInclude "b.h" def
```

This is not completely trivial, because the declarations in `b.h` might depend on `a.h`, so we'll probably need to thread a binding specificaiton through.

dschrempf:
Introduces `withHsBindgen` and monadic `hashInclude`.

Closes #728 and #703.

Also improve consistency of system/quote includes (parts of https://github.com/well-typed/hs-bindgen/issues/958).

As stated in detail at https://github.com/well-typed/hs-bindgen/issues/958#issuecomment-3139856539, we also removed options related to `-isystem`, and `-iquote`. Instead, we only use `-I`.

This also fixes a bug: `-I` was used instead of `-iquote`.

### DIFF
--- a/c-expr/test/Main.hs
+++ b/c-expr/test/Main.hs
@@ -127,7 +127,7 @@ main = do
               { Clang.clangTarget =
                   Just (Clang.Target_Linux_X86_64, Clang.TargetEnvDefault)
               , Clang.clangStdInc = isNothing mbHsBindgenDir
-              , Clang.clangExtraSystemIncludeDirs =
+              , Clang.clangExtraIncludeDirs =
                   [ fromString (hsBindgenDir </> "musl-include/x86_64")
                   | hsBindgenDir <- maybeToList mbHsBindgenDir
                   ]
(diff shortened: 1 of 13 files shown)


edsko (approved):
Amazing how subtle this all is! Feel free to merge once the final changes are in.

@TravisCardwell As you will see, we changed our minds once again here. @dschrempf is going to leave a summary of our conclusions; we should merge this into your "the story of #include" doc.

dschrempf:
Link to summary: https://github.com/well-typed/hs-bindgen/issues/958#issuecomment-3139856539

PR #957, opened 2025-07-30
Helpful error message when Prelude was not imported with TH

### LINKED ISSUE #762: Provide helpful error message if user forgets imports in TH mode
We cannot generate the imports, but we _can_  ([I think](https://hackage-content.haskell.org/package/template-haskell-2.23.0.0/docs/Language-Haskell-TH.html#v:lookupTypeName), not 100% sure) check what is in scope (try to resolve the name), and if not, provide a useful error message ("hint: import such-and-such").

We might also want to minimize how many imports are needed by providing some `HsBindgen.Runtime.Prelude` module in `hs-bindgen-runtime`.

dschrempf:
If `import qualified HsBindgen.Runtime.Prelude` is missing in modules using `hs-bindgen` with Template Haskell, we now throw a helpful error message.

Closes #762.


hs-bindgen/test/th/Test/TH/Test01.hs
```diff
@@ -20,6 +20,7 @@ import HsBindgen.TH
 
 -- Used by generated code
 import Foreign.C.Types
+import HsBindgen.Runtime.Prelude qualified
```
phadej, line 17: I'm not sure why the `Prelude` is included, but then at least `SizedByteArray` is probably not needed. Nor `Foreign.C.Types`. Both imports are there because type-checking of generated code needs constructors in scope (which is - if you ask me - a bug in TH).

dschrempf: You are right, we can remove 'Foreign.C.Types', I think it is included in 'HsBindgen.Runtime.Prelude'. I also added the 'SizedByteArray' to the 'Prelude'.

Further, I changed the error message so that it states that we need the constructors to be in scope.

If you consider this a bug in TH, is there an issue we can link to or should we probably create one?

phadej: There are issues like
- https://gitlab.haskell.org/ghc/ghc/-/issues/25774
- https://gitlab.haskell.org/ghc/ghc/-/issues/8510

which are more about extensions, but arguably requiring constructors to be in scope for `coerce` or `CApiFFI` is just another variant of "TH splices care too much about surroundings they are spliced into"

(i.e. perfectly we would not need to ask users to enable dozens of extensions either - we know we need them, we don't need rest of the module to use them).

dschrempf:
I am merging this into the other feature branch to avoid conflicts.

Issue #809, opened 2025-06-26
Make it possible to keep a type opaque


edsko:
We make types opaque when (1) they lack a declaration on the C side or (2) for unions. However, `rust-bindgen` also gives users the [explicit option](https://rust-lang.github.io/rust-bindgen/opaque.html) to make types opaque:

> Sometimes a type definition is simply not translatable to Rust, for example it uses [C++'s SFINAE](https://en.wikipedia.org/wiki/Substitution_failure_is_not_an_error) for which Rust has no equivalent. In these cases, it is best to treat all occurrences of the type as an opaque blob of bytes with a size and alignment. bindgen will attempt to detect such cases and do this automatically, _but other times it needs some explicit help from you_.

(my emphasis). I am not currently aware of any situations in which we cannot detect whether something should be opaque or not, so marking this as a low-priority enhancement issue for now unless we have a specific use-case.

edsko:
@TravisCardwell The only reason we might actually do this ticket sooner rather than later is if it gives a more natural implementation of prescriptive binding specs?

TravisCardwell:
I agree with the current low priority, which should be increased when we have a need for this.  Perhaps it would be useful to simplify exposed bindings even when it is not strictly necessary.

Regarding binding specifications, perhaps a boolean is sufficient?

```yaml
- types:
  - headers: example.h
    cname: struct foo
    opaque: true
```

edsko:
Moving this to high priority, because this has one very important benefit: it means that if we do program slicing, any of the dependencies of this type can then be skipped. This can be quite important to to avoid pulling in a whole lot of definitions. See https://discourse.haskell.org/t/automatically-packaging-a-haskell-library-as-a-swift-binary-xcframework/12418/8 for discussion.

edsko:
Addressing this properly, together with program slicing, will probably also mean fixing #910.

edsko:
@TravisCardwell I noticed that this isn't assigned to you; fixed that.

edsko:
This will be configured as a property of the _Haskell_ type, `emptydata`. See also discussion in https://github.com/well-typed/hs-bindgen/issues/1529.

It is critical that program slicing takes this into account.

edsko:
[...]

Issue #799, opened 2025-06-26
Binding specifications: squashing


TravisCardwell:
When there is a `typedef` for a named `struct`/`union`/`enum` with a tag that differs from the `typedef` name, (by default) we create a distinct type to represent the `typedef` because we want to preserve semantic information.

Example:

```c
typedef struct foo { ... } foo_t;
```

```haskell
data Foo = Foo { ... }
...

newtype Foo_t = Foo_t { ... }
...
```

When the `typedef` name is the same as the tag, (by default) we "squash" the `typedef`.

Example:

```c
typedef struct foo { ... } foo;
```

```haskell
data Foo = Foo { ... }
...
```

A prescriptive binding specification should be able to configure this behavior.

* A user should be able to "squash" a specific `typedef` (when the name differs from the tag).  Note that the `typedef` name must be registered as an alias in this case.  Since there may be any number of `typedef` declarations for a type, there may be any number of aliases.

* A user should be able to "unsquash" (?) a specific `typedef` (when the name equals the tag name).  Note that in this case the user will need to override one of the names so that there is no conflict, and the `typedef` name is *not* registered as an alias.

How should we represent this?  Perhaps each type may specify a list of aliases.

```yaml
- types:
    - headers: example.h
      cname: foo
      aliases: foo_t
      module: Acme.Example
      identifier: Foo
      instances:
        - Eq
        - Show
        - Storable
```

How might this information be useful in external binding specifications?

If we would like to generate C code (tests) that reference external bindings, we must determine valid spellings.  This would be a step toward doing that, but it is *not* sufficient because we do not provide a way to distinguish generated names from "ordinary" names.

TravisCardwell:
Related to #566

edsko:
@TravisCardwell Let's discuss this; I also have some WIP related to 566, which is now undoubtedly very stale.

edsko:
Let's hold off on this until I close #566.

edsko:
After discussing this with @TravisCardwell once more, we have decided to with a slightly different approach, where we separate out properties of C types (specifically, which Haskell types they map to), and properties _of_ those Haskell types (like instances). This would then avoid the need for an explicit "alias" concept (which was previously present already anyway in external binding specs: we could always map multiple C types to the same Haskell type).

For the `vector` example:

```yaml
c_types:
  - headers: vector.h
    cname: vector 
    module: Data.Vector
    identifier: Vector
  - headers: vector.h
    cname: struct @vector 
[...]

Issue #308, opened 2024-11-28
Make choice between `newtype` and `type` configurable


edsko:
We can do this at two levels:

* A _global_ default: do we want to use Haskell  `newtype`s or `type` aliases for C typedefs (which would then apply to `typedef int my_clock_t` as well as `typedef struct foo foo_s` equally). Currently the default here is to use `newtype`s everywhere (see also #173)
* More targeted overrides _of_ the defaults, when using `hs-bindgen` as a library (see below).

Background:

In #306 we discussed what to do with things like

```c
struct foo { int x; int y; };
typedef struct foo foo_s;
```

The conclusion was that we'd want to generate a Haskell `newtype` for `foo_s`, unless `foo_s` and `foo` are spelled identically. 
We could alternatively make `foo_s` a Haskell type alias for `foo`, but this feels somewhat inconsistent with #173; in particular, although it is hard to imagine that we might want different type class instances for `foo` and `foo_s` in this example, it's entirely conceivable we might have something like

```c
typedef foo_s some_special_use_of_foo
```

elsewhere in the header file for which we _do_ want to generate a `newtype`. One could imagine heuristics such as

* if the name `foo` and `foo_s` are "sufficiently similar" (`_s` suffix, ..?)
* if the `foo` and `foo_s` are "suffiently closely" defined (same C definition, consecutive lines, ..)

but unless we can find rules that seem to apply quite universally, it might be better to leave these to individual users (using hs-bindgen`-as-a-library); we _should_ however provide the necessary hooks to implement such rules.

edsko:
Since the choice between `newtype` and `type` should be done on a case by case basis, this should piggy-back on the work done by @TravisCardwell for external bindings, which provides a precise way of distinguishing one C identifier from another.

edsko:
@TravisCardwell I'm going to assign this to you, as this should really be done as part of input binding specifications.

PR #960, opened 2025-07-31
Distinguish anonymous declarations

TravisCardwell:
New type constructor `QualNameAnon` is added to the `QualName` type to distinguish anonymous declarations in binding specifications and select predicates.  The string representation uses an `@` prefix.

Example: `@SC_c`

This character should not cause issues with regular expressions.

This character is reserved for future use by YAML and must therefore be quoted.

Example:

```yaml
cname: '@S1_c'
```

### DIFF
--- a/hs-bindgen/fixtures/anonymous.bindingspec.yaml
+++ b/hs-bindgen/fixtures/anonymous.bindingspec.yaml
@@ -8,33 +8,9 @@ types:
   - Show
   - Storable
 - headers: anonymous.h
-  cname: S1_c
-  module: Example
-  identifier: S1_c
-  instances:
-  - Eq
-  - Show
-  - Storable
-- headers: anonymous.h
   cname: struct S2
   module: Example
   identifier: S2
-  instances:
-  - Eq
-  - Show
-  - Storable
-- headers: anonymous.h
-  cname: S2_inner
-  module: Example
-  identifier: S2_inner
-  instances:
-  - Eq
-  - Show
-  - Storable
-- headers: anonymous.h
-  cname: S2_inner_deep
-  module: Example
-  identifier: S2_inner_deep
   instances:
   - Eq
   - Show
@@ -48,7 +24,31 @@ types:
   - Show
   - Storable
 - headers: anonymous.h
-  cname: S3_c
+  cname: '@S1_c'
+  module: Example
+  identifier: S1_c
+  instances:
+  - Eq
+  - Show
+  - Storable
+- headers: anonymous.h
+  cname: '@S2_inner'
+  module: Example
+  identifier: S2_inner
+  instances:
+  - Eq
+  - Show
+  - Storable
+- headers: anonymous.h
+  cname: '@S2_inner_deep'
+  module: Example
+  identifier: S2_inner_deep
+  instances:
+  - Eq
+  - Show
+  - Storable
+- headers: anonymous.h
+  cname: '@S3_c'
   module: Example
   identifier: S3_c
   instances:
--- a/hs-bindgen/fixtures/attributes.bindingspec.yaml
+++ b/hs-bindgen/fixtures/attributes.bindingspec.yaml
@@ -16,14 +16,6 @@ types:
   - Show
   - Storable
 - headers: attributes.h
-  cname: baz
-  module: Example
-  identifier: Baz
-  instances:
-  - Eq
-  - Show
-  - Storable
-- headers: attributes.h
   cname: struct foo
   module: Example
   identifier: Foo
@@ -32,7 +24,15 @@ types:
   - Show
   - Storable
 - headers: attributes.h
-  cname: qux
+  cname: '@baz'
+  module: Example
+  identifier: Baz
+  instances:
+  - Eq
+  - Show
+  - Storable
+- headers: attributes.h
+  cname: '@qux'
   module: Example
   identifier: Qux
   instances:
--- a/hs-bindgen/fixtures/distilled_lib_1.bindingspec.yaml
+++ b/hs-bindgen/fixtures/distilled_lib_1.bindingspec.yaml
@@ -18,37 +18,9 @@ types:
   - Real
   - Storable
 - headers: distilled_lib_1.h
-  cname: a_typedef_enum_e
-  module: Example
-  identifier: A_typedef_enum_e
-  instances:
-  - Eq
-  - Ord
-  - Read
-  - Show
-  - Storable
-- headers: distilled_lib_1.h
   cname: struct a_typedef_struct
   module: Example
   identifier: A_typedef_struct_t
-  instances:
-  - Eq
-  - Show
-  - Storable
-- headers: distilled_lib_1.h
-  cname: another_typedef_enum_e
-  module: Example
-  identifier: Another_typedef_enum_e
-  instances:
-  - Eq
-  - Ord
-  - Read
-  - Show
-  - Storable
-- headers: distilled_lib_1.h
-  cname: another_typedef_struct_t
-  module: Example
-  identifier: Another_typedef_struct_t
   instances:
   - Eq
   - Show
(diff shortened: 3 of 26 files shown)


TravisCardwell:
In `vector.h`, the vector typed is defined as follows:

```c
typedef struct { ... } vector;
```

Currently, the following binding specification is generated:

```yaml
types:
- headers: vector.h
  cname: '@vector'
  module: Vector
  identifier: Vector
  instances:
  - Eq
  - Show
  - Storable
```

When using this binding specification as an external binding specification to create bindings for `vector_length.h`, we run into an issue.

* After the `NameAnon` pass, there is a `typedef` named `vector` for an anonymous `struct` named `@vector`.
* In the `ResolveBindingSpec` pass, references to `@vector` are correctly replaced with external bindings.
* In the `Select` pass, function `len vector_length(vector* p);` references `vector`, *not* `@vector`, since `typedef`s are not handled/collapsed until the following `HandleTypedefs` pass.  Type `vector` is (correctly) not selected because it is not in the main file, resulting in a missing declaration error.

A pending task is to add alias information to binding specifications.

We already have a `declAliases :: [C.Name]` field in the AST `DeclInfo`.  It is initialized to the empty list, and we currently set it in the `NameAnon` pass.  Perhaps that is too early, since prescriptive binding specifications should be able to configure this (#308).  Perhaps the aliases should be set in the `HandleTypedefs` pass?

When generating a binding specification, we need to output any aliases for each type.  The above would become the following.

```yaml
types:
- headers: vector.h
  cname: '@vector'
  aliases:
  - vector
  module: Vector
  identifier: Vector
  instances:
  - Eq
  - Show
  - Storable
```

When loading an external binding specification, we must make sure that there is not duplicate configuration for the same type.  Configuration for `({vector.h}, @vector)` and `({vector.h}, vector)` may not both exist.

When resolving external bindings, any use of an alias is also replaced with an external binding.  In this case, `vector` would have an external binding, and there would no longer be a missing declaration error.

Thoughts?

TravisCardwell:
> Perhaps the aliases should be set in the `HandleTypedefs` pass?

This does not work, because aliases must be known during the `ResolveBindingSpec` pass.  I think it is indeed appropriate to initialize aliases in `NameAnon`, update them in `ResolveBindingSpec`, and execute them (remove alias declarations) in `HandleTypedefs`.  Perhaps we need to move some of the logic earlier, though, because we must know what types are aliases in `ResolveBindingSpec`.

We need to decide about the binding specification syntax.  How should we specify if a `typedef` should be a `newtype` or `type` declaration in Haskell?

I am thinking through one option below, but I am of course open to alternate ideas!

----

One option is to just use `aliases`, a list of `CName`.  (Since all aliases are `typedef`s, the are always in the ordinary namespace and are never anonymous.)  If `aliases` is specified, it must list every alias.  We can create a `type` declaration for each alias that does not have the same Haskell name as the type being aliased.  Perhaps we need to provide a way to specify a custom Haskell name for an alias.  When `aliases` is specified, any `typedef` alias that is *not* in the list should result in a `newtype` declaration to create a distinct Haskell type.  (`HandleTypeDefs` must not remove that declaration, even when there is only one use.)

----

```h
typedef struct foo { ... } foo_t;
```

**(1)** User wants distinct types

```yaml
types:
  - headers: acme.h
    cname: struct foo
    aliases: []
    module: Acme
    identifier: Foo
  - headers: acme.h
    cname: foo_t
    aliases: []
    module: Acme
    identifier: Foo_t
```

This is the default if `foo_t` is not the only use of `struct foo`.  In cases where there is only one use, users can use this syntax in a prescriptive binding specification to force this behavior.

**(2)** User wants `type Foo_t`

```yaml
types:
  - headers: acme.h
    cname: struct foo
    aliases: foo_t
    module: Acme
    identifier: Foo
```

**(3)** User wants `type FooT` (custom name)

```yaml
types:
  - headers: acme.h
    cname: struct foo
    aliases:
      - cname: foo_t
        identifier: FooT
    module: Acme
    identifier: Foo
```

**(4)** User wants no declaration for the alias

```yaml
types:
  - headers: acme.h
    cname: struct foo
    aliases:
      - cname: foo_t
        identifier: Foo
    module: Acme
    identifier: Foo
```

**(5)** User wants a single declaration using the alias name

```yaml
types:
  - headers: acme.h
    cname: struct foo
    aliases: foo_t
    module: Acme
    identifier: Foo_t
```

This is the default if `foo_t` is the only use of `struct foo`.

----

```h
typedef struct foo { ... } foo;
typedef struct boo { ... } Boo;
```

In cases where the default Haskell names conflict, and the `typedef` alias is not the only use, a prescriptive binding specification must resolve the conflict.  The same basic patterns as above may be used.

**(6)** User wants distinct types

```yaml
types:
  - headers: acme.h
    cname: struct foo
    aliases: []
    module: Acme
    identifier: Foo
  - headers: acme.h
    cname: foo
    aliases: []
    module: Acme
    identifier: FooT
```

----

```h
typedef struct { ... } foo;
```

Anonymous types always have a single use.

**(7)** User wants just `data Foo`

```yaml
types:
  - headers: acme.h
    cname: '@foo'
    aliases: foo
    module: Acme
    identifier: Foo
```

This is the default.

**(8)** User wants distinct types (for some reason)

```yaml
types:
  - headers: acme.h
    cname: @foo
    aliases: []
    module: Acme
    identifier: FooStruct
  - headers: acme.h
    cname: foo
    aliases: []
    module: Acme
    identifier: Foo
```

**(9)** User wants `data Foo` to be opaque

```yaml
types:
  - headers: acme.h
    cname: '@foo'
    aliases: foo
    module: Acme
    identifier: Foo
    opaque: True
```

Configuration of opaque types is a separate issue (#809), but we should consider the relationship between various options.  In this example, the syntax configures a single `Foo` type that is opaque.

edsko:
For 

```c
typedef struct foo { ... } foo_t;
```

```yaml
types:
  - headers: acme.h
    cname: foo_t
    module: Acme
    identifier: Foo_t
    representation: {newtype/synonym/transparent}
```

* `newtype`: `newtype Foo_t = Foo_t Foo`
* `synonym`: `type Foo_t = Foo`
* `transparent`: any reference to `Foo_t` will become a reference to `Foo` ("squashed"). (In this case some fields, such as `identifier` and `instances`, _must_ be omitted).

We'd _also_ have an entry for `foo` (anonymous or not); in the anonymous case, it would look something like

```yaml
types:
  - headers: acme.h
    cname: '@foo'
    module: Acme
    identifier: Foo # can be omitted (then we use the default naming)
```

The user can now pick which identifier we use; without a prescriptive binding specification, we would use `Foo_t` when squashing the typedef; but with the prescriptive binding spec, the user would choose.

(Perhaps `representation` applies in other situations too; for example, an enum could be `transparent` or `synonym`.)

edsko:
If the user wants a type to be opaque:

```yaml
types:
  - headers: acme.h
    cname: '@foo'
    module: Acme
    identifier: Foo
    representation: opaque
```

or

```yaml
types:
  - headers: acme.h
    cname: foo_t
    module: Acme
    identifier: Foo
    representation: opaque-and-omit-anything-this-depends-on-if-not-used-anywhere-else
```

`opaque-and-omit-anything-this-depends-on-if-not-used-anywhere-else` will require some further thought. As is, this is really hard to implement; but unclear what the right solution is. Explicitly listing everything that needs to be omitted is possible, but awkward, users might end up having to list a ton of internal types where all the want to do is make the wrapper public type opaque. Conversely, if we say make "everytihng this depends on opaque" (irrespective or whether it's used anywhere else) is easier to implement, but much less useful. More thought needed. This is not specific to binding specifications; we already have this proble when generating bindings for something like

```c
struct FILE__ { ... };

typedef struct FILE__ FILE;
```

_If_ we can implement this (perhaps it's not that difficult?), then `opaque` should mean `opaque-and-omit-anything-this-depends-on-if-not-used-anywhere-else`. Not sure if that default would need overriding, but unlikely: the point of making it opaque is that the implementation is internal, and if those types are not used anywhere else, there is no point generating bindings for them.

Side note: generated binding specifications do _not_ in general need to record omitted types: this is the purpose of selection and program slicing. Recording everything we don't generate bindings for (say, everything in the sys headers) could be an enormous list. However, explicit omits in prescriptive binding specifications _should_ be included in the generated binding specifications, so that the output binding spec is a valid extension of the input binding spec.

TravisCardwell:
(Rebased)

TravisCardwell:
This PR is blocked because "aliases" are not represented in binding specifications.  We have decided to refactor binding specifications to instead specify C and Haskell types separately (#799).  I need the work already done in this PR, so I am hijacking it.  I just renamed it.

I rebased the branch and changed `QualName` to distinguish the tag kind for anonymous types.  We now write `struct @foo` instead of just `@foo`.

TravisCardwell:
This PR is quite out of date now, so I am restoring the original title and closing it.  I will create a new PR with an updated implementation once this is unblocked.

PR #961, opened 2025-07-31
Cancel the Simple GHA job automatically when pushing to a PR

jorisdral:
If a new version of a branch is pushed to a PR, we currently don't cancel outdated jobs. They run until they finish. With this change, such "obsolete" jobs will be cancelled automatically.

### DIFF
--- a/.github/workflows/simple.yml
+++ b/.github/workflows/simple.yml
@@ -8,6 +8,10 @@ on:
     branches:
       - main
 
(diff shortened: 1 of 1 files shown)


phadej:
note that this will cancel builds for multi-commit pull requests, i.e. not only for "wiggle" "WIP" "hoops" commits in draft pull requests.

Issue #855, opened 2025-07-02
Avoid reference to `hs-bindgen-root` in errors


edsko:
Something like this:

```
[Error  ] [Libclang ] hs-bindgen-root.h:1:10: fatal error: 'foo.h' file not found
```

is confusing. This is a regression, introduced in #792, when we removed `ParseCHeadersException`.

Issue #958, opened 2025-07-30
Consistent handling of system/quote includes


edsko:
Guiding principles for terminology:

1. A "directory" is just that: a directory (which is a `FilePath`)

2. We _never_ refer to a "path". We either refer to a `FilePath` or a
   `SearchPath` (which is a list of directories).

3. When we include a header

   ```c
   #include <foo.h>
   #include "foo.h"
   ```

   it's a `FilePath` combined with an instruction on how to resolve that
   `FilePath`.

   ```hs
   -- | The argument to a @#include@ directive in C.
   data HashIncludeArg =
       -- | @#include "foo.h"@
       Quote FilePath

       -- | @#include <foo.h>@
     | System FilePath
   ```

# `clang`

## Include paths

```hs
newtype CIncludeDir == CIncludeDir FilePath

data ClangArgs = ClangArgs{
    -- This is not a search path: these are /additional/ directories to be
    -- /added/ to the search path.
    clangExtraQuoteIncludeDirs  :: [CIncludeDir]
  , clangExtraSystemIncludeDirs :: [CIncludeDir]
  }
```

## `#include` args

Not relevant in `clang`: move `CHeaderIncludePath` to `hs-bindgen`.

# hs-bindgen

No parsing, but validation (sanity check: forward slash only, no absolute
paths, ..). Don't generate errors, but report warnings (tracing).

Independent: allow CLI flags to disable specific warnings (`-Wno-foo`).

# CLI

## Include paths

1. `--system-include-dir` (alias: `-I`)
2. `--quote-include-dir`

## `#include` args

1. `--input-quote`
2. `--input-system`

No `-i` alias: force people to think about what this argument means.

Independent: we should reserve single letters for very common options only; `-s`
etc. should go.

# TH

## Include paths

```hs
data IncludeDir =
     IncludeDir        FilePath
   | RelativetoPkgRoot FilePath

data HashIncludeOpts = HashIncludeOpts {
      extraQuoteIncludeDirs  :: [IncludeDir]
    , extraSystemIncludeDirs :: [IncludeDir]
    , ..
    }
```

## `#include` args

1. `hashInclude :: HashIncludeArg -> WithHsBindgen ()`

```hs
withHsBindgen def $ do
  hashInclude $ Quote "foo.h"
```

# Binding specs

## Include paths

Not relevant (depend on the local system).

## `#include` args

```yaml
types:
  - cheaders-quote: ["foo.h"]
[...]

PR #963, opened 2025-07-31
Only support "system" (i.e., angular bracket) #include directives

dschrempf:
See issue https://github.com/well-typed/hs-bindgen/issues/958.

### DIFF
--- a/hs-bindgen/app/HsBindgen/App/Common.hs
+++ b/hs-bindgen/app/HsBindgen/App/Common.hs
@@ -22,7 +22,6 @@ module HsBindgen.App.Common (
   , cmd'
   ) where
 
-import Control.Exception (Exception (displayException))
 import Control.Monad.IO.Class (MonadIO)
 import Data.Bifunctor (Bifunctor (bimap), first)
 import Data.Char qualified as Char
@@ -423,7 +422,7 @@ parseInputs = some . argument (eitherReader parseHeader) $ mconcat [
     ]
   where
     parseHeader :: String -> Either String HashIncludeArg
-    parseHeader = first displayException . parseHashIncludeArg
+    parseHeader = first (show . prettyForTrace) . hashIncludeArgEither
 
 {-------------------------------------------------------------------------------
   Auxiliary hs-bindgen functions
--- a/hs-bindgen/clang-ast-dump/Main.hs
+++ b/hs-bindgen/clang-ast-dump/Main.hs
@@ -61,7 +61,7 @@ instance Exception AstDumpException where
 
 clangAstDump :: Options -> IO ()
 clangAstDump opts@Options{..} = do
-    putStrLn $ "## `" ++ renderHashIncludeArg optFile ++ "`"
+    putStrLn $ "## `" ++ getHashIncludeArg optFile ++ "`"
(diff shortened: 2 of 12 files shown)


dschrempf:
@TravisCardwell this removes `Quote` `HashIncludeArg`s since we do not use them. I understand you may have conflicting code. If so, feel free to cherry pick what you like, or just merge what you have and we can rebase relevant changes from this PR onto your work.

dschrempf:
This still needs some more work. As far as I can see, the current directory is only in the "quote include search path" , leading to weird traces which we would have to change. For example,

```console
(C) + cabal run -- hs-bindgen-cli preprocess hs-bindgen/examples/simple.h
```
leads to
```console
[Error  ] [Libclang ] hs-bindgen-root.h:1:10: error: 'hs-bindgen/examples/simple.h' file not found with <angled> include; use "quotes" instead
An error happened (see above)
```

One can easily circumvent this by adding `-I ./`:
```console
(C) + cabal run -- hs-bindgen-cli preprocess -I ./ hs-bindgen/examples/simple.h
```

edsko:
Hmm I almost think that's a feature, not a bug. The bindings we would generate would have `#include <hs-bindgen/examples/simple.h>` as their input, which is probably not what you want (not even with quotes either, since "current working directory" is not really well-defined when compiling the bindings).

TravisCardwell:
```
cabal run -- hs-bindgen-cli preprocess hs-bindgen/examples/simple.h

cabal run -- hs-bindgen-cli preprocess -I ./ hs-bindgen/examples/simple.h
```

Both of these are indeed questionable usage.  The last one works as long as there are no (system) includes that need `examples` to be in the C include search path.  That is the case for our examples.  I generally consider `-I .` to be a red flag, but upon further consideration perhaps it does not matter as long as it works.

```
[Error  ] [Libclang ] hs-bindgen-root.h:1:10: error: 'hs-bindgen/examples/simple.h' file not found with <angled> include; use "quotes" instead
```

This error message definitely needs to change.

* `hs-bindgen-root.h` should not be exposed to users
* "file not found" is pretty misleading.  Perhaps something like "unable to resolve `#include <hs-bindgen/examples/simple.h>`" would be better?

dschrempf:
@TravisCardwell let us know if you also fix the error trace, or if this is a task to be done later. Thanks for picking this up!

TravisCardwell:
I went ahead and fixed the error trace as well.  "File not found" errors are formatted as I suggested above, but feel free to change it if different wording would be easier to understand.  In case there are any other errors, `hs-bindgen-root.h` is stripped from the formatted message.

TravisCardwell:
I went ahead and cherry-picked the commits into this branch for easy review. :bow:

edsko:
Does that mean we can close https://github.com/well-typed/hs-bindgen/issues/855 aftre this PR?

TravisCardwell:
Yes!

closes #855

edsko:
Excellent, glad we have reached some kind of conclusion this. 

I'm sure it'll last at least 24 hours before we realize this is still completely wrong :grimacing:

dschrempf:
(Maybe it lasts until Monday :-)).

PR #969, opened 2025-08-04
Support `const` global variables

### LINKED ISSUE #41: Generate bindings for constants
When a C header defines

```c
extern const int x;
```

we should generate a binding `x :: CInt`.

jorisdral:
Resolves #41


edsko (approved):
Great, chipping away at the final missing C features :) Feel free to merge once the final changes we discussed are in.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -1308,31 +1308,79 @@ functionDecs mu typedefs info f _spec =
...
+      -- We are generating a binding for a global variable here. This binding
+      -- must be marked NOINLINE, so that it will be evaluated at most once.
+      -- /If/ we have a Storable instance, but that storable instance has a
+      -- superclass constraint, then we could _in principle_ add that superclass
+      -- constraint to as a constraint to the type of the global, but this would
+      -- then turn the global into a function instead.
+    , Hs.Storable
+        `elem`
+          getInstances instsMap "unused" (Set.singleton Hs.Storable) [typ ty]
     ]
```
jorisdral: @TravisCardwell do you have any thoughts on the approach here? Can we somehow ensure that the `Storable` instance we're finding here has no (unsolvable) superclass constraints?

TravisCardwell: The motivation and design both sound good to me. :+1:

The `InstanceMap` in the current code is a map from the type to instances.

```haskell
type InstanceMap = Map (HsName NsTypeConstr) (Set HsTypeClass)
```

I implemented constraints right when the `Frontend` refactor transitioned from experimental to top priority, and I have yet to integrate it into the new codebase.  The `InstanceMap` in my aborted constraints branch is a map from the type to instances, where each instance is associated with a set of types to be used as constraints (of the same type class).

```haskell
type InstanceMap =
  Map (HsName NsTypeConstr) (Map HsTypeClass (Set (HsName NsTypeConstr)))
```

Discussing with Edsko at the time, we decided that it would be wise to change this so that we can have arbitrary constraints, not just constraints of the same type class (even though there is no need for this now).  Perhaps it will look something like the following.

```haskell
type InstanceMap =
  Map
    (HsName NsTypeConstr)          -- type the instance is generated for
    ( Map
        HsTypeClass                -- instance type class
        ( Set
            ( HsTypeClass          -- constraint type class
            , HsName NsTypeConstr  -- constraint type
            )
        )
    )
```

The return type of `getInstances` is the type of elements in `InstanceMap`.  In the current code, `getInstances` returns a `Set HsTypeClass`.  In the aborted branch, it returns `Map HsTypeClass (Set (HsName NsTypeConstr))`.  When supporting arbitrary constraints, perhaps it will return `Map HsTypeClass (Set (HsTypeClass, HsName NsTypeConstr))`.

Your comment is very clear.  Instead of just checking for membership, we could use a predicate to check that `Storable` is a member of a map *and* has no constraints.

```haskell
    ...
    | instsMap <- toList mInstsMap
    , isUnconstrainedStorable instsMap
    ]
  where
    isUnconstrainedStorable :: InstanceMap -> Bool
    isUnconstrainedStorable instanceMap =
        (== Just Set.empty)
      . Map.lookup Hs.Storable
      $ getInstances instanceMap ...
```

This would ensure that the `Storable` instance has no constraints at code generation time.  The getter would not be generated if there are any constraints, even if the user provides instances to satisfy the constraints.  Is that acceptable?  Am I missing anything?

jorisdral: Thanks for the detailed description! Yes, your `isUnconstrainedStorable` suggestion sounds acceptable. So is my understanding correct that you still aim to proceed with the refactoring of the instance map, or was it aborted indefinitely?

edsko: Definitely not indefinitely, but I guess unclear at the moment when it will be reintroduced. Other higher priority things keep coming up :)

jorisdral: Alright, then I'll create an issue to revisit this check for instances without superclases later, and link to these comments

jorisdral: See https://github.com/well-typed/hs-bindgen/issues/993

Issue #903, opened 2025-07-11
Add `-D` global option


TravisCardwell:
From the `clang` manual:

> `-D<macroname>=<value>`
> Adds  an  implicit `#define` into the predefines buffer which is read before the source file is preprocessed.

Providing this option is a convenience, so that users do not have to specify it via Clang arguments.

Issue #592, opened 2025-04-18
External binding generation should support "private" headers


edsko:

TravisCardwell:
The include graph for the example in the manual is as follows.

```mermaid
graph BT
  GI[game_internal.h]
  GW[game_world.h]
  GP[game_player.h]

  GW --> GI
  GP --> GI
```

There are various ways that Haskell bindings could be organized:

1. Haskell modules could be created for each of these headers, *including* `game_internal.h`.  The external bindings created for `game_internal.h` would be used in the translation of `game_world.h` and `game_player.h`.  Perhaps the Haskell modules would be named `Game.Internal`, `Game.World`, and `Game.Player`, respectively.  Use of the term "internal" is not even very unfortunate, as it reflects the C organization.

2. A single Haskell module could be created for both `game_world.h` and `game_player.h`.  (This is not supported by the preprocessor or TH APIs yet, but the C parser already supports it.)  In this case, any needed declarations in `game_internal.h` are included in the generated module.  The external bindings for the generated module would include the bindings for those declarations.

3. Separate Haskell modules could be created for `game_world.h` and `game_player.h` without creating a module for `game_internal.h`.  Assuming that there is a single `Game_state` type, this does not work well because the modules would have separate declarations.

The example in the manual has the following configuration.

```yaml
types:
- headers:
  - game_world.h
  - game_player.h
  cname: game_state
  package: hs-game
  module: Game.State
  identifier: Game_state
```

If `game_world.h` is translated to `Game.World` and `game_player.h` is translated to `Game.Player`, where does `Game.State` come from?  Perhaps it is the translation of `game_internal.h` but renamed?

```
$ hs-bindgen-cli preprocess \
    -I cbits \
    -i game_internal.h \
    --package game \
    --module Game.State \
    --output src/Game/State.hs \
    --gen-external-bindings bindings/game-state.yaml
$ hs-bindgen-cli preprocess \
    -I cbits \
    -i game_world.h \
    --external-bindings bindings/game-state.yaml \
    --package game \
    --module Game.World \
    --output src/Game/World.hs \
    --gen-external-bindings bindings/game-world.yaml
$ hs-bindgen-cli preprocess \
    -I cbits \
    -i game_player.h \
    --external-bindings bindings/game-state.yaml \
    --package game \
    --module Game.Player \
    --output src/Game/Player.hs \
    --gen-external-bindings bindings/game-player.yaml
```

[...]

Issue #500, opened 2025-03-12
Manual


edsko:
This is a parent issue for all manual related tickets.

TravisCardwell:
I have a note that we may want to include a warning that GHC and C handle floating point numbers differently:

* GHC completely ignores the floating-point environment.  [Reference](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/ffi.html#floating-point-and-the-ffi)
* The `Show` instances format floating-point numbers differently than how they are formatted using standard C formatters.  In particular, the Haskell implementation finds the shortest string representation that parses to the specified floating-point value.  (This is why formatting them with the `Show` instance is so slow.)

TravisCardwell:
I am getting ready to integrate the includes documentation into the manual, and I think that the "Invocation" section should probably be split up.  How about splitting the "Generated low-level bindings" documentation into "Usage" and "Translation" categories?

In the following outline, some proposed/WIP sections are detailed to give an idea of what each page includes.  Thoughts?

----

<details><summary>Introduction</summary>

* ...

</details>

**Usage**

<details><summary>Installation</summary>

* Linux
* macOS
* Windows
* Nix

</details>

<details><summary>Invocation</summary>

* Preprocessor
* Template Haskell
* Literate Haskell
* Haskell library

</details>

<details><summary>Settings</summary>

* Command-line options
* Library options
* Environment variables

</details>

<details><summary>Includes</summary>

* Terminology
* Clang
    * C include paths
    * Command-line options
    * Environment variables
    * Header resolution
    * Debugging
* `hs-bindgen`
    * Header resolution
    * Command-line options
    * Library options
    * Environment variables
    * Header arguments
    * CAPI
    * Debugging

</details>

<details><summary>Predicates and slicing</summary>

* Parse predicates
* Select predicates
* Slicing

</details>

<details><summary>Binding specifications</summary>

* Binding specification files
* Binding specification generation
* External bindings
* Standard library bindings
* Prescriptive binding specifications

</details>

<details><summary>Test generation</summary>

* ...

</details>

<details><summary>Cross-compilation</summary>

* ...

</details>

**Translation**

<details><summary>Generated names</summary>

* ...

</details>

<details><summary>Structs</summary>

* ...

</details>

<details><summary>Enums</summary>

* ...

</details>

<details><summary>Unions</summary>

* ...

</details>

<details><summary>Functions</summary>

* ...

</details>

<details><summary>Global variables and constants</summary>

* ...

[...]

Issue #909, opened 2025-07-11
Improve handling of `clang` arguments


dschrempf:
At the moment, the function `withExtraClangArgs` _appends_ the `clang` arguments obtained from the environment. That is, the arguments specified in the `BINDGEN_EXTRA_CLANG_ARGS` environment variable supersede other arguments.

This behavior seems to be concordant with `rust-bindgen` (briefly analyzing the `rust-bindgen` code).

However, we may want to differentiate between _internal_ `clang` arguments, BINDGEN_EXTRA_CLANG_ARGS, and other `clang` arguments provided on the command line. In particular, the precedence should be as follows:

> internal-options < bindgen-extra-clang-args < command-line-options

That is, `internal-options` should have the lowest precedence, while `command-line-options` should have highest precedence.

A possibly clean solution would be:

- Amend the data type `ClangArgs` so that it has separate fields `internal-options`, `bindgen-extra-clang-args`, and `command-line-options`.
- Do not append BINDGEN_EXTRA_CLANG_ARGS in the function `withExtraClangArgs`, but only fill the specific field
- Assemble the `clang` arguments with correct precedence in the function `fromClangArgs` which convert `ClangArgs` to the actual list of command line argument strings.

TravisCardwell:
Related to #763

dschrempf:
I assigned this to myself, since we discussed it in #970.

PR #970, opened 2025-08-05
Add includes documentation, small fixes

TravisCardwell:
Summary of changes:

* The low-level documentation is organized to have separate pages for various aspects of usage, as proposed in [this comment](https://github.com/well-typed/hs-bindgen/issues/500#issuecomment-3131172020).
* A "Clang options" page is added, including the existing `BINDGEN_EXTRA_CLANG_ARGS` documentation.
* The game example is updated so that it does not expose the internal header.  [Comment](https://github.com/well-typed/hs-bindgen/issues/592#issuecomment-3148827178) in #592.  The external binding documentation is updated.
* The `--clang-option` restrictions that forced use of `hs-bindgen-cli` options are removed, as they no longer make sense.
* C language CLI options are re-ordered so that they are together.
* The includes documentation is added, updated for current usage.
* Typos found by [`typos`](https://github.com/crate-ci/typos) are fixed.
* The Haddock documentation for `DeclIndex` is improved.
* The `-D`/`--define-macro` option is added.  #903
* Generated C source now uses system includes.


hs-bindgen/src-internal/HsBindgen/SHs/Translation.hs
```diff
@@ -47,7 +47,7 @@ csources decls = unlines $ headers ++ bodies
     -- It is important that we don't include the same header more than once,
     -- /especially/ for non-extern non-static globals.
     headers = ordNub
-      [ "#include \"" ++ header ++ "\""
+      [ "#include <" ++ header ++ ">"
```
dschrempf, line 50: Ah, thanks for spotting this!

manual/LowLevel/ClangOptions.md
```diff
@@ -0,0 +1,59 @@
...
+`hs-bindgen` passes [Clang command-line options][] to `libclang`.  Some common
+options are managed by `hs-bindgen`, but you can configure arbitrary options as
+well.
+
+[Clang command-line options]: <https://clang.llvm.org/docs/ClangCommandLineReference.html>
+
+When using `hs-bindgen-cli preprocess`, Clang options may be specified as
+command-line options.  Common options are exposed as `hs-bindgen` command-line
+options, while arbitrary Clang options may be passed using `--clang-option`.
+You may specify this option any number of times.
```
dschrempf: We may want to add a short example here.

manual/LowLevel/ClangOptions.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Environment variables
+
+Clang options may also be set using environment variables.  This is
+particularly useful when setting environment-specific configuration that may
+not be hard-coded in the source code.  When used with the CLI, options passed
+with the CLI take precedence.  It is the only way to configure Clang options
```
dschrempf: I am not entirely sure if the precedence is correct. We append BINDGEN_EXTRA_CLANG_ARGS. So precedence is as you state for `-I`, `-isystem`, and `-iquote` (they are parsed left to right). However, I am not sure if other command line options take precedence when they come after earlier specified options. In this case, other options in BINDGEN_EXTRA_CLANG_ARGS would take precedence.

I will have a look a the clang documentation.

manual/LowLevel/GeneratedNames.md
```diff
@@ -184,7 +184,7 @@ data Config_Deref = Config_Deref {
 #### Top-level anonymous types
 
 For top-level anonymous type declarations we don't generate any code(such
```
dschrempf: I think a space is missing before the parenthesis.

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
+# Includes
+
+Includes are a surprisingly common source of issues.  Understanding the details
```
dschrempf: Maybe "the resolution of includes"?

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+  processed content of the referenced header is inserted into the source.  Note
+  that there may be whitespace between the `#` and `include`, used to visually
+  format blocks within macro logic.
+
+* __System include__: A system include is the inclusion of a header using an
+  include directive of the form `#include <stdio.h>`.  `hs-bindgen` only uses
+  system includes.
+
+* __Quote include__: A quote include is the inclusion of a header using an
+  include directive of the form `#include "stdio.h"`.  While `hs-bindgen` does
```
dschrempf, line 28: Should we use a different header name here? That is, a header name not clashing with a system header name?

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+* __Header resolution__: The preprocessor must locate the source of a header
+  specified in an include directive, which we call header resolution.  Header
+  resolution is implementation-specific: it may be done differently in different
+  toolchains.  See below for details about how Clang does header resolution.
+
+* __C include search path__: A C include search path is a list of directories to
+  search when resolving a header.  Similar to how the more well-known `PATH`
+  environment variable is a list of directory paths used to search for
+  executables, `C_INCLUDE_PATH` is a list of directory paths used to search for
+  headers.  The multiple meanings of the word "path" is an unfortunate source of
```
dschrempf:
```suggestion
  headers.  The multiple meanings of the word "path" are an unfortunate source of
```

(plural sounds better to me here, but I am not a native)

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+
+The following are used to add a directory to a C include search path:
+
+* `-I <directory>` adds a directory to the system C include search path, with
+  precedence over the following options.
+* `-isystem <directory>` adds a directory to the system C include search path.
+* `-iquote <directory>` adds a directory to the quote C include search path.
+
+These options can be passed multiple times, and the directories are added to
+the C include search paths in the same order.
```
dschrempf, line 83: (Maybe state "from left to right", which may be clearer than "in the same order")/

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+
+* `-nostdinc` disables the default include directories (both the standard system
+  directories and the compiler builtin directories).
+* `-nostdlibinc` disables the standard system directories.
+* `-nobuiltininc` disables the compiler builtin directories.
+
+Reference:
+
+* [Include path management](https://clang.llvm.org/docs/ClangCommandLineReference.html#include-path-management)
+
```
dschrempf, line 95: We could also link to the GCC documentation here, it is clearer than the Clang documentation, in my opinion.

https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html

https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+Note that there are also system-specific environment variables that may add to
+the system C include search path.
+
+### Clang header resolution
+
+Header resolution for a system include is done by searching for the specified
+header relative to the following directories:
+
+1. Directories in the system C include search path, in order
+2. Directories in the quote C include search path, in order
```
dschrempf, line 129: Could you give a citation for this second step? I was totally unaware of it, and do not understand why this is the case.

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+
+## `hs-bindgen`
+
+`hs-bindgen` uses `libclang` to resolve headers when generating bindings.  Note
+that this is separate from compilation of the generated bindings, described in
+the CAPI section below.
+
+`libclang` constructs C include search paths like Clang, but compiler builtin
+directories are *not* included by default.  Unless using an alternate standard
+library, the compiler builtin directory must be configured using an
```
dschrempf:
```suggestion
library, the compiler builtin directory must be configured using a
```

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+`libclang` are created based on the `hs-bindgen` command-line options and
+environment variables, described below.  In addition, `libclang` process the
+Clang environment variables, described above.
+
+[Clang options]: <ClangOptions.md>
+
+### `hs-bindgen` command-line options
+
+`hs-bindgen-cli` provides a `-I <DIRECTORY>` option that adds a directory to the
+system C include search path.  It is translated to a Clang `-I` option.
```
dschrempf:
```suggestion
system C include search path.  It is translated to the Clang `-I` option.
```

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+
+`hs-bindgen-cli` also provides a `--no-stdinc` option that disables the default
+include directories.  It is translated to a Clang `-nostdinc` option.
```
dschrempf:
```suggestion
include directories.  It is translated to the Clang `-nostdinc` option.
```

(Maybe I am wrong, but definite article sounds better to me here).

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+Note that `hs-bindgen-cli preprocess include/foo.h` does not work in general,
+even though it is convenient to type with tab-completion.  It translates to
+include directive `#include <include/foo.h>`.  Header resolution searches for
+`include/foo.h` in the system C include search path, and it will not find
+`foo.h` unless you add the current working directory to the system C include
+search path.
+
+A C include search path is used to search for C headers, so only adding
+directories that contain C header files is best practice.  Adding a directory
+that contains C source as well as C header files is a not optimal.  Adding a
```
dschrempf:
```suggestion
that contains C source as well as C header files is suboptimal.  Adding a
```

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+directory that contains many types of files (such as a Haskell project
+directory) is dirty.  Command `hs-bindgen-cli preprocess -I . include/foo.h`
```
dschrempf:
```suggestion
directory) is bad practice.  Command `hs-bindgen-cli preprocess -I . include/foo.h`
```
(?)

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+is therefore a dirty hack, but it works as long as `foo.h` (and any
```
dschrempf:
```suggestion
is therefore a hack, but it works as long as `foo.h` (and any
```

manual/LowLevel/ClangOptions.md
TravisCardwell: This documentation is written assuming that #909 is implemented.  You wrote:

```
internal-options < bindgen-extra-clang-args < command-line-options
```

> That is, `internal-options` should have the lowest precedence, while `command-line-options` should have highest precedence.

I agree that CLI options should take precedence over environment variables, in general as well as for consistency with Clang.  This is still the plan?

dschrempf: Ah great, that is after #909 we have either fixed or confirmed the behavior is correct. I will resolve this comment.

manual/LowLevel/ClangOptions.md
```diff
@@ -0,0 +1,59 @@
...
+when using the Template Haskell API.
```
edsko: "It is the only way to configure clang in TH": that should not be the case. We should extend the TH options type with clang options. @dschrempf perhaps something for you to look at?

manual/LowLevel/Includes.md
edsko: I don't think "normal" people think in terms of "resolving" includes, but I do agree that that first sentence could be improved a bit (though not essential). Perhaps something like "Correctly dealing with `#include`s in C is surprisingly tricky" or something like that?

manual/LowLevel/ClangOptions.md
dschrempf: The statement is true, but it is an easy fix (`HashIncludeOpts`). The reason why I didn't include `ClangArgs` etc. is that there are many options we could add, but probably we do not need all of them.

dschrempf: Edsko and I had a chat, see issue https://github.com/well-typed/hs-bindgen/issues/971

dschrempf: This will be fixed in https://github.com/well-typed/hs-bindgen/pull/972. If you want @TravisCardwell , you can already changethe manual, otherwise I add the change to #972. The user can now fully specify the `Config` using `withHsBindgen`.

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+
+* __Include__: An include is the inclusion of a header using a `#include`
+  directive, which is processed by the C preprocessor.  The recursively
+  processed content of the referenced header is inserted into the source.  Note
+  that there may be whitespace between the `#` and `include`, used to visually
+  format blocks within macro logic.
+
+* __System include__: A system include is the inclusion of a header using an
+  include directive of the form `#include <stdio.h>`.  `hs-bindgen` only uses
+  system includes.
```
edsko: Is this the right terminology? I just wonder/worry, because "system headers" are something very specific (headers for which clang emits no warnings).

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+  directories that are in a C include search path by default.  They generally
+  include *standard system directories* and *compiler builtin directories*.
+
+## Clang
+
+Clang maintains two separate C include search paths:
+
+* The *system C include search path* initially contains some default include
+  directories.
+* The *quote C include search path* is initially empty.
```
edsko: One thing that I keep being confused about is the difference between `-I` and `-isystem`. They both add a directory to the "system include path", but then `-isystem` _also_ marks it as a special kind of directory...?

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+
+`hs-bindgen` uses `libclang` to resolve headers when generating bindings.  Note
+that this is separate from compilation of the generated bindings, described in
+the CAPI section below.
+
+`libclang` constructs C include search paths like Clang, but compiler builtin
+directories are *not* included by default.  Unless using an alternate standard
+library, the compiler builtin directory must be configured using an
+command-line option or environment variable.  The compiler builtin directory
+should be for the version of `libclang` being used.
```
edsko: Under which circumstances are these compiler built-ins needed? Is this the behaviour that we want..?

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+
+`hs-bindgen-cli` also provides a `--no-stdinc` option that disables the default
+include directories.  It is translated to a Clang `-nostdinc` option.
+
+Clang has many more include options, which may be passed via `--clang-option`
+options.  For example, a Clang option can be used to configure the quote C
+include search path.  Since all `hs-bindgen` includes are system includes, there
+is no need to configure the quote include search path for `hs-bindgen`, but
+configuration could be necessary for the C project that bindings are being
+generated for.
```
edsko: Perhaps a more important example is `-idirafter`, which _does_ affect the system include path. (Indeed, it's like `-isystem`, in that it also marks it as a "system" directory, suppressing warnings; there does not seem to be an equivalent of `-I` but that appends rather than prepends..?).

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+$ hs-bindgen-cli preprocess -I include foo.h
+```
+
+This translates to include directive `#include <foo.h>`.  Header resolution
+searches for `foo.h` in the system C include search path and finds it in the
+`include` directory.
+
+Note that `hs-bindgen-cli preprocess include/foo.h` does not work in general,
+even though it is convenient to type with tab-completion.  It translates to
+include directive `#include <include/foo.h>`.  Header resolution searches for
```
edsko, line 232: I wonder if it would be slightly confusing that we have `#include <include/foo.h>` vs `#include <anythingElse/foo.h>`.

manual/LowLevel/Includes.md
```diff
@@ -0,0 +1,304 @@
...
+path.  These options are determined by configuration in the `.cabal` files for
+the package and all transitive dependencies, the `cabal.project` and
+`cabal.project.local` files for the project, and any command-line configuration
+(passed via `cabal` for example).  The order of the `-I` options passed to the C
+compiler is determined by GHC.  Environment variables supported by the C
+compiler (such as `C_INCLUDE_PATH`) are also relevant.
+
+Note that `hs-bindgen` configuration is not relevant.  Header resolution at this
+stage is *not* affected by `hs-bindgen-cli` options or
+`BINDGEN_EXTRA_CLANG_ARGS` environment variables.
```
edsko, line 265: This part worries me. Related to https://github.com/well-typed/hs-bindgen/issues/734

edsko:
Thanks for doing this. Left some remarks/questions. It will be great to have an authoritative reference on this topic!

manual/LowLevel/Includes.md
TravisCardwell: In this sentence, an "option" is a type of argument passed on the command-line.  Any number of options may be specified, so an indefinite article is used on purpose.  Use of the definite article does not sound right to me.  I changed to wording to avoid articles altogether.

manual/LowLevel/Includes.md
TravisCardwell: Use of the definite article does not sound right to me.  I changed to wording to avoid articles altogether.

manual/LowLevel/Includes.md
TravisCardwell: How about this?

> Configuring command-line options and environment variables so that C `#include` directives resolve to the intended headers can be surprisingly tricky.

manual/LowLevel/ClangOptions.md
TravisCardwell: I updated it, but feel free to add details if needed.

manual/LowLevel/Includes.md
TravisCardwell: I updated the wording to make this clearer.

> `libclang` constructs C include search paths like Clang, but compiler builtin directories are *not* included by default.  When using the GNU C library, a compiler builtin directory must be configured using a command-line option or environment variable.  The compiler builtin directory should be for the version of `libclang` being used.  When using Musl, configuring a compiler builtin directory should not be necessary.

It is the behavior we are stuck with, due to the implementation of `libclang`.

manual/LowLevel/Includes.md
TravisCardwell: I changed the example to use `-idirafter`.

Making `-idirafter` suppress warnings is not in the CLI documentation, which lists a separate `-isystem-after` option.  Testing, however, the `-isystem-after` option does not work for me at all, and I confirmed that `-idirafter` does indeed suppress warnings.  What a mess.

manual/LowLevel/Includes.md
TravisCardwell: Directory name `include` is the best practice.  (An include directory *should* only transitively contain `.h` files, and the convention is to name the directory `include`.  Many Haskell projects follow this convention, not just C projects.)

I updated the example to make this clear:

> For example, a best practice for Haskell projects is to store project C code in `include` and `cbits` directories, where the `include` directory contains the headers and the `cbits` directory contains the `.c` source.  That `include` directory should be added to the system C include search path, and header arguments should be relative to it.

When we create an example that includes C code (#873), we should use directories `include`, `cbits`, and `src` to promote best practices, not put all `.h` and `.c` files in the same directory.

Note that the current example is fine because shortcuts are made to create a minimal example.  The `c` directory is not part of the Haskell project; it is just a place where we put the mocked external dependencies.

manual/LowLevel/Includes.md
TravisCardwell: Indeed, getting information from Cabal would allow users to just configure the include search paths for Cabal, instead of have to duplicate the configuration for Cabal and `hs-bindgen`.

manual/LowLevel/Includes.md
TravisCardwell: How about we avoid the name *system* for `#include <foo.h>` includes?  Since we use *quote* for `#include "foo.h"` includes, we can use *bracket* for `#include <foo.h>` includes.  Since we only support `#include <foo.h>` includes, this terminology is just used in documentation anyway.

I will go ahead and add information about warning suppression of "system headers" by Clang in the documentation, since the terminology is a source of confusion.

Since we are now mentioning the Clang `-isystem` and `-idirafter` options, I rewrote the documentation about how Clang constructs the C include search paths.

manual/LowLevel/Includes.md
TravisCardwell: The updated documentation should make this clear.

manual/LowLevel/Includes.md
edsko: Yes, let's avoid "system" for `<..>`, since "system headers" are a special thing.

manual/LowLevel/Includes.md
edsko: A simple example here will clarify "resolve".

manual/LowLevel/Includes.md
edsko: https://github.com/well-typed/hs-bindgen/issues/975

PR #979, opened 2025-08-07
Mangle "use quotes instead" Clang diagnostics

TravisCardwell:
When a user creates a header file to translate, the best practice is as follows.

* Save the header in a directory that just contains header files, usually named `include`.
* Configure the C include search path to search that directory.
* Specify the header relative to that directory.

```
$ mkdir -p include
$ echo 'typedef char sym;' > include/example.h
$ cabal run hs-bindgen-cli -- preprocess -I include example.h
...
```

A user may instead configure the C include search path to search the current directory, so that they can use tab-completion to specify the header path.  (This is bad practice, but it works as long as there are no bracket includes that require the `include` directory to be in the C include search path, as described in the includes documentation in the manual.)

```
$ cabal run hs-bindgen-cli -- preprocess -I . include/example.h
...
```

This issue is reproduced if the user does not configure the C include search path.

```
$ cabal run hs-bindgen-cli -- preprocess include/example.h
[Error  ] [Libclang ] error: 'include/example.h' file not found with <angled> include; use "quotes" instead
An error happened (see above)
```

What is happening?  Clang attempts to resolve `#include <include/example.h>`.  It first checks the directories in the bracket C include search path and does not resolve the header since the project root was not added.  It then checks the directories in the quote C include search path and does not resolve the header since none are configured.  It then checks the current working directory and sees that the header would be resolved with a quote include, because quote includes attempt to resolve using the current working directory.  The header is still not resolved because a bracket include is used, but the diagnostic is emitted to provide a hint about what might be wrong.  This is helpful in general, but it is misleading in our case.

Note that attempting to preprocess a header that does not exist at all results in the following error message, which I implemented recently so that we do not disclose the internal root header name.

```
$ cabal run hs-bindgen-cli -- preprocess does_not_exist.h
[Error  ] [Libclang ] unable to resolve #include <does_not_exist.h>
An error happened (see above)
```

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Clang.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Clang.hs
@@ -124,13 +124,22 @@ instance PrettyForTrace ClangMsg where
         | RootHeader.isInRootHeader diagnosticLocation -> textToCtxDoc $
             case getFileNotFound diagnosticSpelling of
               Just header -> "unable to resolve #include <" <> header <> ">"
-              Nothing     ->
-                Text.stripStart $ Text.dropWhile (/= ' ') diagnosticFormatted
+              Nothing     -> case getFileNotFoundQ diagnosticSpelling of
+                Just header ->
+                  "unable to resolve #include <" <> header
+                    <> "> (must specify header relative to directory in C include search path)"
+                Nothing     ->
+                  Text.stripStart $ Text.dropWhile (/= ' ') diagnosticFormatted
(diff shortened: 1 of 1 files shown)


TravisCardwell:
I mangled the trace message as follows:

```
$ cabal run hs-bindgen-cli -- preprocess include/example.h
[Error  ] [Libclang ] unable to resolve #include <include/example.h> (must specify header relative to directory in C include search path)
An error happened (see above)
```

I considered showing the usage hint in both cases, but I think it is best to only display it when it is relevant.

PR #980, opened 2025-08-07
Update `--select-by-decl-name` usage info

TravisCardwell:
The CLI usage info for both `--select-by-decl-name` and `--select-except-by-decl-name` is updated to specify that *C* names are matched, not Haskell names.


dschrempf:
Probably not matching here, but I was wondering why we use the word `except` instead of `exclude`? Is there a particular reason for this choice of nomenclature? I ask because I was confused about this option at some point.

edsko:
@dschrempf there were two reasons for this choice

* having all selection options start with `--select` we felt was helpful, as it made it more obvious that they belonged together
* `--select-exclude-..` is quite awkward. `--select-except-..` suggests we're seeing whatever we were selecting before, _except_ some things; we felt that this matches the (`or ... && and (map not ..)`) semantics very closely.

Issue #937, opened 2025-07-24
Include C parameter names in haddocks


edsko:
In #933 we discuss why this would be useful. That ticket is just about extending the AST; this ticket is about including the names in the Haskell documentation.

Issue #935, opened 2025-07-24
Associate translated Doxygen comments with arguments


edsko:
In #924 Doxygen parameter declarations are just listed in the function documentation. We should change that and associate them with the function parameter instead. This will depend on having the names of those parameters available, so #933 must be done first. 

It _might_ make sense to do #937 before this one, though they can really be done in either order.

Issue #933, opened 2025-07-24
Parse C function parameter names


edsko:
C function signatures typically, but not necessarily, include parameter names, [such as](https://clang.llvm.org/doxygen/group__CINDEX.html#ga86d822034407d60d9e1f36e07cbc0f67)

```c
CINDEX_LINKAGE CXSourceLocation clang_getLocation	(	CXTranslationUnit 	tu,
CXFile 	file,
unsigned 	line,
unsigned 	column 
)	
```

These argument names can be helpful; sometimes they are the only form of documentation that exists. 

We currently don't parse these argument names; we should change that. Note that when we do, the argument name must be **optional** in the AST, because they are optional in signatures:

```c
void f(int, char);
```

is a perfectly fine signature in C.

Issue #113, opened 2024-08-23
Take advantage of Doxygen annotations to improve configuration-free high level API generation


edsko:
This is not about documentation per se (#26), but rather to take advantage of things such as [`param[in]` versus `param[out]`](https://www.doxygen.nl/manual/commands.html#cmdparam) to guide high-level API generation.

bolt12:
See: https://github.com/well-typed/hs-bindgen/blob/a994f0abbc05bd1631f5a55e3ede9577d6ea25e3/hs-bindgen/src-internal/HsBindgen/Hs/Haddock/Translation.hs#L131

PR #985, opened 2025-08-11
Parse C function parameter names

bolt12:
Tackles #933, #937, #935

Note: Missing TH mode which I am working on

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST/Type.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST/Type.hs
@@ -1,6 +1,8 @@
 module HsBindgen.Backend.Hs.AST.Type (
   HsPrimType (..),
   HsType (..),
+  ResultType(..),
+  extractResultType,
   hsPrimIntTy,
   hsPrimFloatTy
 ) where
@@ -66,6 +68,24 @@ data HsType =
   | HsSizedByteArray Natural Natural
   | HsBlock HsType
   deriving stock (Generic, Show, Eq)
+
+-- | When translating a 'C.Type' there are C types which we
+-- cannot pass directly using C FFI. We need to distinguish these.
+--
+-- Result types can be heap types, which are types we can't return by value
+-- due to Haskell FFI limitation. Or they can be normal types supported by
+-- Haskell FFI.
+--
+data ResultType a =
+    NormalResultType a
+    -- ^ Normal result type.
+  | HeapResultType a
+    -- ^ Heap type that is not supported by Haskell FFI
+  deriving stock (Generic, Show, Functor)
+
+extractResultType :: ResultType a -> a
+extractResultType (NormalResultType t) = t
+extractResultType (HeapResultType t)   = t
 
 hsPrimIntTy :: C.Type.IntegralType -> HsPrimType
 hsPrimIntTy = \case
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST/Type.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST/Type.hs
@@ -74,7 +74,8 @@ data HsType =
 --
 -- Result types can be heap types, which are types we can't return by value
 -- due to Haskell FFI limitation. Or they can be normal types supported by
--- Haskell FFI.
+-- Haskell FFI. This is also true for function arguments as well, result types
+-- are a special case where unsupported result types become arguments.
 --
 data ResultType a =
     NormalResultType a
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Haddock/Translation.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Haddock/Translation.hs
@@ -1,18 +1,19 @@
 module HsBindgen.Backend.Hs.Haddock.Translation where
 
 import Data.Char (isDigit)
-import Data.List (partition)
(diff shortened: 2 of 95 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST/Type.hs
```diff
@@ -67,6 +69,24 @@ data HsType =
   | HsBlock HsType
   deriving stock (Generic, Show)
 
+-- | When translating a 'C.Type' there are C types which we
+-- cannot pass directly using C FFI. We need to distinguish these.
+--
+-- Result types can be heap types, which are types we can't return by value
+-- due to Haskell FFI limitation. Or they can be normal types supported by
+-- Haskell FFI.
+--
```
bolt12, line 79: Mention that this is true for arguments as well

Results are a special case - unsupported results become arguments

hs-bindgen/src-internal/HsBindgen/Backend/SHs/AST.hs
```diff
@@ -328,15 +329,23 @@ data Newtype = Newtype {
...
+    , foreignImportComment    :: Maybe Comment
     }
   deriving stock (Show)
 
+data FunctionParameter = FunctionParameter
+  { functionParameterName    :: Maybe (HsName NsVar)
+  , functionParameterType    :: ClosedType
+  , functionParameterComment :: Maybe Comment
+  }
+  deriving stock (Show)
```
bolt12, line 359: Add a comment saying that we might want to reconsider the decision of having `functionParameterType` as well as `foreignImportResultType` be a `ClosedType` if we want to ever generate polymorphic type signatures

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Haddock/Translation.hs
```diff
@@ -1,18 +1,39 @@
...
+-- - If a parameter name matches one in the provided list, attach the comment to it
+-- - If no match is found, include the 'ParamCommand' in the resulting comment
+--
+-- Returns the processed comment and the updated parameters list
+--
+generateHaddocksWithParams ::
+     Maybe C.Comment
+  -> [Hs.FunctionParameter]
+  -> (Maybe Hs.Comment, [Hs.FunctionParameter])
+generateHaddocksWithParams Nothing params             = (Nothing, params)
```
bolt12: align '='

bolt12: Associate the argument name with the parameters if the parameters are named.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Haddock/Translation.hs
```diff
@@ -25,24 +46,83 @@ generateHaddocks C.Comment{..} =
...
+      , updatedParams
+      )
+  where
+    -- Matches only 'C.ParamCommand' blocks that match a function parameter.
+    --
+    isMatchingParamCommand :: C.CommentBlockContent -> Bool
+    isMatchingParamCommand C.ParamCommand{..} =
+      any ( (\paramName -> not (Text.null paramCommandName)
+                        && paramName == Just paramCommandName
+            )
```
bolt12: maybe we don't need the null check

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Haddock/Translation.hs
```diff
@@ -25,24 +46,83 @@ generateHaddocks C.Comment{..} =
...
+                               else Just (Text.strip commentCName)
+        , commentChildren = finalChildren
+        }
+      , updatedParams
+      )
+  where
+    -- Matches only 'C.ParamCommand' blocks that match a function parameter.
+    --
+    isMatchingParamCommand :: C.CommentBlockContent -> Bool
+    isMatchingParamCommand C.ParamCommand{..} =
```
bolt12: This would simplify things if we returned some kind of evidence. See `partitionEithers`

bolt12: Also we can return the direction information of the particular param command in order to (in the future) be able to tackle #113

edsko (changes requested):
Nice work. Some higher level changes we discussed:

* Associate C name with function parameters even if they have no Doxygen comments
* Get function names in `functionDecl` rather than `function`, so that we don't need to pass the `CXCursor` to `fromCXType`

Once you feel it's ready, ping me, and I'll go through the fixtures; as we discussed, it's inevitable that they will change, so I'll just have to make sure that there are no unexpected changes.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Haddock/Translation.hs
```diff
@@ -80,49 +90,69 @@ generateHaddocksWithParams (Just C.Comment{..}) params =
...
+
+-- | If the function parameter doesn't have any comments then add a simple
+-- comment with just its name (if exists).
+--
+addFunctionParameterComment :: Hs.FunctionParameter -> Hs.FunctionParameter
+addFunctionParameterComment fp@Hs.FunctionParameter {..} =
+  case functionParameterName of
+    Nothing -> fp
+    Just hsName
+      | Text.null (getHsName hsName) -> fp
```
bolt12: error here

PR #986, opened 2025-08-11
Resolve headers in bulk

TravisCardwell:
Headers listed in binding specification files are now resolved using a single translation unit.

A benefit of this approach is that Clang parsing of common headers (included by more than one of the resolved headers) is not duplicated.  Note that the previous implementation was able to `break` as soon as the header was resolved, which allowed us to avoid traversing the full Clang AST but did not prevent Clang from creating that AST.  In this implementation, we must traverse the full Clang AST.

We are concerned that there might be a case where some fatal error could be raised that prevents further header resolution.  We tried various ideas but were unable to get this to happen, as all errors are diagnostics (which we ignore when resolving headers).  The implementation is changed to detect this and emit a `ResolveHeaderNotResolved` trace message.

The `hs-bindgen-cli resolve` command is updated to suppress `ResolveHeader` tracing since it outputs results.  The command exit with exit status `1` if any header does not resolve.

### DIFF
--- a/hs-bindgen/app/hs-bindgen-cli.hs
+++ b/hs-bindgen/app/hs-bindgen-cli.hs
@@ -103,10 +103,10 @@ execResolve GlobalOpts{..} ResolveOpts{..} = do
         case Map.lookup header includes of
           Just path -> (False <$) . putStrLn $
             "#include <" ++ getHashIncludeArg header
-              ++ "> resolves to " ++ show path
+              ++ "> resolved to " ++ show path
           Nothing   -> (True  <$) . putStrLn $
             "#include <" ++ getHashIncludeArg header
-              ++ "> does not resolve (header not found)"
+              ++ "> could not be resolved (header not found)"
     case mErr of
       Just False -> exitSuccess
       Just True  -> exitFailure
@@ -119,7 +119,7 @@ execResolve GlobalOpts{..} ResolveOpts{..} = do
 
     customLogLevel :: CustomLogLevel Level TraceMsg
     customLogLevel = CustomLogLevel $ \case
-      TraceResolveHeader ResolveHeaderSuccess{}  -> Just Debug
+      TraceResolveHeader ResolveHeaderFound{}    -> Just Debug
       TraceResolveHeader ResolveHeaderNotFound{} -> Just Debug
       _otherTrace -> Nothing
 
--- a/hs-bindgen/app/hs-bindgen-cli.hs
+++ b/hs-bindgen/app/hs-bindgen-cli.hs
@@ -4,10 +4,12 @@ module Main (main) where
 
 import Control.Exception (Exception (..), SomeException (..), fromException,
                           handle, throwIO)
-import Control.Monad (forM_, void, (<=<))
+import Control.Monad (forM, void, (<=<))
 import Data.ByteString qualified as BS
+import Data.Map.Strict qualified as Map
+import Data.Set qualified as Set
 import Optics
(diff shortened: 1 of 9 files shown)


TravisCardwell:
(Rebased)

TravisCardwell:
The frontend creates the include graph before parsing, and the headers that we translate are resolved then.  They do not need to be resolved separately using `resolveHeaders`.

dschrempf:
I was asking because we are thinking to create an artefact for "resolve headers". But since we are _not_ doing things lazily yet, this would mean the header is actually parsed by the frontend (even if the only thing we are interested in is the file the header resolves to). @edsko

edsko:
> I was asking because we are thinking to create an artefact for "resolve headers". But since we are _not_ doing things lazily yet, this would mean the header is actually parsed by the frontend (even if the only thing we are interested in is the file the header resolves to). @edsko

@dschrempf , Travis and I discussed this just now. This was a misunderstanding on my part. The `resolveHeaders` function is called in the CLI on an arbitrary set of headers (user specified). I _thought_ that we would call this on the main headers during the normal execution of `hs-bindgen`, but it turns out that that is not the case. We call this only on the headers that appear in binding specifications. We could _maybe_ shoehorn the `resovleHeaders` CLI function into that, but it feels a bit ugly. Travis suggests that we instead move the "resolve headesr' CLI to `hs-bindgen-dev`, and use internal API instead. It's not _really_ an artefact of the normal hs-bindgen processing pipeline.

hs-bindgen/app/hs-bindgen-cli.hs
```diff
@@ -90,15 +92,36 @@ execBindingSpec GlobalOpts{..} BindingSpecCmdStdlib{..} = do
...
+          clangArgs
+          (Set.fromList hashIncludeArgs)
+      fmap or . forM hashIncludeArgs $ \header ->
+        case Map.lookup header includes of
+          Just path -> (False <$) . putStrLn $
+            "#include <" ++ getHashIncludeArg header
+              ++ "> resolves to " ++ show path
+          Nothing   -> (True  <$) . putStrLn $
+            "#include <" ++ getHashIncludeArg header
+              ++ "> header not found"
```
edsko:
```suggestion
              ++ "> could not be resolved (header not found)"
```

hs-bindgen/src-internal/HsBindgen/Resolve.hs
```diff
@@ -27,15 +32,19 @@ import Text.SimplePrettyPrint (hang, hsep, string)
 
 data ResolveHeaderMsg =
     ResolveHeaderClang ClangMsg
-  | ResolveHeaderSuccess  HashIncludeArg SourcePath
+  | ResolveHeaderFound HashIncludeArg SourcePath
   | ResolveHeaderNotFound HashIncludeArg
+
+    -- | Header not resolved, /perhaps/ due to an error while parsing previous
+    -- headers
+  | ResolveHeaderNotResolved HashIncludeArg
```
edsko:
```suggestion
  | ResolveHeaderNotResolved HashIncludeArg
```
```suggestion
    -- | @clang@ not attempt to resolve the header at all, /perhaps/ due to an error while parsing previous
    -- headers
    --
    -- NOTE: We have so far been unable to construct examples that cause this to happen, but we detect when this happens.
  | ResolveHeaderNotAttempted HashIncludeArg
```

edsko (approved):

Issue #502, opened 2025-03-13
Is there a way to get unit-id for preprocessor


phadej:
... so it can generate unit-local (i.e. somewhat unique) names?

This is relevant for `userland-capi` like usage in preprocessor setup.

edsko:
@phadej Am I right in thinking that https://github.com/well-typed/hs-bindgen/pull/575 means we can close this ticket?

phadej:
@edsko no, #575 is no help for preprocessor. In preprocessor we need to generate code with unique names already.

edsko:
Marking this as high priority because without this we might pollute the global namespace.

dschrempf:
This is still somewhat of an issue, but less so since we merged https://github.com/well-typed/hs-bindgen/pull/989, which introduces `UniqueId` and corresponding traces.

Should we close this issue?

edsko:
Yes, we can close. Thanks for reminding me!

PR #989, opened 2025-08-12
Rethink implementation of unique module identifiers

### LINKED ISSUE #987: Decide what to do with `ModuleUnique`
Description provided in part by @TravisCardwell, thanks.

- `HsModuleOpts` is the user-provided options for pretty-printing the module, currently just the module name.
- `ModuleUnique` is an identifier that is unique per module, needed to generate unique names in the generated C code. `ModuleUnique` is used to disambiguate function declarations (see `wrapperName` in `Translation.hs`).

The module name in `HsModuleOpts `is Haskell syntax (example: `Foo.Bar`) while `ModuleUnique` is C syntax (example: `Foo_Bar`).

When using the client, `ModuleUnique` is generated from the module name in `HsModuleOpts`. The `HsModuleOpts` is also used when translating the reified C declarations to an `HsModule`.

When using the TH interface, `ModuleUnique` is generated from the "location" obtained from Template Haskell.

At the moment, the library interface requires specification of both, `ModuleUnique` and `HsModuleOpts`, which is suboptimal.

For example, we could always generate `ModuleUnique` from `HsModuleOpts`.

Related: #502.

dschrempf:
The unique identifier is ipmortant to discriminate between C symbols that are using a the global namespace.

We seek to avoid random identifiers because they lead to irreproducible builds (and large source code changes, if the code is checked in).

However, we really want to encourage users to provde a _unique_ identifier.

We came up with a good solution, see comment https://github.com/well-typed/hs-bindgen/pull/989#issuecomment-3182643561.

Closes #987.

Related: https://github.com/well-typed/hs-bindgen/issues/992.


phadej:
This is wrong. The wrapper "unique prefix" (i.e. module unique) **needs** to be unique.

If the same header is processed in different modules, their wrappers **must** be different, or you'll get linker errors. In other words, the unique wrapper **cannot only depend on generated content**.

(Even module name is not good enough, as there can be the same module name in different packages).

dschrempf:
Exactly @phadej, Thanks for chiming in. That's what we decided in a meeting. We now assemble the unique module identifier using a user-provided module name, and a hash obtained from the declarations.

This is already much better than before, where `ModuleUnique` had to be provided at top level.I think the chances for clashes are much lower now.

However, if need be, we can revert to using an additional ID (in addition to the `HsModuleName` and the hash, I guess).

phadej:
> This is already much better than before

It's strictly worse for TH generated code.

EDIT: You can generate better identifier in the preprocessor code / provide better API in the preprocessor part of the lib, but please don't make TH worse, because currently it works "perfectly".

TravisCardwell:
The hash is computed over `show` of the list of C declarations.

```haskell
hashStr = showHex' $ hash $ XXH3 $ pack $ show decls
```

This means that the hash is sensitive to any changes we make to the AST, prescriptive binding specification changes for any (selected) declaration at all (not just those in the module), changes of parse/selection predicates, etc.

A drawback of this is that it results in unhelpful noise in diffs.  Users who keep generated source in revision control in order to audit all changes will have to ignore this noise, which can obfuscate meaningful differences.  Users who make use of revision control to iteratively refine configuration would also have to ignore the always changing hashes.

Speaking of which, this is not very compatible with our golden tests.  This PR includes many changes to fixtures, and we would continue to have such noise every time something in the AST changes.

TravisCardwell:
> (Even module name is not good enough, as there can be the same module name in different packages).

I was considering something like `mangle (packageName <> "_" <> moduleName <> "_")`, but perhaps even that is technically insufficient since there can be the same module name in different components of a package.  (It would also require specifying the package name, which is annoying.)

Perhaps it would be acceptable to provide a simple default (such as `mangle moduleName <> "_"`) but allow users to override it in cases where it is necessary?

----

BTW, this PR mangles using `mapHead toLower . filter isLetter`, but I do not think that `filter isLetter` is appropriate.  For example, a package may implement varying standards/versions where some modules are named with a common prefix (such as `RFC` or `V`) followed by numbers, and this would not provide a unique prefix when the numbers are filtered out.  We should keep the numbers and either adjust `mapHead toLower` appropriately or alternatively add a constant prefix (such as `"fn_" <> mangle moduleName <> "_"`).  Also it is generally "safer" to translate invalid characters (to `_` for example) instead of filter them out.

dschrempf:
I think it is really good we discuss this, and thanks for your many convincing arguments.

After this discussion, my view is that
- users should be able to provide a _truly unique_ module ID (like we have on `main`),
- BUT: by default, we must make sure that the ID is _unique_ (like we have on this PR), because, in my opinion, careless users will not provide unique strings.

So what about we use an ADT:

```haskell
data UniqueModelIdSetting =
    -- | Note to users: Make sure this string is UNIQUE.
    FixedUniqueModelId String
    -- | Compute hash with provided salt from C declarations,
    -- and prefix hash with sanitized module name.
  | HashWithSalt Int
```

- We use `FixedUniqueModelId` in tests to avoid fixture changes.
- We use `FixedUniqueModelId` in TH, like Oleg suggested.
- The _default_ is `HashWithSalt 0`.

I will also change the _sanitize_ function of the module name like Travis suggested.

Another option I came up with:

``` haskell
data UniqueModelIdConfig = UniqueModelIdConfig {
    uniquePrefix   :: Maybe Int -- 'Nothing': Use mangled Haskell module name.
    uniqueHashSalt :: Maybe Int -- 'Nothing': Do not use hash.
  }
```

with default

``` haskell
defaultUniqueModelIdConfig = {
    uniquePrefix   = Nothing
    uniqueHashSalt = Just 0
  }
```

Thoughts?

TravisCardwell:
I think that it is ultimately a design decision that must take account various trade-offs.

* `mangle moduleName <> "_"` or `"fn_" <> mangle moduleName <> "_"`
    * Generated function names are stable.  It causes differences in generated code only when the function is moved to a different module or the C function is renamed, both of which cause `diff` changes anyway.
    * Generated function names are unique for the vast majority of cases.  Problems only arise in edge cases, such as when different packages/components use functions of the same name in modules of the same name.  In such cases, users can specify a custom prefix.

* `UniqueModuleIdSetting`
    * With `HashWithSalt`, generated function names are unstable.  `hs-bindgen` implementation details as well as any change of usage configuration causes many differences in generated code.  This issue is not an edge case, as it affects anybody who audits generated code.  Generated function names are unique, with an extremely low probability of (hash) collision.
    * Users who need stability may use `FixedUniqueModuleId`.  With the proposed design, they must manually specify a unique prefix for each module.  An option that generates prefixes based on the module name could be provided, however.

My opinion is that the benefit of stability outweighs the increased uniqueness, given that users can specify a custom prefix when needed.

BTW: `s/Model/Module/g`

dschrempf:
> Generated function names are unique for the vast majority of cases. Problems only arise in edge cases, such as when different packages/components use functions of the same name in modules of the same name. In such cases, users can specify a custom prefix.

That is true only if you are managing the package. If two other users provide packages with bindings generated using @hs-bindgen@ with clashing identifiers, I believe you can not do anything about it.

What about we write good Haddocks stating that using the hash will lead to large diffs?

Suggestion:

```haskell
-- | C uses a global namespace. We must ensure that identifier generated by
-- @hs-bindgen@ are unique.
data UniqueModuleIdSetting =
      -- | Use (mangled) Haskell module name.
      HaskellModuleNameOnly
      -- | Make sure the ID is /unique/ and a valid C identifier.
    | UniqueModuleId String
      -- | Compute hash with salt from C declarations. Prefix hash with
      -- (mangled) Haskell module name.
      --
      -- NOTE: Hash changes may lead to large source code changes (i.e.,
      -- "diff"s).
    | HashWithSalt Int
  deriving stock (Show)

instance Default UniqueModuleIdSetting where
  def = HaskellModuleNameOnly
```

or

```haskell
instance Default UniqueModuleIdSetting where
  def = HashWithSalt 0
```

edsko:
Conclusions of discussing this with @TravisCardwell and @dschrempf :

1. When wrapping a C function called `fn`, we will generate a function called

```c
hs_bindgen_<unique_id>_hash<module,fn>
```

The reason for the `hash<module,fn>` rather than just `<module,fn>` is that only the first 64 (?) characters are used to distinguish C names by the linker, and Haskell module names can be quite long.

2. In TH, that `<unique_id>` can be constructed from `loc_package <$> TH.location`, as we do now. In this case we must query TH for the module name.

3. In PP, the module name is passed on the command line; in this case, by default the unique ID will be the empty string. In many cases, this will be fine: the combination of the Haskell module name and the C function name should already be unique "in most cases".

4. Nonetheless, we would encourage as good practice to pass a `--unique-id` on the CLI. As example, we could give something like `com.well-typed.foo`; bit Java-esque, but the point here is that this is something that should be "globally" unique, and things like Hackage don't not have a good answer to that (everything is one flat namespace); domain names however _are_ unique. Of course, users could use anything here (although we should probably insist on a maximum length).

5. Especially when `hs-bindgen` is incorporated into a build process (nix derivation, custom cabal setup script, ..), users _really_ should use `--unique-id`. To encourage them to do so, we should issue a warning when no unique ID is provided, and we should offer `-Werror` or `-Werror=unique-id`.

dschrempf:
I think we should also state that in TH mode, we default to using the `UniqueId`, if it is provided. Only if the `UniqueId` is not provided, we use the `location` to generate a unique ID.

phadej:
> "globally" unique

THe [package name](https://hackage-content.haskell.org/package/template-haskell-2.23.0.0/docs/Language-Haskell-TH-Syntax.html#t:Loc) given in TH is actually a unit id (blame GHC for inconsistent naming), which itself is *unique* in the build plan. It's different for different components of the same package etc.

What I have implemented originally resembles CApiFFI, and CApiFFI approach works. See [mkWrapperName](https://github.com/ghc/ghc/blob/master/compiler/GHC/Tc/Utils/Env.hs#L1169) in GHC, `hs-bindgen` doesn't need to number imports as it doesn't import the same thing twice (like you can manually). Needing to specify `--unique-id` on command line don't feel good at all.

edsko:
@phadej Yes, we will use the package name as the unique ID in TH mode, but we don't have access to it in the PP. Do you have a suggestion for what we could do there instead of just asking the user for a unique ID?

phadej:
@edsko no i have no suggestion, that's why I opened https://github.com/well-typed/hs-bindgen/issues/502

dschrempf:
TLDR: The C standard says
- at least 63 significant characters for internal ids,
- at least 31 significant characters for external ids;
- but only for historical reasons, and all new implementations support at least 255 significant characters for external ids.

*** For reference:

I was digging a bit deeper into the minimum numbers of "significant characters".
The reason was that there are different restrictions for "internal" and
"external" identifiers.

The following are quotes from the C standard n3220:

> 5.2.5.2 Translation limits:
>
> — 63 significant initial characters in an internal identifier or a macro name
> (each universal character name or extended source character is considered a
> single character)
>
> — 31 significant initial characters in an external identifier (each universal
> character name specifying a short identifier of 00FFFF or less is considered 6
> characters, each universal character name specifying a short identifier of
> 010000 or more is considered 10 characters, and each extended source character
> is considered the same number of characters as the corresponding universal
> character name, if any)

Further,

> Implementation limits:
>
> As discussed in 5.2.5.2, an implementation may limit the number of significant
> initial characters in an identifier; the limit for an external name (an
> identifier that has external linkage) may be more restrictive than that for an
> internal name (a macro name or an identifier that does not have external
> linkage). The number of significant characters in an identifier is
> implementation-defined.

And then,

> 6.11.3 External names:
>
> Restriction of the significance of an external name to fewer than 255
> characters (considering each universal character name or extended source
> character as a single character) is an obsolescent feature that is a
> concession to existing implementations.

Pages:

> length
>   external name, 22, 55, 190
>   function name, 22, 55, 190
>   identifier, 54
>   internal name, 22, 55

PR #991, opened 2025-08-12
Add a TODO and accompanying test for macro "newtypes"

jorisdral:
NOTE: this builds on top of #969, which should be merged into `main` first

These macro "newtypes" redefine an existing identifier as a macro, but the macro simply desugars to the pre-exesting identifier. Such cases exis in the wild, for example in `stdio.h`:

```c
typedef int FILE;

extern FILE *const stdin;
extern FILE *const stdout;
extern FILE *const stderr;

#define stdin  (stdin)
#define stdout (stdout)
#define stderr (stderr)
```

The current behaviour of `hs-bindgen` is to not generate a binding and emit a "redeclaration" trace message. Should `hs-bindgen` instead allow this corner-case of a redeclared identifier, and generate a binding for the non-macro declaration?

### DIFF
--- a/.github/workflows/check-stylish-haskell.yml
+++ b/.github/workflows/check-stylish-haskell.yml
@@ -50,7 +50,7 @@ jobs:
 
     - name: Save cache
       uses: actions/cache/save@v4
-      if: ${{ steps.cache-cabal-restore.outputs.cache-hit == 'false' }}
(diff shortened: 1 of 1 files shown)


jorisdral:
I'm sneaking in a fix for caching behaviour related to the `stylish-haskell` GHA job. I changed a conditional that broke the caching behaviour in #1153

PR #998, opened 2025-08-18
Command line client options -Werror, Werror=<...>

### LINKED ISSUE #992: Provide `-Werror` and `-Werror=<..>`
It would be rather useful for users to be able to say that their codebase must be "-Wall clean", especially when using CI.

This was initially motivated by [`-Wunique-id`](https://github.com/well-typed/hs-bindgen/pull/989#issuecomment-3182643561), but would also be useful for other warnings.

Once we have the `-Werror=<foo>` syntax, we should probably also offer `-Wno-<foo>`, to disable specific warnings.

Marking this as part of release-0.1 primarily because if users start using `hs-bindgen`, we should encourage them to use unique IDs from the very start.

dschrempf:
Closes #992.

I refrained from implementing the single dash options because it departs from standard syntax, and is a bit more involved to implement using `optparse-applicative`.

Right now, the options have the following names:

``` text
Options with user-provided trace IDs:
--log-as-info TRACE_ID
--log-as-warning TRACE_ID
--log-as-error TRACE_ID

Options affecting specific traces:
--log-as-info-uchar-resolution-trace
--log-as-warning-macro-traces
--log-as-error-warnings
```

The tracer infrastructure has changed quite a bit. Notably
- custom log level functions can now amend the current log level (and not only the default one)
- multiple trace-related classes have been combined into `class PrettyForTrace a => IsTrace l a | a -> l`, where `l` is the log level (right now `Level`, or `SafeLevel`).

### Issues/notes
- [x] The uniquenes of `TraceId`s is not ensured. Actually, multiple traces can have the same trace ID. In my opinion it makes sense to collect a set of traces into a single category (e.g., traces related to header resolution). However, when introducing new trace IDs, the implementor needs to check manually that the new ID does not exist yet. We could rename `TraceId` to `TraceCategory` or `TraceGroup`.
- [ ] `TraceId` is a simple newtype around `String`. We also may want to think about error codes (then we could distinguish a bit more strictly between this is the trace with error code `XYZ` which belongs into category "header-resolution".

### DIFF
--- a/hs-bindgen/app/HsBindgen/App/Common.hs
+++ b/hs-bindgen/app/HsBindgen/App/Common.hs
@@ -5,8 +5,6 @@ module HsBindgen.App.Common (
     -- * Global options
     GlobalOpts(..)
   , parseGlobalOpts
-    -- * Macro warnings
-  , MacroLogLevel(..)
     -- * HsBindgen configuration
   , parseBindgenConfig
     -- * Clang-related options
@@ -26,6 +24,7 @@ import Data.Bifunctor (Bifunctor (bimap))
 import Data.Char qualified as Char
 import Data.Either (partitionEithers)
 import Data.List qualified as List
+import Data.Maybe (catMaybes)
 import Data.Text (Text)
 import Data.Text qualified as Text
 import Optics (set)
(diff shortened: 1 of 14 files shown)


edsko:
I agree that diverging from unix conventions is not the best idea. What about `--disable-warning=foo,bar,baz` (and a similar `--enable-warning=..`)?

dschrempf:
In a meeting, we agreed on providing the following command line options:

```
--log-as-info foo
--log-as-warning foo
--log-as-error foo
--log-as-error-warnings
```

edsko (approved):

PR #1000, opened 2025-08-19
Differentiate between prefix and suffix `clang` arguments

### LINKED ISSUE #909: Improve handling of `clang` arguments

dschrempf:
Closes #909.

### DIFF
--- a/c-expr/test/CallClang.hs
+++ b/c-expr/test/CallClang.hs
@@ -307,8 +307,8 @@ clangWithTranslationUnit userClangArgs srcContents k =
 
     args :: Clang.ClangArgs
     args = userClangArgs
-      { Clang.clangOtherArgs =
-          Clang.clangOtherArgs userClangArgs
+      { Clang.clangArgsBefore =
+          Clang.clangArgsBefore userClangArgs
             ++
             [ "-Werror=pointer-integer-compare"
             , "-Werror=compare-distinct-pointer-types"
--- a/clang/src/Clang/Args.hs
+++ b/clang/src/Clang/Args.hs
@@ -75,10 +75,27 @@ data ClangArgs = ClangArgs {
       -- @Block.h@ header.
     , clangEnableBlocks :: Bool
 
-      -- | Other arguments
+      -- | Prepend these arguments when calling @clang@. See also
+      -- 'clangArgsInner'.
+    , clangArgsBefore :: [String]
+
+      -- | Arguments passed to @clang@.
+      --
+      -- The complete list of arguments passed to @clang@ is
+      --
+      -- @
+      --   'clangArgsBefore' ++ internalArgs ++ 'clangArgsInner' ++ 'clangArgsAfter'
+      -- @
+      --
+      -- where we assemble @internalArgs@ using the specialized 'ClangArgs'
+      -- records above.
       --
       -- See https://clang.llvm.org/docs/ClangCommandLineReference.html
-    , clangOtherArgs :: [String]
+    , clangArgsInner :: [String]
+
+      -- | Append these arguments when calling @clang@. See also
+      -- 'clangArgsInner'.
+    , clangArgsAfter :: [String]
     }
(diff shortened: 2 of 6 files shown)


dschrempf:
This is issue number 1000 :rocket: !

dschrempf:
@TravisCardwell since you are an expert in `clang` header resolution and command line arguments :-):

We introduce two new lists of arguments passed to `clang`:
- `clangPrefixArgs`
- `clangSuffixArgs`

As their name suggests, these lists of arguments are prepended, and appended to the other arguments (see `fromClangArgs`, and `getClangInternalArgs`).

The contents of `BINDGEN_EXTRA_CLANG_ARGS` are appended to `clangSuffixArgs`.

The existing CLI option `--clang-option` affects `clangPrefixArgs`.

A new CLI option `--clang-option-after` affects `clangSuffixArgs`.

We ask you in particular because:
- Include directories are parsed from left to right; so, `BINDGEN_EXTRA_CLANG_ARGS` defines fallback include directories superseded by user-provided ones (e.g., using `-I`). Is this correct? We desire this way of precedence.
- Are there other command line arguments with opposite precedence (we came up with `-Wno-foo -Wall`; here `-Wall` supersedes `-Wno-foo`)? If so, is this bad since we do not desire this way of precedence? (That's also why we provide the new option to cover all use cases...).

TravisCardwell:
> Include directories are parsed from left to right; so, `BINDGEN_EXTRA_CLANG_ARGS` defines fallback include directories superseded by user-provided ones (e.g., using `-I`). Is this correct? We desire this way of precedence.

:+1: That is consistent with Clang and GCC: options specified via the command-line take precedence over options specified via environment variables.

> Are there other command line arguments with opposite precedence (we came up with `-Wno-foo -Wall`; here `-Wall` supersedes `-Wno-foo`)? If so, is this bad since we do not desire this way of precedence? (That's also why we provide the new option to cover all use cases...).

It is indeed annoying that composition of such options is not commutative.

The `BINDGEN_EXTRA_CLANG_ARGS` come after the "suffix" arguments!

1. `--clang-option` options
2. "Internal" options (including `-I` options)
3. `--clang-option-after` options
4. `BINDGEN_EXTRA_CLANG_ARGS` options

If `BINDGEN_EXTRA_CLANG_ARGS` specifies `-Wall`, then there is no way to add `-Wno-foo` via the command line.

I guess we *could* do something like the following...

1. `--clang-option-before` options
2. "Internal" options (including `-I` options)
3. `--clang-option` options
4. `BINDGEN_EXTRA_CLANG_ARGS` options
5. `--clang-option-after` options

dschrempf:
Thanks! I agree with you @TravisCardwell.

@edsko, should we go ahead and implement a third option, as well as ammend the documentation and manual?

edsko:
Yes, let's.

PR #1011, opened 2025-08-22
Manual: Parse, select, slice

### LINKED ISSUE #785: Manual section(s) about selection and program slicing
Should probably be written after #784.

dschrempf:
Closes #785.

Link to rendered document: https://github.com/well-typed/hs-bindgen/blob/dom/785/manual-selection-slicing/manual/LowLevel/PredicatesAndSlicing.md


hs-bindgen/src-internal/HsBindgen/Frontend/Predicate.hs
```diff
@@ -88,6 +88,29 @@ instance Default ParsePredicate where
 
 -- | Predicates for the @Select@ pass select based on header file paths or the
 -- declarations themselves
+--
+-- NOTE: The 'ParsePredicate' and the 'SelectPredicate' both allow matching
```
dschrempf: Update or remove.

dschrempf: Done; referring to the manual.

manual/LowLevel/PredicatesAndSlicing.md
```diff
@@ -1,15 +1,211 @@
...
+   and exclude declarations not matching the parse predicate. That is, the parse
+   predicate dictates which declarations `hs-bindgen` reifies into
+   `hs-bindgen`-specific data structures.
+
+3. Given the set of declarations parsed and reified by `hs-bindgen`, [_select
+   predicates_](#select-predicates) dictate which declarations to generate
+   bindings for.
+
+4. Now we have the set of selected declarations to generate bindings for. If
+   [_program slicing_](#program-slicing) is disabled, `hs-bindgen` only generate
```
TravisCardwell: `s/only generate/only generates/`

manual/LowLevel/PredicatesAndSlicing.md
```diff
@@ -1,15 +1,211 @@
...
+
+```c
+enum FileOperationStatus read_file_chunk(FILE *file_ptr, void *buffer,
+                                         size_t bytes_to_read);
+```
+
+The translation of the `read_file_chunk` function declaration is useful only if
+the _transitive dependencies_ `FileOperationStatus`, `FILE`, and `size_t` are
+available as well. For a given set of selected declarations, _program slicing_
+determines and selects the corresponding transitive dependencies.
```
TravisCardwell: Note that both `FILE` and `size_t` have standard external bindings that are used by default.  These therefore do not need to be "made available" (parsed/reified), making the example potentially confusing.

manual/LowLevel/PredicatesAndSlicing.md
```diff
@@ -1,15 +1,211 @@
...
+We use the following command line
+
+```
+hs-bindgen-cli preprocess -I /path/to/simple main.h
+```
+
+By default, `hs-bindgen` the command-line client uses the following configuration:
+
+- Parse predicate: Parse all declarations in header files in
+  `/path/to/simple/**`.
```
TravisCardwell: This could be easily misinterpreted.  We parse/reify declarations in `/path/to/simple/main.h` as well as any header in `/path/to/simple/**` that is (transitively) included via that main header.

manual/LowLevel/PredicatesAndSlicing.md
```diff
@@ -1,15 +1,211 @@
...
+Example 2: Assume we have external binding specifications for `dependency.h`.
+Only the first scenario works out:
 
-## Slicing
+- Parse `dependency.h` and `main.h`, select `main.h`; the external binding
+  specifications for `dependency.h` can be applied.
 
-TODO
+- Parse `main.h`, select all; the external binding specifications for
+  declarations in `dependency.h` __can not__ be applied.
```
TravisCardwell: This is incorrect.  Named declarations do not need to be parsed/reified to be used as external bindings.

manual/LowLevel/PredicatesAndSlicing.md
```diff
@@ -1,15 +1,211 @@
-# Predicates and slicing
+# Parsing, selecting and program slicing
```
TravisCardwell: If we change the title, we need to also change it in `README.md` and change the filename for consistency.

manual/LowLevel/PredicatesAndSlicing.md
dschrempf: Thanks, I renamed those types to avoid confusion.

manual/LowLevel/PredicatesAndSlicing.md
dschrempf: Yes, thanks for pointing this out; I think that's the origin of confusion. The statement is correct in that the parse predicate is `/path/to/simple/**`. But that doesn't mean we actually parse all headers matching this predicate, but only the ones that happen to be transitively included by the main headers we have specified. I will clarify this in the text.

dschrempf: Actually, the statement was incorrect. It was

> - Parse predicate: Parse all declarations in header files in `/path/to/simple/**`.

But it should be

> - Parse predicate: Match all declarations in header files in
  `/path/to/simple/**`.
  
  I have improved this section.

manual/LowLevel/PredicatesAndSlicing.md
dschrempf: I removed this example, it was not very enlightening anyways.

dschrempf:
No problem! Thanks for the review!

PR #1012, opened 2025-08-25
Automatic builtin include path

TravisCardwell:
This PR adds automatic configuration of the builtin include directory.  An overview of usage is in the manual [here](https://github.com/well-typed/hs-bindgen/blob/tcard/builtin-include-path/manual/LowLevel/Includes.md#hs-bindgen).

I will refactor `ClangArgs`/`ClangArgsConfig` in a separate PR.

### DIFF
--- a/manual/LowLevel/Includes.md
+++ b/manual/LowLevel/Includes.md
@@ -179,20 +179,20 @@ does not have to be done manually.
     1. `${LLVM_PATH}/{{RESOURCE_DIR}}/include`
     2. `$(${LLVM_CONFIG} --prefix)/{{RESOURCE_DIR}}/include`
     3. `$(llvm-config --prefix)/{{RESOURCE_DIR}}/include`
-    4. `$(${LLVM_PATH}/bin/clang -print-file-name=include)`
-    5. `$($(${LLVM_CONFIG} --prefix)/bin/clang -print-file-name=include)`
-    6. `$($(llvm-config --prefix)/bin/clang -print-file-name=include)`
-    7. `$(clang -print-file-name=include)`
+    4. `$(${LLVM_PATH}/bin/clang -print-resource-dir)/include`
+    5. `$($(${LLVM_CONFIG} --prefix)/bin/clang -print-resource-dir)/include`
+    6. `$($(llvm-config --prefix)/bin/clang -print-resource-dir)/include`
+    7. `$(clang -print-resource-dir)/include`
 
 * Clang configuration (`clang`) determines the builtin include directory using
   the Clang compiler, which must match the version of `libclang` being used.
   `hs-bindgen` compares the version strings and only proceeds when there is a
   match.
 
-    1. `$(${LLVM_PATH}/bin/clang -print-file-name=include)`
-    2. `$($(${LLVM_CONFIG} --prefix)/bin/clang -print-file-name=include)`
-    3. `$($(llvm-config --prefix)/bin/clang -print-file-name=include)`
-    4. `$(clang -print-file-name=include)`
+    1. `$(${LLVM_PATH}/bin/clang -print-resource-dir)/include`
+    2. `$($(${LLVM_CONFIG} --prefix)/bin/clang -print-resource-dir)/include`
+    3. `$($(llvm-config --prefix)/bin/clang -print-resource-dir)/include`
+    4. `$(clang -print-resource-dir)/include`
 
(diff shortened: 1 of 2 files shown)


manual/LowLevel/Includes.md
```diff
@@ -163,12 +156,69 @@ Clang environment variables, described above.
...
+  determines the builtin include directory by getting the relative resource
+  directory from `libclang`.  The LLVM prefix can be set using the `LLVM_PATH`
+  environment variable or determined using `llvm-config` (optionally configured
+  using the `LLVM_CONFIG` environment variable).  Automatic configuration
+  tries to determine the builtin include directory using `clang` if all else
+  fails.
+
+    1. `${LLVM_PATH}/{{RESOURCE_DIR}}/include`
+    2. `$(${LLVM_CONFIG} --prefix)/{{RESOURCE_DIR}}/include`
+    3. `$(llvm-config --prefix)/{{RESOURCE_DIR}}/include`
```
dschrempf, line 181: What is RESOURCE_DIR? For me, `include` is directly in `llvm-config --prefix`:
```
(C) $ ls $(llvm-config --prefix)
 bin   include   lib   nix-support
```

TravisCardwell: There is more than one `include` directory.

`$(llvm-config --prefix)/include` is the include directory for LLVM/Clang itself.  It contains the headers for LLVM/Clang, which we need for binding with `libclang` for example.

`$(llvm-config --prefix)/{{RESOURCE_DIR}}/include` is the builtin include directory.  It contains the compiler-specific headers, which are needed to use the standard library.

```
$ /usr/lib/llvm15/bin/llvm-config --prefix
/usr/lib/llvm15
$ ls "$(/usr/lib/llvm15/bin/llvm-config --prefix)/include"
clang  clang-c  llvm  llvm-c
$ /usr/lib/llvm15/bin/clang -print-file-name=include
/usr/lib/llvm15/lib/clang/15.0.7/include
$ ls "$(/usr/lib/llvm15/bin/clang -print-file-name=include)/stddef.h"
/usr/lib/llvm15/lib/clang/15.0.7/include/stddef.h
```

dschrempf: Hm, interesting, thank you. I think we should describe or define RESOURCE_DIR in the manual.

BTW, I get GCC headers when issuing `clang --print-file-name=include`:
```
(C) $ clang --print-file-name=include
/nix/store/qdknxw57cwy1jkrhq7fzmiis73j42jv6-gcc-14.3.0/lib/gcc/x86_64-unknown-linux-gnu/14.3.0/include
```
Do you think this is a problem?

TravisCardwell: LOL, that is unexpected!

Could you please share the output of `clang -E - -v </dev/null`, so that I can see all of the include path directories?  Perhaps the LLVM/Clang builtin include directory is (also) used, and that option is misbehaving...

I did indeed run into an error once when (accidentally) using the wrong builtin include directory, but I do not remember if it was a GCC directory or just for the wrong LLVM/Clang version.  I remember that it was only an issue when using LLVM/Clang 20.1, but I do not remember which standard header triggered it.  (Configuring the correct Clang builtin include directory was the "obvious" solution, and I now regret that I did not keep better notes.)

I added some description about the resource directory.  Note that I implemented it this way to mirror how LLVM/Clang determines the directory (in C++).  It is an unfortunate leak of an implementation detail, but I think it is better to explain so that users can better understand what is going on.  The hack is very noticeable due to the overflow message.

edsko (approved):
Approving, though I'd prefer to use `-print-resource-dir` consistently rather than `-print-resource-dir` in one place and `-print-file-name=include` in another, if we can. And if we can't, it would be good to know why.

manual/LowLevel/Includes.md
TravisCardwell: @dschrempf Could you please also verify the output of `clang -print-resource-dir`?  Thanks!

TravisCardwell:
There is a PR in the merge queue already, and there are conflicts, so I will merge after a rebase.

manual/LowLevel/Includes.md
TravisCardwell: [Analysis](https://github.com/well-typed/hs-bindgen/issues/975#issuecomment-3251838612)

PR #1013, opened 2025-08-25
Delay parse-related traces

### LINKED ISSUE #905: Delay emission of warnings about unsupported declarations until they are truly selected
When using program slicing, we (1) parse all declarations, and then (2) select declarations according to selection predicate and the use-decl graph.

Often, we only select a small fraction of declaration in step (2). However, we report __all__ warnings about unsupported declarations encountered in step (1), even when they are skipped in step (2).

Ideally, we delay emission of declaration-specific warnings after we know if the declaration has been selected or not. Without program slicing, this is already during the Parse pass, but with program slicing enabled, we only have this information available after the Slice pass.

dschrempf:
Closes #905. 

Also differentiate between delayed traces, and traces related to missed selections. A missed selection is when the user intended to select a declaration, but we failed to reify the declaration.

I added some tests that we
- emit traces for skipped, but selected declarations (missed declarations),
- do not emit traces for skipped, unselected declarations.

Status: Most issues have been fixed. I think the remaining ones are more of aesthetic nature.

- [x] Test that we can detect missed declaration when selecting name kinds. For example, test that we can get traces when a struct fails, when only selecting PCRE "struct".

#### DeclMeta

The `DeclMeta` types don't line up well. We have `ParseDeclMeta`, `DeclMeta p` used from Sort to ResolveBindingSpec, and then `SelectDeclMeta` used by Select and later passes:

```haskell
data ParseDeclMeta = ParseDeclMeta {
      parseDeclNonParsed :: NonParsedDecls
    , parseDeclParseMsg  :: ParseMsgs Parse
    }

data DeclMeta p = DeclMeta {
      declIndex     :: DeclIndex
    , declUseDecl   :: UseDeclGraph
    , declDeclUse   :: DeclUseGraph
    , declNonParsed :: NonParsedDecls
    , declParseMsgs :: ParseMsgs p
    }

data SelectDeclMeta = SelectDeclMeta {
      selectDeclIndex     :: DeclIndex
    , selectDeclUseDecl   :: UseDeclGraph
    , selectDeclDeclUse   :: DeclUseGraph
    , selectDeclNonParsed :: NonParsedDecls
    }
```

#### ParseMsgKey

We store parse traces in a map with key
```haskell
data ParseMsgKey p = ParseMsgKey {
      parseMsgDeclLoc  :: SingleLoc
    , parseMsgDeclId   :: Id p
    , parseMsgDeclKind :: NameKind
    }
```
The type of `Id p` depends on the pass, and so we need convert the key in various places. We can not really avoid doing so, because we need to be able to match the select predicate on these keys. The reason is that we need to know if we wanted to select declarations missing from the reified list of declarations because reification has failed.

Actually, the `ParseMsgKey` as defined above is not enough. See below.

#### Trace messages

We now have parse messages in multiple locations of our `TraceMsg` data type:
```haskell
TraceFrontend (FrontenParse (Msg Parse))
TraceFrontend (FrontendSelect (SelectMissedParseMsg (Msg Parse)))
```
We might want to separate `Msg Parse` to reflect this change.


edsko:
@TravisCardwell Can you take a look at this as well please, and in particular, share your opinion on `ParseMsgKey` (for example, in relation to `QualDeclId`).

_Edit_: more general observation about unifiying with existing code: #1028.

edsko (approved):

TravisCardwell:
Previously, we decided to *not* remove the `NonParsedDecls` from the annotation even though they are no longer needed.  We should probably do the same for the parse messages, or go ahead and remove the `NonParsedDecls`.

Regarding `ParseMsgKey`, it does not look right.  I do not understand why a key includes a source location.  The whole purpose of `Id p` is to uniquely identify declarations.  `Id Parse` with `NameKind` is `QualPrelimDeclId`.  A `QualDeclId` can be converted to a `QualPrelimDeclId` to perform a lookup.

The `Ord` instances for `DynGraph`, `IncludeGraph`, and `TranslationUnit` set off alarm bells. :rotating_light:

dschrempf:
Yes, we need to rework the IDs. I am confused as to which IDs/keys we have, and why. We will rework them as part of #1028.

I designed the parse message key so that it contains all information required to match select predicates, and we need the source location to do so.

In retrospect, we could move the source location to the value part of the map like in `NonParsedDecls`, (I think the types of `NonParsedDecls` and `ParseMsgs` should be the same; that is, they are the same thing, the message for `NonParsedDecl` could be "skipped reification of declaration", and we would not emit it by default, only when the declaration is selected). We made all these connections while thinking about #1028.

With respect to the Ord instances, I think we can remove those.

dschrempf:
See https://github.com/well-typed/hs-bindgen/pull/1042 @TravisCardwell

PR #1008, opened 2025-08-21
Generate address stubs for functions

jorisdral:
Resolves #46 

This PR builds on top of #1007, which should be merged first

See the commit message and updated manual section for details about the changes

PR #1024, opened 2025-08-28
Make foreign imports of address stubs return in `IO`

jorisdral:
... and make all these foreign imports `unsafe` too. The `IO` change is required for subtle reasons. We plan to use address stubs for function declarations in the future as well, but without the `IO` `GHC` will complain about "dodgy foreign imports".

These dodgy import warnings were found in #1008. The current PR fixes this so that #1008 can be merged after.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
@@ -1348,11 +1348,18 @@ functionDecs opts moduleName typedefs info f _spec =
 -- stub function is mangled, though the original name of the stub function is
 -- included in a comment before the stub.
 --
--- > foreign import capi safe "abc949ab" simpleGlobal_ptr :: Ptr CInt
+-- > foreign import ccall unsafe "abc949ab" abc949ab :: IO (Ptr CInt)
 --
 -- Note that stub function also has a @const@ function attribute to emphasise
 -- that the function always returns the same address throughout the lifetime of
--- the program, which means we can omit the 'IO' from the foreign import.
+-- the program. This means we could omit the 'IO' from the foreign import to
+-- make it a pure foreign import. Instead, we make the foreign import impure and
+-- we generate an additional pure Haskell function that safely unsafely runs the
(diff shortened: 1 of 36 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -1458,15 +1467,18 @@ addressStubDecs opts moduleName info tyQual ty _spec = (,stubImportName) $
         , foreignImportCallConv = CallConvUserlandCAPI
         , foreignImportOrigin   = Origin.Global ty
         , foreignImportComment  = generateHaddocks (C.declComment info)
-        , foreignImportSafety   = SHs.Safe
+          -- We're binding to simple address stubs, so there are no callbacks
+          -- into Haskell code, meaning we can make these imports unsafe.
```
jorisdral: And short running

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -1496,6 +1508,26 @@ addressStubDecs opts moduleName info tyQual ty _spec = (,stubImportName) $
...
+
+    runnerDecl :: Hs.Decl
+    runnerDecl = Hs.DeclSimple $ SHs.DVar $ SHs.Var {
+          varName    = runnerName
+        , varType    = runnerType
+        , varExpr    = runnerExpr
+        , varComment = Nothing
+        }
+
+    runnerName = HsName $ getHsIdentifier (C.nameHsIdent (C.declId info)) <> "_ptr"
```
jorisdral, line 1533: Use `stubNameMangled`

edsko (approved):
Nice, feel free to merge once the Haskell name of the stub is set to the C name (and therefore unique).

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
jorisdral: Done

PR #1040, opened 2025-09-01
Honor `const` type qualifiers

### LINKED ISSUE #922: Wrapper generated code is missing `const`
I found the following bug:

```c
bool list_example(const char** items, size_t count);
```

generates:

```haskell
foreign import ccall safe "Generated_list_example" list_example :: (F.Ptr (F.Ptr FC.CChar)) -> HsBindgen.Runtime.Prelude.CSize -> IO FC.CBool
```

Which looks okay, however the wrapper that is generated is:

```c
_Bool Generated_list_example (char **arg1, size_t arg2) { return list_example(arg1, arg2); }
```

which leads to the following error when compiling:

```
src/../doxygen_test.h:172:32: error:
     note: expected ‘const char **’ but argument is of type ‘char **’
      172 | bool list_example(const char** items, size_t count);
          |                   ~~~~~~~~~~~~~^~~~~
    |
172 | bool list_example(const char** items, size_t count);
    |                                ^

<no location info>: error:
    `cc' failed in phase `C Compiler'. (Exit code: 1)
Error: [Cabal-7125]
Failed to build doxygen-example-0.1.0.0.
```

The wrapper is missing the `const`.

### LINKED ISSUE #994: Include type qualifiers in C types of the frontend's internal and external ASTs
The `Type` datatype in the `HsBindgen.Frontend.AST.Internal` and `HsBindgen.Frontend.AST.External` modules does not carry type qualifier information, like `const`, `volatile`, and `restrict`. Since #898 and #969 we pretty-print the `Type` datatype as C code, including outermost type qualifiers. By outermost type qualifier we mean, for example:

```c
// outermost const
extern const int a; // type is const-int
// no outermost const
extern const int * b; // type is pointer-to-const-int
// outermost const
extern int * const c; // type is const-pointer-to-int
// outermost const
extern const int * const d; // type is const-pointer-to-const-int
```

The second and fourth example also include const type qualifiers *inside* the type, and these are not represented in the `Type` datatype, so they are not printed. This loss of type information can lead to C compiler warnings.

Concretely this problem shows up for bindings of global (const) variables. For the example `global` below we generate and pretty-print a stub C function called `get_global_ptr`:

```c
// global
extern const int * const global; // type is const-pointer-to-const-int

// stub
__attribute__ ((const)) const signed int **get_global_ptr (void) { return &global; }
```

This leads to a warning:

```
/tmp/ghc29646_0/ghc_17.c:14:87: error:
     warning: return discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
```

This warning can be observed when running the manual.

If we would want to prevent these warnings, we should update the `Type` datatype to include type qualifiers.

dschrempf:
Closes #922
Closes #994


dschrempf:
Tests are failing because before this change we forced a `const` qualifier on global constant variables (`globalConst` function).

Now that we properly support `const` type qualifiers, I added a check if the type provided to `globalConst` is really qualified by `const`, and this seems not to be the case.

I think we need to discuss how to proceed.

dschrempf:
Putting this on the queue since it solved the `const` issue with `pcap`.

jorisdral:
Nice :)

PR #1043, opened 2025-09-02
Dom/1002/minimize api 2

### LINKED ISSUE #1002: Minimize API of public library before first release of `hs-bindgen-cli`
The release of `hs-bindgen-cli` also entails public release of `hs-bindgen` the public library. We only want to freeze part of the API which is realy required at this point, and remove other parts of the interface not required at the moment.

dschrempf:
- Minimize exposed API from `HsBindgen.Lib` and `HsBindgen.TH`.
- Minimize internal exports used from `hs-bindgen-cli` (some are required because we have the `dev` command).

Should be merged after #1041 

Closes #1002


edsko (approved):
Approving the modified version of this PR that we discussed over meet.

dschrempf:
For this PR:
- Move public libary (Lib and Common) to internal.
- Add comment Lib is intended to be public API but we are not there yet.
- Only module left: TH module.

Issue #1028, opened 2025-08-29
Should `nonParsedDecls` contain failed-to-parse declarations?


edsko:
The declarations that we process fall into three categories:

* _Skipped_: we didn't try to parse them at all. These are recorded (currently) in `nonParsedDecls` (which is possibly a bad name, part of this issue)
* _Attempted to parse, but failed_: these are not recorded anywhere
* _Attempted to parse, and succeeded_: this is the main result from the `Parse` pass

I am wondering if this is correct (I think it's not). Specifically, if we fail to parse a type `struct Foo`, say (maybe because it contains a `long double`), then we should still be able to match it against an external binding spec for `Foo`, but for that we need to have an entry in `nonParsedDecls`. Currently (I think) this will not be possible, which is all the more unfortunate because one of the reasons for a (prescriptive) binding spec might precisely _be_ that we cannot parse it.

@TravisCardwell thoughts?

edsko:
We could potentially unify `ParseMsgs` and `NonParsedDecls`, as follows:

```hs
newtype ParseStatus = ParseStatus {
    getParseStatus :: Map QualDeclId (SourceLoc, DeclStatus)
  }

data DeclStatus =
    ParseSkipped 
  | ParseSucceeded [ParseMsg]
  | ParseFailed [ParseMsg]
```

dschrempf:
```
Status quo:
Set of all declarations is split into
    NonParsedDecls
    ParsedAndSucceededDecls
    ParseMsgs

Goal (?):
Set of all declarations is split into
    SkippedDecls   -> Skipped and not selected [OK]   | Skipped and selected (via program slicing) [ERROR]
    FailedDecls    -> Failed and not selected  [OK]   | Failed and selected                        [EMIT TRACES]
    SucceededDecls -> Succeeded and not selected [OK] | Succeeded and selected                     [EMIT TRACES]
```
(The split corresponds to the `ParseStatus` type from above).

edsko:
Assigned this to both @TravisCardwell and @dschrempf right now; Travis because this affects binding specs, Dominik because this affects the `ParseMsg` work; let's figure out what to do first, and then who will do it.

TravisCardwell:
Indeed, being able to provide external bindings for declarations that we do not support sounds desirable.  Such bindings would need to be created manually, so we would never generate them, right?  The binding specification would never be used as a prescriptive binding specification, as we cannot even parse/reify the unsupported declarations.

[...]

PR #1060, opened 2025-09-04
Add `TagKind` to `QualPrelimDeclIdAnon`

TravisCardwell:
This is done for two reasons:

* We need to name `QualPrelimDeclId`, resulting in a `QualDeclId`.
  See #1028
* We plan on distinguishing the tag kind in the spelling of anonymous types.  Example: `struct @foo`

To make sure I am not missing anything, I implemented `nameQualPrelimDeclId` in `NameAnon` as well as a preliminary `nameParseStatus` function.  Both are commented out for now since they are not used yet.  The `nameParseStatus` implementation needs to be updated to use the `newtype` wrappers when they are added.


edsko (approved):

TravisCardwell:
Yet again builds fail due to Ubuntu server issues.

PR #1063, opened 2025-09-05
Fix translation of String macros

dschrempf:
- Use `GHC.Ptr.Ptr (Ptr)` instead of the the non-existing (not re-exported) `Foreign.Ptr.Ptr (Ptr)`.
- Add the `MagicHash` extension.


dschrempf:
Sorry, I merged this without explicit approval. Edsko and I talked about the code change in the morning.

PR #1065, opened 2025-09-08
Add `ClangArgsConfig`

TravisCardwell:
There are two reasons for using a separate configuration type:

* Clang arguments are determined based on configuration impurely.  Arguments may be added depending on `BINDGEN_CLANG_EXTRA_ARGS` environment variables.  Builtin include directory arguments may also be added.  Both should be done once (during "boot").

* `Clang.Args` implemented many `hs-bindgen` features/limitations.  We plan on making the `clang` package separate/independent from `hs-bindgen`, so those features/limitations should be implemented in `hs-bindgen`.

The `ClangArgs` type is changed to be a trivial `newtype` wrapper around raw arguments (`[String]`).

`CStandard` is still defined in `Clang.Args`, as it is *not* specific to `hs-bindgen`.  Function `getStdClangArg` gets the appropriate Clang option, depending on the version of Clang.  Sum type `Gnu = DisableGnu | EnableGnu` is added to support this.  Note that this functionality is used in the `c-expr-runtime` tests, which do not depend on `hs-bindgen`.

Following our configuration convention, `ClangArgsConfig` is defined in `HsBindgen.Config.ClangArgs`.  The `BuiltinIncDirConfig` type is moved into this module, as this configuration is part of `ClangArgsConfig`.

The `BINDGEN_EXTRA_CLANG_ARGS` functionality is moved to `HsBindgen.Clang.ExtraClangArgs`, alongside the `BuiltingConfigDir` module.  Note that I ended up refactoring `getExtraClangArgs` because there was a bug: the environment variable for the target was always reported in the trace message even when falling back to `BINDGEN_EXTRA_CLANG_ARGS`.

Resolving the Clang arguments with processing of `BINDGEN_CLANG_EXTRA_ARGS` environment variables and builtin include directory configuration is a common action, needed in development commands that cannot rely on the whole `boot` configuration/artifact.  I implemented it in `HsBindgen.Boot` as function `getClangArgs`.  This function is used within `boot` as well as in the development commands that need it.  Note that exception `InvalidClangArgs` was previously handled deep inside of our `clang` library, using `throwIO`.  We can now handle it in this `getClangArgs` function.  I kept `throwIO` for now, but we can change this if desired.

This PR also adds support for the `BINDGEN_BUILTIN_INCLUDE_DIR` environment variable.


hs-bindgen/src-internal/HsBindgen/Config/ClangArgs.hs
```diff
@@ -0,0 +1,289 @@
...
+      --   , 'clangArgsInner'
+      --   , 'clangArgsAfter'
+      --   ]
+      -- @
+      --
+      -- where we assemble @clangArgsInternal@ using the specialized
+      -- 'ClangArgsConfig' records above.
+      --
+      -- See https://clang.llvm.org/docs/ClangCommandLineReference.html
+    , clangArgsInner :: [String]
```
dschrempf, line 96: Now that we have this type in `hs-bindgen`, we should probably mention where we put BINDGEN_EXTRA_CLANG_ARGS. We could make it a separate record. (In my opinion not really necessary, just explaing the original choice of option names).

```haskell
      -- @
      -- concat [
      --     'clangArgsBefore'
      --   ,  clangArgsInternal
      --   , 'clangArgsInner'
      --   ,  BINDGEN_EXTRA_CLANG_ARGS
      --   , 'clangArgsAfter'
      --   ]
      -- @
      --
      --where ...
```

dschrempf: (That would also make it clear why we have three options, not two).

dschrempf (commented):
When running `hs-bindgen-cli` on this branch, I get
```
[Warning] [HsBindgen] [builtin-include-dir] builtin include directory not found using llvm-config: /nix/store/pximyq2ycfc8z738wlvca6dfivx8gnby-llvm-19.1.7-dev/lib/clang/19/include
```
I am not getting this warning on the main branch.

I also suggest reporting the actual found builtin include directory using an `Info`-level trace. At the moment, I get
```
[Debug  ] [HsBindgen] [builtin-include-dir] builtin include directory found using clang: /nix/store/6hyah654xshls0qdr5gqk5mpa5rvd43f-clang-wrapper-19.1.7/resource-root/include
```

I also tested setting the env var and it works perfectly!

TravisCardwell:
Thank you very much for testing, @dschrempf!

I investigated, but I still do not know what could be causing that warning in this branch if it is not emitted in `main`.  The only change to the builtin include directory logic is the processing of the environment variable at the top of the function, so nothing related to how the builtin include directory is determined has been changed.  Perhaps there is a difference in the environment?

The directory reported is indeed incorrect, though, and it is fortunate that we see it!  With POSIX systems that following the Filesystem Hierarchy Standard, the LLVM prefix is consistent even though many distributions separate the installation into separate packages (so that users do not have to install all of LLVM/Clang when doing so it not needed).  Windows just installs all of LLVM/Clang to one place (like `opt` in FHS), so it is also consistent.  With Nix, however, the separate packages do *not* have a consistent LLVM prefix.

This warning indicates that `llvm-config --prefix` is outputting the prefix for the `llvm` package, which is sensible.  The problem is that the resource directory package installs to a separate directory under `/nix/store` (and is not linked), so this prefix cannot be used to determine the resource directory.

Since the LLVM implementation assumes a consistent prefix, it is not surprising that Nix has issues.  In the Nix section of the Installation page of the manual, I think that we should document that Nix users should set `BINDGEN_BUILTIN_INCLUDE_DIR=disable`.  Nix users must use `BINDGEN_CLANG_EXTRA_ARGS` anyway.  Sound good?

TravisCardwell:
Automatic configuration of the builtin include directory should "just work" on all platforms (*cough* aside from Nix).  The various trace messages reporting how the directory is determined have level `Debug` because they should only ever be needed for debugging.  I do not really mind, though, so I will go ahead and change these to `Info`.  Perhaps this information would be useful to have in logs if running with `-v3` in CI.

hs-bindgen/src-internal/HsBindgen/Config/ClangArgs.hs
TravisCardwell: Sounds good.  I updated to the following:

```
-- Complete list of arguments passed to @libclang@:
      --
      -- @
      -- concat [
      --     'clangArgsBefore'
      --   , clangArgsInternal  -- other ClangArgsConfig options
      --   , 'clangArgsInner'
      --   , clangArgsExtra     -- BINDGEN_EXTRA_CLANG_ARGS
      --   , 'clangArgsAfter'
      --   , clangArgsBuiltin   -- builtin include directory
      --   ]
      -- @
```

I also updated the CLI footer accordingly.

dschrempf:
> Since the LLVM implementation assumes a consistent prefix, it is not surprising that Nix has issues. In the Nix section of the Installation page of the manual, I think that we should document that Nix users should set BINDGEN_BUILTIN_INCLUDE_DIR=disable. Nix users must use BINDGEN_CLANG_EXTRA_ARGS anyway. Sound good?

Yes, that's also why I advocated for the `BINDGEN_BUILTIN_INCLUDE_DIR` environment variable. I think it is best practice to refer Nix users to a Nix Flake, and they can go from there. This Flake can be the one I am managing at the moment, but does not have to be.

With respect to the discrepancy between resource directories of `llvm-config` and `clang` with Nix: That's unfortunate, but I don't think it is a problem as we have a good solution (see above).

> Automatic configuration of the builtin include directory should "just work" on all platforms (cough aside from Nix). The various trace messages reporting how the directory is determined have level Debug because they should only ever be needed for debugging. I do not really mind, though, so I will go ahead and change these to Info. Perhaps this information would be useful to have in logs if running with -v3 in CI.

Yes, you are completely right, I was only referring to a final trace message with `Info` level. This final trace message may report the final automatically determined builtin include directory (if it was automatically determined). Does this make sense? (All other messages can and should be `Debug`, as you say).

Ad Warning. I will investigate now, and report back.

dschrempf:
I tried again, and the warning only shows up on the feature branch. However, I also saw that there is a nice Info message stating how the builtin include directory was finally found, so my comment above can probably marked as done.

Let me know, if I can provide more information!

[log-branch.txt](https://github.com/user-attachments/files/22226263/log-branch.txt)
[log-main.txt](https://github.com/user-attachments/files/22226265/log-main.txt)

TravisCardwell:
Thank you very much!  Having the full log really helps a lot!

Now I understand where that out-of-place GCC path is coming from.  It is passed as a command-line argument via `BINDGEN_EXTRA_CLANG_ARGS`!

I also now understand the difference in behavior.  Even getting the resource directory using `-print-resource-dir` is affected by the configuration in your `BINDGEN_EXTRA_CLANG_ARGS`.

I can resolve the issue by applying any `BINDGEN_EXTRA_CLANG_ARGS` to the `ClangArgsConfig` and passing it (or the corresponding `ClangArgs`) to `getBuiltinIncDir`, so that it can be threaded through the functions and used when calling `-print-resource-dir`.  The result would then be applied to the intermediate `ClangArgsConfig` to produce a final `ClangArgsConfig` that is translated to the `ClangArgs` artifact.

The only worry I have about this is that users can put *anything* in `BINDGEN_EXTRA_CLANG_ARGS`, and who knows what might break the functionality.  I guess that is already true, though.  Users may use `BINDGEN_EXTRA_CLANG_ARGS` as well as `--clang-option` at their own risk.

I will go ahead and implement this fix.

dschrempf:
Thanks for your analysis!

Just one last question:

> Now I understand where that out-of-place GCC path is coming from. It is passed as a command-line argument via BINDGEN_EXTRA_CLANG_ARGS!

Do you think this is a bug? I am using `BINDGEN_EXTRA_CLANG_ARGS` as it is assembled by `rustPlatform.bindgenHook` available in Nixpkgs. Should I report or at least ask around upstream?

TravisCardwell:
I do not understand enough to know if it is a bug or not.

In [`rust-bindgen-hook.sh`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/rust/hooks/rust-bindgen-hook.sh):

```bash
BINDGEN_EXTRA_CLANG_ARGS="$(< @clang@/nix-support/cc-cflags) $(< @clang@/nix-support/libc-cflags) $(< @clang@/nix-support/libcxx-cxxflags) $NIX_CFLAGS_COMPILE"
```

The `@clang@/nix-support/cc-cflags` flags are defined [here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/compilers/llvm/common/default.nix).

The `@clang@/nix-support/libc-cflags` are more complicated, and it looks like they include `@clang@/nix-support/cc-cflags`.  Perhaps that is where the duplicate settings come from...?  Perhaps they also include non-Clang stuff, which may be where the GCC paths are coming from.  I have not confirmed this, though.

I imagined that they must be configuring GCC paths for a reason, but the duplicate settings indicate that not much attention is being to it.  Perhaps it is a bug.

dschrempf:
~~Hm, maybe I ask around on Discourse.~~

I did some research, and it turns out the we'd have to change the `stdenv` proper to only use `clang`, and not `gcc`. I tried doing so, but it requires recompilation of the complete toolchain, a no-go in my opinion.

The reason is that `bindgenHook` just collects all library dependencies in the closure and adds their respective C compilation flags (including their include directories). Since we have GCC in the closure, the relevant include folders end up in BINDGEN_EXTRA_CLANG_ARGS. I think we should just live with that.

References:
- https://nixos.wiki/wiki/Using_Clang_instead_of_GCC

Edit: I will add a note to the Flake.

TravisCardwell:
When we had modes `auto` (default), `clang`, and `disable`, use of `clang` mode was something that a user had to explicitly request.  The implementation detects this, and any `clang` traces are made `Error`s.  Now that `clang` is the default, we can no longer do this!  I am out of time this evening, but this will be a quick fix that I will do in the morning.

Issue #407, opened 2025-02-07
Update unqualified imports and/or reserved identifiers


TravisCardwell:
Unqualified imports may collide with mangled names.  Since changing the name mangler to do minimal name mangling by default (#331), mangled names no longer have a `C` prefix.  Aside from operators, which we do not generate, all identifiers that are imported unqualified should be reserved ([here](https://github.com/well-typed/hs-bindgen/blob/07b182bfb981666a51e7bff1dc7535e123bcf288/hs-bindgen/src/HsBindgen/Hs/AST/Name.hs#L623-L639)).

Looking at the current implementation, here are the current problematic unqualified imports:

* `Bounded`
* `Enum`
* `Eq`
* `FiniteBits`
* `Floating`
* `Fractional`
* `IO`
* `Int`
* `Integral`
* `Num`
* `Ord`
* `Read`
* `RealFloat`
* `RealFrac`
* `Real`
* `Show`
* `Void`

One option is to minimize use of unqualified imports in generated code.  FWIW, I am a fan of this option.

We may want to prevent translations to some of these names anyway, for "[sanity](https://github.com/well-typed/hs-bindgen/blob/07b182bfb981666a51e7bff1dc7535e123bcf288/hs-bindgen/src/HsBindgen/Hs/AST/Name.hs#L641-L676)."

[`sanityReservedTypeNames`](https://github.com/well-typed/hs-bindgen/blob/07b182bfb981666a51e7bff1dc7535e123bcf288/hs-bindgen/src/HsBindgen/Hs/AST/Name.hs#L641-L676) only has `C`-prefixed types, but perhaps we should now add many more reserved type names?

[`hsBindgenReservedTypeNames`](https://github.com/well-typed/hs-bindgen/blob/07b182bfb981666a51e7bff1dc7535e123bcf288/hs-bindgen/src/HsBindgen/Hs/AST/Name.hs#L623-L629) currently reserves `~`, but I do not think this is necessary because the name mangler should never generate `~` even with custom rules.

edsko:
Marking this as a bug, because we can currently actually generate invalid code:

```c
struct Eq {
  int x;
  int y;
};
```

results in Haskell code that cannot be compiled 

```
Ambiguous occurrence ‘Eq’
```

PR #1069, opened 2025-09-08
Update reserved names

TravisCardwell:
See #407

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Config/FixCandidate/ReservedNames.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Config/FixCandidate/ReservedNames.hs
@@ -135,20 +135,39 @@ ghcNonReservedKeywords = [
 
 -- | Names in the type namespace that @hs-bindgen@ may use unqualified
 --
--- * '~'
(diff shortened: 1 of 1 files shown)


dschrempf (approved):
This looks good, thank you. Each day I learn more about the intricacies of `hs-bindgen` :-D. Out of curiousity, how did you get the list of reserved type names? Many more declarations are exported from `Prelude`.

TravisCardwell:
We generate modules using `NoImplicitPrelude` and import identifiers from `Prelude` explicitly.  We can also import `Prelude` qualified, but it looks like we are currently not doing that.

To get the list, I did two things:

* In `HsBindgen.Backend.Artefact.HsModule.Names`, search for all uses of `importU`.  Note that tuples are an edge case.

* In `hs-bindgen/fixtures`, confirm the `.pp.hs` files.  They should be a *subset* of the set of names found above.

    ```
    $ find hs-bindgen/fixtures -type f -name '*.pp.hs' \
    | xargs grep '^import Prelude' \
    | sed -e 's/^[^(]*(//' -e 's/)$//' -e 's/, /\n/g' \
    | sort --unique
    ```

Issue #94, opened 2024-08-15
Allow users to choose between `safe`/`unsafe`/both


edsko:
The choice between `safe` and `unsafe` is not really something the tool can make on behalf of the user,  though in the absence of any configuration, `safe` should be the default. There are also examples where we want to make _two_ bindings (one `safe` and one `unsafe`) to the same C function; for example, see [cryptohash](https://github.com/haskell-hvr/cryptohash-sha256/blob/8094c77650ed36b813ff7ea3eeeaedf5edbdfee6/src/Crypto/Hash/SHA256/FFI.hs#L58-L62) (where the choice between the two is [based on input size](https://github.com/haskell-hvr/cryptohash-sha256/blob/8094c77650ed36b813ff7ea3eeeaedf5edbdfee6/src/Crypto/Hash/SHA256.hs#L167), though this is of course not directly relevant to `hs-bindgen`).

edsko:
One option here might be to follow the Vulkan bindings, and generate two modules `.Safe` and `.Unsafe`, so that programmers can just import the module they want. (Somewhat related: #75).

phadej:
> One option here might be to follow the Vulkan bindings, and generate two modules .Safe and .Unsafe

That will be terrible for interfaces with structs.

edsko:
Yes, this will certainly not be a good option in all cases.

TravisCardwell:
What if we only put functions in `Safe` and `Unsafe` submodules?  Module `Foo` could define the types and perhaps even re-export the functions from `Foo.Safe`.  Users would be able to use `Foo.Unsafe` at their own discretion.

edsko:
Yes, the three module struct is what we agreed on I think? Not sure I like the re-export option; I think it's good if we make users every explicitly pick a safety level; it might also be awkward to use, because you probably want the types module imported unqualified. But I could see the argument either way, I don't feel _terribly_ strongly.

edsko:
@TravisCardwell I've asked Joris to work on this, my understand is that you haven't done any work on this ticket yet right?

TravisCardwell:
Correct: I have not done any work on this ticket yet.

jorisdral:
@edsko and I were discussing whether we should add an additional module that contains the bindings to addresses of declarations. Currently, we generate bindings to the addresses of global variables, and we might soon generate bindings to the addresses of functions too (see #1008). Putting the address imports in their own module might be cleaner, and it prevents name clashes to a certain degree

[...]

PR #1071, opened 2025-09-09
Attach C-side wrappers to calling convention type

dschrempf:
This is a preparation for #94.

- We attach the required C sources directly to the userland CAPI calling convention constructor.
- We split the translation of function declarations into creating the declaration itself, and into creating the corresponding function pointer.

I am confident that these changes are OK, because they only change the Hs AST, not the pretty-printed Haskell sources.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/CallConv.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/CallConv.hs
@@ -2,27 +2,47 @@
 --
 -- Intended for unqualified import.
 module HsBindgen.Backend.Hs.CallConv (
-    CallConv(..)
+    UserlandCapiWrapper(..)
+  , CallConv(..)
   , ImportStyle(..)
   ) where
 
 import GHC.Generics (Generic)
 
+import HsBindgen.Frontend.RootHeader
+
 {-------------------------------------------------------------------------------
   Definition
 -------------------------------------------------------------------------------}
 
+-- | The 'CallConvUserlandCAPI' requires a wrapper on the C side with a
+-- corresponding import.
+data UserlandCapiWrapper = UserlandCapiWrapper {
+      capiWrapperDefinition :: String
+    , capiWrapperImport     :: HashIncludeArg
+    }
+  deriving (Show, Generic)
+
 data CallConv =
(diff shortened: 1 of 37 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/Hs/CallConv.hs
```diff
@@ -2,27 +2,43 @@
...
     --
-    -- The header is not required in this case (it is only used by the wrapper).
-    CallConvUserlandCAPI
+    -- There is no need to import the C header into the Haskell source file in
+    -- this case (the C header is only used by the wrapper on the C side).
+    --
+    -- We directly attach the C-side wrappers.
+    CallConvUserlandCAPI UserlandCapiWrapper
 
+    -- TODO_PR: Not sure if we use the GHC @capi@ calling convention. Remove?
```
dschrempf: Can we remove support for the standard CAPI calling convention? I think we do not use it.

I am not sure about `ccall`.

dschrempf: We decided to not remove support, I have added an appropriate NOTE into the code.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -334,7 +333,17 @@ generateDecs opts haddockConfig moduleName typedefs (C.Decl info kind spec) =
...
       C.DeclFunction f ->
-        return $ functionDecs opts haddockConfig moduleName typedefs info f spec
+        let funDecls = functionDecs opts haddockConfig moduleName typedefs info f spec
+            funType  = (C.TypeFun (snd <$> C.functionArgs f) (C.functionRes f))
+            -- Declare a function pointer. We can pass this 'FunPtr' to C
+            -- functions that take a function pointer of the appropriate type.
+            --
+            -- For now, we generate function pointers for /all/ function
+            -- declarations. We might be able to be smart and generate function
+            -- pointers only for a subset of /useful/ functions.
```
edsko: Comment is not needed; there is no way to know which addresses the user might want (it might be passed to a different function from another library for all we know).

Issue #1039, opened 2025-09-01
Update CI test strategy


TravisCardwell:
In https://github.com/well-typed/hs-bindgen/issues/996#issuecomment-3190747376, @dschrempf suggested:

> Testing a 3D matrix (OS, GHC version, LLVM version) will be too cumbersome. Maybe we should test all combinations of GHC and LLVM versions for a single OS, and a "recommended" tuple of (GHC version, LLVM version) for the other OSes?

The tests that we currently run are shown below, per platform.

Testing arbitrary versions of LLVM/Clang is easiest on macOS.  How about we update the configuration to test the following?

* The earliest release of LLVM/Clang that we support (14.0.0), tested using the latest minor version of each major release of GHC that we support, on Ubuntu
* The latest release of LLVM/Clang supported by [KyleMayes/install-llvm-action][], tested using the latest minor version of each major release of GHC that we support, on macOS
* The latest minor version of each major release of LLVM/Clang supported by [KyleMayes/install-llvm-action][], tested using our "recommended" version of GHC (what version?), on macOS
* The latest release of LLVM/Clang supported by [KyleMayes/install-llvm-action][], tested using our "recommended" version of GHC (what version?), on Windows
* The version of LLVM/Clang that is bundled with our "recommended" version of GHC (what version?) when installed via GHCup, on Windows

Once LLVM/Clang 21 is supported, that would be a total of 21 jobs, not including the manual jobs.  Too many?

## macOS

| GHC   | 14   | 15   | 16   | 17   | 18   | 19   | 20   | 21   |
| :---  | :--: | :--: | :--: | :--: | :--: | :--: | :--: | :--: |
| 9.2   |      |      |      |      |      |      |      |      |
| 9.4   | ✔    |      | ✔    |      | ✔    |      | ✔    |      |
| 9.6   |      |      |      |      |      |      |      |      |
| 9.8   |      |      |      |      |      |      |      |      |
| 9.10  |      |      |      |      |      |      |      |      |
| 9.12  |      |      |      |      |      |      |      |      |

Shall we update the tests to test the latest minor releases for all major versions of LLVM/Clang that we support?  Shall we keep testing with GHC 9.4, or should we select a different GHC version?

Details:

* The [macOS runner image][] currently has LLVM/Clang versions 16.0.0 and 18.1.8 pre-installed.  We do not use these.
* We use versions installed by [KyleMayes/install-llvm-action][], which come from [llvm-project/releases][].
[...]

PR #1072, opened 2025-09-09
Reduce CI by using a two-tier approach

dschrempf:
Related to, but not solving https://github.com/well-typed/hs-bindgen/issues/1039

This PR seeks to reduce the strain on our CI pipeline.

First, we do not run CI on normal pushes. We only run checks on pull requests (PRs) targeting `main`, in the merge queue (MQ) or when PRs reach `main`.

Second, we use a two-tiered approach:

  1. We provide fast feedback on PR status changes.
  2. We perform comprehensive checks in the MQ.

  In particular:
  - PR targeting `main`: We run the "Simple Haskell CI" only on PR status changes.
  - MQ: We build the manuals only in the MQ.
  - MQ: We run the full "Haskell CI" only in the MQ.

Other checks that are fast run always (PR, MQ, PR reaches `main`):
- `fix-whitespace`
- `hlint`

This contains manual changes to `haskell-ci.yml`. I am not familiar with `haskell-ci`, are there appropriate options to keep these changes after recreating the GitHub Actions script?


edsko:
> This contains manual changes to `haskell-ci.yml`. I am not familiar with `haskell-ci`, are there appropriate options to keep these changes after recreating the GitHub Actions script?

See https://github.com/haskell-CI/haskell-ci/pull/796

dschrempf:
I also realized that the "Simple CI" spreads across LLVM versions, but only uses GHC 9.4. We should probably test the earliest and latest GHC we support on PR status changes.

dschrempf:
Some more feedback from Armando: Can we run CI conditionally? For example, can we run build the manuals when the manual was changed?

dschrempf:
@bolt12 We could use something like https://github.com/dorny/paths-filter.

```yaml
jobs:
  # Detect changes.
  changes:
    runs-on: ubuntu-latest
    permissions:
      pull-requests: read
    outputs:
      manual: ${{ steps.filter.outputs.manual }}
    steps:
    - uses: dorny/paths-filter@v3
      id: filter
      with:
        filters: |
          manual:
            - 'manual/**'

  # Build and test manual.
  manual:
    needs: changes
    if: ${{ needs.changes.outputs.manual == 'true' }}
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
```

What do you think?

bolt12:
That sounds exactly what we need! Have you tested it? Perhaps you can do a quick check in this PR and then remove the changes to the manual

dschrempf:
@bolt12 do we still need `build-manual-ubuntu-llvm.yml`, or can it go altogether?

bolt12:
I think it can go @dschrempf !

TravisCardwell:
I tried to remove it before, but CI still considered it a "required check" even with the `.yml` file gone.  Perhaps that is configured in some GitHub project settings (which I do not have access to).

dschrempf:
Decision log from meeting:

Simple CI:
- Extend to build the manual. (Remove separate build manual jobs).
- Test all LLVM versions on Ubuntu only (except the one version tested by Haskell-CI).
- Test one LLVM version on MacOS, Windows.

PR targeting main:
- Fast jobs.
- Simple on Ubuntu (one GHC, one LLVM) plus fast jobs.

MQ:
- All jobs.

TravisCardwell:
Perhaps we could specify `--unique-id` when building the manual, to get rid of those warnings in the logs?

TravisCardwell:
I investigated the Windows issue.

The previous Windows manual tests used the version of LLVM/Clang that is installed with GHC by GHCup.

```
clang version 14.0.6
Target: x86_64-w64-windows-gnu
```

The updated job uses a version of LLVM/Clang that we install from the builds distributed on GitHub, via `KyleMayes/install-llvm-action`, which has a different target.

```
clang version 14.0.6
Target: x86_64-pc-windows-msvc
```

The MSVC target includes Microsoft's implementation of complex numbers, *not* the C standard.  ([Reference](https://learn.microsoft.com/en-us/cpp/c-runtime-library/complex-math-support))

dschrempf:
Nice, thank you for finding the target which I didn't see. ~~How did you obtain the target, is it also in the logs?~~ (Found it).

I am aware that the two jobs use different Clang versions. The problem is, I can not run both, the "`hs-bindgen` tests" and "build the manual" with the same Clang version on Windows. Somehow it did work before, because the tests were run with one versin while the manual was built with another.

I will see what I can do!

TravisCardwell:
You can see the target in the logs for the `Debug (hs-bindgen)` step.  The target displayed is the default target for the `libclang` shared library that is dynamically linked at runtime.

If you would like to run the Windows tests and build the manual using the version of LLVM/Clang that is installed with GHC by GHCup, you need to prepend the `bin` directory to the Windows system path.

From [`build-manual-windows.yml`](https://github.com/well-typed/hs-bindgen/blob/77044c352fe1ac747f597a3cb72b1b549fa32532/.github/workflows/build-manual-windows.yml#L84-L93):

```yaml
# Environment variables:
#
# * PATH - The PATH is used for locating DLLs, so it is critical that the
#   LLVM/Clang bin directory is prepended to the system PATH.  Otherwise,
#   hs-bindgen will dynamically load the libclang.dll for a different
#   version of LLVM/Clang.  Note that this is a change to the system path,
#   using Windows path syntax.
- name: Configure environment variables
  shell: bash
  run: echo "C:\ghcup\ghc\${{matrix.ghc}}\mingw\bin" >> "$GITHUB_PATH"
```

If you want to test using `KyleMayes/install-llvm-action`, perhaps we should configure the `complex` test so that it does not run on Windows.  (That action includes the necessary update to the Windows system path.)

dschrempf:
Yes, thank you. I tried setting this variable on Friday before and after the build/test/manual, to no avail. I will summarize the different errors below, depending on which setup we use:

1. Install Clang 14 using `install-llvm-action`; do not temper with `GITHUB_PATH`.
   - Clang version: 14.0.6
   - Target: `x86_64-pc-windows-msvc`
   - Build: ✅
   - Test: ✅
   - Manual: ❌ "libclang `complex` error"

2. Do not use `install-llvm-action`; do not prepend `mingw/bin` to `GITHUB_PATH`.
   - Clang version: 20.1.8
   - Target: `x86_64-pc-windows-msvc`
   - Build: ❌ "configure: error: could not find llvm-config; LLVM_PATH is also not set" <- (We can set this path, I will try later)
   - Test: ❔
   - Manual: ❔

3. Do not use `install-llvm-action`; prepend `C:\ghcup\ghc\${{ inputs.ghc }}\mingw\bin` to `GITHUB_PATH`
   - Clang version: 20.1.8 <- This is interesting. ~~Why does it not use the `mingw/bin` provided one which should be 14.0.6?~~ The path did not exist, GHC does not find `llvm-config` and fails.
   - Target: `x86_64-pc-windows-msvc`
   - Build: ❌ "configure: error: could not find llvm-config; LLVM_PATH is also not set"
   - Test: ❔
   - Manual: ❔

4. Do not use `install-llvm-action`; prepend `C:\tools\ghc-${{ inputs.ghc }}\mingw\bin` to `GITHUB_PATH`
   - Clang version: 14.0.6 (yay!)
   - Target: `x86_64-pc-windows-gnu` (interesting!)
   - Build: ✅
   - Test: ❌ "   (ty_1 *,double) - computed type: int -  Clang's type: <n/a>"
   - Manual: ❔

5. Install Clang 20 using `install-llvm-action`; do not temper with `GITHUB_PATH`
   - Clang version: 20.1.4
   - Target: `x86_64-pc-windows-msvc`
   - Build: ❌ "ld.lld: error: undefined symbol: __declspec(dllimport) clang_getUnqualifiedType"
   - Test: ❔
   - Manual: ❔

6. Install Clang 14 using `install-llvm-action`; **only for manual**: prepend `C:\tools\ghc-${{ inputs.ghc }}\mingw\bin` to `GITHUB_PATH`

   Build/test:
   - Clang version: 14.0.6
   - Target: `x86_64-pc-windows-msvc`
   - Build: ✅
   - Test: ✅

   Manual:
   - Clang version: 14.0.6
   - Target: `x86_64-pc-windows-gnu` <- note the tiny little difference here
   - Manual: ✅

   Drawback: Need to recompile `clang` and `hs-bindgen-cli`! This is not a big deal because these two components compile relatively fast.

TravisCardwell:
Oh!  I did not realize that you were running into the issue in the manual!

Perhaps this is yet another case of bad header filenames?  We should make all of our header filenames unique.  This might be problematic:

```
$ find . -type f -name complex_test.h | sort
./hs-bindgen/examples/golden/complex_test.h
./manual/c/complex_test.h
```

(I am wondering about this because the observed errors look like those that happen when the wrong header is included.)

dschrempf:
OK, the directory is wrong:
 ls: cannot access 'C:\ghcup\ghc\9.4.8': No such file or directory
I will investigate.

OMG. So it looks like `ghcup` comes with some default ghc versions, and then these files are available. It does not come with GHC 9.4.8, but it does come with GHC 9.12.2, that's why the original action _did work_.

TravisCardwell:
> OMG. So it looks like `ghcup` comes with some default ghc versions, and then these files are available. It does not come with GHC 9.4.8, but it does come with GHC 9.12.2, that's why the original action *did work*.

This is incorrect.

The directory was not found because the job installed GHC 9.4.8 via [Chocolatey](https://chocolatey.org/), *not* [GHCup](https://www.haskell.org/ghcup/).  Investigating, [`haskell-actions/setup`](https://github.com/haskell-actions/setup) installs via Chocolatey if the requested version is available and falls back to GHCup only when it is not ([source](https://github.com/haskell-actions/setup/blob/0cb0c9e0efc43f53d8a524915d3c6367c3fb7803/lib/installer.js#L182-L187)).

dschrempf:
Interesting though, that the `ghcup` directory is existing, and contains some other versions of GHC. Why is that?

TravisCardwell:
> Interesting though, that the `ghcup` directory is existing, and contains some other versions of GHC. Why is that?

According to the [Windows 2022 README](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md), GHC 9.12.2 is installed in the runner image.  Investigating a job for my PR that still uses GHC 9.12.2 on Windows, the `Set up Haskell` step indeed detects that it is already installed and just sets it as the default.  (It does not re-install.)

I see `9.10.2` also listed in your test.  I do not know where that comes from.  We could figure it out if it is already on the image or if it is a side-effect of something we do with more logging.

TravisCardwell:
> It seems that we can compile `hs-bindgen` with LLVM 14, but *not* the manual.

This is incorrect.  Our old manual tests use LLVM/Clang 14.

The errors are likely due to one of the following:

* We have multiple files named `complex_test.h`.  We should make our header filenames unique.
* It is possible the MSVC incompatibilities are causing issues.  If this is so, then I think that we should investigate and document it.  We could work around the issue by using the LLVM/Clang that targets `x86_64-w64-windows-gnu` so that standard headers are used.

TravisCardwell:
> It seems that we can not compile the `hs-bindgen` with newer versions of LLVM.

This sounds like something that we should resolve, perhaps in a separate issue.

I found [this log](https://github.com/well-typed/hs-bindgen/actions/runs/17724903510/job/50363665314) and see that it was tested with GHC 9.4.8 (installed via Chocolatey), using LLVM/Clang 20.1.4 targeting MSVC.

The logs indicate that `7z` can extract the files directly from the `.exe`!  I tried it, and it works (even on Linux).  I dumped the symbols of `libclang.dll` and confirmed that [`clang_getUnqualifiedType`](https://clang.llvm.org/doxygen/group__CINDEX__TYPES.html#ga8adac28955bf2f3a5ab1fd316a498334) exists:

```
$ objdump -p libclang.dll | grep clang_getUnqualifiedType
        [ 360] +base[ 361]  0168 clang_getUnqualifiedType
```

The logs for the `configure` step for our `clang` package look fishy...  Perhaps it is a configuration issue...

Unfortunately, this is not something that I can test locally with my current environment.

TravisCardwell:
> I also added some fixes the `Invocation` section of the manual, but in my opinion, we should reword this section so it is more concise.

FWIW, the added information belongs in the "Installation" section of the manual.  The "Invocation" section is about how to *use* the low-level preprocessor, Template Haskell, Literate Haskell, and library APIs.  Note that the "Installation" section is not low-level, as it concerns installing `hs-bingen` in general, for low-level or (future) high-level usage.

There are indeed a number of things that should be reworded/fixed.

We should probably improve the manual in a separate issue/PR.

cabal.project.ci
```diff
@@ -1,3 +1,6 @@
+-- The index state should match the one used for building the manual.
+index-state: 2025-09-16T04:55:58Z
+
```
jorisdral: `cabal.project` files can import other project files. Maybe that's the easiest way to enforce that the manual uses the same index state: create a separate `cabal.project.index-state` file containing the index state that both `cabal.project.ci` and the manual project file import

.github/workflows/haskell-simple-base.yml
```diff
@@ -0,0 +1,147 @@
...
+        run: |
+          wget http://security.ubuntu.com/ubuntu/pool/universe/n/ncurses/libtinfo5_6.3-2ubuntu0.1_amd64.deb
+          sudo apt install ./libtinfo5_6.3-2ubuntu0.1_amd64.deb
+
+      - name: Setup Haskell
+        id: setup-haskell
+        uses: haskell-actions/setup@v2
+        with:
+          ghc-version: ${{ inputs.ghc }}
+          cabal-version: latest
```
jorisdral: I'd probably set this to a major version like `3.12` so that the jobs can't break when a new cabal version comes out, and so that the caches don't change unexpectedly. Sometimes new cabal versions require dependencies to be rebuilt

.github/workflows/haskell-simple-base.yml
```diff
@@ -0,0 +1,147 @@
...
+          ls "$LLVM_PATH/bin"
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: cabal.project
+        run: cp cabal.project.ci cabal.project
+
+      - name: Dry build (for cache key)
+        run: cabal build all --enable-tests --dry-run
```
jorisdral: Maybe `Tests: True` should be added to `cabal.project.ci` instead of added on the command line. And then we can remove `--enable-tests` from other steps in this workflow too

.github/workflows/haskell-simple-base.yml
```diff
@@ -0,0 +1,147 @@
...
+      - name: Build manual (Windows)
+        if: runner.os == 'Windows'
+        uses: ./.github/actions/build-manual-windows
+
+      - name: Save cache
+        uses: actions/cache/save@v4
+        if: always()
+        with:
+          path: ${{ steps.setup-haskell.outputs.cabal-store }}
+          key: ${{ inputs.runner }}-${{ inputs.ghc }}-${{ inputs.llvm}}-${{ hashFiles('dist-newstyle/cache/plan.json') }}
```
jorisdral: Ideally, the "Save cache" step should come right after the step where we build all the dependencies, but I see that we're also building the manual and such which can build even more dependencies. Maybe we should add a comment as to why we use the `always`?

.github/workflows/haskell-simple-base.yml
```diff
@@ -0,0 +1,147 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: cabal.project
+        run: cp cabal.project.ci cabal.project
+
+      - name: Dry build (for cache key)
+        run: cabal build all --enable-tests --dry-run
+
+      - name: Restore cache
```
jorisdral:
```suggestion
      - name: Restore cache
        id: cache-cabal
```

.github/workflows/haskell-simple-base.yml
jorisdral:
```suggestion
      - name: Save cache
        uses: actions/cache/save@v4
        if: always()
        with:
          path: ${{ steps.setup-haskell.outputs.cabal-store }}
          key: ${{ steps.cache-cabal.outputs.cache-primary-key }}
```

.github/workflows/haskell-simple-merge-queue.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  test:
+    name: Test
+    # PRs can only be added to the MQ when all "required" checks pass.
+    # PRs can only be merged into `main` when all "required" checks pass in the MQ.
+    #
+    # So we need to always run all checks, but skip PR view jobs in the MQ, and
+    # MQ jobs in the PR view.
+    if: github.event_name == 'merge_group'
```
jorisdral: Right now, I'm seeing in the PR view:

image

So there is a single skipped job. I think that instead we should change the workflows so that we see multiple jobs skipped, one for each matrix combination. I think we could achieve this by adding a new input to the base workflow that determines whether the base workflow should run or not, and move the conditional from here into the base workflow.

.github/workflows/haskell-simple-pull-request.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  test:
+    name: Test
+    # PRs can only be added to the MQ when all "required" checks pass.
+    # PRs can only be merged into `main` when all "required" checks pass in the MQ.
+    #
+    # So we need to always run all checks, but skip PR view jobs in the MQ, and
+    # MQ jobs in the PR view.
+    if: github.event_name == 'pull_request'
```
jorisdral: Same comment about skipped jobs here

.github/workflows/haskell-simple-merge-queue.yml
```diff
@@ -0,0 +1,48 @@
...
+    strategy:
+      matrix:
+        include:
+          - runner: 'macos-latest'
+            ghc:    '9.4.8'
```
jorisdral, line 23: You can use `9.4` if you want

.github/actions/build-manual-macos/action.yml
jorisdral, line 1: Correct me if I'm wrong, but at first glance it seems that the three manual actions are very similar (if not idential) save for the configuration steps. Maybe the actions should only contain the configuration steps, and the steps for building the manual can be inlined in the base workflow?

dschrempf:
> This is incorrect. Our old manual tests use LLVM/Clang 14.

I was referring to the manual not compiling with Clang version 14 installed by `install-llvm-action`, and that _is correct_. We need to use the Clang provided by GHC. Both have the same versions, but different targets, as you pointed out.

EDIT: I have changed the top-level note so it is clearer! Sorry for being a bit blunt and please bear with me, I hope we can get this in soon, I am through many iterations of trying to get Windows to work already ;-).

> FWIW, the added information belongs in the "Installation" section of the manual.

This information was added in a separate PR, I tried to improve the section by removing duplicate content. Yes, let us improve the manual in a separate PR.

> Perhaps it is a configuration issue...

I agree. Especially the paths look strange mixing forward and backward slashes. I am a bit tired of fixing CI by now, maybe we can analyze the failures with newer Clang on Windows in a separate ticket/PR.

jorisdral (commented):
Nice :)

.github/workflows/haskell-simple-base.yml
jorisdral: Does `always` mean that we save the cache *even if* we didn't build any dependencies because an earlier step failed? If so, that could cause a bit of churn

cabal.project.ci
dschrempf: Yes! I was thinking the same. The problem is that we do not want to use all stanzas from the main cabal project file. I will see if we can do some form of selective import (or if importing the complete file is actually OK).

.github/workflows/haskell-simple-base.yml
dschrempf: `always` means that we save the cache, even when a previous job failed. I am used to saving the cache at the end with an `always` clause.

I do not know how this behaves when we actively "cancel" the job (e.g., interactively, or when pushing a new reference).

I can add a note that we save the cache here at the end because it includes build artefacts from the manual. [DONE]

dschrempf: Nice, thanks. I was wondering why you added an id in the previous job step.

.github/actions/build-manual-macos/action.yml
dschrempf: Hm, that's an interesting idea. Previously, they were large workflow files themselves, and we slimmed them down to what we really need to build the manual exclusively.

I will try to do so!

Actually, the base workflow file also contains some "<YourOSHere> quirks" steps. Maybe we can put those into compositve actions...

.github/workflows/haskell-simple-merge-queue.yml
dschrempf: Yes, that does work! Nice!

.github/actions/build-manual-macos/action.yml
dschrempf: I had a look and there are some other subtle changes like forward/backward slashes. I will leave them separate for now, if that is OK.

.github/workflows/haskell-simple-base.yml
dschrempf: Actually, I checked what happens when we "cancel" the job, and "always()" steps are still run. That's definitely not what we want.

So I will change this to the recommended alternative: "if: ${{ !cancelled() }}". Thanks for the detailed review!

dschrempf: This may also affect `haskell-ci`. I will change the relevant line now, but it will be overwritten by `haskell-ci`.

cabal.project.ci
dschrempf: I added a `cabal.project.ci.base` file defining the index state and some other flags we want to be consistent to avoid rebuilds.

.github/actions/build-manual-macos/action.yml
TravisCardwell: > I had a look and there are some other subtle changes like forward/backward slashes. I will leave them separate for now, if that is OK.

I have looked into such inconsistent path separators before, and I found that LLVM/Clang processes paths leniently in that regard, and it always joins with POSIX path separators (`/`).

I think that we should be consistent on our side, using Windows path separators in Windows paths, to avoid/minimize issues with Windows itself.  We resolve headers using Clang, however, so those resolved paths unfortunately mix separators.  There should be no problems as long as we just log such paths, not actually access them.  (Within a MinGW environment, perhaps even accessing them would work?  I am not sure.)

Windows is a fractal of bad design.

TravisCardwell:
> I have changed the top-level note so it is clearer! Sorry for being a bit blunt and please bear with me, I hope we can get this in soon, I am through many iterations of trying to get Windows to work already ;-).

No worries!  I feel your pain.

(I intensely dislike Windows, but) I am happy to take up the Windows issues in a separate PR.

PR #1078, opened 2025-09-11
Detect anonymous structs in unions

### LINKED ISSUE #764: Detect anonymous structs in unions
We have a ticket for _supporting_ this (#682), but currently we don't even detect when they are used, and simply generate incorrect code. We should do something similar as we do for structs.

bolt12:
This closes #764

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
@@ -249,7 +249,7 @@ unionDecl info = simpleFold $ \curr -> do
         -- Local declarations inside unions that are not used by any fields
         -- result in implicit fields. Unfortunately, @libclang@ does not make
         -- these visible <https://github.com/llvm/llvm-project/issues/122257>.
-        -- This matters, because we need the offsets of these implicit fields.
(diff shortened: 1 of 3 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -244,21 +244,39 @@ unionDecl info = simpleFold $ \curr -> do
...
-        -- TODO (#682): Support anonymous structures in unions.
-        -- See 'partitionChildren' in 'structDecl'.
+        -- Separate out nested declarations from regular struct fields
+        --
+        -- Local declarations inside unions that are not used by any fields
+        -- result in implicit fields. Unfortunately, @libclang@ does not make
+        -- these visible <https://github.com/llvm/llvm-project/issues/122257>.
+        -- This matters, because we need the offsets of these implicit fields.
+        -- For now we only try to detect the situation and report an error when
+        -- it happens. Hopefully this is anyway very rare.
```
bolt12:
```suggestion
        -- We could in principle support it but currently we don't. See #682.
        -- For now we only try to detect the situation and report an error when
        -- it happens. Hopefully this is anyway very rare.
```

Issue #1068, opened 2025-09-08
Type unavailable when generating bindings for `wlroots`


dschrempf:
When generating bindings for `wlroots`, I get the following error:

```
PANIC!: the impossible happened
Please report this as a bug at https://github.com/well-typed/hs-bindgen/issues/
type not found: "Wl_notify_func_t"
CallStack (from HasCallStack):
  panicPure, called at src-internal/HsBindgen/Backend/Hs/Translation.hs:172:28 in hs-bindgen-0.1.0-inplace-internal:HsBindgen.Backend.Hs.Translation
```

Relevant `Info`-level log messages:
```
[Info   ] [HsBindgen] [select] 'wl_notify_func_t' (in source at "/nix/store/rk79php7mqv0q83fmb3vhgrkar3z3rbx-wayland-1.24.0-dev/include/wayland-server-core.h:180:16") excluded
[Info   ] [HsBindgen] [select] Selected 'wl_notify_func_t' (in source at "/nix/store/rk79php7mqv0q83fmb3vhgrkar3z3rbx-wayland-1.24.0-dev/include/wayland-server-core.h:180:16")
```

In particular, `wl_notify_func_t` is defined as
```
typedef void (*wl_notify_func_t)(struct wl_listener *listener, void *data);
```
which is a typedef for a complete function signature (Why on our beloved earth would anyone do that; there must be a good reason, I am curious).

Also, why do we _exclude_ __and__ _select_ the bespoken declaration?

edsko:
What does "exclude" even mean?

dschrempf:
I assume this is a rethorical question! I think it refers to "not selected", and the trace is logged when a declaration ends up in the "unmatched declarations batch" (i.e., the batch where the select predicate match failed).

I have no idea how a declaration can be in both batches. But I assume it is some interaction with it being a weird typedef defining the return type of a function, I guess?

dschrempf:
Maybe the typedef is not that weird? I think it means:

`wl_notify_func_t` ~ "pointer to a function with two arguments (a `struct` and a `void` pointer), and return type `void`"

edsko:
No, it wasn't rhetorical. Does it mean, "parse not attempted"? "parse failed"? 
It could end up in one of those two categories if program slicing is enabled I guess?

dschrempf:
I think, in this case, it is unrelated to the parse pass. These messages are emitted in the select pass.

As far as I can tell, this declaration was
- parsed by `libclang`
[...]

PR #1095, opened 2025-09-16
Fix select message duplicates

dschrempf:
Fix part 1 of #1068.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select/IsPass.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select/IsPass.hs
@@ -93,7 +93,7 @@ data SelectConfig = SelectConfig {
 -- | Select trace messages
 data SelectMsg =
     SelectTransitiveDependencyUnavailable C.NsPrelimDeclId
-  | SelectExcluded (C.DeclInfo Select)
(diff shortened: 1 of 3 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select/IsPass.hs
```diff
@@ -93,7 +93,7 @@ data SelectConfig = SelectConfig {
 -- | Select trace messages
 data SelectMsg =
     SelectTransitiveDependencyUnavailable C.NsPrelimDeclId
-  | SelectExcluded (C.DeclInfo Select)
+  | SelectNotSelected (C.DeclInfo Select)
```
dschrempf: Report reason of why something was selected (either by predicate, or because it is a transitive dependency).

dschrempf: If easy, add the reason why it is a transitive case.

dschrempf: Reporting the root declaration causing the transitive dependency is hard, because we unified the resolution of transitive dependencies into one function call (traversing all roots in one go). We had the resolution separate (i.e., per declaration root) in the beginning but decided to optimize the algorithm.

PR #1096, opened 2025-09-16
Fix `preprocess --path-style` option

TravisCardwell:
Options should be listed before arguments in the `--help` output.  I added a comment to help avoid making this mistake again.

This could alternatively be resolved by using `ApplicativeDo` and configuring the order in the `Parser` only, but there is no reason to do this here, as `Opts` is just used for CLI arguments and it is trivial to keep the order consistent.


TravisCardwell:
This change is trivial.  I will go ahead and merge while the our queue is empty.

Issue #1070, opened 2025-09-09
Test compilation of fixtures


dschrempf:
I do think we have a ticket for testing compilation of fixtures, but I could not find it.

_We should compile all generated fixtures in our CI/CD pipeline._

I know, the pipeline is already taking too long. However, I think it is important that we produce working code, and that we do not introduce regressions, that are otherwise invisible to us.

For example, I have a branch https://github.com/well-typed/hs-bindgen/tree/dom/test-fixture-compilation with a  relatively simple [compilation script](https://github.com/well-typed/hs-bindgen/blob/dom/test-fixture-compilation/hs-bindgen/fixtures/compile).

Last update: 2025-10-01.

This script now reports a total of 9 failures (which is a big improvement to earlier times):

```
Compilation of bool_c23.pp.hs failed. (C-phase)
Compilation of decls_in_signature.pp.hs failed. (C-phase)
Compilation of fun_attributes.pp.hs failed. (C-phase)
Compilation of macro_strings.pp.hs failed. (GHC-phase)
Compilation of iterator.pp.hs failed. (C-phase)
Compilation of program_slicing_simple.pp.hs failed. (GHC-phase)
Compilation of redeclaration.pp.hs failed. (GHC-phase)
Compilation of typenames.pp.hs failed. (GHC-phase)
Compilation of reparse.pp.hs failed. (GHC-phase)

```

We can, for example, do this compilation only for one combination of `(CompilerVersion,OS,LLVMVersion)`.

edsko:
Perhaps we should do this as part of #1159.

PR #1106, opened 2025-09-17
Fix associated type indentation (PP)

TravisCardwell:
The pretty-printer for associated types was missing a `nest`, resulting in incorrect indentation when the type name is long.

### DIFF
--- a/hs-bindgen/fixtures/enum_cpp_syntax.pp.hs
+++ b/hs-bindgen/fixtures/enum_cpp_syntax.pp.hs
@@ -44,7 +44,7 @@ instance F.Storable Foo_enum where
 instance HsBindgen.Runtime.CEnum.CEnum Foo_enum where
 
   type CEnumZ Foo_enum =
-  HsBindgen.Runtime.Prelude.Word32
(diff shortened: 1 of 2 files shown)


TravisCardwell:
This is a trivial bug fix.  I shall go ahead and merge.

dschrempf:
Thanks @TravisCardwell! I was thinking, can we maybe add a test for this? Or do you think it is trivial enough to not test the indentation fix?

TravisCardwell:
The `enum_cpp_syntax` test already tests this case.  I doubt it is worth adding a new one.

dschrempf:
Hm, but it didn't catch the bug, did it? (Or we just didn't compile it, and that's why we missed it?)

TravisCardwell:
Correct: we did not catch the bug because we did not compile it.

edsko:
At some point we should compile all golden tests, ideally. Perhaps we are close to that ideal already?

dschrempf:
Relevant issue: https://github.com/well-typed/hs-bindgen/issues/1070

PR #1111, opened 2025-09-19
Move to `hs-bindgen-cli internal frontend`

TravisCardwell:
The name `parse` is potentially confusing.  This command outputs the declarations after all of the `Frontend` passes, not just the `Parse` pass.


TravisCardwell:
This is a tiny change that we discussed yesterday.  I will go ahead and merge.

Issue #1123, opened 2025-09-19
We need to import _data_ constructors of all types in foreign imports


dschrempf:
(This problem only arises when creating multiple modules, one for each binding category).

If we do not import the _data_ constructor of all types involved, we get errors of the form

```
generated/Vector/Length/Safe.hs:21:1: error: [GHC-72317]
    • Unacceptable result type in foreign declaration:
        ‘ghc-internal-9.1202.0:GHC.Internal.Foreign.C.Types.CDouble’ cannot be marshalled in a foreign call
          because its data constructor is not in scope
    • When checking declaration:
        foreign import ccall safe "hs_bindgen_192e69881bd1d6b4" vector_length
          :: Ptr.Ptr Vector.Vector -> IO Len
    Suggested fix: Import the data constructor ‘CDouble’ of ‘CDouble’
   |
21 | foreign import ccall safe "hs_bindgen_192e69881bd1d6b4" vector_length
```

In the particular case I am observing, the error is thrown at a foreign import in `Vector.Length.Safe` and `Vector.Length.Unsafe`
```haskell
foreign import ccall safe "hs_bindgen_192e69881bd1d6b4" vector_length
  :: Ptr.Ptr Vector.Vector
     {- ^ __C declaration:__ @p@
     -}
  -> IO Len
```

We have (in `Vector.Length`)
```haskell
newtype Len = Len
  { un_Len :: FC.CDouble
  }
  deriving stock (Eq, Ord, Read, Show)
  deriving newtype (F.Storable, Enum, Floating, Fractional, Num, Real, RealFloat, RealFrac)
```

`Vector.Length` compiles just fine, because it does import `Foreign.C` as `FC`. The `Safe`/`Unsafe` modules however, only know about `Len` as `TApp (TGlobal IO_type) (TCon "Len")`, so they do not know what `Len` actually _is_. That is, they can not possibly know what to import (at the moment).

At the moment, I am hacking around this issue by importing our `HsBindgen.Runtime.Prelude` which exports exactly these constructors, and which we want to migrate to in the future, anyways (see issue https://github.com/well-typed/hs-bindgen/issues/1061). However, right now this is leading to many "redundant import" errors.

dschrempf:
I forgot to discuss this with you @edsko. Maybe we should briefly have a look at it in one of our next meetings.

dschrempf:
Also probably relevant: Discussion in https://github.com/well-typed/hs-bindgen/pull/957

Issue #1061, opened 2025-09-05
Discrepancy between imports  in TH and preprocessor mode


dschrempf:
In TH mode, we use `HsBindgen.Runtime.Prelude`, in preprocessor mode we use hand-crafted separate imports (`moduleOf`).

The advantage of the hand-crafted import list is that we may not pull in unnecessary dependencies. Actually, we may not even need `hs-bindgen-runtime`.

The drawback of the hand-crafted import list is, for example, that we have to take care of qualifiers (ambiduous ocurrence errors, e.g., https://github.com/well-typed/hs-bindgen/issues/1055), redundant imports (e.g., https://github.com/well-typed/hs-bindgen/issues/1056).

Some more remarks from a chat with Edsko:
- [ ] Bindgen Prelude: Split Global into TypeConstr TermVar TermConstr (remove panics in mkGlobalExpr).
- [ ] Unify mkGlobal in PP and TH mode.
- [ ] Use our own Prelude qualified.
- [ ] Use a specific set of symbols from the Haskell Prelude unqualified (these symbols must be reserved).

dschrempf:
FYI, I am assigning this to myself, and will implement this import strategy which seems much more stable.

dschrempf:
I have had a closer look:

#### Why do have two ways of defining how globals are to be imported?

For preprocessor mode (PP), we have `HsBindgen.Backend.Artefact.HsModule.Names.resolveGlobal`, which uses `moduleOf` (ultimately causing the redundant import issues).

For Template Haskell (TH), we have function `mkGlobal` and `mkGlobalExpr` in `HsBindgen.Backend.TH.Translation`. (Why do we have two for TH, and only one for PP?).

Before I go on implementing the common `Prelude`, we should discuss this!

TravisCardwell:
With Template Haskell, we are able to use the compiler: we use quotes to lookup names of types (`''`) and values (`'`) within the current scope.  The imports must be outside of the Template Haskell splice, after all.  There are a number of functions used to traverse different parts of the AST.

We are not able to do that with the preprocessor.  We generate a new module at runtime, so we cannot query the compiler.  We have to generate the imports.

Trying to change these to do the same thing is a bad idea.

dschrempf:
@edsko Pulling you in here since we discussed this. How should we move on? I fixed the redundant import in #1066, but there will be more issues like these.

Maybe using fully qualified names throughout (like @TravisCardwell suggested in #1064) could fix this issue?

[...]

PR #1125, opened 2025-09-22
Import CAPI via HsBindgen Prelude (fixes redundant import warnings)

dschrempf:
Fixes https://github.com/well-typed/hs-bindgen/issues/1123, and removes the redundant import warnings.

I think we should prioritize #1061. A common Prelude that we consistently use would make things easier.


dschrempf:
Putting this on the queue since it is a small bug fix, that also removes report redundancy warnings.

Issue #1127, opened 2025-09-23
Function wrappers do not have haddock documentation


bolt12:
@TravisCardwell said:

> I noticed that the foreign import functions are also included in the rendered Haddock documentation, and these are the ones annotated with Haddock documentation.
>
> In a minimal example on main, `foreign import ccall unsafe ... print_point_wrapper` has Haddock documentation, while `print_point` does not.  Function pointer `foreign import ccall unsafe ... hs_bindgen_ex_d87e49118b66a303` has Haddock documentation, while `print_point_ptr` does not.  Note that in this case, even the hash is displayed in the rendered Haddock documentation.
>
> Perhaps we can improve on this situation?
>
> <img width="286" height="431" alt="Image" src="https://github.com/user-attachments/assets/60476608-ecd0-41f0-bdb3-721643b19cfc" />
>
> <img width="474" height="378" alt="Image" src="https://github.com/user-attachments/assets/ebb4af27-99e8-4981-a609-dc5201012d1a" />

TravisCardwell:
Here are the aspects of the above behavior that could be considered unfortunate:

* Two functions are exported for every function in the API.
    * Hopefully users will recognize that they only use the second of each pair of functions.  Perhaps we should mention this in documentation?
    * This significantly clutters rendered Haddock documentation, especially in cases where there is detailed Doxygen documentation.
    * This should be resolved when using the preprocessor, using an export list.  (#76)
    * We do not write the export list for TH mode.  TH users need to maintain their own export list (or live with exporting everything).
    * One idea is to omit all documentation for the foreign import functions.  This would minimize repetition in the generated documentation as well as make it pretty clear to users which functions they should use.

* Internal function names containing both the unique ID and hash are exposed.
    * Perhaps we should name the Haskell functions differently?

* Documentation is displayed for the foreign import functions but not the public API functions.
    * This one should be fixed by #1128.

edsko:
Regarding @TravisCardwell 's three points [above](https://github.com/well-typed/hs-bindgen/issues/1127#issuecomment-3325604224)

1.  Once we make sure that only public-facing functions have haddocks, let's use

```
{-# OPTIONS_HADDOCK prune #-}

```

to avoid including non-public functions in the export list. @bolt12 That does mean _all_ public functions (and _only_ public functions, see point 3) should have haddocks, but I think that is already the case?

[...]

PR #1128, opened 2025-09-23
Fix #1104 and #1127

### LINKED ISSUE #1104: Forward slashes in haddocks attached to bindings are interpreted as markup
The following comment, which is part of generated bindings:

```hs
{-| __C declaration:__ @BOTAN_FFI_ERROR@

    __defined at:__ @botan-3/botan/ffi.h:114:6@

    __exported by:__ @botan-3/botan/ffi.h@
-}
newtype BOTAN_FFI_ERROR = BOTAN_FFI_ERROR
  { un_BOTAN_FFI_ERROR :: FC.CInt
  }
  deriving stock (Eq, Ord)
```

Renders as:

Image

It seems that the forward slashes are interpreted as markup. This is corroborated by `haddock`'s documentation:

https://haskell-haddock.readthedocs.io/latest/markup.html#code-blocks

We should make sure that forward slashes are rendered as forward slashes instead of markup.

bolt12:


TravisCardwell:
Here are the updated screenshots:

example-point-unsafe

example-point-funptr

This PR adds the Haddock documentation to the public API functions.  Note, however, that it does *not* add the parameter documentation.  That should probably be fixed.

bolt12:
Thank you @TravisCardwell , I'll fix that!

bolt12:
@TravisCardwell the problem is that we can only add parameter documentation to FFI functions, the public API function generation does not have any infrastructure in place, currently to do this.

See:

```hs
type Decl :: Star
data Decl where
    DeclData            :: SNatI n => Struct n -> Decl
    DeclEmpty           :: EmptyData -> Decl
    DeclNewtype         :: Newtype -> Decl
    DeclPatSyn          :: PatSyn -> Decl
    DeclDefineInstance  :: DefineInstance -> Decl
    DeclDeriveInstance  :: DeriveInstance -> Decl
    DeclForeignImport   :: ForeignImportDecl -> Decl
    DeclVar             :: VarDecl -> Decl
    DeclUnionGetter     :: UnionGetter -> Decl
    DeclUnionSetter     :: UnionSetter -> Decl
    DeclSimple          :: SHs.SDecl -> Decl
```

There's no `DeclFunction`, the functions are generated directly with `DeclSimple` and `DVar`. This should be improved. Not sure if I want/should address this in this PR. 

I can open a separate issue. What do you think (cc @edsko ) ?

TravisCardwell:
I see!  Thanks!

We should discuss the three possible issues described [here](https://github.com/well-typed/hs-bindgen/issues/1127#issuecomment-3325604224) with @edsko and decide what our goal is.  Indeed, a separate PR is likely warranted.

edsko:
Let's make sure that documentation is _only_ attached to public functions, not to internal wrappers.

bolt12:
I have added a new issue, regarding the function parameters: https://github.com/well-typed/hs-bindgen/issues/1143 . 

I will address the `{-# OPTIONS_HADDOCK prune #-}` and only add haddocks to public functions in this PR.

bolt12:
Could I get an extra round of reviews, please? cc @jorisdral @TravisCardwell @edsko

edsko:
This is not yet correct, is it? I'm still seeing documentation attached to internal wrapper functions?

bolt12:
Maybe I misunderstood your comment here: https://github.com/well-typed/hs-bindgen/issues/1127#issuecomment-3332213450 I thought it was okay to leave it for wrappers?

bolt12:
Or perhaps you were looking at the fixtures? I forgot to update those

bolt12:
I will currently leave documentation for "function_wrapper" because those generate function parameter arguments. At least until #1143 is done

TravisCardwell:
By only putting the currently generated documentation on the functions like `print_point`, those functions naturally attract attention.  This is good since they are the functions that correspond to the C API.

The wrapper functions may be useful for performance optimization.  We should give them concise documentation so that they do not distract attention.  How about something like the following?

```haskell
-- | Wrapper for 'print_point' that takes 'C.Ptr's
```

bolt12:
I understand that but for now we can't have function argument documentation if not on the wrappers. I think it is unfortunate to remove them until we have fixed the situation for functions

bolt12:
This is fully addressed in #1196

PR #1136, opened 2025-09-24
Take into account `const` function attribute in function address stubs

### LINKED ISSUE #1032: Take into account `const` function attributes for function address stubs
For a C function like:

```c
int square (int x) __attribute__ ((const));
```

We are generating a pure function because the `const` attribute tells us that the C function is Haskell pure.

```hs
{-| __from C:__ @square@ -}
foreign import ccall safe "hs_bindgen_test_fun_attributes_8effe939268709e4" square
  :: FC.CInt
  -> FC.CInt
```

Since #1008, we are generating a binding for the address of the function as well

```hs
foreign import ccall unsafe "hs_bindgen_test_fun_attributes_631c7b52d4d4fe3a" hs_bindgen_test_fun_attributes_631c7b52d4d4fe3a
  :: IO (F.FunPtr (FC.CInt -> IO FC.CInt))

{-# NOINLINE square_ptr #-}

square_ptr :: F.FunPtr (FC.CInt -> IO FC.CInt)
square_ptr =
  GHC.IO.Unsafe.unsafePerformIO hs_bindgen_test_fun_attributes_631c7b52d4d4fe3a
```

But these bindings include `IO` in the return type inside the function pointer. We should remove this `IO` if possible

jorisdral:
Resolves #1032


hs-bindgen-runtime/src/HsBindgen/Runtime/FunPtr.hs
jorisdral, line 1: We could also leave out this module entirely and tell the user to use `castFunPtr` instead of this safe version. Feels a bit like `castFunPtrSafe` is over-engineered

edsko:
After discussing this with @jorisdral , we realized that actually a `FunPtr (Int -> Int)` is not useful: C functions will always want a `FunPtr (Int -> IO Int)` (there is no way to declare that an argument must be pure). We therefore decided _not_ to merge this PR, and leave the `IO` in the types of function addresses.

Issue #1135, opened 2025-09-24
Change types we generate for typedefs around function pointers


edsko:
See analogy with structs:

# Non-pointer case

## Structs

```c
typedef struct {
  int x;
  int y;
} point;
```

```hs
data Point {
    x :: Int
  , y :: Int
  }
```

## Functions

```c
typedef int (DataValidator)(int value);
```

```hs
newtype DataValidator = DataValidator (Int -> IO Int)

instance ToFunPtr DataValidator where
  toFunPtr :: DataValidator -> IO (FunPtr DataValidator)
```

# Pointer-case

## Structs

```c
typedef struct {
  int x;
  int y;
} *point;
```

```hs
data Point_Deref {
    x :: Int
  , y :: Int
  }

newtype Point = Point (Ptr Point_Deref)
```

## Functions

```c
typedef void (*ProgressUpdate)(int percentComplete);
```

```hs
newtype ProgressUpdate_Deref = ProgressUpdate_Deref (Int -> IO ())

newtype ProgressUpdate = ProgressUpdate (FunPtr ProgressUpdate_Deref)

instance ToFunPtr ProgressUpdate_Deref where
  toFunPtr :: ProgressUpdate_Deref -> IO (FunPtr ProgressUpdate_Deref)
```

### Without implicit conversion (after #1033)

```c
void onProgressChanged(ProgressUpdate update);
```

```hs
onProgressChanged :: ProgressUpdate -> IO ()
```

(Caution: this _looks_ like implicit conversion, superficially, but is not.)

### With implicit conversion (after #1033 and #1035)

```c
// implicit function to function pointer conversion
int validateInput(DataValidator validator, int rawValue);
```

```hs
validateInput :: FunPtr DataValidator -> Int -> IO Int
```

Issue #1035, opened 2025-08-29
Support implicit function pointer conversion


jorisdral:
In most contexts, a function (type) is implicitly converted to a function pointer (type). For example, a function can either be declared to take a function pointer argument or just a function argument, though they are equivalent because of the implicit conversion. `hs-bindgen` handles a similar
implicit conversion for arrays, but not yet for functions.

See https://en.cppreference.com/w/c/language/conversion.html#Function_to_pointer_conversion for info on function-to-pointer conversion

=== Example

```c
extern int apply_pointer (int (*f)(int), int x);
extern int apply_nopointer (int f(int), int x);
```

are equivalent but the foreign imports are different, and the foreign import for the latter has an additional bug where it is missing some parentheses in the type

```hs
foreign import ccall safe "hs_bindgen_5553e6f9246adc97" apply_pointer
  :: F.FunPtr (FC.CInt -> IO FC.CInt)
  -> FC.CInt
  -> IO FC.CInt

foreign import ccall safe "hs_bindgen_d00aec9ae0c7c9a1" apply_nopointer
  :: FC.CInt -> IO FC.CInt
  -> FC.CInt
  -> IO FC.CInt
```

edsko:
Not marking this for release-0.1 after all because I think these non-pointer functions are actually rare (I'm surprised they're allowed at all). If I am wrong about that, let me know and we should bump the priority.

edsko:
Marking this as a bug because we're generating invalid syntax.

edsko:
```c
typedef int int2int(int);

extern int apply_pointer (int (*f)(int), int x);
extern int apply_nopointer (int f(int), int x);

extern int apply_pointer_td (int2int* f, int x);
extern int apply_nopointer_td (int2int f, int x);
```

Option 1:

```hs
newtype Int2Int = Int2Int { getInt2Int :: Int -> IO Int }

apply_pointer :: FunPtr (Int -> IO Int) -> Int -> IO Int
apply_nopointer :: FunPtr (Int -> IO Int) -> Int -> IO Int

apply_pointer_td :: FunPtr Int2Int -> Int -> IO Int
apply_nopointer_td :: FunPtr Int2Int -> Int -> IO Int
```

Option 2:

```hs
newtype Int2Int = Int2Int { getInt2Int :: FunPtr (Int -> IO Int)  } -- note the FunPtr

apply_pointer :: FunPtr (Int -> IO Int) -> Int -> IO Int
[...]

Issue #1033, opened 2025-08-29
Support function argument types that are typedefs of function types


jorisdral:
If we have this header file:

```c
typedef int int2int(int);
extern int apply (int2int * f, int x);
```

Then currently we would generate bindings along the lines of:

```hs
newtype Int2int = Int2int
  { un_Int2int :: FC.CInt -> IO FC.CInt
  }
foreign import ccall safe "hs_bindgen_e89baabc0ea659ed" apply
  :: F.Ptr Int2int
  -> FC.CInt
  -> IO FC.CInt
```

The foreign import for `apply` should be using a `FunPtr Int2Int` instead of a `Ptr Int2Int`. Given that we know the definition of the typedef `int2int`, it should be possible to unwrap the typedef to check whether it wraps a function type, and then generate a `FunPtr` or `Ptr` accordingly.

In some cases, we can not unwrap the `int2int` typedef, for example when the typedef came from a binding spec. In such cases, we can currently not do better than generate a `Ptr Int2Int`.

edsko:
Let's do this after #1077 is merged.

jorisdral:
See also https://github.com/well-typed/hs-bindgen/issues/1035#issuecomment-3327241522

PR #1139, opened 2025-09-25
Change how newtypes for function ptr are generated

bolt12:
- Addresses #1135
- Reuse `handleDecl` logic that is used for tagged type, i.e. Struct,
  Enum and Unions.


hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -1047,8 +1053,12 @@ typ' ctx = go ctx
...
         Hs.HsPrimType (goPrim p)
     go _ (C.TypePointer t) = case t of
-        C.TypeFun {} -> Hs.HsFunPtr (go CPtrArg t)
-        _            -> Hs.HsPtr (go CPtrArg t)
+        C.TypeTypedef (C.TypedefRegular C.NamePair { nameC })
+          | Just C.TypeFun{} <- Map.lookup nameC typedefs
+                      -> Hs.HsFunPtr (go CPtrArg t)
+          | otherwise -> Hs.HsPtr (go CPtrArg t)
+        C.TypeFun {}  -> Hs.HsFunPtr (go CPtrArg t)
+        _             -> Hs.HsPtr (go CPtrArg t)
```
jorisdral: I think this way of looking through typedefs is fine -- it's also one of the options I considered while working on #1035 and #1033. I might have a slightly different approach that I want to experiment with, but I'm not yet sure if it is going to work, so IMO this is good as is!

edsko: Is that really the case though? This does a _single_ lookup; shouldn't we recursively look through? What if we get another typedef? Or does that not matter here? I feel like some kind of "unfold typedefs" function, perhaps with an explicit argument indicating whether we want to "unfold all the way" or "just one later" would be good.

bolt12: Good point. I looked into the code and indeed this `Map` only strips down 1 layer of typedefs. Maybe we need a better abstraction rather than just a `Map`. I will say this is fine because it should be the most common case, but for sake of soundness, I think some more work should be put into this `Map` 😄

jorisdral: True, I overlooked recursive unfolding. That would still be good to include

bolt12: https://github.com/well-typed/hs-bindgen/issues/1142

edsko: No, the `Map` _itself_ fine, it just gives definitions.  We need Joris' proposed "normalize".

bolt12: where's that proposal?

jorisdral: I don't have a written down proposal just yet

jorisdral: Btw, Maybe this bit of code should be using `getUnderlyingType`:

```suggestion
go _ (C.TypePointer t) = case getUnderlyingType typedefs t of
        C.TypeFun {}  -> Hs.HsFunPtr (go CPtrArg t)
        _             -> Hs.HsPtr (go CPtrArg t)
```

bolt12: oh so there's already a function that does what we want? 😄

hs-bindgen/examples/golden/typedefs.h
```diff
@@ -1,2 +1,5 @@
 typedef int myint;
 typedef int * intptr;
+
+typedef void (*FunctionPointer_Function)(void);
+typedef int (NonFunctionPointer_Function)(int value);
```
jorisdral, line 9: If you add the following example here:
```suggestion

typedef void (*FunctionPointer_Function)(void);
typedef int (NonFunctionPointer_Function)(int value);

//! Pointer-to-function type
typedef void (*f1)(void);

typedef void g1(void);
//! Pointer-to-function type, but with one typedef indirection
typedef g1 * g2;

typedef void h1(void);
typedef h1 h2;
//! Pointer-to-function type, but with two typedef indirections
typedef h2 * h3;
```
Then you can see that `getUnderlyingType` works as expected

hs-bindgen/fixtures/distilled_lib_1.pp.hs
```diff
@@ -495,14 +495,24 @@ pattern ENUM_CASE_2 = A_typedef_enum_e 2
 pattern ENUM_CASE_3 :: A_typedef_enum_e
 pattern ENUM_CASE_3 = A_typedef_enum_e 3
 
+{-| __C declaration:__ @callback_t_Deref@
```
edsko: This isn't right, I think?

hs-bindgen/fixtures/doxygen_docs.pp.hs
```diff
@@ -174,6 +174,26 @@ pattern COLOR_GREEN = Color_enum 1
 pattern COLOR_BLUE :: Color_enum
 pattern COLOR_BLUE = Color_enum 2
 
+{-|
+
+  Callback function type
```
edsko: We probably don't want to copy all of these haddocks from `Event_callback_t`?

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -1046,9 +1052,13 @@ typ' ctx = go ctx
         Hs.HsPrimType (goVoid c)
     go _ (C.TypePrim p) =
         Hs.HsPrimType (goPrim p)
-    go _ (C.TypePointer t) = case t of
-        C.TypeFun {} -> Hs.HsFunPtr (go CPtrArg t)
-        _            -> Hs.HsPtr (go CPtrArg t)
+    go _ (C.TypePointer t) = case getUnderlyingType typedefs t of
+        C.TypeTypedef (C.TypedefRegular C.NamePair { nameC })
```
edsko: This should not have a `TypeTypedef` left after `getUnderlyingType`?

hs-bindgen/fixtures/distilled_lib_1.pp.hs
```diff
@@ -495,14 +495,18 @@ pattern ENUM_CASE_2 = A_typedef_enum_e 2
 pattern ENUM_CASE_3 :: A_typedef_enum_e
 pattern ENUM_CASE_3 = A_typedef_enum_e 3
 
+newtype Callback_t_Deref = Callback_t_Deref
+  { un_Callback_t_Deref :: (Ptr.Ptr Void) -> HsBindgen.Runtime.Prelude.Word32 -> IO HsBindgen.Runtime.Prelude.Word32
+  }
```
bolt12, line 501: add a comment saying something like:

> Auxiliary type used by 'Callback_t'

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/HandleTypedefs.hs
```diff
@@ -41,23 +44,52 @@ handleTypedefs C.TranslationUnit{..} = (
...
+          let derefDecl, mainDecl :: C.Decl HandleTypedefs
+              derefDecl = C.Decl {
+                  declInfo = declInfo' { C.declId = C.DeclId (curName <> "_Deref") (C.NameOriginGenerated (C.AnonId declLoc)) }
+                , declKind = handleUseSites td
+                           $ C.DeclTypedef $ C.Typedef {
+                               typedefType = C.TypeFun args res
+                             , typedefAnn  = NoAnn
+                             }
+                , declAnn  = defaultTypeSpec
+                }
```
bolt12, line 72: Add the comment about the auxiliary type here

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Haddock/Translation.hs
```diff
@@ -64,20 +64,25 @@ generateHaddocksWithParams ::
...
-     , Hs.commentLocation   = Just (updateSingleLoc pathStyle declLoc)
-     , Hs.commentHeaderInfo = mHeaderInfo
-     , Hs.commentChildren   = []
-     }
-  , map addFunctionParameterComment params)
+generateHaddocksWithParams HaddockConfig{..} declLoc mHeaderInfo declId Nothing params
+    -- If it is a dereferenced newtype we should not include any comment to it
+    --
+  | Text.isSuffixOf "Deref" (getName (nameC declId)) =
+    (Nothing, map addFunctionParameterComment params)
```
bolt12: We should look at the Origin from `declId` instead. If it's generated do something special.

jorisdral:
> Feel free to merge once the final changes we discussed are in.
> 
> This will clash with some of the work @jorisdral is doing (unwrapping typedefs will work differently) but it shouldn't be too difficult to resolve (in either direction) I think. 

I don't mind resolving merge conflicts if they come up, so feel free to merge once the PR is ready

PR #1140, opened 2025-09-25
Support implicit function to pointer conversion

### LINKED ISSUE #1035: Support implicit function pointer conversion

jorisdral:
Resolves #1035

#1035 mentions also supporting function to pointer conversion in the presence of typedefs, but I'm going to implement that as part of #1033


edsko (approved):
Feel free to merge this once this is done in the parse step rather than the translation step.

jorisdral:
I'm going to merge this PR by bypassing the merge queue because GitHub is having problems with their MacOS runners. I did run all the other merge queue checks before merging by hand, see https://github.com/well-typed/hs-bindgen/actions/runs/18165434332

Issue #1050, opened 2025-09-04
Should we panic on unbound typedefs?


dschrempf:
Right now, when we get the underlying type of a typedef, we panic if the typedef is not available (i.e., unbound). This happens in the function `getUnderlyingType` in `Translation.hs`.

However, there could be situations where this is actually the case. Consider the minimal example

```c
typedef int foo;
extern const foo x;
```

If we _do not select_ `foo`, the typedef is unavailable. At the moment, we do not emit the panic because `hs-bindgen` already fails earlier reporting:
```
$ cabal run -- hs-bindgen-cli preprocess -I ./ c.h --select-by-decl-name x
...
[Error  ] [HsBindgen] [mangle-names] Missing declaration: 'foo'; did you select the declaration?
An error happend (see above)
...
```

However, we should still make a conscious decision if we want to keep this panic. There is also a corresponding TODO note in the code.

jorisdral:
While in the `hs-bindgen` backend we might not have information available about all typedefs in a header, `clang` probably does have a lot of information that we could use. For example, we could use `clang_getCanonicalType` to erase typedefs. And maybe `clang_getTypeDeclaration` can help us construct mappings of typedefs to types

* [`clang_getTypeDeclaration`](https://clang.llvm.org/doxygen/group__CINDEX__TYPES.html#ga0aad74ea93a2f5dea58fd6fc0db8aad4)
* [`clang_getCanonicalType`](https://clang.llvm.org/doxygen/group__CINDEX__TYPES.html#ga0aad74ea93a2f5dea58fd6fc0db8aad4)

dschrempf:
I think both of these links pint to `clang_getTypeDeclaration`;
[clang_getCanonicalType](https://clang.llvm.org/doxygen/group__CINDEX__TYPES.html#gaa9815d77adc6823c58be0a0e32010f8c) sounds very interesting though!

AFAIK, we only want to use these function in the parse pass. Do you think we should annotate all typedefs with their underlying type already in that pass?

jorisdral:
I'm on the fence about how best to use these functions. I've tried annotating typedefs with their canonical type, but this wasn't as straightforward as I had hoped when passing these typedefs through all the frontend passes. Also, `getCanonicalType` removes all sugar for us. It might be more useful if we can control how much sugar and which sugar (typedefs, const qualifiers) to remove.  It's sometimes hard to recover the connection between a type and its canonical version after the fact.

[...]

PR #1145, opened 2025-09-25
Jdral/erase typedefs

### LINKED ISSUE #1142: Come up with a better abstraction for looking through Typedef inner types
Is that really the case though? This does a _single_ lookup; shouldn't we recursively look through? What if we get another typedef? Or does that not matter here? I feel like some kind of "unfold typedefs" function, perhaps with an explicit argument indicating whether we want to "unfold all the way" or "just one later" would be good.

_Originally posted by @edsko in https://github.com/well-typed/hs-bindgen/pull/1139#discussion_r2378457030_

jorisdral:
Resolves #1142. #1050 is related

This PR builds on top of #1135, so that one should be merged first.

This PR is an experiment to come up with a more principle approach to erasing `typedef`s from C types where necessary. One such example occurs when we want to check if a type is a function type so that we can decide whether to use `Ptr` or `FunPtr` in the Haskell code for the bindings. To do that reliably we should erase `typedefs` and `const` qualifiers.

Most of the effort in this PR is related to adding some type safety when inspecting C types. Meaning, if you want to check that a type is a *canonical* function type, one would first map the full C type to a canonical type and then check if it is a function type. There is currently nothing preventing us from inspecting the full C type when we should be inspecting the corresponding canonical C type instead, but the types should hopefully make it clearer that we should do the mapping before the inspecting.

There are two commits in this PR, each with a separate attempt at implementing this. Attempt 1 was with higher-kinded datatypes, which I didn't like so much, and it is not so complete as Attempt 2. Attempt 2 uses a Trees That Grow style, though it's more like a Trees That Shrink style.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
@@ -904,22 +904,6 @@ typedefDecs opts haddockConfig typedefs info typedef spec = do
           , clss `Set.member` insts
           ]
 
--- | Recursively unwrap the underlying type of a typedef.
---
--- TODO https://github.com/well-typed/hs-bindgen/issues/1050: Should we panic on
--- unbound typedefs?
-getUnderlyingType :: Map C.Name C.Type -> C.Type -> C.Type
-getUnderlyingType typedefs = go
-  where
-    go (C.TypeTypedef ref) = case ref of
-      C.TypedefRegular n ->
-        let err = panicPure $ "Unbound typedef " ++ show n
-            t'  = Map.findWithDefault err (C.nameC n) typedefs
-        in  go t'
-      C.TypedefSquashed _ t' ->
-        go t'
-    go t = t
-
 {-------------------------------------------------------------------------------
   Macros
 -------------------------------------------------------------------------------}
@@ -1036,9 +1020,9 @@ typ' :: HasCallStack => TypeContext -> Map C.Name C.Type -> C.Type -> Hs.HsType
 typ' ctx typedefs = go ctx
   where
     go :: TypeContext -> C.Type -> Hs.HsType
-    go _ (C.TypeTypedef (C.TypedefRegular name)) =
+    go _ (C.TypeTypedef (C.Full (C.TypedefRegular name))) =
         Hs.HsTypRef (C.nameHs name)
-    go c (C.TypeTypedef (C.TypedefSquashed _name ty)) =
+    go c (C.TypeTypedef (C.Full (C.TypedefSquashed _name ty))) =
         go c ty
     go _ (C.TypeStruct name _origin) =
         Hs.HsTypRef (C.nameHs name)
@@ -1052,13 +1036,9 @@ typ' ctx typedefs = go ctx
         Hs.HsPrimType (goVoid c)
     go _ (C.TypePrim p) =
         Hs.HsPrimType (goPrim p)
-    go _ (C.TypePointer t) = case t of
-        C.TypeTypedef (C.TypedefRegular C.NamePair { nameC })
-          | Just C.TypeFun{} <- Map.lookup nameC typedefs
-                      -> Hs.HsFunPtr (go CPtrArg t)
-          | otherwise -> Hs.HsPtr (go CPtrArg t)
-        C.TypeFun {}  -> Hs.HsFunPtr (go CPtrArg t)
-        _             -> Hs.HsPtr (go CPtrArg t)
+    go _ (C.TypePointer t) = case C.eraseTypedefsRec typedefs t of
+        C.TypeFun{} -> Hs.HsFunPtr (go CPtrArg t)
+        _ -> Hs.HsPtr (go CPtrArg t)
     go _ (C.TypeConstArray n ty) =
         Hs.HsConstArray n $ go CTop ty
     go _ (C.TypeIncompleteArray ty) =
@@ -1335,13 +1315,13 @@ functionDecs safety opts haddockConfig moduleName typedefs info f _spec =
     wrapType ty = go ty
       where
         go = \case
-          C.TypeStruct {}             -> HeapType ty
-          C.TypeUnion {}              -> HeapType ty
-          C.TypeComplex {}            -> HeapType ty
-          (C.TypeConstArray n ty')    -> CAType   ty n ty'
-          (C.TypeIncompleteArray ty') -> AType    ty ty'
-          (C.TypeTypedef _)           -> go $ getUnderlyingType typedefs ty
-          _                           -> WrapType ty
+          C.TypeStruct {}              -> HeapType ty
+          C.TypeUnion {}               -> HeapType ty
+          C.TypeComplex {}             -> HeapType ty
+          (C.TypeConstArray n ty')     -> CAType   ty n ty'
+          (C.TypeIncompleteArray ty')  -> AType    ty ty'
+          (C.TypeTypedef (C.Full ref)) -> go $ C.eraseTypedef typedefs ref
+          _                            -> WrapType ty
 
     resType :: ResultType HsType
     resType =
@@ -1471,13 +1451,14 @@ global ::
   -> C.DeclSpec
   -> [Hs.Decl]
 global opts haddockConfig moduleName instsMap typedefs info ty _spec =
-  let underlyingType = case ty of
-        C.TypeConstArray _ ty' -> ty'
-        otherType -> getUnderlyingType typedefs otherType
-  in case underlyingType of
-    -- Generate getter if the underlying type is @const@.
-    C.TypeConst _ -> stubDecs ++ getConstGetterOfType ty
-    _             -> stubDecs
+    case C.eraseTypedefsRec typedefs ty of
+      -- Generate getter if the erased type is @const@-qualified.
+      C.TypeConst _ -> stubDecs ++ getConstGetterOfType ty
+      -- Generate getter if the erased type of array elements is @const-qualified@
+      C.TypeConstArray _ (C.TypeConst _) -> stubDecs ++ getConstGetterOfType ty
+      C.TypeIncompleteArray (C.TypeConst _) -> stubDecs ++ getConstGetterOfType ty
+      -- Otherwise, do not generate a getter
+      _ -> stubDecs
   where
     -- *** Stub ***
     stubDecs :: [Hs.Decl]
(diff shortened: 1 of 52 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
```diff
@@ -307,6 +332,132 @@ isVoid :: Type -> Bool
...
+instance GetCanonicalType (Map C.Name Type) Type where
+  getCanonicalType env = go
+    where
+      go ty = case ty of
+        TypePrim pt -> TypePrim pt
+        TypeStruct np no -> TypeStruct np no
+        TypeUnion np no -> TypeUnion np no
+        TypeEnum np no -> TypeEnum np no
+        TypeTypedef ref -> go (eraseTypedef env ref)
+        TypeMacroTypedef np no -> TypeEnum np no
```
jorisdral: Wow! typo

hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
```diff
@@ -307,6 +332,132 @@ isVoid :: Type -> Bool
...
+instance GetErasedType (Map C.Name Type) Type where
+  getErasedType env = go
+    where
+      go ty = case ty of
+        TypePrim pt -> TypePrim pt
+        TypeStruct np no -> TypeStruct np no
+        TypeUnion np no -> TypeUnion np no
+        TypeEnum np no -> TypeEnum np no
+        TypeTypedef ref -> go (eraseTypedef env ref)
+        TypeMacroTypedef np no -> TypeEnum np no
```
jorisdral: Wow! typo

bolt12:
#1139 merged :)

hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
```diff
@@ -265,19 +278,22 @@ data CheckedMacroType = CheckedMacroType {
...
 
--- | Type use
+-- | C types
 --
 -- For type /declarations/ see 'Decl'.
-data Type =
+type Type = FullType
+
+-- | C types in Trees That Shrink style
+data TypeF p =
```
jorisdral: Replace `p` by `tag` to avoid confusion with this not being a `Pass`, but a different kind of type-level tag

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/IsPass.hs
```diff
@@ -79,6 +81,38 @@ emptyParseDeclMeta = ParseDeclMeta {
...
+-- In Haskell, the type of @x@ will be the name of the referenced typedef @foo@.
+-- The type of @x@ will /also/ record the underlying type of @foo@. The Haskell
+-- type roughly looks like:
+--
+-- > 'RegularTypedef' "foo" "int"
+--
+-- The underlying type is included so that it can be inspected locally. For
+-- example, in the backend we inspect whether a typedef wraps a function type.
+-- Having the underlying type available there makes it possible to check this.
+data TypedefRefRegular p = TypedefRefRegular
```
jorisdral: Rename this to `OrigTypedefRef`. This name makes more sense, since this is how the typedef reference appears in C code prior to renaming/squashing

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/IsPass.hs
```diff
@@ -79,6 +81,38 @@ emptyParseDeclMeta = ParseDeclMeta {
...
+-- to the typedef @foo@
+--
+-- > typedef foo int;
+-- > foo x;
+--
+-- In Haskell, the type of @x@ will be the name of the referenced typedef @foo@.
+-- The type of @x@ will /also/ record the underlying type of @foo@. The Haskell
+-- type roughly looks like:
+--
+-- > 'RegularTypedef' "foo" "int"
```
jorisdral: Extend this example with an additional, intermediate typedef. This should emphasise that the underlying type does not erase all typedefs, just the one layer.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ResolveBindingSpecs.hs
```diff
@@ -378,8 +379,11 @@ instance Resolve C.Type where
         auxU C.TypeEnum         uid (C.NameKindTagged C.TagKindEnum)
       C.TypeMacroTypedef uid ->
         auxU C.TypeMacroTypedef uid C.NameKindOrdinary
-      C.TypeTypedef      nm  ->
-        auxN C.TypeTypedef      nm  C.NameKindOrdinary
+      C.TypeTypedef      (TypedefRefRegular nm uTy) -> do
+        (uTyDepIds, uTy') <- resolve uTy
+        let mkTypeTypedef = C.TypeTypedef . flip TypedefRefRegular uTy'
+        (tdDepIds, td) <- auxN mkTypeTypedef nm  C.NameKindOrdinary
+        pure (uTyDepIds `Set.union` tdDepIds, td)
```
jorisdral:
```suggestion
      C.TypeTypedef      (TypedefRefRegular nm uTy) -> do
        (_uTyDepIds, uTy') <- resolve uTy
        let mkTypeTypedef = C.TypeTypedef . flip TypedefRefRegular uTy'
        (tdDepIds, td) <- auxN mkTypeTypedef nm  C.NameKindOrdinary
        pure (tdDepIds, td)
```

jorisdral: @TravisCardwell does it make sense that we do not include the dependencies of the *underlying type* in the result?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames.hs
```diff
@@ -475,8 +475,10 @@ instance Mangle C.Type where
       C.TypeComplex prim   -> return $ C.TypeComplex prim
 
 instance Mangle RenamedTypedefRef where
-  mangle (TypedefRegular C.DeclId{..}) = TypedefRegular <$>
-    mangleQualName (C.QualName declIdName C.NameKindOrdinary) declIdOrigin
+  mangle (TypedefRegular C.DeclId{..} uTy) = do
+    uTy' <- mangle uTy
+    flip TypedefRegular uTy' <$>
+      mangleQualName (C.QualName declIdName C.NameKindOrdinary) declIdOrigin
```
jorisdral, line 484: Add a comment that it is slightly dangerous to recurse into the underlying type, though it's currently fine

hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
```diff
@@ -265,19 +278,22 @@ data CheckedMacroType = CheckedMacroType {
...
     TypePrim C.PrimType
   | TypeStruct Int.NamePair C.NameOrigin
   | TypeUnion Int.NamePair C.NameOrigin
   | TypeEnum Int.NamePair C.NameOrigin
-  | TypeTypedef TypedefRef
+  | TypeTypedef (TypedefRefF p)
   | TypeMacroTypedef Int.NamePair C.NameOrigin
```
jorisdral, line 294: Maybe this should have a type annotation as well, so that we can erase macro typedefs as well?

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -1174,17 +1134,22 @@ anyFancy types = any p types where
...
       (C.TypeConstArray n ty')    -> CAType   ty n ty'
       (C.TypeIncompleteArray ty') -> AType    ty ty'
-      (C.TypeTypedef _)           -> go $ getUnderlyingType typedefs ty
+      -- Note: we're only interested in finding the first non-typedef type
+      -- to determine which method of wrapping to use (.e.g, heap type or
+      -- primitive type or array type). As such, we erase typedefs only one
+      -- at a time. If we used 'C.getErasedType' we would erroneously remove
+      -- typedefs in other places as well, such as in the element types of
+      -- arrays. We want to keep typedefs there!
+      (C.TypeTypedef ref)         -> go $ C.eraseTypedef ref
```
jorisdral, line 1152: We should erase macro typedefs here too.

hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
jorisdral: See #1200

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
jorisdral: See #1200

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ResolveBindingSpecs.hs
jorisdral: Some context: references to typedefs are annotated with the type `Type p` underlying the typedef. Applying pass `p'` after `p`, we still have to convert the annotation from `Type p` to `Type p'`. We do this here by calling `resolve` on the annotation, but not returning the dependencies.

TravisCardwell:
The `Set C.QualPrelimDeclId` returned are dependencies of the declaration currently being processed by `resolveDeep`, as represented by edges in the `UseDeclGraph`.

Example:

```c
typedef struct point {
    double x;
    double y;
} point;

struct rect {
    point tl;
    point br;
};
```

```mermaid
graph TD;
  v1["PrelimDeclIdNamed &quot;point&quot;"]
  v0["PrelimDeclIdNamed &quot;point&quot;"]
  v2["PrelimDeclIdNamed &quot;rect&quot;"]
  v1-->|"UsedInTypedef ByValue"|v0
  v2-->|"UsedInField ByValue &quot;br&quot;"|v1
  v2-->|"UsedInField ByValue &quot;tl&quot;"|v1
```

Imagine the `resolveDeep` call for the `struct rect` declaration.  The two fields are `typedef`s, so they are processed by this code.  Why are we recursing on the underlying type?  I do not think doing so is correct, as we are processing the use sites of external bindings in `struct rect`, not type `typedef`s themselves.

Returning the dependencies of the underlying type is indeed not necessary, as there are no edges between `PrelimDeclIdNamed "rect"` and the bottom `PrelimDeclIdNamed "point"`.

Hmmm, that `UseDeclGraph` does not distinguish between `struct point` and `point`.  Perhaps it is an issue in what we are displaying?  Hopefully it is not a bug.

BTW, I implemented the `hs-bindgen-cli info use-decl-graph` command in order to get the above.  I will throw it into a PR really quick so that you can use it as well.

edsko:
Not traversing isn't really an option, unless we do some type trickery. Perhaps we should have a chat about this with the three of us.

jorisdral:
I don't understand the resolve bindings spec pass well enough to understand why it would be incorrect to recurse into the underlying type. Is it harmful? We're mostly piggy-backing off of the recursive call to `resolve` so that we get a `Type ResolveBindingSpecs` from `Type NameAnon`

TravisCardwell:
Ack!  Recursing on the underlying type is indeed needed when processing the `typedef` itself, such as `point` in the above example. :bow:

PR #1152, opened 2025-09-30
Refactor `qualified as Hs` usage, rename to `CommentRef`

TravisCardwell:
I worked on this as part of #799 and would like to get it merged separately so that I can rebase that branch and avoid having to rebase this change more than necessary.  The change is simple, but it touches many lines.  From the commit messages:

**Refactor `qualified as Hs` usage**

Long ago, we did *not* import language types qualified, so we named types like `CName` and `HsName`.  We have since decided to use qualified imports for such types.  The C types have already been refactored: `HsBindgen.Frontend.Naming` defines the naming types, while `HsBindgen.Language.C` defines the primitive types.  Both are imported `qualified as C`.  We did not refactor the Haskell language types yet, however.  This commit implements that change, so that `HsBindgen.Language.Haskell` is imported `qualified as Hs`.

These modules are designed to work alongside the AST modules.  The C type modules and C AST modules are all imported `qualified as C`.  Similarly, `HsBindgen.Language.Haskell` and Haskell AST modules are all imported `qualified as Hs`.

Documentation modules should be imported with different names, because some type names conflict.  Documentation modules have been imported as `C` and `Hs` to distinguish C documentation from Haskell documentation.  Following this lead, they are now imported as `CDoc` and `HsDoc`.  This still distinguishes C documentation from Haskell documentation as well as distinguishes them from core/AST types.  For example, `Hs.Bold` is now `HsDoc.Bold`.

**Rename to `CommentRef`**

The comment reference type was previously called just `Reference`.  This name is too general.  This commit renames the type to `CommentRef` to make it clear that it is a comment reference, not one of the other reference types that we use.  Abbreviation `Ref` is consistent with the other reference types.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/HsModule/Translation.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/HsModule/Translation.hs
@@ -9,6 +9,7 @@ module HsBindgen.Backend.HsModule.Translation (
   , HsModule(..)
     -- * Translation
   , HsModuleOpts(..)
+  , defModuleName
   , translateModuleMultiple
   , translateModuleSingle
   , mergeDecls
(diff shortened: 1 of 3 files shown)


TravisCardwell:
**Move default generated module name**

It was defined in a `Default` instance for `ModuleName` in `HsBindgen.Language.Haskell`.  This module is for general Haskell, and the `ModuleName` type is used for more than just specifying the names of generated modules.  The default belongs in `HsBindgen.Backend.HsModule.Translation`, but a complication is that the `ConfigPP` uses `ModuleName` directly instead of `HsModuleOpts`.  I resolved this by defining and exporting `defModuleName`, which is used in the `Default` instance for both `HsModuleOpts` and `ConfigPP`.

edsko (approved):

PR #1153, opened 2025-09-30
Run GHA jobs on the `main` branch

jorisdral:
And a few smaller changes

### DIFF
--- a/.github/workflows/check-stylish-haskell.yml
+++ b/.github/workflows/check-stylish-haskell.yml
@@ -7,6 +7,10 @@ on:
   pull_request:
   merge_group:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 env:
   # Fix the index state. It rarely has to change unless we want to use a new
   # version of stylish-haskell.
--- a/.github/workflows/check-hlint.yml
+++ b/.github/workflows/check-hlint.yml
@@ -7,6 +7,10 @@ on:
   pull_request:
   merge_group:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   hlint:
     name: HLint
--- a/.github/workflows/check-whitespace.yml
+++ b/.github/workflows/check-whitespace.yml
@@ -7,6 +7,10 @@ on:
   pull_request:
   merge_group:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   whitespace:
     name: Whitespace


.github/workflows/check-stylish-haskell.yml
```diff
@@ -23,35 +28,37 @@ jobs:
...
-        key: ${{ steps.setup-haskell.outputs.ghc-version }}-${{ hashFiles('dist-newstyle/cache/plan.json') }}
-        restore-keys: ${{ steps.setup-haskell.outputs.ghc-version }}-
+        key: ${{ env.key }}-plan-${{ hashFiles('dist-newstyle/cache/plan.json') }}
+        restore-keys: ${{ env.key }}-
 
+    # Ignores the project file, since installing stylish-haskell does not have
+    # to use the same configuration as the rest of the hs-bindgen project.
     - name: Install stylish-haskell
       run: |
-        cabal update
```
jorisdral, line 41: `cabal update` is performed by `haskell-actions/setup@v2`

.github/workflows/check-stylish-haskell.yml
```diff
@@ -23,35 +28,37 @@ jobs:
...
 
     - name: Save cache
       uses: actions/cache/save@v4
       # Even save the cache, even when a previous job failed. Do not save the
       # cache when the job gets canceled.
-      if: ${{ !cancelled() }}
+      if: ${{ steps.cache-cabal-restore.outputs.cache-hit != 'true' }}
       with:
         path: ${{ steps.setup-haskell.outputs.cabal-store }}
         key: ${{ steps.cache-cabal-restore.outputs.cache-primary-key }}
```
jorisdral, line 56: If the "Install stylish-haskell" step succeeded, and the cache we restored did not exactly match the primary key, then saving the cache is not necessary. The cabal store did not change

.github/actions/base/action.yml
```diff
@@ -84,10 +84,12 @@ runs:
     - name: Restore cache
       id: cache-cabal-restore
       uses: actions/cache/restore@v4
+      env:
+        key: build-and-test-${{ inputs.runner }}-ghc-${{ steps.setup-haskell.outputs.ghc-version }}-cabal-${{ steps.setup-haskell.outputs.cabal-version }}-llvm-${{ inputs.llvm}}
```
jorisdral, line 88: Cabal version updates can also cause rebuilds, so I included it in the cache name

.github/workflows/check-stylish-haskell.yml
dschrempf: Can we use a positive predicate (i.e., `== 'false')? Or is this not what we ant here?

Also, the cache should _not_ be saved when the action is cancelled.

.github/workflows/check-stylish-haskell.yml
```diff
@@ -22,35 +32,30 @@ jobs:
...
-        chmod +x scripts/ci/run-stylish-haskell.sh
-        scripts/ci/run-stylish-haskell.sh -c
+        cabal install stylish-haskell --ignore-project --index-state=${{ env.hackage-index-state }}
 
     - name: Save cache
       uses: actions/cache/save@v4
-      # Even save the cache, even when a previous job failed. Do not save the
-      # cache when the job gets canceled.
-      if: ${{ !cancelled() }}
+      if: ${{ steps.cache-cabal-restore.outputs.cache-hit != 'true' }}
```
dschrempf: Also here (see above).

edsko (approved):
So glad somebody understands this stuff who is not me :grimacing:

.github/workflows/check-stylish-haskell.yml
jorisdral: > Can we use a positive predicate (i.e., `== 'false')? Or is this not what we ant here?

Ah yes, makes sense, I'll change it

> Also, the cache should _not_ be saved when the action is cancelled.

The ordering of steps is now
* restore cache
* install stylish-haskell
* save cache
* check formatting

The "save cache" step is only performed if all the previous steps succeeded. Cancelling before we get to the "save cache" step would mean the save cache step does not happen. I put "save cache" before "check formatting" because the "check formatting" step will not build any additional dependencies. So if we cancel while the "check formatting" step is happening, then we will have still saved the cache, but that is okay because the contents of the cache are correct. The next rerun of the job can then restore the cache we just saved. Part of the motivation for developing separate `cache/restore` and `cache/save` actions was so that caches can be saved even if later steps (that do not affect the cache) fail

jorisdral: Wait.. Maybe I'm wrong about the `cancelled` stuff. Investigating now

jorisdral: Right, so if I'm reading https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#status-check-functions correctly, every step has the `succeed()` status check automatically applied. This means that every step is cancelled unless you explicitly use one of the status check functions in the conditional after `if:`.

This is an example where I cancelled during the install step, and the save step was skipped, which is what we want:
https://github.com/well-typed/hs-bindgen/actions/runs/18125238102/job/51578852967

dschrempf: Thanks.

PR #1154, opened 2025-09-30
Actually run GHA jobs on the `main` branch

jorisdral:
I forgot to update some conditionals in #1153 :see_no_evil:


jorisdral:
I added this PR to the merge queue since I think it's uncontroversial

PR #1156, opened 2025-09-30
Select predicate to only select "available" declarations

### LINKED ISSUE #1062: Selection predicate "exclude declarations with attribute deprecated"
This ticket originated from translating `libpcap`, which has some deprecated declarations that are _not provided by the compiled shared library provided by my OS (NixOS)_.

With `pcap`, I had to manually exclude those declarations from binding generation. This is tedious.

Edsko came up with a nice idea: Provide a selection predicate that allows to exclude declarations labelled with the attribute "deprecated" (or for that matter, any attribute)?

dschrempf:
In particular, the predicate deselects deprecated declarations.

Closes #1062.

Issues:
- [ ] We use `CXAvailabilityKind` (which is a low-level Clang type) in `hs-bindgen`. Do we want to translate this to our own type? I am not even sure, because it encodes exactly the information we need.
- [ ] We add `CXAvailabilityKind` to the `ParseMsgKey`, which is, at least coneptionally, a bit weird. Since we actually need to match against the availability it makes sense to leave it as a key, but we could also move it to the value-part of the map.
- [ ] Command line option.


edsko (approved):
Feel free to merge once the changes we discussed are in.

PR #1185, opened 2025-10-04
Calculation of instance maps with circular dependencies

### LINKED ISSUE #1068: Type unavailable when generating bindings for `wlroots`

dschrempf:
Closes #1068:

```
struct forward_declaration;

typedef void (*fun_ptr)(struct forward_declaration *s);

struct forward_declaration {
  fun_ptr f;
};
```

Fails to translate because of a circular dependency. However, this circular dependency can be resolved by dropping "by-reference" (`ByRef`) edges in the use-decl graph. This solution fixes the underlying problem and is more elegant than the workaround I had implemented (see below).

### Workaround using retries

Low and behold, I have fixed this error using a wrapper that retries failed translations of declarations. Note the circular dependency:

When translating `C.Decl`s in `Translation.Hs`, we now retry to translate failed declarations.

TODO:
- [x] Add a test case for the above example.
- [x] I think retries warrant trace messages. Will this be our first true safe backend trace message? -> Since we have so many pressing things before release 0.1, I postpone this and generate a separte issue: https://github.com/well-typed/hs-bindgen/issues/1186

EDIT: Sitting on this branch, I can generate bindings for `wlroots` `backend.h`. Isn't this exciting :-)? (I didn't try to compile them though :-P.

EDIT: If a declaration fails to translate in an earlier stage it will also be sorted to succeed declarations that were successfully translated at earlier stages. That is, the order of declarations may change. Is this an issue? The advantage of doing so is that we don't have to traverse the complete list of declarations at each stage and look for translations in a failed state. (Of course we could think of a better algorithm, but simplicity usually wins).


dschrempf:
This is the use-decl graph of the example provided in the PR description:

image

dschrempf:
We also tried another example
```c
struct a;

struct b {
  struct a *toA;
};

struct a {
  struct b toB;
};
```

with use-decl graph
```mermaid
graph TD;
  v1["PrelimDeclIdNamed &quot;a&quot;"]
  v0["PrelimDeclIdNamed &quot;b&quot;"]
  v0-->|"UsedInField ByRef &quot;toA&quot;"|v1
  v1-->|"UsedInField ByValue &quot;toB&quot;"|v0
```

edsko (changes requested):
We discussed this, and will opt for a different solution: the bug is in sorting, true circular dependencies cannot exist.

dschrempf:
Fixed, should be ready for review. @edsko

edsko (approved):

PR #1189, opened 2025-10-07
Refactor binding specifications

TravisCardwell:
Binding specifications are going through a major refactoring.  There are still design issues that need to be resolved (see #799), but this PR contains various commits that work toward our goal.

## Refactor `AOmittable`

In preparation for stricter parsing, `AOmittable` now has *two* type arguments.  This makes it possible to strictly parse a subset of fields of an omitted object.

## Use `Default`

Binding specifications were implemented before we started using `Default`.  `defaultTypeSpec` is replaces with a `Default` instance.

## Check source paths when checking C types

When checking that we used all types listed in a prescriptive binding specification, the main headers that expose the type (determined using the source path) are now taken into account.

## Use `hsname` instead of `identifier` (in YAML only)

The Haskell identifier is now specified using `hsname` in YAML/JSON binding specification files.  Note that the type used is still `Hs.Identifier`.  (Type `Hs.Name` has a phantom namespace, which we cannot parse from YAML/JSON.)

## Rename pass to `ResolveBindingSpecs`

This emphasizes that it resolves both external and prescriptive binding specifications.

Depending on what we decide about making binding specifications module-specific, we may also need to process many separate binding specifications.

## Refactor C AST to use `DeclOpaque` constructor

Instead of separate `DeclStructOpaque`, `DeclUnionOpaque` and `DeclEnumOpaque`, there is a single `DeclOpaque C.NameKind`.  The name kind is required to determine the correct spelling for the declaration.

This is preparation for being able to specify type declarations as opaque in prescriptive binding specifications.

## Reimplement `getMainHeadersAndInclude`

This was implemented using `DynGraph.findTargets`, which I implemented to get exactly what is needed but was a bit out-of-place because it is not a generally-useful graph function.  I replaced it with `DynGraph.findEdges`, which is a bit more general.  It finds edges from the specified starting vertex and edges to terminal vertices in paths from that starting vertex.  Taking advantage of the structure of the graph, where all paths terminate at the root header, it can get the same information with a much simpler implementation.

## Implement omitted types in binding spec generation

We already tracked omitted types in `ResolveBindingSpecs`, in a `Set C.QualName`.  This commit changes the type to `Map C.QualName SourcePath`, because the source path is needed to determine the headers.  `resolveBindingSpecs` returns this map along with the translation unit and trace messages.

We already implemented a `GetMainHeadersAndInclude` function, which is used to get header information for Haddock documentation.  This commit adds `GetMainHeaders`, which just gets the main headers, implemented using the existing function.

For now, both the omitted type map and the `GetMainHeaders` function are added as `Frontend` artifacts.  After we move instance resolution into the frontend, we will be able to refactor this so that binding specification generation is done within the frontend, which would allow us to instead provide a generated binding specification artifact.

Binding specification generation is refactored to use `GetMainHeaders` to determine headers for all types, including the omitted types.  A list of main headers is no longer passed in.

Golden test `binding_spec_simple` is added, using a prescriptive binding specification.  A new `testPrescriptiveBindingSpec` field is added to `TestCase` to support this.

## Order type classes lexicographically

Type classes were output in (arbitrary) definition order in generated binding specifications.


TravisCardwell:
(Deleted)  Wait a minute here.  The headers are set using the include graph...  I am investigating.

```
cabal run hs-bindgen-cli -- info include-graph -I manual/c game_world.h game_player.h
```

```mermaid
graph TD;
  v2["manual/c/game_internal.h"]
  v3["manual/c/game_player.h"]
  v0["manual/c/game_world.h"]
  v0-->|"#include &quot;game_internal.h&quot;"|v2
  v3-->|"#include &quot;game_internal.h&quot;"|v2
```

Found it!  I had a stupid mistake in my `findEdges` code: I forgot to take into account the ST masking.  Fixed!

Issue #1178, opened 2025-10-02
Select pass panics when transitive dependencies are declarations that we did not attempt to parse


dschrempf:
In particular, when generating bindings for `libpcap`, we have to provide `--parse-all`, otherwise, we get the following panic:

```
PANIC!: the impossible happened
Please report this as a bug at https://github.com/well-typed/hs-bindgen/issues/
Unavailable transitive dependencies:
'sockaddr' (tag)
'timeval' (tag)
'u_char' (ordinary)
'u_int' (ordinary)
'u_short' (ordinary)
This is an indication that declarations have been removed after parse.
Please remove unsupported declarations in the parse pass, and not later!

CallStack (from HasCallStack):
  panicPure, called at src-internal/HsBindgen/Frontend/Pass/Select.hs:68:13 in hs-bindgen-0.1.0-4wqc19Y8vzYLYKQTF1RxiI-internal:HsBindgen.Frontend.Pass.Select
```

- [ ] Ensure quality of the error messages: "A not selected; did you select A?" "A not parsed; did you parse A?"

dschrempf:
Prerequsite: https://github.com/well-typed/hs-bindgen/issues/1028

Issue #1038, opened 2025-08-31
Probably replace `SelectDeclMeta` with `DeclMeta`


dschrempf:
After PR https://github.com/well-typed/hs-bindgen/pull/1013 we have three "decl meta" types: `ParseDeclMeta`, `DeclMeta p`, and `SelectDeclMeta`.

`SelectDeclMeta` is just `DeclMeta p` without parse messages. We wanted to merge these two types, and drag along the parse messages until the last pass. However, we can not mangle names because many missing declarations are reported (the parse messages tend to refer to failed parses, so these are expected, but hard to fix).

For now, I left `SelectDeclMeta` in place, but we may want to still merge the two types mentioned above.

dschrempf:
Related to #1028

edsko:
For reference, we currently (df479f314d70660a417dc561987f71d02d1c6120) have

```hs
-- constructed in "Pass.Parse" 
data ParseDeclMeta = ParseDeclMeta {
      parseDeclNonParsed :: NonParsedDecls
    , parseDeclParseMsg  :: ParseMsgs Parse
    }

-- constructed in "Parse.Sort", extends 'ParseDeclMeta'
data DeclMeta p = DeclMeta {
      declIndex     :: DeclIndex
    , declUseDecl   :: UseDeclGraph
    , declDeclUse   :: DeclUseGraph
    , declNonParsed :: NonParsedDecls
    , declParseMsgs :: ParseMsgs p
    }

-- constructed in "Pass.Select", drops some fields from 'DeclMeta'
data SelectDeclMeta = SelectDeclMeta {
      selectDeclIndex     :: DeclIndex
    , selectDeclUseDecl   :: UseDeclGraph
    , selectDeclDeclUse   :: DeclUseGraph
    , selectDeclNonParsed :: NonParsedDecls
    }
```

I think `ParseDeclMeta` will look a bit different after #1028, where it will become `ParseStatus` (@dschrempf correct me if I'm wrong here). I think that `ParseStatus` can simply be included as a field of `DeclMeta`, and then we do not need to drop it from `SelectDeclMeta`.

@dschrempf , I think perhaps we can simply remove the `p` parameter from `ParseDeclMeta`/`ParseStatus`, essentially fixing `p = Parse`; in other words, no need to apply name mangling.

Issue #1037, opened 2025-08-31
Provide feedback to users when selecting declarations


dschrempf:
~~This is an intermediate step between https://github.com/well-typed/hs-bindgen/pull/1013 and https://github.com/well-typed/hs-bindgen/issues/1028.~~ (Things have changed since then).

In the `Select` pass, we report delayed parse messages. After PR #1013, we report traces for declarations that we have failed to parse.

We also want to report a trace when we select a declaration we did not parse. (This is epecially useful when program slicing is enabled). We may also want to distinguish if we _failed to parse_, or if we _did not attempt to parse_ the missng declarations.

dschrempf:
On Friday we discussed implementing this next. However, after pondering on this ticket for some time, I think we should first decide and implement #1028. The reason is that even for skipped but selected messages, we need to resolve binding specs etc. So we can not directly look through the non-parsed declarations.

I think we have to discuss the three of us. @TravisCardwell @edsko

edsko:
The purpose of this ticket is to ensure that when users enable program slicing, but some transitive dependency is not available to bindgen because it was not parsed, that we can give the user a more informative error message explaining what has gone wrong and what they can do to address it.

dschrempf:
See TODO items in Select.hs. These are leftovers after #1192.

### Solution
1. Start from a list of unavailable declarations.
2. Traver decl-use graph to use sites.
3. Remove declarations that have unmet dependencies from list of handled declarations.
4. Emit trace messages for all declarations that we can not handle due to missing dependencies.

### Golden test!

```c
struct Foo {
  long double x;
  struct Baz y;
};

struct Bar {
  struct Foo *x;
};

struct Baz {
  int x;
};
```
- We can not generate bindings for `Bar`, because we fail to parse `Foo`.
- We are not aware that `Foo` depends on `Baz`, but that is not a problem, because we can not generate bindings for `Foo` anyways.
- We can generate bindings for `Baz`.

dschrempf:
Stuff that PR #1240 will not cover:

- Test of select messages for OK case (`selection_ok.h`). (Same variants as with 
`selection_fail.h`).

```c
// selection_ok.h
struct B {
  int i;
};

struct A {
  struct B *b;
};
```

- Quirks with selection anonymous declarations: #1273.

- Determine if all transitive dependencies are also selected. If a transitive dependency is not available, suggest changing the select predicate or adding external bindings specification(s).

dschrempf:
[...]

PR #1192, opened 2025-10-07
Store parse status of all declarations

dschrempf:
We get a CI error: https://github.com/well-typed/hs-bindgen/actions/runs/18343147007/job/52243182474?pr=1192

> anonymous declaration without tag kind: (unnamed at /usr/include/x86_64-linux-gnu/bits/cmathcalls.h:130:1)

I can not reproduce the error locally, I think because I am using a newer version of LLVM.

Will close
- #1028 
- ~~#1037~~ 
- #1038 
- #1178

Finally, this PR is ready for discussion:
- [x] CI failure (not reproducible locally, LLVM version?): Conversion to `QualPrelimDeclId` fails. -> Moved to separate issue #1220.
- [x] Similarly: Panic due to anonymous declaration without use site in `Select` pass.
- [x] Do we want to check for `unavailableTransitiveDeps`?
- [x] `DeclIndex`: Assembly is complicated.
- [x] `DeclIndex`: For failures, we do not overwrite existing, previously inserted key-value pairs.
- [x] Binding specifications: Should we also lookup failed parses?
- [x] Binding specifications: We still emit (Error?) traces when replacing declarations.

EDIT: This PR does not yet solve #1037, there is more work to do (we have to find declarations that we fail to generate bindings for because transitive dependencies are unavailable).

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Naming.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Naming.hs
@@ -57,6 +57,7 @@ module HsBindgen.Frontend.Naming (
     -- ** QualPrelimDeclId
   , QualPrelimDeclId(..)
   , qualPrelimDeclId
+  , qualPrelimDeclIdSafe
 
     -- * NameOrigin
   , NameOrigin(..)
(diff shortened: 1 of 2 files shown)


dschrempf:
This PR will be parked for a while, see https://github.com/well-typed/hs-bindgen/issues/1028#issuecomment-3385091801.

dschrempf:
In agreement with Edsko, I am putting this on the queue. Two issues are still unresolved:
- #1220 
- #1037

Issue #1201, opened 2025-10-09
Unify "category" selection between TH and CLI


edsko:

dschrempf:
Related to
- https://github.com/well-typed/hs-bindgen/issues/1262, and
- https://github.com/well-typed/hs-bindgen/issues/1263

dschrempf:
This ticket is about fine-grained selection of which binding category to create bindings for. At the moment, we only allow the specification of the `Safety`; instead, we want to specify something like

```haskell
newtype ByCategory a = ByCategory { unByCategory :: Map BindingCategory a }
  deriving newtype (Functor, Foldable, Show)

type Predicate = ???

type SelectByCategory = ByCategory Predicate
```

dschrempf:
I am pondering on this ticket. Some observations:

I suggest performing the "category selection" (not to be confused with the `Select` pass) as late as possible (right before translating the module to a string-like object).

Hence, the types we want are something like

```haskell
-- At the latest stage we have `SDecl`s.
type Predicate = SDecl -> Maybe SDecl
type SelectByCategory = ByCategory Predicate

-- `SDecl` may be too complicated, or we may not want to expose it, then we
-- could use names (I guess `HsBindgen.Language.Haskell.Name`). By the way, is
-- there a function `f :: SDecl -> Name`?
type SimplePredicate = Name -> Bool
type SimpleSelectByCategory = ByCategory SimplePredicate
```

Then, we would apply the `SelectByCategory` or `SimpleSelectByCategory` before translating the module/modules to string-like form. For consistency, I suggest always doing this (i.e., when translating to a single module, and when translating to multiple modules).

The defaults will be:
- When translating to multiple modules, the predicate will always be `Just decl` (`True`).
- When translating to a single module, the predicate will be `Just decl` (`True`) for all categories but `Unsafe`.

The predicate will be user-configurable. That is, we add the predicate to  `HsBindgen.Config.Config`, and the `BackendConfig`.

This means, that
- the artifact `FinalDecls` still contains _all_ declarations, but
- `FinalModule` _will not_ contain all declarations anymore, and
- `FinalModules` _may not_ contain all declarations anymore.

The above makes me think that we have to __distinguish between the single module and multiple module predicates__ (i.e., the predicates may be different, even for a single run).

PR #1182, opened 2025-10-03
Introduces c-minisat bindings example

bolt12:

PR #1203, opened 2025-10-09
Introduces c-rogueutil bindings example

### LINKED ISSUE #1168: Create bindings for `rogueutil`

bolt12:
This commit was split from #1182 due to the fact that the header file for `rogueutil` defines the function implementation on the header file without using `static` or `extern` which leads to duplicate symbol errors.

Right now a hacky workaround is employed by adding `shared: False` to the `cabal.project` file but in truth this can be fixed after #1201 is addressed.

Closes #1168

### DIFF
--- a/examples/c-rogueutil/hs-project/app/Main.hs
+++ b/examples/c-rogueutil/hs-project/app/Main.hs
@@ -5,9 +5,9 @@ module Main where
 import Control.Monad (forM_)
 import Foreign.C (newCString, withCString)
 
-import RogueUtil.Generated qualified as RU
-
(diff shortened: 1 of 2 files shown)


dschrempf:
FYI, the #1201 was fixed a while ago. I think this would be a nice test and use-case!

dschrempf:
Command line options are not yet available, see https://github.com/well-typed/hs-bindgen/issues/1328, and https://github.com/well-typed/hs-bindgen/issues/1449.

Binding category choice should work fine when using Template Haskell mode.

Issue #1046, opened 2025-09-02
Allow generation of bindings specs in TH mode, and adapt configuration types (Config, ConfigPP)


dschrempf:
Let us think about how we treat the _generation_ of binding specifications in Template Haskell mode. Is this even a use case? If so, do we want TH to write external files, and what should the user interface be?

TravisCardwell:
My initial impression was that one needs to use the preprocessor for projects that generate multiple modules with interdependence via binding specifications, while the Template Haskell API is just appropriate for simple projects.  I get the feeling that the sentiment has changed, however.  Perhaps the TH API will make it easy to get started quickly, while the preprocessor will be used by power users.

If we do not mind if TH writes files, it is possible to add support to the API.  In the configuration for a generated module, users would need to specify the path of the output binding specification in addition to the paths of all external binding specifications, perhaps all relative to the package root.  The imports should naturally make the order work out...

A related topic is test generation.  Should the TH API be able to create/update generated tests?

edsko:
Having discussed this with @dschrempf , the preprocessor currently (df479f314d70660a417dc561987f71d02d1c6120) uses

```hs
data Opts = Opts {
      config            :: Config
    , configPP          :: ConfigPP
    , hsOutputDir       :: FilePath
    , outputBindingSpec :: Maybe FilePath
    , inputs            :: [UncheckedHashIncludeArg]
    -- NOTE inputs (arguments) must be last, options must go before it
    }
```

We will move `outputBindingSpec` to `Config`, thereby making it available in TH as well, and we will move `hsOutputDir` and `inputs` to `ConfigPP`.

edsko:
Created a separate ticket to track test generation in TH mode at https://github.com/well-typed/hs-bindgen/issues/1170 .

dschrempf:
I started working on this ticket.

At the moment, I am moving the configuration options around, as requested above. The problem is that `ConfigPP` is used in almost all our client commands (not only in `preprocess`).

Adding configuration options to `ConfigPP` would also expose them in these other client commands although they do not use them (and, in turn, require other configuration options not in `ConfigPP`). I think the best solution is to have different `ConfigXY` types for the different client commands. However, such a proliferation would also mean that we need more code...

dschrempf:
I think `ConfigPP` is a misnomer (at least at the moment).

Given:
`TH` = Configuration options used by Template Haskell
`PP` = Configuration options used by preprocessor client command
[...]

PR #1207, opened 2025-10-10
Move module from `BackendConfig` to `BootArtefact`

TravisCardwell:
We need the module name in the frontend as well as the backend.  Since the configuration should specify "how" and not "what," this commit removes it from the configuration altogether.  It is instead passed as an argument, like the input headers, and stored in the boot artifact.


TravisCardwell:
Issue #1046 discusses moving input headers and output directory configuration into the configuration (stopping the "how not what" organization principle).  Note that the module is a special case, however, because users should *not* be able to configure the module when using the Template Haskell API.

TravisCardwell:
(Rebased)

PR #1208, opened 2025-10-10
Add `libpcap` example project

### LINKED ISSUE #1169: Create bindings for `libpcap`

dschrempf:
I tried to stay as close to the `minisat` example of @bolt12. However, there are some changes:

- Establish convention to use module names `Generated.<lib>`.
- I didn't use a Git submodule, because I couldn't get the development version of `libpcap` to work.
- We should only refer to the upstream `libclang` reference _once_. I split the `cabal.project.base` file into two (one defining upstream references, and `cabal.project.base.dev` defining development flags).
- I originally thought that `minisat-c-bindings` needed to be downloaded. (My `git pull` didn't pick up the git sub-module, and I think others won't perform a deep clone by default either). So I mention this in the `minisat` readme file.
- I added sub-shells (to avoid repetitive change of the current working directory).
- I added `--project-dir="${PROJECT_ROOT}"` to the `cabal-install` invocation. I think this is more intuitive.
- I fixed the header include. (It should not be a path but a header file as used in `#include` statements).
- I fixed the `LD_LIBRARY_PATH` (the `$` was escaped, and the path was not expanded).

Other issues I noticed but have _not_ fixed:
- `minisat-c-bindings` generates bindings against `minisat-c`. We install `minisat-c-bindings` from GitHub, but `minisat-c` using the system repositories. There might be a version mismatch that we do not see nor control. I tried avoiding that by not installing `libpcap` using system repositories, but only via the external code. (Of course, this is easier for `libpcap`, because it is not binding itself to another library that must be installed...)

Closes #1169

### DIFF
--- a/.github/actions/base/action.yml
+++ b/.github/actions/base/action.yml
@@ -137,6 +137,10 @@ runs:
       if: runner.os == 'Linux'
       uses: ./.github/actions/examples/c-minisat
 
+    - name: Build libpcap example (Ubuntu)
+      if: runner.os == 'Linux'
+      uses: ./.github/actions/examples/libpcap
+
     # Save the cache after building the project AND the manual.
     - name: Save cache
       uses: actions/cache/save@v4
--- a/.github/actions/examples/libpcap/action.yml
+++ b/.github/actions/examples/libpcap/action.yml
@@ -1,0 +1,14 @@
+name: libpcap example
+description: Build and run the libpcap example
+
+runs:
+  using: composite
+  steps:
+    - name: Install dependencies
+      shell: bash
+      run: sudo apt-get update && sudo apt-get install -y cmake flex bison
+
+    - name: Run generation and build script
+      shell: bash
+      working-directory: ./examples/libpcap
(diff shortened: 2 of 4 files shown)


examples/c-minisat/README.md
```diff
@@ -6,7 +6,14 @@ open-source SAT solver.
 
 ## Prerequisites
 
-You need the MiniSat library installed on your system. For example, on NixOS:
+Ensure you have download the MiniSat source code:
+
+```bash
+git submodule init
+git submodule update
+```
```
bolt12: Ah good point, I missed that!

examples/c-minisat/generate-and-run.sh
bolt12, line 1: I think we probably want to add the 

```
git submodule init
git submodule update
```

to the script too no?

if you are already changing this could you do it (if you agree?)

then we can simplify CI

dschrempf:
I am putting this on the queue since it only adds an example project, and does not change `hs-bindgen`-related code.

PR #1209, opened 2025-10-10
Refactor FunPtr module to expose the TH instances

bolt12:


hs-bindgen-runtime/src/HsBindgen/Runtime/FunPtr.hs
```diff
@@ -1,44 +1,18 @@
...
+-- | Function pointer utilities and type classes with pre-generated instances.
+--
+-- This module provides the 'ToFunPtr' and 'FromFunPtr' type classes along with
+-- Template Haskell generated instances for common function signatures.
 --
--- This module provides a type class 'ToFunPtr' that allows for a uniform
--- interface to convert Haskell functions to C function pointers.
+-- Users should import this module to get both the type classes and the instances.
 --
 module HsBindgen.Runtime.FunPtr (
```
jorisdral, line 8: Orphan instances exposed by a library can get kind of hairy, so I wonder if we should always export the orphan instances from places where we export the `ToFunPtr` class. If so, it means that even though the instances are orphans internally, they are non-orphan instances for any users of the `hs-bindgen-runtime` library. Moreover, do we foresee use cases where the user really does not want to have access to the default instances?

hs-bindgen-runtime/src/HsBindgen/Runtime/TH/Types.hs
jorisdral, line 1: I feel like it would be nice if we could expose `generateInstance` to the user as well, so they can use that directly instead of writing their own instances by hand

edsko: I don't mind this, but if we want to do that, we'd have to think more carefully about the API design. Let's keep it internal for now and reconsider later. 

Mind, we also don't really expect most users to be writing instances here anyway; most instances should be generated by hs-bindgen, or come standard in the runtime.

hs-bindgen-runtime/src/HsBindgen/Runtime/FunPtr.hs
edsko: I agree, but I _think_ that's precisely what this PR attempts to achieve?

hs-bindgen-runtime/hs-bindgen-runtime.cabal
```diff
@@ -53,6 +53,7 @@ library
       HsBindgen.Runtime.ConstantArray
       HsBindgen.Runtime.FlexibleArrayMember
       HsBindgen.Runtime.FunPtr
+      HsBindgen.Runtime.FunPtr.Class
```
bolt12: This module should not be exposed. Once #1061 is addressed this (i.e. the resolution mechanism picking a not exported module) should no longer be a problem

hs-bindgen-runtime/src/HsBindgen/Runtime/FunPtr.hs
jorisdral: The version of the PR that I reviewed exposed the `FunPtr.Class` module separately, which meant you could import the class without the instances. That's not longer the case now.

hs-bindgen-runtime/src/HsBindgen/Runtime/TH/Types.hs
jorisdral: > I don't mind this, but if we want to do that, we'd have to think more carefully about the API design. Let's keep it internal for now and reconsider later.

:+1: 

> Mind, we also don't really expect most users to be writing instances here anyway; most instances should be generated by hs-bindgen, or come standard in the runtime.

The runtime instances probably don't cover all use cases because if we tried that we'd get combinatorial explosion. I suppose `hs-bindgen` itself could generate instances by scanning all the types in the AST to see which instances are probably necessary

PR #1210, opened 2025-10-10
Add `hs-bindgen-cli info use-decl-graph`

TravisCardwell:


TravisCardwell:
(Rebased)

TravisCardwell:
Note that this currently just outputs the full `use-decl` graph.  We will likely want to add options that enable users to filter what is output.

edsko:
@dschrempf ^^^ We've been wanting this :)

dschrempf:
Yes! The reason was that the full use-decl graph was not available, and `hs-bindgen` panicked. If I remember correctly, we resorted to only manually printing the declarations names only.

edsko:
Oh right! I had forgotten about that. @TravisCardwell we wanted to be able to show the use-decl graph, but just with names, not declarations. But I tried to lookup the definition, and I'm now confused:

https://github.com/well-typed/hs-bindgen/blob/35ed16eaa0fd9be07b657846cb37b8efcc1a44d4/hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/UseDeclGraph.hs#L163-L173

Why do we get the ID from the use-decl graph, then look it up in the decl index, only to then just show only the ID....? Has rthi changed recently? If we're just showing the ID anyway, why do we need the declindex at all here?

TravisCardwell:
This PR just added the CLI for the existing implementation; I did not change the implementation.  When we update the command to support filtering, we may also want to add options for what to display per node. :+1:

PR #1211, opened 2025-10-10
Only generate high-level wrappers for const-qualified arrays 

### LINKED ISSUE #1202: Incorrect bindings generated for non-const arrays
Consider a function such as [qrcodegen_encodeText](https://github.com/nayuki/QR-Code-generator/blob/2c9044de6b049ca25cb3cd1649ed7e27aa055138/c/qrcodegen.h#L187):

```c
bool qrcodegen_encodeText(const char *text, uint8_t tempBuffer[], uint8_t qrcode[],
	enum qrcodegen_Ecc ecl, int minVersion, int maxVersion, enum qrcodegen_Mask mask, bool boostEcl);
```

We currently generate

```hs
foreign import ccall safe "hs_bindgen_a3a47bae1577c790" qrcodegen_encodeText_wrapper ::
     Ptr.Ptr FC.CChar
  -> Ptr.Ptr HsBindgen.Runtime.Prelude.Word8
  -> Ptr.Ptr HsBindgen.Runtime.Prelude.Word8
  -> Qrcodegen_Ecc
  -> FC.CInt
  -> FC.CInt
  -> Qrcodegen_Mask
  -> FC.CBool
  -> IO FC.CBool

qrcodegen_encodeText :: 
   Ptr.Ptr FC.CChar) 
-> IncompleteArray Word8 
-> IncompleteArray Word8 
-> Qrcodegen_Ecc 
-> FC.CInt 
-> FC.CInt 
-> Qrcodegen_Mask 
-> FC.CBool 
-> IO FC.CBool
```

This is wrong: `qrcodegen_encodeText` _writes_ into these buffers; the wrapper is unusable. Using the `IncompleteArray` type here is incorrect, we should just a pointer; using `IncompleteArray` would only be valid if the elements of the array are `const`.

I haven't looked, but it's likely that the same problem also applies to arrays of known size.

jorisdral:
Resolves #1202.

See the commit messages for more information


hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
```diff
@@ -287,7 +287,7 @@ data TypeF tag =
   | TypeStruct Int.NamePair C.NameOrigin
   | TypeUnion Int.NamePair C.NameOrigin
   | TypeEnum Int.NamePair C.NameOrigin
-  | TypeTypedef (TypedefRefF tag)
+  | TypeTypedef !(TypedefRefF tag)
```
jorisdral: Add a comment about why the `!` is here

hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
```diff
@@ -313,7 +313,7 @@ data TypeF tag =
     -- See <https://en.cppreference.com/w/c/language/array#Arrays_of_unknown_size>
   | TypeIncompleteArray (TypeF tag)
   | TypeBlock (TypeF tag)
-  | TypeQualified (TypeQualifierF tag) (TypeF tag)
+  | TypeQualified !(TypeQualifierF tag) (TypeF tag)
```
jorisdral: Add a comment about why the `!` is here

hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
```diff
@@ -363,6 +369,61 @@ isCanonicalTypeFunction ty = case getCanonicalType ty of
...
+    -- | Array of known size
+    CAType
+      -- | Array size
+      Natural
+      -- | Array element type
+      t
+    -- | Array of unkown size
+  | AType
+      -- | Array element type
+      t
```
jorisdral: Rename to `ArrayType`

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -1135,25 +1145,25 @@ anyFancy types = any p types where
 
 -- | Types that we cannot directly pass via C FFI.
 wrapType ::  C.Type -> WrappedType
```
jorisdral, line 1161: Open up a new PR to rename/refactor wrap/unwrap. Maybe we should add type utilities in a separate module as well.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
jorisdral, line 1: Should the `Hs` step be split into separate passes?

hs-bindgen/src-internal/HsBindgen/Frontend/AST/External.hs
jorisdral: Renamed to `ArrayClassification` instead

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
jorisdral: Created new issue #1214

PR #1212, opened 2025-10-14
Parse status, Step 2: Select predicates on use sites of anonymous declarations

dschrempf:
Step 2 of https://github.com/well-typed/hs-bindgen/issues/1028#issuecomment-3385091801.

- [ ] We query the use-decl graph and the decl-index in the `Select` pass for anonymous declarations. We perform the same query in the `NameAnon` pass (but we do not store the required information, so have to query again).
- [x] At the moment, we do not ensure that we break out of infinite dependency loops (anon <- named <- anon <- ...).

The next step is the last one implementing parse status into the declaration index. This last step will also clean up the `Select` pass quite a bit (hopefully).

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
@@ -9,6 +9,10 @@ import Data.Set qualified as Set
 import Clang.HighLevel.Types
 
 import HsBindgen.Errors (panicPure)
+import HsBindgen.Frontend.Analysis.DeclIndex (DeclIndex)
+import HsBindgen.Frontend.Analysis.DeclIndex qualified as DeclIndex
+import HsBindgen.Frontend.Analysis.DeclUseGraph (DeclUseGraph)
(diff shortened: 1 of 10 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
```diff
@@ -98,31 +102,55 @@ selectDecls isMainHeader isInMainHeaderDir SelectConfig{..} unitRBS =
...
+        -- Never select builtins.
+        C.QualPrelimDeclIdBuiltin _ -> NotSelected
+      where
+        declUseGraph :: DeclUseGraph
+        declUseGraph = declDeclUse (C.unitAnn unitRBS)
+
+        index :: DeclIndex
+        index = declIndex (C.unitAnn unitRBS)
+
+        -- TODO_PR: May be infinite loop (anon <- named <- anon <- ...).
```
dschrempf: This is the possible infinite loop. I already had a check in there breaking the loop, but it was a bit uglier, and I wanted to hear other opinions.

dschrempf: Resolved (panic).

PR #1213, opened 2025-10-14
Scan function types to generate FFI stubs and FunPtr instances

### LINKED ISSUE #1179: Scan for function types
As of #1094, we create `ToFunPtr` and `FromFunPtr` instances for typedef'd function types, something like

```c
typedef void (f)(int);
```

or indeed

```c
typedef void (*f)(int);
```

(in the second case we generate instances for `F_Deref`).

We should also scan function arguments (and perhaps struct fields?) for function types that are _not_ typedef'd, like

```c
struct CXCursor;

void f(void (*g)(CXCursor* cursor));
```

and generate `ToFunPtr` and `FromFunPtr` instances for those, _provided_ that those are not orphans (that is, at least one type must be defined in the module that we are currently generating).

bolt12:
Closes #1179

### DIFF
--- a/hs-bindgen/examples/golden/callbacks.h
+++ b/hs-bindgen/examples/golden/callbacks.h
@@ -90,3 +90,6 @@ void processMeasurementWithValidation(
         DataValidator validator
     )
 );
+
+typedef int foo;
+void f(void (*callback)(foo x));
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
@@ -183,8 +183,6 @@ scanAllFunctionPointerTypes =
                 foldMap (scanTypeForFunctionPointers . C.structFieldType) structFields
               C.DeclUnion C.Union {..}   ->
                 foldMap (scanTypeForFunctionPointers . C.unionFieldType) unionFields
-              C.DeclEnum C.Enum {..} ->
-                scanTypeForFunctionPointers enumType
               C.DeclFunction fn ->
                 foldMap scanTypeForFunctionPointers ((C.functionRes fn) : (snd <$> C.functionArgs fn))
               _ -> Set.empty
@@ -207,13 +205,16 @@ scanAllFunctionPointerTypes =
 
 -- | Check if a type is defined in the current module
 isDefinedInCurrentModule :: DeclIndex -> C.Type -> Bool
-isDefinedInCurrentModule declIndex ty = case C.getCanonicalType ty of
+isDefinedInCurrentModule declIndex ty = case ty of
   C.TypeStruct namePair origin                 -> isInDeclIndex namePair origin (C.NameKindTagged C.TagKindStruct)
   C.TypeUnion namePair origin                  -> isInDeclIndex namePair origin (C.NameKindTagged C.TagKindUnion)
   C.TypeEnum namePair origin                   -> isInDeclIndex namePair origin (C.NameKindTagged C.TagKindEnum)
   C.TypePointer (C.TypeStruct namePair origin) -> isInDeclIndex namePair origin (C.NameKindTagged C.TagKindStruct)
   C.TypePointer (C.TypeUnion namePair origin)  -> isInDeclIndex namePair origin (C.NameKindTagged C.TagKindUnion)
   C.TypePointer (C.TypeEnum namePair origin)   -> isInDeclIndex namePair origin (C.NameKindTagged C.TagKindEnum)
+  C.TypeTypedef (C.TypedefRegular namePair _)  -> isInDeclIndex namePair C.NameOriginInSource C.NameKindOrdinary
+  C.TypeTypedef (C.TypedefSquashed name _)     -> isInDeclIndex (C.NamePair name (Hs.Identifier (C.getName name)))
+                                                               C.NameOriginInSource C.NameKindOrdinary
   _                                            -> False
   where
     isInDeclIndex :: C.NamePair -> C.NameOrigin -> C.NameKind -> Bool
--- a/hs-bindgen/fixtures/callbacks.bindingspec.yaml
+++ b/hs-bindgen/fixtures/callbacks.bindingspec.yaml
@@ -126,3 +126,21 @@ types:
   cname: SampleBufferFull_Deref
   module: Example
   hsname: SampleBufferFull_Deref
+- headers: callbacks.h
+  cname: foo
+  module: Example
+  hsname: Foo
+  instances:
+  - Bits
+  - Bounded
+  - Enum
+  - Eq
+  - FiniteBits
+  - Integral
+  - Ix
+  - Num
+  - Ord
+  - Read
+  - Real
+  - Show
+  - Storable
--- a/hs-bindgen/fixtures/callbacks.hs
+++ b/hs-bindgen/fixtures/callbacks.hs
@@ -6114,6 +6114,319 @@
                       (Idx 1)])))},
       defineInstanceComment =
       Nothing},
+  DeclNewtype
+    Newtype {
+      newtypeName = Name
+        "@NsTypeConstr"
+        "Foo",
+      newtypeConstr = Name
+        "@NsConstr"
+        "Foo",
+      newtypeField = Field {
+        fieldName = Name
+          "@NsVar"
+          "un_Foo",
+        fieldType = HsPrimType
+          HsPrimCInt,
+        fieldOrigin = GeneratedField,
+        fieldComment = Nothing},
+      newtypeOrigin = Decl {
+        declInfo = DeclInfo {
+          declLoc = "callbacks.h:94:13",
+          declId = NamePair {
+            nameC = Name "foo",
+            nameHsIdent = Identifier "Foo"},
+          declOrigin = NameOriginInSource,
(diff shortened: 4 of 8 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
jorisdral, line 1: Maybe double-check if you need to convert `Type`s to `ErasedType` or `CanonicalType` when pattern matching

hs-bindgen/examples/golden/callbacks.h
```diff
@@ -24,3 +24,9 @@ void onNewMeasurement2(MeasurementReceived2 handler);
 
 typedef void (*SampleBufferFull)(int samples[10]);
 void onBufferReady(SampleBufferFull handler);
+
+// Function pointers that take other function pointers as arguments
+void transformMeasurement(struct Measurement *data, void (*transformer)(struct Measurement *m, double (*scale)(double, int), int factor));
```
bolt12: We should add another example but with a 3rd order function that has a domain specific type argument

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -1418,6 +1453,75 @@ functionDecs safety opts haddockConfig moduleName info f _spec =
...
+    , Hs.DeclDefineInstance $ Hs.DefineInstance
+      { defineInstanceDeclarations = Hs.InstanceFromFunPtr
+        Hs.FromFunPtrInstance
+        { fromFunPtrInstanceType = ftHsType
+        , fromFunPtrInstanceBody = ffiStubFrom
+        }
+      , defineInstanceComment = Nothing
+      }
+    ]
+  | tf@(C.TypePointer ft@(C.TypeFun args' res')) <- C.functionRes f : (map snd (C.functionArgs f))
```
edsko: @jorisdral I forget, should we worry about just a raw `TypeFun` here, or has that already been taken care of?

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -149,7 +149,27 @@ generateDeclarations' ::
   -> [WithCategory Hs.Decl]
 generateDeclarations' opts haddockConfig moduleName decs =
     flip State.evalState Map.empty $
-      concat <$> mapM (generateDecs opts haddockConfig moduleName) decs
+      concat <$> mapM (generateDecs opts haddockConfig moduleName isDefinedInCurrentModule) decs
```
bolt12: We should scan all types, not only functions. Construct a Set of them and then call `functionTypeFFIStubsAndFunPtrInstances`

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -149,7 +149,27 @@ generateDeclarations' ::
...
+  where
+    -- | Check if a type is defined in the current module
+    isDefinedInCurrentModule :: C.Type -> Bool
+    isDefinedInCurrentModule ty = case ty of
+      C.TypeStruct namePair _   -> namePairInDecs namePair
+      C.TypeUnion namePair _    -> namePairInDecs namePair
+      C.TypeEnum namePair _     -> namePairInDecs namePair
+      C.TypeTypedef ref         -> typedefRefInDecs ref
+      C.TypePointer t           -> isDefinedInCurrentModule t
+      _                         -> False
```
bolt12: This should function should be moved into `DeclIndex`.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -149,7 +149,27 @@ generateDeclarations' ::
   -> [WithCategory Hs.Decl]
 generateDeclarations' opts haddockConfig moduleName decs =
     flip State.evalState Map.empty $
-      concat <$> mapM (generateDecs opts haddockConfig moduleName) decs
+      concat <$> mapM (generateDecs opts haddockConfig moduleName isDefinedInCurrentModule) decs
+  where
+    -- | Check if a type is defined in the current module
+    isDefinedInCurrentModule :: C.Type -> Bool
+    isDefinedInCurrentModule ty = case ty of
```
bolt12:
```suggestion
    isDefinedInCurrentModule ty = case canonicalize ty of
```

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -1418,6 +1453,75 @@ functionDecs safety opts haddockConfig moduleName info f _spec =
...
+    -- one custom non-orphan types or pointers to these.
+  , case argsType of
+      C.TypeStruct _ _                   -> isDefinedInCurrentModule argsType
+      C.TypeEnum _ _                     -> isDefinedInCurrentModule argsType
+      C.TypeUnion _ _                    -> isDefinedInCurrentModule argsType
+      C.TypePointer t@(C.TypeStruct _ _) -> isDefinedInCurrentModule t
+      C.TypePointer t@(C.TypeEnum _ _  ) -> isDefinedInCurrentModule t
+      C.TypePointer t@(C.TypeUnion _ _ ) -> isDefinedInCurrentModule t
+      C.TypePointer t@(C.TypeTypedef _ ) -> isDefinedInCurrentModule t
+      _                                  -> False
```
bolt12: This pattern match is redundant

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
jorisdral: The parser should have already put a pointer in front of all function types that are missing one. So you don't have to worry about the case that it is just a `TypeFun`, indeed

bolt12: I have confirmed with him that this is not something we should care since it is taken care of in the Parse pass.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -140,37 +144,88 @@ generateDeclarations opts config name =
...
+scanAllFunctionPointerTypes :: [C.Decl] -> Set C.Type
+scanAllFunctionPointerTypes =
+  foldMap (\(C.Decl _ kind _) ->
+            case kind of
+              C.DeclStruct C.Struct {..} ->
+                foldMap (scanTypeForFunctionPointers . C.structFieldType) structFields
+              C.DeclUnion C.Union {..}   ->
+                foldMap (scanTypeForFunctionPointers . C.unionFieldType) unionFields
+              C.DeclEnum C.Enum {..} ->
+                scanTypeForFunctionPointers enumType
```
bolt12: Remove this

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -140,37 +144,88 @@ generateDeclarations opts config name =
...
+      C.TypeConstArray _ t                  -> scanTypeForFunctionPointers t
+      C.TypeBlock t                         -> scanTypeForFunctionPointers t
+      C.TypeQualified _ t                   -> scanTypeForFunctionPointers t
+      C.TypeTypedef (C.TypedefRegular _ t)  -> scanTypeForFunctionPointers t
+      C.TypeTypedef (C.TypedefSquashed _ t) -> scanTypeForFunctionPointers t
+      _                                     -> Set.empty
+
+-- | Check if a type is defined in the current module
+isDefinedInCurrentModule :: DeclIndex -> C.Type -> Bool
+isDefinedInCurrentModule declIndex ty = case C.getCanonicalType ty of
```
bolt12: We shouldn't call `getCanonicalType` here

edsko (approved):
Good stuff, I like how the manual example code looks! Feel free to merge once the `getCanonicalType` problem we discussed is resolved.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -140,17 +144,89 @@ generateDeclarations opts config name =
...
+      C.TypeConstArray _ t                  -> scanTypeForFunctionPointers t
+      C.TypeBlock t                         -> scanTypeForFunctionPointers t
+      C.TypeQualified _ t                   -> scanTypeForFunctionPointers t
+      C.TypeTypedef (C.TypedefRegular _ t)  -> scanTypeForFunctionPointers t
+      C.TypeTypedef (C.TypedefSquashed _ t) -> scanTypeForFunctionPointers t
+      _                                     -> Set.empty
+
+-- | Check if a type is defined in the current module
+isDefinedInCurrentModule :: DeclIndex -> C.Type -> Bool
+isDefinedInCurrentModule declIndex ty = case ty of
```
edsko, line 208: @bolt12 Should this have a case for `const` now? What if we have

```c
typedef int foo;
void f(void (*handler)(const foo x))
```

bolt12: You are right I missed that pattern matching, ugh

PR #1217, opened 2025-10-15
Add `botan` example project

### LINKED ISSUE #1171: Create bindings for `botan`

jorisdral:
Resolves #1171

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,3 @@ unversioned
 
 *.dll
 *.log
-
-/examples/c-minisat/hs-project/src/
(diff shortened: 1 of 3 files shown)


.github/workflows/examples.yml
```diff
@@ -5,8 +5,6 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
```
jorisdral, line 9: This change should be reverted before merging. This change was only included so that we could see the `botan` example run in CI even when the PR was not targetting the `main` branch

PR #1218, opened 2025-10-15
Run example projects only in one GHA job

jorisdral:


examples/README.md
bolt12, line 1: Nice :)

bolt12 (approved):
LGTM! Great work!

examples/README.md
```diff
@@ -0,0 +1,103 @@
+# Examples
+
+This directory contains runnable example projects using `hs-bindgen`. These
+examples include bindings and toy programs for the following C libraries:
+
+* `c-minisat`
+* `libcap`
```
dschrempf: `libpcap`

cabal.project.base
```diff
@@ -1,3 +1,5 @@
+index-state: 2025-09-16T04:55:58Z
```
dschrempf, line 1: Do we want to freeze the index state also for devs?

.github/actions/base/action.yml
```diff
@@ -42,36 +33,9 @@ runs:
...
-      if: ${{ startsWith(inputs.runner, 'macos') }}
-      run: |
-        echo "LLVM_CONFIG=$LLVM_PATH/bin/llvm-config" >> "$GITHUB_ENV"
-        echo "LIBCLANG_PATH=$LLVM_PATH/lib/" >> "$GITHUB_ENV"
-        # https://stackoverflow.com/questions/63342521/clang-on-macos-having-problems-with-its-includes
-        export SDKROOT=$(xcrun --show-sdk-path --sdk macosx)
-        echo $SDKROOT
-        echo "SDKROOT=$SDKROOT" >> "$GITHUB_ENV"
-
-    - name: Debug (LLVM/Clang)
```
dschrempf, line 62: Is there a reason why we removed the debug output? It is quite useful to know which version of the N installed versions is actually used :-P

dschrempf (approved):
Thank you very much. I was worried about running all examples on all PR/MQ jobs!

.github/actions/base/action.yml
jorisdral: It's probably an oversight on my side, I'll change it back!

cabal.project.base
jorisdral: I usually do -- the upside is that there are no surprising differences between your local dev builds and GHA CI builds. Are there downsides you see to doing this?

dschrempf: I do not know. I picked on this because: I am not sure how it affects the Nix setup, and the Nix Flake. At the moment, Nix users use the binary cache for all upstream derivations, and only have to build the `hs-bindgen`-related packages. The binary cache is also used by the tutorial (and we advertize it).

On the other hand, I think freeze files are the way to go (that's one of the reasons I am using Nix in the first place), so I welcome fixing the index. It would be even better if we had proper lock files.

Not sure what the conclusion is. Maybe we try it and see if we run into problems with the tutorial? In the worst case we Nix users download packages from the binary cache when entering the development shell, and then rebuild everything using the index state when running `cabal`.

jorisdral: I can't comment on the interaction with Nix unfortunately, I have very little experience with that

In past projects I've worked on, the nix setup would take into account the index state. For example: https://github.com/IntersectMBO/ouroboros-consensus/blob/80e345822afd51d2627eec25592e1ebc8027152b/CONTRIBUTING.md?plain=1#L272

jorisdral: Do we know what is included in the binary cache? Does the binary cache save multiple versions of the same package? I'd expect it not to store only a single version

PR #1227, opened 2025-10-28
Run `cabal-fmt`

jorisdral:
The first two commits prepare the documentation in cabal files for running `cabal-fmt`. In particular, comments should come before fields like `build-depends`, not inside them. The last commit runs `cabal-fmt`


jorisdral:
> If we can have those ghc options to be on a line per option, that would be nice, but either way feel free to merge.

I checked and it doesn't seem like I can change it.

I've also added a CI job that checks formatting with `cabal-fmt`

PR #1229, opened 2025-10-28
Compile fixtures

### LINKED ISSUE #1070: Test compilation of fixtures

bolt12:
This PR closes #1070.

We only ignore 4 fixtures:

- iterator.pp.hs: Makes use of Apple block extension which would require clang (see #913)
- decls_in_signature.pp.hs: Unusable struct (see #1128)
- redeclaration.pp.hs: Same as typenames.pp.hs
- typenames.pp.hs: hs-bindgen namespace possible bug/feature

And fixed 6 fixtures that weren't compiling.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/HsModule/Names.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/HsModule/Names.hs
@@ -19,6 +19,7 @@ import Data.Ix qualified
 import Data.List qualified as L
 import Data.List.NonEmpty qualified as NonEmpty
 import Data.Map.Strict qualified as Map
+import Data.Maybe qualified
 import Data.Set qualified as Set
 import Data.Void qualified
 import Foreign qualified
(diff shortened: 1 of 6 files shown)


jorisdral:
Quick drive-by comment: the script only picks up fixtures that are in the `hs-bindgen/fixtures` directory, but not any of it's sub-directories, like `hs-bindgen/fixtures/manual`. Maybe the script should search through the folders recursively

bolt12:
@jorisdral thanks! Good point! I'll fix it!

hs-bindgen/src-internal/HsBindgen/Backend/HsModule/Names.hs
```diff
@@ -204,6 +204,7 @@ resolveGlobal = \case
     Applicative_seq               -> importU '(<*>)
     Monad_return                  -> importU 'return
     Monad_seq                     -> importU '(>>)
+    Maybe_type                    -> ResolvedName "Maybe(..)" IdentifierName (Just (UnqualifiedHsImport iPrelude))
```
bolt12: Use Prelude.Just and Prelude.Nothing

PR #1235, opened 2025-10-29
Update Nix Flake

dschrempf:
- Add `cabal-fmt` to the Nix development shell
- Update Nix Flake lock file


dschrempf:
Merging since this only affects Nix-related files.

PR #1237, opened 2025-10-29
Add a configure step to the `examples` workflow file

jorisdral:
This prevents having to pass `--disable-*` options to cabal everywhere in the
workflow file. As a result, I've also refactored the `generate-and-run.sh`
scripts for the examples to not just overwrite a `cabal.project.local` file with
paths to the C libraries and binaries, but instead append the paths to the file
if it already exists.


jorisdral:
Merging this because it is a fairly small change

PR #1238, opened 2025-10-29
Pattern match on qualifier const

bolt12:
Fixes https://github.com/well-typed/hs-bindgen/pull/1213#discussion_r2474263219


edsko (approved):
Thanks :) Feel free to merge once you have a test case, it's useful I think to ensure we don't introduce any regressions later; it's an easy thing to miss.

PR #1240, opened 2025-10-30
Inform user about selected but unavailable declarations

dschrempf:
These may be because
1. the selected declarations is unavailable, or
2. a transitive dependency of the selected declaration is unavailable.

The second point is much harder to detect.

Further, this PR changes the way selection works on anonymous declarations. Now, we apply the select predicate to the _use site_ of the anonymous declaration.

This commit also contains various other changes:
- Add omitted and external to DeclIndex
- Handle failed macros
- Update DeclMeta so it is in sync with the list of declarations threaded through the passes
- Error when selected declaration itself is unavailable
- ResolveBindingSpecs: RWS -> ReaderT/State to avoid leaks

#### State

- [x] The use-decl and decl-use graphs are still not in sync with the resolved binding specifications.

We still have two failing golden tests:
- [x] `type_qualifiers.h` ~~(Interaction with binding specifications, we will discuss tomorrow with @TravisCardwell).~~ should be solved by https://github.com/well-typed/hs-bindgen/pull/1248
- [x] `reparse.h` ~~(`size_t` and probably other declarations in the standard library binding specifications are considered unavailable).~~ Should be solved by https://github.com/well-typed/hs-bindgen/pull/1248

Other Todos:
-   [X] Document that `enable-program-slicing` can cause declarations to be included even if they are explicitly excluded by a select predicate.
-   [X] In `ResolveBindingSpecs`:
    -   [X] Try to reconstruct the decl-use graph by reversing the use-decl graph
    -   [X] Add `external` map to `DeclIndex`, removing the decls from `succeeded`, `notAttempted`, `failed` due to external binding specifications.
    -   [X] Add `omitted` map to `DeclIndex`, which contains declarations omitted by prescriptive binding specifications.
-   [X] In `HandleMacros`:
    -   Remove macros that fail to resolve from `succeeded` to `failedMacros`.
-   [x] Add golden tests (see below, tests may change)
-   [X] Check expectations for test variants of `selection_fail.h` tests
-   [x] Also determine if transitive dependencies are selected. If a transitive dependency is not available, suggest changing the select predicate or adding external bindings specification(s)

See also https://github.com/well-typed/hs-bindgen/issues/1037#issuecomment-3516788536).

#### Golden test 1

```c
// my_size_t.h:
typedef char *size_t; // bad programmer, bad
```

```c
#include <my_size_t.h>

#define A int

void f(A x, size_t y); // use of macro here, so must be reparsed
```

#### Golden test 2

```c
// my_foo.h
typedef long double foo;
```

```c
#include <my_foo.h>

#define A foo
```

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclIndex.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclIndex.hs
@@ -159,31 +159,16 @@ fromParseResults results =
           ParseResultNotAttempted x ->
             over
               ( #index % #notAttempted )
-              ( alter qualPrelimDeclId x )
+              ( insertFailure qualPrelimDeclId x )
               oldIndex
           ParseResultFailure x ->
             over
               ( #index % #failed )
-              ( alter qualPrelimDeclId x )
+              ( insertFailure qualPrelimDeclId x )
               oldIndex
       where
         qualPrelimDeclId :: C.QualPrelimDeclId
         qualPrelimDeclId = getQualPrelimDeclId parse
-
-    alter :: (Ord k, Show a, Eq a) => k -> a -> Map k a -> Map k a
-    alter key x =
-      Map.alter ( \case
-        Nothing -> Just x
-        Just x'
-          -- TODO https://github.com/well-typed/hs-bindgen/issues/1283: We parse
-          -- declarations multiple times.
-          | x == x'   -> Just x
-          -- TODO_PR: We also encounter conflicting values. In particular for
-          -- the `hsb_complex_test`, `cimag` is defined twice (function and
-          -- macro).
-          -- | otherwise -> panicPure $ "conflicting values: " <> show (x, x')
-          | otherwise -> traceShow ("conflicting values: " <> show (x,x')) $ Just x
-        ) key
 
     insert ::
          ParseSuccess
@@ -231,6 +216,14 @@ fromParseResults results =
        -- In case of an error, /remove/ the value from the map
        failure :: e -> (Maybe a, Maybe e)
        failure err = (Nothing, Just err)
+
+    -- For failures, we just stick with the first failure.
+    insertFailure :: Ord k => k -> a -> Map k a -> Map k a
+    insertFailure key x =
+      Map.alter ( \case
+        Nothing -> Just x
+        Just x' -> Just x'
+        ) key
 
 sameDefinition :: C.DeclKind Parse -> C.DeclKind Parse -> Bool
 sameDefinition a b =
@@ -299,7 +292,7 @@ keysSet DeclIndex{..} = Set.unions [
     , Map.keysSet notAttempted
     , Map.keysSet failed
     , Map.keysSet failedMacros
-    -- TODO https://github.com/well-typed/hs-bindgen/issues/799: Also add
+    -- TODO https://github.com/well-typed/hs-bindgen/issues/1301: Also add
     -- 'omitted' here and deal with the 'omitted' case in selection.
     ]
 
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
@@ -79,31 +79,32 @@ selectDecls
   isInMainHeaderDir
   SelectConfig{..}
   C.TranslationUnit{..} =
-    let -- | Directly match the select predicate on the 'DeclIndex', obtaining
+    let -- Directly match the select predicate on the 'DeclIndex', obtaining
         -- information about succeeded _and failed_ selection roots.
         selectedIndex :: DeclIndex
         selectedIndex = selectDeclIndex match index
 
-        -- | Identifiers of selection roots. Some of them may be unavailable
+        -- Identifiers of selection roots. Some of them may be unavailable
         -- (i.e., not in the 'succeeded' map, and hence, not in the list of
         -- declarations attached to the translation unit).
         rootIds :: Set DeclId
         rootIds = DeclIndex.keysSet selectedIndex
 
-        -- | Identifiers of transitive dependencies including selection roots.
+        -- Identifiers of transitive dependencies including selection roots.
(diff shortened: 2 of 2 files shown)


TravisCardwell:
Using the current `dom/1037/select-but-unavailable-message` branch (`43c47a97`), I tried to create a minimal example using `size_t` that I can experiment with using `hs-bindgen-cli preprocess`.

For my first try, bindings are generated without issue, correctly using `HsBindgen.Runtime.Prelude.CSize` according to the `stdlib` binding specification.

```c
#include <stddef.h>

typedef size_t ExSize;
```

I tested using my system headers as well as using the Musl headers, to match the CI environment:

```
    --builtin-include-dir disable \
    --clang-option "-isystem$(pwd)/hs-bindgen/musl-include/x86_64" \
    -I "$(pwd)" \
    --unique-id ex \
    --hs-output-dir ex-src \
    ex.h
```

Looking at `type_qualifiers.h`, the `size_t` is used in a function parameter, so I tried that.  Bindings are again generated without issue, using my system headers as well as the Musl headers.

```c
#include <stddef.h>

void foo(size_t);
```

The golden tests for both `type_qualifiers` and `reparse` tests seem straightforward; no selection predicates are specified.  I am still trying to figure out what is going on.

TravisCardwell:
Looking at the `-v3` output for my minimal test:

```
[Info   ] [HsBindgen] [handle-macros-reparse] Failed to reparse:  Encountered unexpected node in the language-c AST:  .../hs-bindgen/musl-include/x86_64/bits/alltypes.h:50: (column 24) [ERROR]  >>> Syntax Error !
  Syntax error !
  The symbol `size_t' does not fit here.


  unexpected, called at src-internal/HsBindgen/Frontend/LanguageC.hs:118:9 in hs-bindgen-0.1.0-inplace-internal:HsBindgen.Frontend.LanguageC
...
[Info   ] [HsBindgen] [binding-spec-resolve] during resolution of external binding specification:
  header stddef.h resolved to .../hs-bindgen/musl-include/x86_64/stddef.h
...
[Info   ] [HsBindgen] [select] 'ExSize' (in source at ".../ex.h:3:16")selected (direct select predicate match)
...
```

We indeed fail to reparse the `size_t` declaration.  Here is the relevant code:

```c
#define _Addr long
typedef unsigned _Addr size_t;
```

I would like to add additional logging for `ResolveBindingSpecs`.  With some quick ones in place, I get the following additional logs:

```
[Info   ] [HsBindgen] [resolve-binding-specs] declaration with external binding dropped: 'size_t'
[Info   ] [HsBindgen] [resolve-binding-specs] type replaced with external binding: 'size_t'
```

The first message indicates that we still have a `Decl` for `size_t`, even though reparsing fails.  This log is probably fine to keep as-is.

The second message is the resolution of the external binding in `typedef size_t ExSize`.  This log should probably also show the context, which involves threading some state.  Perhaps it should have level `Debug` since there may be *many* such logs.

TravisCardwell:
(Updated to `d543ea8a`)

I tried `type_qualifiers.h` as-is, and I get an error:

```
[Error  ] [HsBindgen] [mangle-names] Missing declaration: 'bool'; did you select the declaration?
```

Strangely, bindings are still written to files even though there is an error.

The error makes sense, though: there is no `bool` external binding, and I did not select the `bool` type.  There is no `bool` in `stdlib` because when I worked on `stdlib` the `bool` type was already being interpreted as `CBool`.  Perhaps this has changed.

Adding a `bool` external binding to `stdlib` resolves the issue with my manual test.  The `preprocess` command generates bindings without issue.  The tests still fail; I will look at those next.

BTW, I recently wrote some documentation about C boolean types in the module documentation for [`HsBindgen.Runtime.CBoolSem`](https://github.com/well-typed/hs-bindgen/blob/a019c0d1b569521bee91fe04094491e1ed79e558/hs-bindgen-runtime/src/HsBindgen/Runtime/CBoolSem.hs).

TravisCardwell:
Running the preprocessor test for `type_qualifiers.h`

```
$ cabal test hs-bindgen:test:test-hs-bindgen \
    --test-option '--pattern=/type_qualifiers/&&/pp/'
```

results in the following unexpected traces (reformatted):

```
[select][pretty] 'list_example' (in source at ".../type_qualifiers.h:14:6")
  selected (direct select predicate match) but unavailable (parse of
  transitive dependency 'list_example' not attempted: (!) adjust parse
  predicate)
[select][show  ] ...
[select][pretty] 'list_example' (in source at ".../type_qualifiers.h:14:6")
  selected (direct select predicate match) but unavailable (parse of
  transitive dependency 'size_t' not attempted: (!) adjust parse predicate)
[select][show  ] ...
```

Since `size_t` is replaced with an external binding, there should be no transitive dependency on `size_t`.  When we talked last, we talked about how the `DeclUseGraph` was not being updated along with the `UseDeclGraph` to reflect the removal of dependencies when using external bindings.  I see that it is indeed being recalculated in `ResolveBindingSpecs` as follows:

```haskell
declDeclUse = DeclUseGraph.fromUseDecl stateUseDecl
```

When test manually using `hs-bindgen-cli preprocess`, the `UseDeclGraph` and `DeclUseGraph` correctly have the dependency between `list_example` and `size_t` removed.  When testing using `cabal test`, however, I see quite different graphs with the dependency *not* removed!

@dschrempf just taught me the reason for this discrepancy: the tests now default to using C23!  That explains a lot!

It looks like we represent `size_t` as a primitive when using C23, so there is no external binding in `ResolveBindingSpecs`.  We still compute the dependency, however, and that dependency is *not* removed since there is no external binding.  Now I finally understand why we see the dependency in the `Select` pass.

dschrempf:
FYI, this can be reproduced using, e.g.,
```
cabal run -- hs-bindgen-cli preprocess --unique-id pwnd -I ./hs-bindgen/musl-include/x86_64 --builtin-include-dir disable --gnu --hs-output-dir tmp --create-output-dirs --standard c23 -I ./hs-bindgen/examples/golden type_qualifiers.h -v3

```

dschrempf:
Thanks @TravisCardwell for your input and for debugging the issue. Edsko and I have had some discussions, and we know the culprit (macro reparsing). We are working on a fix.

Issue #1099, opened 2025-09-17
Build and test `hs-bindgen` with newer versions of Clang on Windows


dschrempf:
At the moment, we observe linker issues with newer versions of Clang on Windows. For details, see #1072.

We should make sure that we can compile and test `hs-bindgen`, as well as build and run the manual of `hs-bindgen` with newer versions of Clang.

Relevant CI error:
```
ld.lld: error: undefined symbol: __declspec(dllimport) clang_getUnqualifiedType
```

edsko:
Related: Use Clang provided by GHC in Windows CI #1098

edsko:
Removing the `CI` label for now, as we should fix the problem, independently from whether or not we test this on CI.

TravisCardwell:
From https://github.com/well-typed/hs-bindgen/issues/1098#issuecomment-3857136118:

> The (official) builds of GHC that we install come with a MinGW environment that includes a GNU toolchain with a corresponding version of LLVM/Clang.  That is *the* happy path for using GHC on Windows, and we need to test it.  Note that the version of LLVM/Clang used is determined by the version of GHC, as each build of GHC includes a specific version of LLVM/Clang.
>
> It looks like GHC does *not* support MSVC targets at all.  Users can use `hs-bindgen` to generate bindings targeting MSVC, but doing so is somewhat dubious since GHC itself cannot target MSVC when compiling those bindings.  We decided to not check for this and emit an error/warning, however, because people may want to attempt to work around such limitations.  That is a big project in itself and is (IMHO) not something that we should test.

Installing newer version of LLVM/Clang using [KyleMayes/install-llvm-action](https://github.com/KyleMayes/install-llvm-action) is *not* viable because they target MSVC.

*Perhaps* one could build a newer version of LLVM/Clang for the MinGW environment used by GHC.  This is likely not trivial, and I do not think we should worry about testing such cases unless there is a definite need to.

edsko:
Re-opening as low priority.

PR #1243, opened 2025-11-03
Update compile-fixtures script to generate haddocks

### LINKED ISSUE #1159: More comprehensive optional CI "Release" flow
This should be triggered manually, and is intended as a more comprehensive test suite. It should test

- [ ] All LLVM versions 14-21
- [ ] `cabal check`
- [ ] Haddocks
- [x] #1070

.. and possibly others.

Marking this for `release-0.1` as it would be quite useful to have it at that point, though of course we can release without this ticket being done.

bolt12:
This PR updates the compile fixtures script to build the fixtures and haddocks at the same time using `haddock`. The script now also generates a single index.html file where one can see the haddocks of each individual fixture.

This PR also adds a new CI job that can be manually triggered, this CI job will essentially run some more comprehensive checks on top of the current ones, i.e. it will:

- Build the manual on all platforms for all LLVM versions
- Run `cabal check` for all packages
- Run `cabal sdist` and try to see if it is buildable for all packages
- Build haddocks for all packages

This PR closes #1159 

Some nice features that would be nice to have: compile and generate haddocks for TH fixtures.


bolt12:
hmm not sure if this needs to be merge in order for `comprehensive-ci.yml` to appear in the Actions tab.

bolt12:
I have changed the comprehensive-ci.yml workflow to trigger on push so that I can see it works, then I will change to to `workflow_dispatch` type of event (See: https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow) so that it can be triggered manually from the UI

edsko:
Blocked on https://github.com/well-typed/hs-bindgen/pull/1256.

bolt12:
Only build the haddocks on the comprehensive CI

.github/workflows/comprehensive-ci.yml
```diff
@@ -0,0 +1,382 @@
...
+
+  test-all-llvm-and-ghc-versions:
+    name: Test (${{ matrix.runner }}, GHC ${{ matrix.ghc }}, LLVM ${{ matrix.llvm }})
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 60
+
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: [ubuntu-latest, macos-latest]
```
bolt12, line 25: Add `windows`

.github/workflows/comprehensive-ci.yml
```diff
@@ -0,0 +1,382 @@
...
+    name: Test (windows-latest, GHC ${{ matrix.ghc }})
+    runs-on: windows-latest
+    timeout-minutes: 60
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # GHC 9.2 uses GCC, not LLVM/Clang, so we do not test it.
+        # On Windows, LLVM is bundled with GHC — no separate LLVM matrix.
+        ghc: ['9.4', '9.6', '9.8', '9.10', '9.12', '9.14']
```
bolt12, line 101: We should try and test with other LLVM versions (pair GHC, LLVM)

bolt12: Check if we can test GHC 9.2 with an LLVM version that is installed.

.github/workflows/haskell.yml
```diff
@@ -193,6 +193,9 @@ jobs:
     - name: 🛠️ Build Cabal dependencies
       run: cabal build all --only-dependencies
 
+    - name: 🏗️ Build (--prefer-oldest, lower bounds check)
+      run: cabal build all --prefer-oldest
```
bolt12: Make this part of the comprehensive CI job

.github/workflows/comprehensive-ci.yml
TravisCardwell: `comprehensive-ci.yml` uses `.github/actions/setup-llvm`, which installs LLVM/Clang using `KyleMayes/install-llvm-action`.  On Windows, this installs LLVM/Clang builds that target MSVC.  GHC does not support MSVC, so we decided to *not* test this (unless an explicit requirement to do so arises).

.github/workflows/comprehensive-ci.yml
TravisCardwell: While we should not use `KyleMayes/install-llvm-action` on Windows, there are ways to test more versions.  For example, we could test GHC 9.2 by also installing GHC 9.4 and using the LLVM/Clang that is distributed with GHC 9.4 to test with GHC 9.2.  GHC 9.14 comes with a more recent version of LLVM/Clang, and we could use it to test other versions.  We may be able to install a different version of LLVM/Clang via MSYS2.  We may be able to build different versions of LLVM/Clang within MSYS2.

We have many more higher priority tasks, though, so we decided to only test the straightforward/easy setups that users will likely use.  We can of course test other options if requested by a client.

.github/workflows/comprehensive-ci.yml
edsko: @bolt12 Sorry, this is on me, I forgot the details here. Thanks @TravisCardwell for reminding me. I had forgotten that installing more recent `clang` versions on Windows is non-trivial, because they target msvc. We have a ticket for this, which got closed (prematurely, that might have been on me as well): https://github.com/well-typed/hs-bindgen/issues/1099.

PR #1246, opened 2025-11-04
Add `bool` to `stdlib` external binding spec

TravisCardwell:
I ran into this while looking into #1240.  When I worked on `stdlib`, the `bool` type was already being translated to `CBool`.  This has apparently changed.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/Stdlib.hs
+++ b/hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/Stdlib.hs
@@ -46,7 +46,8 @@ bindingSpec = BindingSpec.BindingSpec{..}
     bindingSpecCTypes  :: CTypeMap
     bindingSpecHsTypes :: HsTypeMap
     (bindingSpecCTypes, bindingSpecHsTypes) = mkMaps $
-         integralTypes
+         boolTypes
+      ++ integralTypes
       ++ floatingTypes
       ++ mathTypes
       ++ stdTypes
@@ -55,6 +56,11 @@ bindingSpec = BindingSpec.BindingSpec{..}
       ++ timeTypes
       ++ fileTypes
       ++ signalTypes
+
+    boolTypes :: [(CTypeKV, HsTypeKV)]
+    boolTypes = [
+        mkTypeN "bool" "CBool" cD intI (Just $ BindingSpec.Builtin BFT.CBool) ["stdbool.h"]
+      ]
 
     integralTypes :: [(CTypeKV, HsTypeKV)]
     integralTypes =


edsko:
@dschrempf , the consequences of this PR are not visible on `main`, but they might become visible on your branch where you change our invocation of `language-c`. You might need to cherry-pick this into that branch.

I'm leaving this unmerged until we see that it actually affects something.

We might need to add one or two tests that use C17 rather than C23 to make this more visible.

TravisCardwell:
(Rebased)

dschrempf:
Just bumped into this. Is there anything to do on my side?

TravisCardwell:
(Rebased)

@dschrempf Did "your branch where you change our invocation of `language-c`" get merged yet?  If so, I guess it did not need this change.

Indeed, we should have some tests for other C standards, not only test C23.

TravisCardwell:
(Rebased)

dschrempf:
Sorry, I missed that. I think, yes. The mentioned branch got merged a while ago. I think it was mostly about requiring a C standard (instead of having it as optional).

There is a discrepancy though:
- We use C17 by default on the command line (at least this is what the help states).
- We use C23 in our tests (`mkTestClangArgsConfig`).

TravisCardwell:
Our tests need to cover more than one C standard.  Using the latest C standard that we support (currently C23) for the majority of tests is easy, but perhaps using the earliest C standard possible per test would provide better testing and would naturally test a wide range of C standards.  The vast majority of C in the world is not C23 after all.

To test this issue, we at least need a way to use a different C standard (C17) for the test case.  This requires some refactoring of the test harness code.

Regarding the C standard used by default outside of tests (with the CLI and TH API), this behavior will change when I finish #1372 (currently on hold because the identifier redesign takes priority).  If a user does not specify a C standard, we will use `libclang` to determine an appropriate C standard, similar to how we determine the target when one is not specified.

TravisCardwell:
I wonder if we could organize the test harness so that tests that cannot run with the current version of `libclang` are skipped?

In the majority of cases, the C standard (minimum required for the test) is sufficient.  We have a function that gets the Clang argument for a C standard, returning an `Either` type that indicates if the C standard works with the current `libclang` version.  (Note that this function is significantly refactored in my WIP branch.)  We can skip a test if this function returns a `Left` value.

In some cases, we will need to skip tests based on the `libclang` version, however.  C23 is "supported" from Clang 9, but only a fraction of the standard is implemented in older versions.  When testing with LLVM/Clang 16, we would like to be able to skip any C23 tests that do not work with that version for this reason.

edsko:
After discussing this with @TravisCardwell , let's make C95 our default in the tests, selecting a newer version only where needed. (For the CLI, we'll query clang instead, rather than choosing our own default.)

dschrempf:
I am runnin into this again when creating bindings for `wlroots`. If I do not use `--standard c23`, I get 

```
Missing declaration: DeclId {name = DeclName {text = "bool", kind = NameKindOrdinary}, origDeclId = OrigDeclId (PrelimDeclIdNamed (DeclName {text = "bool", kind = NameKindOrdinary})), haskellId = ()}
  panicPure, called at src-internal/HsBindgen/Frontend/Pass/MangleNames.hs:148:18 in hs-bindgen-0.1.0-1K6BmsuR8ZO937zeo1KSkn-internal:HsBindgen.Frontend.Pass.MangleNames
```

edsko:
@dschrempf If this PR fixes your issue, feel free to merge it :)

dschrempf:
I rebased, and added a small change wrt FFI types. This PR indeed fixes the issues I am having with C standards and `wlroots`.

PR #1250, opened 2025-11-06
Fix evaluation of error traces

### LINKED ISSUE #1197: Exit code is zero, even with Error traces
I think we defeated our own failure mechanism by using references to cached computations and running them later than expected. I get exit code `0` although I have error traces.

```console
$ cabal run -- hs-bindgen-cli preprocess --hs-output-dir unused not-found
Configuration is affected by the following files:
- cabal.project
[Warning] [HsBindgen] [unique-id] empty unique identifier ('UniqueId'):
  C uses a global namespace.
  We encourage using a unique identifier to avoid duplicate symbol names.
  For example, use and adapt 'com.example.package'.
[Error  ] [Libclang ] [clang] unable to resolve #include <not-found>
[1] 11:19:26 dominik@schwarzbaer:/home/dominik/work/hs-bindgen

$ echo $?
0
```

I think we return from the frontend before running the computations, and so, there are not `Error` traces, yet.

I expect our program fails when we trace to files, the handles will be closed before the traces are emitted.

As far as I understand we are anyways moving towards not failing, even when there was an error; instead we just omit the affected declarations.

bolt12:
This PR fixes #1197

### DIFF
--- a/hs-bindgen/hs-bindgen.cabal
+++ b/hs-bindgen/hs-bindgen.cabal
@@ -346,10 +346,10 @@ library test-common
     , utf8-string                  >=1.0.2  && <1.1
 
 test-suite test-hs-bindgen
-  import:         lang
-  type:           exitcode-stdio-1.0
-  main-is:        test-hs-bindgen.hs
-  hs-source-dirs: test/hs-bindgen
+  import:             lang
+  type:               exitcode-stdio-1.0
+  main-is:            test-hs-bindgen.hs
+  hs-source-dirs:     test/hs-bindgen
   other-modules:
     Test.HsBindgen.Golden
     Test.HsBindgen.Golden.Check.BindingSpec
--- a/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Integration/ExitCode.hs
+++ b/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Integration/ExitCode.hs
@@ -1,8 +1,10 @@
 module Test.HsBindgen.Integration.ExitCode (tests) where
 
 import Control.Exception (fromException, try)
+import System.Exit (ExitCode (..))
 import System.FilePath ((</>))
 import System.IO.Temp (withSystemTempDirectory)
+import System.Process (readProcessWithExitCode)
 import Test.Common.HsBindgen.TracePredicate
 import Test.HsBindgen.Golden.TestCase
 import Test.HsBindgen.Resources
@@ -12,9 +14,6 @@ import Test.Tasty.HUnit
 import HsBindgen.Artefact (Artefact (..), NP (..))
 import HsBindgen.TraceMsg
 import HsBindgen.Util.Tracer
-import System.Process (readProcessWithExitCode)
-import System.Exit (ExitCode(..))
-import Debug.Trace qualified as Debug
 
 {-------------------------------------------------------------------------------
   Tests


hs-bindgen/src-internal/HsBindgen/Util/Tracer.hs
```diff
@@ -31,8 +31,10 @@ module HsBindgen.Util.Tracer (
   , TracerConfig (..)
     -- * Tracers
   , withTracer
+  , withTracerRef
```
dschrempf: Can we replace `withTracer` by `withTracerRef`? I.e., only export one function doing the right thing?

hs-bindgen/src-internal/HsBindgen/Util/Tracer.hs
```diff
@@ -31,8 +31,10 @@ module HsBindgen.Util.Tracer (
...
   , TracerState(..)
   , withTracer'
+  , withTracerRef'
```
dschrempf: Also here. I think this needs some adaptation in the test suite, but should otherwise be fine.

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -107,32 +110,44 @@ runArtefacts
...
-
-    runArtefact :: Artefact a -> ArtefactM a
+    go :: Artefacts x -> ArtefactM (Either (TraceException msg) (NP I x))
+    go Nil       = pure (Right Nil)
+    go (a :* as) = do
+      artefactResult <- runArtefact a
+      case artefactResult of
+        Left errors -> pure (Left errors)
+        Right artefact -> do
+          tracerState <- liftIO $ readIORef tracerStateRef
```
dschrempf: Can we instead export a function from `Tracer.hs` performing this check, and only use this function here? `runArtefacts` does not need to know how to detect an error, only that an error has happend (with a list of error traces I guess).

dschrempf: I guess we will also need the function when adapting the tests.

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -107,32 +110,44 @@ runArtefacts
...
+          case tracerState of
+            TracerState Error errors -> pure (Left (TraceException errors))
+            _ -> do
+              results <- go as
+              case results of
+                Left err -> pure (Left err)
```
dschrempf: You could move the `pure` before the `case`, if you want.

dschrempf: It is then maybe better to use `fmap` to handle the `Right` case (but that's a detail).

hs-bindgen/src-internal/HsBindgen.hs
```diff
@@ -73,14 +74,14 @@ hsBindgen
...
-    backendArtefact <- withTracerSafe tracerConfigSafe $ \tracer -> do
-      backend tracer bindgenBackendConfig bootArtefact frontendArtefact
-    -- 4. Artefacts.
-    withTracerSafe tracerConfigSafe $ \tracer -> do
-      runArtefacts tracer bootArtefact frontendArtefact backendArtefact artefacts
+      -- 3. Backend.
+      backendArtefact <- withTracerSafe tracerConfigSafe $ \tracer' -> do
+        backend tracer' bindgenBackendConfig bootArtefact frontendArtefact
+      -- 4. Artefacts.
+      withTracerSafe tracerConfigSafe $ \tracer' -> do
```
dschrempf: Yes, I agree. It is suboptimal that we have the safe trace intertwined here. This makes me think we should probably combine the two. Edsko and I thought about this quite a lot, and the beauty of the current approach is that the backend truly only can omit safe traces. On the other hand, I am pretty sure we can lift safe traces into the other tracers at this stage too.

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -107,32 +110,39 @@ runArtefacts
...
-      FinalModules        -> liftIO backendFinalModules
+      HsDecls             -> Right <$> liftIO backendHsDecls
+      FinalDecls          -> Right <$> liftIO backendFinalDecls
+      FinalModuleBaseName -> pure (Right backendFinalModuleBaseName)
+      FinalModuleSafe     -> Right <$> liftIO backendFinalModuleSafe
+      FinalModuleUnsafe   -> Right <$> liftIO backendFinalModuleUnsafe
+      FinalModules        -> Right <$> liftIO backendFinalModules
       -- Lift and sequence.
-      (Lift as' f)        -> go as' >>= f
+      (Lift as' f)        -> go as' >>= \x -> mapM f x
```
dschrempf:
```suggestion
      (Lift as' f)        -> go as' >>= mapM f
```

PR #1256, opened 2025-11-06
Windows CI jobs

TravisCardwell:
This *draft* PR exists to experiment with Windows CI jobs.

### DIFF
--- a/cabal.project.base
+++ b/cabal.project.base
@@ -5,7 +5,7 @@ index-state: 2026-02-05T03:27:53Z
 source-repository-package
     type: git
     location: https://github.com/well-typed/libclang
-    tag: 155642a4a4a9f0414a058a8f08f39aa6c7bb57ed
+    tag: ad0134b24a78fba4ef7c7f68d97d0feeded66be1
 
 --
 -- Temporary package bounds overrides for ghc 9.14
--- a/flake.lock
+++ b/flake.lock
@@ -21,17 +21,17 @@
     "libclang-bindings-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1770274896,
-        "narHash": "sha256-JnxJBo2L4URFD8JbpjnPG/ej/xKFe7y5ZpjnvIztwAM=",
+        "lastModified": 1771549116,
+        "narHash": "sha256-aweqLM5Y4odpY2XNAxNZDkuyp4b0rjIAEoMcDwpP8y8=",
         "owner": "well-typed",
         "repo": "libclang",
-        "rev": "155642a4a4a9f0414a058a8f08f39aa6c7bb57ed",
+        "rev": "ad0134b24a78fba4ef7c7f68d97d0feeded66be1",
         "type": "github"
       },
       "original": {
         "owner": "well-typed",
         "repo": "libclang",
-        "rev": "155642a4a4a9f0414a058a8f08f39aa6c7bb57ed",
+        "rev": "ad0134b24a78fba4ef7c7f68d97d0feeded66be1",
         "type": "github"
       }
     },
--- a/flake.nix
+++ b/flake.nix
@@ -5,7 +5,7 @@
     flake-parts.url = "github:hercules-ci/flake-parts";
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
     libclang-bindings-src = {
-      url = "github:well-typed/libclang?rev=155642a4a4a9f0414a058a8f08f39aa6c7bb57ed";
+      url = "github:well-typed/libclang?rev=ad0134b24a78fba4ef7c7f68d97d0feeded66be1";
       flake = false;
     };
   };


TravisCardwell:
I created this `tcard/windows` branch from `bolt12/1159` (#1243) for experimentation.  I disabled all workflows except `comprehensive-ci.yml`.  Within `comprehensive-ci.yml`, I commented out the non-Windows jobs.  To start with, I changed the configuration for the remaining job to only test with GHC 9.4.8 and LLVM/Clang 14.  These versions are selected just because I see that those are the versions that we already (successfully) test on Windows.

TravisCardwell:
I am currently trying to figure out why `libclang-bindings` is using the incorrect version of LLVM/Clang, even though I am setting `LLVM_PATH`, `LLVM_CONFIG`, and `PATH`.  I am now able to get logs from the `configure` step:

```
checking for llvm-config... C:/tools/ghc-9.4.8/mingw/bin/llvm-config.exe
checking LLVM version... 14.0.6
checking LLVM library directory... C:/tools/ghc-9.4.8/mingw/lib
checking LLVM include directory... C:/tools/ghc-9.4.8/mingw/include
...
checking libclang version... clang version 20.1.8
```

Those first four messages show that it is using the configured installation of LLVM/Clang, as desired, but that last message shows that it is not linking to the correct library!  Now I know what is going wrong and can start working on it.

TravisCardwell:
I am unable to reproduce the `libclang` linking issue locally, so I created a `libclang` draft PR (https://github.com/well-typed/libclang/pull/29) so that I can debug it with `tmate` there, without blocking the `hs-bindgen` CI.  The issue is reproduced in that PR.

TravisCardwell:
One annoyance with debugging failed tests is that Cabal interleaves output, which can occasionally make it difficult to understand what is failing.  I wonder how much longer it takes to run tests sequentially?

Testing locally:

```
$ time cabal test all
...
real    0m23.317s
user    0m36.351s
sys     0m7.064s
```

```
$ time cabal test -j1 all
...
real    0m35.149s
user    0m29.338s
sys     0m6.036s
```

CI is slower, but I do not think that the difference is too significant.  I will always use `-j1` for now.  If we decide that the difference is significant, we can add a manual dispatch option to run tests sequentially when requested.

TravisCardwell:
The `c-expr-runtime` tests all failed with `Clang's type: <n/a>`.

Investigating, I noticed that the target is `x86_64-pc-windows`, which is suspicious.  While our `hs-bindgen` tests target `x86_64-pc-linux-musl` (using the vendored `musl` headers) in order to have matching golden test fixtures across all platforms, I think that `c-expr-runtime` tests should target Windows.  Target `x86_64-pc-windows` is not a MinGW target, so it probably defaults to MSVC, which GHC does not support.  GHC targets `x86_64-unknown-mingw32`, so I changed the Windows target to that.

The test setup code always tries to determine the vendored `musl` directory and issues a warning when the directory is not found, even on Windows, where the `musl` headers are not used even when found.  I refactored the code so that it only tries to determine the vendored `musl` directory on POSIX platforms, to make it clear that it is not (intended to be) used on Windows.

Note that this code only distinguishes between `Windows` and `Posix` platforms, and target `x86_64-pc-linux` is always used in the `Posix` case.  Perhaps this causes insufficient testing on macOS?

With these changes, `c-expr-runtime` tests now pass, in addition to the `hs-bindgen` tests.

TravisCardwell:
We are using symbolic links to link some manual headers to files in the examples:

```
$ find . -type l | xargs ls -l | sed 's|^[^/]*/||' | sort
manual/c/arrays.h -> ../../hs-bindgen/examples/golden/manual/arrays.h
manual/c/function_pointers.h -> ../../hs-bindgen/examples/golden/manual/function_pointers.h
manual/c/globals.h -> ../../hs-bindgen/examples/golden/manual/globals.h
manual/c/zero_copy.h -> ../../hs-bindgen/examples/golden/manual/zero_copy.h
```

It is working in CI, but this does not work in many Windows installations.  In my test VM, I will manually copy the files to work around it, but perhaps we should avoid using symbolic links.

TravisCardwell:
I am changing the manual `Makefile`, which builds the shared libraries, to always use `clang` (*not* `gcc`).  We use the `-fPIC` option, which notoriously does not work on Windows, but my tests indicate that we *can* use it with the MinGW target.  It just breaks with MSVC targets.

TravisCardwell:
On macOS, `hs-bindgen` can automatically set `SDKROOT`, but we still need to set it for `clang` to build shared libraries.  I documented this in the action.

TravisCardwell:
I updated the workflow based on the feedback from @jorisdral in https://github.com/well-typed/libclang/pull/29, as well as made changes for consistency with changes in https://github.com/well-typed/libclang/pull/32.  Thank you Joris!

* Removed branch restrictions on `pull_request` and `merge_group`
* Configured tested GHC versions so that we test *all* versions in the MQ as well as when pushing to `main`
* Changed the "Save cache" condition to match the new design: save on cache miss regardless of cancellation status

Note that there are two changes to the `libclang` workflow that are not made here:

* `libclang` now uses `cabal configure` to configure the project, but `hs-bindgen` still uses a `cabal.project.ci` file.
* `libclang` saves the cache immediately after building dependencies, but `hs-bindgen` saves the cache at the end because the `manual` section may also add things to the Cabal store.

PR #1259, opened 2025-11-07
Some small updates to the `c-qrcode` example

jorisdral:
See the commit messages


jorisdral:
Merging this because it's only small changes to improve QoL

PR #1264, opened 2025-11-07
Added `SelectNoDeclarationsMatched` Warning

### LINKED ISSUE #1194: Should we inform the user when nothing is selected?
At the moment, when the select predicate does not match any declaration, we go ahead and do not produce any output files (in preprocessor mode). I think this is desired behavior, but maybe we want to inform the user issuing a 'Warning' (or a 'Notice' since it will be a backend trace)?

bolt12:
Fixes #1194

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
@@ -4,6 +4,7 @@ module HsBindgen.Frontend.Pass.Select (
 
 import Data.Foldable qualified as Foldable
 import Data.List.NonEmpty qualified as NonEmpty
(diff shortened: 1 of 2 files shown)


bolt12:
Hmm I didn't check this, something must not be right in my PR? It seems I issue this warning even when the user didn't specify that he wants to select a predicate. That's why the tests are failing, I think?

edsko:
> Hmm I didn't check this, something must not be right in my PR? It seems I issue this warning even when the user didn't specify that he wants to select a predicate. That's why the tests are failing, I think?

No, there is always a selection predicate (by default, `--select-from-main-files`). These are either files that have only unsupported definitions (say a file with only a `typedef` around a `long double`), or something is wrong.

PR #1271, opened 2025-11-10
Nix: Resolve intricacies when overriding the version of LLVM used

dschrempf:
The dependency graph involving LLVM versions is complex. We want to override the used version of LLVM in some places, but not all. (If we completely change the LLVM version, even the Rust compiler itself is recompiled).

This commit carefully overrides the version of LLVM used in places where we need the specific LLVM version, but not in places were the default LLVM version is alright, avoiding unnecessary rebuilds.

- Also update the Nix Flake lock file.
- I also tested and updated the tutorial. There have been quite some user-facing changes.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,7 +31,9 @@
         "aarch64-darwin"
         "x86_64-darwin"
       ];
-      perSystem = import ./nix/hs-bindgen.nix { inherit nixpkgs overlays; };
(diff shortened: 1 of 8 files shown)


dschrempf:
I am putting this on the queue because it only affects Nix files; I have tested the derivations using the tutorial.

dschrempf:
(Fixed the tests so they can execute `hs-bindgen-cli`.)

Issue #827, opened 2025-06-28
Finish support for function pointers and write tests


edsko:
We mostly already support function pointers; for example, 

```c
typedef int (*foo)(int);
```

results in

```hs
newtype Foo = Foo
  { un_Foo :: F.FunPtr (FC.CInt -> IO FC.CInt)
  }
```

However, the _construction_ of such values is not yet really supported; we can either 

- [x] construct them by taking the address of a C function (https://github.com/well-typed/hs-bindgen/issues/46)
- [x] or from Haskell functions (#51). 

We also need to 

- [x] Go in the other direction: from function pointer to Haskell function ((#49) 
- [x] experiment with this and write some tests,
- [ ] add a manual section

edsko:
Marking this as high priority because without this feature function pointers, an important feature, are not all that useable.

edsko:
Technically #81 is part of this also, but leaving that as a separate ticket as we'll only be able to tackle that one with the high-level API (if at all).

edsko:
See also https://github.com/well-typed/hs-bindgen/issues/91#issuecomment-3135681148 .

edsko:
@jorisdral Maybe you can get back to this after @bolt12 finishes #1094 ?

edsko:
After discussing this with @bolt12 and @jorisdral again, we realised we need a slightly different design, inspired by how we deal with structs. I'll create a new ticket for that, which we will need to do before we can do some of the other issues here (in particular, it will be required for #51).

edsko:
I think this can be closed once we have a manual section, which @bolt12 is going to do.

PR #1300, opened 2025-11-18
Updated LowLevel/Functions manual documentation

bolt12:
This PR closes [#827](https://github.com/well-typed/hs-bindgen/issues/827).

### DIFF
--- a/manual/LowLevel/Functions.md
+++ b/manual/LowLevel/Functions.md
@@ -15,20 +15,16 @@ conventions: `safe` and `unsafe`. The distinction is important:
   because they avoid this overhead, but using `unsafe` incorrectly can lead to
   undefined behavior.
 
-Currently, `hs-bindgen` generates all function imports using the `safe` calling
-convention:
+To give users control over this choice, `hs-bindgen` generates two separate
+modules for function bindings:
 
-```hs
-foreign import ccall safe "B_function_name"
-  function_name :: CInt -> IO CInt
-```
+* `ModuleName.Safe` - contains all function imports using the `safe` calling
+  convention
+* `ModuleName.Unsafe` - contains all function imports using the `unsafe`
+  calling convention
 
-This is a conservative choice that ensures correctness in all cases. While it
-means we pay the cost of the safe calling convention even for functions that
-do not need it, this cost is typically negligible unless the function is called
-in a tight loop. Users who require the performance of `unsafe` imports can
-manually wrap the generated bindings with their own `unsafe` imports after
-verifying that the C function does not call back into Haskell.
+Both modules export identical APIs, differing only in their calling convention.
+Users can import from whichever module best suits their needs.
 
 ## Function addresses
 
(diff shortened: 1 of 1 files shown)


manual/LowLevel/Functions.md
```diff
@@ -193,6 +193,162 @@ apply1_union :: Apply1Union
...
+instance FromFunPtr ProgressUpdate_Deref where
+  fromFunPtr = fromProgressUpdate_Deref
+```
+
+> [!NOTE]
+> The `hs-bindgen-runtime` library provides Template Haskell utilities for
+> generating these instances manually when needed. See
+> `HsBindgen.Runtime.TH.Instances` for details. This can be useful when
+> working with function pointer types that `hs-bindgen` doesn't automatically
+> generate instances for, or when writing custom high-level bindings.
```
bolt12: Remove this.

manual/LowLevel/Functions.md
```diff
@@ -193,6 +193,162 @@ apply1_union :: Apply1Union
...
+
+Not every function pointer type receives these imports. `hs-bindgen` analyzes
+how each function pointer type is used in the API:
+
+- Types appearing as function parameters receive `"wrapper"` imports (you need
+  to create function pointers to pass as arguments)
+- Types for function pointer values receive `"dynamic"` imports (you
+  need to call returned function pointers)
+- Types appearing in struct or union fields receive both (they may flow in
+  either direction)
```
bolt12: Rephrase this, and mention that these instances are only generated if the function signature has at least one domain specific type (because otherwise they would be orphans and to avoid repeating the instances)

manual/LowLevel/Functions.md
```diff
@@ -193,6 +193,162 @@ apply1_union :: Apply1Union
...
+
+-- Preferred: automatic cleanup with withToFunPtr
+withToFunPtr myCallback $ \funPtr -> do
+  onProgressChanged (ProgressUpdate funPtr)
+
+-- Or manually manage the function pointer lifetime
+do
+  funPtr <- toFunPtr myCallback
+  onProgressChanged (ProgressUpdate funPtr)
+  freeHaskellFunPtr (un_ProgressUpdate funPtr)
```
bolt12: Change this example to use `bracket`

manual/LowLevel/Functions.md
```diff
@@ -193,6 +193,162 @@ apply1_union :: Apply1Union
...
+  void (*onReceived)(struct Measurement *data);
+  int (*validate)(struct Measurement *data);
+  void (*onError)(int errorCode);
+};
+
+void registerHandler(struct MeasurementHandler *handler);
+```
+
+Since these function pointer types appear in struct fields, they receive both
+wrapper and dynamic imports. This allows you to populate the struct with
```
bolt12: This is not quite right. Maybe say something like "Here we need both instances"

manual/LowLevel/Functions.md
bolt12, line 1: 1. Introduction
2. Safe vs Unsafe
3. Function addresses
3.1. current "Function pointers" _beceomes_ section 3
3.2. "Implicit function to pointer conversion" becomes 3.2
4. Conversion between Haskell functions and C functions
5. Userland CAPI: Introduction (we generate a C wrapper, yadayayada)
5.1 Structs by value
5.2 Static inline functions

manual/LowLevel/Functions.md
```diff
@@ -2,13 +2,37 @@
...
+foreign import ccall safe "B_function_name"
+  function_name :: CInt -> IO CInt
+```
+
+This is a conservative choice that ensures correctness in all cases. While it
+means we pay the cost of the safe calling convention even for functions that
+do not need it, this cost is typically negligible unless the function is called
+in a tight loop. Users who require the performance of `unsafe` imports can
+manually wrap the generated bindings with their own `unsafe` imports after
+verifying that the C function does not call back into Haskell.
```
edsko: This is not correct. We generate two modules: `.Safe` and `.Unsafe`; users can (must) choose whichever calling convention they want.

manual/LowLevel/Functions.md
```diff
@@ -193,13 +217,225 @@ apply1_union :: Apply1Union
...
+instance ToFunPtr ProgressUpdate_Deref where
+  toFunPtr = toProgressUpdate_Deref
+
+instance FromFunPtr ProgressUpdate_Deref where
+  fromFunPtr = fromProgressUpdate_Deref
+```
+
+A function pointer will have a `ToFunPtr` and `FromFunPtr` instance if at
+least one of its arguments contains at least one domain specific type. This
+check is done recursively so higher order functions will be inspected
```
edsko, line 278: I think an example of what constitutes a domain specific type, and what does not (i.e., for which we would _not_ generate an instance) would help clarify this.

manual/LowLevel/Functions.md
```diff
@@ -193,13 +217,225 @@ apply1_union :: Apply1Union
...
+
+The generated wrappers are named with a prefix based on the Haskell module
+name. For example, a C function `foo` might have a wrapper `B_foo` that we
+import instead:
+
+```c
+// Generated C wrapper
+void B_print_point ( struct point * arg1 ) {
+  print_point ( arg1 );
+}
```
edsko, line 400: That's not right, is it..? We use some kind of hash, to avoid name collisions?

bolt12: right, I was taking this from the paper, but that's no longer true

manual/LowLevel/Functions.md
bolt12: Right!

edsko (approved):
Looks great!

PR #1308, opened 2025-11-19
`HasBaseForeignType`

### LINKED ISSUE #1282: FFI Constructor Import Issue

For example:

1. Module RPM.Types defines:

```hs
```

2. Module RPM.Argv defines:

```hs
import qualified RPM.Types

```

```hs
import RPM.Argv
 ...
```

is still possible:

generate:

```hs
module RPM.Ds where
import qualified RPM.Types

```

```hs
module RPM.Ds (
  ) where
```

list.

jorisdral:
Resolves #1282

This is a WIP, not ready for review


jorisdral:
This is now superseded by  #1417

PR #1315, opened 2025-11-21
Deselect declarations transitively depending on omitted declarations

dschrempf:
Warn user in such cases.

Based on #1314 which should be merged first.

Travis and I agreed that trace messages should start with captial letters.

### DIFF
--- a/hs-bindgen/examples/golden/program-analysis/selection_omit_external.yaml
+++ b/hs-bindgen/examples/golden/program-analysis/selection_omit_external.yaml
@@ -1,0 +1,9 @@
+version:
+  hs_bindgen: 0.1.0
+  binding_specification: '1.0'
+target: any
+hsmodule: Example
+ctypes:
+- omit:
+    headers: program-analysis/selection_omit_external_root.h
+    cname: struct Omitted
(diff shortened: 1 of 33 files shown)


hs-bindgen/examples/golden/program-analysis/selection_omit.h
```diff
@@ -0,0 +1,7 @@
+struct Omitted {
```
dschrempf: Add another level here.

dschrempf:
Test 1: Prescriptive binding with "omit `Ommited`", but we use `Omitted` (directly or
transitively). Check for correct/expected message.

Test 2: External binding specification with "omit `Omitted`".
- Header `root`: Does define `Omitted` and matches against external binding specification.
- Header `a`: Does not use `Omitted` -> No messages at all.
- Header `b`: Uses `Omitted` -> We generate bindings for the declaration `Omitted` that
  was omitted by an external binding specification. At the moment don't issue a
  warning but create a ticket: We could emit a warning in this case.

dschrempf:
@TravisCardwell this is touching the ResolveBindingSpec module. In particular, it only emits the trace "ResolveBindingSpecOmittedType" when a type is omitted (and not on a use site).

Could you have a short look please? We also added a test with a trace expectation to check for this message.

Issue #1244, opened 2025-11-04
Cached `libclang-bindings` may not match current configuration


TravisCardwell:
I did not upgrade my development system for quite some time, just to make sure that I did not break anything before ICFP and Poland.  (I miss NixOS generations!)  I finally upgraded this weekend, and the upgrade changed my default LLVM/Clang from version 20 to 21.  (Clang/LLVM 20 is no longer available, until I install the packages for it.)

Attempts to build `hs-bindgen` failed with the following error:

```
libclang.so.20.1: cannot open shared object file: No such file or directory
```

Since our `libclang-bindings` package is now in a separate project, build artifacts are stored in the Cabal store (in `~/.cabal/store`), *not* locally (in `dist-newstyle`).  The version of `libclang.so` that is linked when we build our `libclang-bindings` package is not reflected in the package hash.  The version of `libclang.so` that is linked when first building for a given GHC version (and revision of the code) is therefore always used, regardless of any environmental changes that configure a different LLVM/Clang version.

In this case, I upgraded from LLVM/Clang 20 to 21.  My environment was configured to use LLVM/Clang 21, but old builds of our `libclang-bindings` package still linked to `libclang.so.20.1`, which no longer exists on my system.

I have previously tested against the many version of LLVM/Clang available on my system by simply adjusting environment variables.  Now that our `libclang-bindings` package is in a separate project, this no longer works correctly.  The configuration only affects the current project: when working on `hs-bindgen`, the packages in `hs-bindgen` are built with the current configuration, but the cached (stored) `libclang-bindings` dependency is used regardless of any configuration changes.

`pkg-config` dependency configuration is reflected in a package hash, but we do not use `pkg-config` to configure LLVM/Clang.  Might there be another way to make the package hash reflect the LLVM/Clang version?

I think that this affects anybody who uses Cabal normally.  (Nix builds that do not use a persistent Cabal store should not be affected, as dependency configuration is a proper input and not just global state.)

If we are unable to make the `libclang-bindings` package hash reflect the FFI dependencies, perhaps we should document this, telling users how to resolve issues.  My understanding, however, is that there is no tooling available for removing specific build artifacts from the Cabal store.  One can manually remove stuff, but this makes the `package.cache` invalid and could cause problems.  Is removing the whole store (for each GHC version that contains a problematic artifact) the only "safe" option?

[...]

PR #1316, opened 2025-11-21
Add information text to  output

### LINKED ISSUE #1312: `info libclang` prints nothing?
```
$ cabal run -v0 -- hs-bindgen-cli info libclang
$
```

@TravisCardwell , any thoughts on this?

@bolt12 , could I ask you to look into this once Travis has commented?

bolt12:
This PR fixes #1312

### DIFF
--- a/hs-bindgen/app/HsBindgen/Cli/Info/Libclang.hs
+++ b/hs-bindgen/app/HsBindgen/Cli/Info/Libclang.hs
@@ -34,12 +34,7 @@ import HsBindgen.Util.Tracer
 info :: InfoMod a
 info = mconcat [
       progDesc "Run libclang with empty input"
-    , footerDoc . Just . PP.reflow $ mconcat [
-          "This command provides a way to get output from libclang."
-        , " For example, use --clang-option=-v to see version and include"
-        , " search path information, taking into account any other Clang"
-        , " options and environment variables."
-        ]
+    , footerDoc . Just . PP.reflow $ mconcat infoHelpMessage
     ]
 
 {-------------------------------------------------------------------------------
--- a/hs-bindgen/src-internal/HsBindgen/Clang.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Clang.hs
@@ -8,6 +8,7 @@ module HsBindgen.Clang (
   , withClang'
     -- * Trace messages
   , ClangMsg(..)
+  , infoHelpMessage
   ) where
 
 import Data.Text qualified as Text
@@ -152,7 +153,7 @@ instance PrettyForTrace ClangMsg where
         | otherwise -> PP.textToCtxDoc diagnosticFormatted
       ClangSetupMsg   x -> prettyForTrace x
       ClangInvokedWithoutOptions ->
-        "libclang invoked successfully (no output produced)."
+        PP.cat $ map PP.textToCtxDoc infoHelpMessage
     where
       getFileNotFound :: Text -> Maybe Text
       getFileNotFound =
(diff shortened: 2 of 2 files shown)


hs-bindgen/app/HsBindgen/Cli/Info/Libclang.hs
```diff
@@ -58,9 +61,40 @@ exec :: GlobalOpts -> Opts -> IO ()
...
+            && unwrapVerbosity (tVerbosity tracerConfig) >= Info) $ do
+        traceWith (contramap (TraceFrontend . FrontendClang) tracer)
+                  ClangInvokedWithoutOptions
+
+      -- Use a filtering tracer that suppresses CXError_ASTReadError.
+      -- This error is expected when diagnostic options like -v or
+      -- -print-target-triple are used, as Clang exits early without
+      -- creating a full AST after printing the requested information.
+      --
+      let clangTracer = squelchUnless (not . isASTReadError)
```
dschrempf: Is the `isASTReadError` a trace (or is it simply output from `libclang`)? If it is a trace, this can be done in a simpler way by using `CustomLogLevel`.

There are some examples (`CustomLogLevelSetting`) in `TraceMsg.hs`.

bolt12: hmmm I will have to check it out! Thank you!

bolt12: Much better! Thank you!

hs-bindgen/app/HsBindgen/Cli/Info/Libclang.hs
```diff
@@ -56,11 +54,42 @@ parseOpts = Opts <$> parseClangArgsConfig
...
+    -- logging
+    --
+    tracerConfigWithoutASTReadError =
+      tracerConfig
+        { tCustomLogLevel = CustomLogLevel $ \trace actualLevel ->
+            case trace of
+              TraceFrontend (FrontendClang (ClangErrorCode (SimpleEnum x)))
+                | Just CXError_ASTReadError <- simpleFromC x -> Debug
+              _ -> unCustomLogLevel (tCustomLogLevel tracerConfig) trace actualLevel
+        }
```
edsko, line 95: @TravisCardwell I've approved this PR and told @bolt12 to go ahead and merge this, but this `tracerConfigWithoutASTReadError` does feel a bit weird to me. It's only on the CLI, and only for this info command, so not a big deal, but nonetheless - do you have any thoughts on this?

TravisCardwell: I still do not understand why we are getting the error.

I did not think I got the error before, but I just checked with LLVM/Clang 14 and see it with that version as well.  I likely just forgot about the error.

Does this error happen because the header is empty?  I tried adding `#define HS_BINDGEN_NOP_H` to the header, but I still see the error.

Perhaps `libclang` is trying to parse the printed output?  Surely not, but I do not know how to easily check this.

Suppressing the error (in this CLI command only) by changing the level to debug sounds like a suitable mitigation to me.

----

BTW, when I first tried building with LLVM/Clang 14, it failed because of the cached build for LLVM/Clang 21 in the Cabal store:

```
Template Haskell error: [Warning] [HsBindgen] [builtin-include-dir] clang version mismatch:
  libclang version: clang version 21.1.6
  clang version:    clang version 14.0.6
```

This issue (#1244) is annoying, and it is frustrating that there is nothing we can do about it.

bolt12: From what I understood, Clang exits early without fully processing/parsing the input file, i.e. no AST is created and `clang_parseTranslationUnit` returns `CXError_ASTReadError` because no AST was built.

So in essence, it prints info and exits without completing the AST construction, regardless of file content

PR #1322, opened 2025-11-25
Introduce --overwrite-files flag

### LINKED ISSUE #1117: We should only overwrite output files, if the user specifies `--force`
Of force, this is up to discussion. Let me know, if you do not want this implemented.

Self assigning.

bolt12:
This closes #1117

### DIFF
--- a/hs-bindgen/app/HsBindgen/Cli/Info/UseDeclGraph.hs
+++ b/hs-bindgen/app/HsBindgen/Cli/Info/UseDeclGraph.hs
@@ -17,8 +17,8 @@ import Options.Applicative hiding (info)
 
 import HsBindgen
 import HsBindgen.App
-import HsBindgen.Artefact
 import HsBindgen.Config
+import HsBindgen.DelayedIO
 import HsBindgen.Frontend.RootHeader
 import HsBindgen.Imports
 
--- a/hs-bindgen/app/HsBindgen/Cli/ToolSupport/Literate.hs
+++ b/hs-bindgen/app/HsBindgen/Cli/ToolSupport/Literate.hs
@@ -24,9 +24,9 @@ import Text.Read (readMaybe)
 
 import HsBindgen
 import HsBindgen.App
-import HsBindgen.Artefact
 import HsBindgen.Backend.SHs.AST
 import HsBindgen.Config
+import HsBindgen.DelayedIO
 import HsBindgen.Errors
 import HsBindgen.Frontend.RootHeader
 
--- a/hs-bindgen/src-internal/HsBindgen.hs
+++ b/hs-bindgen/src-internal/HsBindgen.hs
@@ -21,14 +21,10 @@ module HsBindgen
   , hsBindgenE
   ) where
 
-import Control.Monad.Except (MonadError (..))
-import Control.Monad.Trans.Except (ExceptT, runExceptT)
+import Control.Monad.Except (MonadError (..), withExceptT)
+import Control.Monad.Trans.Except (runExceptT)
 import Data.Map qualified as Map
-import System.Directory (createDirectoryIfMissing, doesDirectoryExist,
-                         doesFileExist)
 import System.Exit (ExitCode (..), exitWith)
-import System.FilePath (takeDirectory, (</>))
-import Text.SimplePrettyPrint ((<+>))
 import Text.SimplePrettyPrint qualified as PP
 
 import HsBindgen.Artefact
@@ -36,9 +32,9 @@ import HsBindgen.Backend
 import HsBindgen.Backend.HsModule.Render
 import HsBindgen.Backend.SHs.AST
 import HsBindgen.BindingSpec.Gen
-import HsBindgen.BindingSpec.Private.V1 qualified as BindingSpec
 import HsBindgen.Boot
 import HsBindgen.Config.Internal
+import HsBindgen.DelayedIO
 import HsBindgen.Errors (throwPure_TODO)
 import HsBindgen.Frontend
 import HsBindgen.Frontend.Analysis.IncludeGraph qualified as IncludeGraph
@@ -103,7 +99,7 @@ hsBindgenE
           Right (r, as) -> do
             -- Before creating directories or writing output files, we verify
             -- adherence to the provided policies.
-            mapM_ checkPolicy as
+            withExceptT (BindgenDelayedIOError) $ mapM_ checkPolicy as
             liftIO $ executeFileSystemActions tracer as
             pure r
 
@@ -117,40 +113,6 @@ hsBindgenE
       , tShowCallStack  = tShowCallStack tracerConfig
       }
 
-    checkPolicy :: DelayedIO -> ExceptT BindgenError IO ()
-    checkPolicy (PutStrLn   _) = pure ()
-    checkPolicy (WriteFile fd) = case fd.location of
-      UserSpecified path -> do
-        let baseDir = takeDirectory path
-        dirExists  <- liftIO $ doesDirectoryExist baseDir
-        fileExists <- liftIO $ doesFileExist path
-        unless dirExists $
-          throwError $ BindgenFileSystemError $ DirectoryDoesNotExist baseDir
-        when (fileExists && fd.fileOverwritePolicy == DoNotOverwriteFiles) $
-          throwError $ BindgenFileSystemError $ FileAlreadyExists path
-      RelativeFileLocation RelativeToOutputDir{..} -> do
-        let path = outputDir </> localPath
-        dirExists  <- liftIO $ doesDirectoryExist outputDir
-        fileExists <- liftIO $ doesFileExist path
-        unless (dirExists || outputDirPolicy == CreateOutputDirs ) $
-          throwError $ BindgenFileSystemError $ DirectoryDoesNotExist outputDir
-        when (fileExists && fd.fileOverwritePolicy == DoNotOverwriteFiles) $
-          throwError $ BindgenFileSystemError $ FileAlreadyExists path
-
-    executeFileSystemActions :: Tracer BindgenMsg -> [DelayedIO] -> IO ()
-    executeFileSystemActions tracer as =
-      forM_ as $ \case
-        PutStrLn   x -> putStrLn x
-        WriteFile fd -> do
-          let path = fileLocationToPath fd.location
-          traceWith tracer $ BindgenWriteFile path fd.description
-          -- Creating the directory is justified by checking the policy first.
-          createDirectoryIfMissing True (takeDirectory path)
-          case fd.content of
-            TextContent str        -> writeFile path str
-            BindingSpecContent ubs -> BindingSpec.writeFile path ubs
-
-
 {-------------------------------------------------------------------------------
   Custom build artefacts
 -------------------------------------------------------------------------------}
@@ -161,7 +123,7 @@ writeIncludeGraph pol mPath = do
     (predicate, includeGraph) <- IncludeGraph
     let rendered = IncludeGraph.dumpMermaid predicate includeGraph
     case mPath of
-      Nothing   -> Lift $ delay $ PutStrLn rendered
+      Nothing   -> Lift $ delay $ WriteToStdOut $ TextContent rendered
       Just path -> Lift $ write pol "include graph" (UserSpecified path) rendered
 
 -- | Write @use-decl@ graph to file.
@@ -171,7 +133,7 @@ writeUseDeclGraph pol mPath = do
     useDeclGraph <- UseDeclGraph
     let rendered = UseDeclGraph.dumpMermaid index useDeclGraph
     case mPath of
-      Nothing   -> Lift $ delay $ PutStrLn rendered
+      Nothing   -> Lift $ delay $ WriteToStdOut $ TextContent rendered
       Just path -> Lift $ write pol "use-decl graph" (UserSpecified path) rendered
 
 -- | Get bindings (single module).
@@ -237,7 +199,7 @@ writeBindingSpec fileOverwritePolicy path = do
             , fileOverwritePolicy
             , content     = BindingSpecContent  bindingSpec
             }
-    Lift $ delay $ WriteFile fileDescription
+    Lift $ delay $ WriteToFile fileDescription
 
 -- | Create test suite in directory.
 writeTests :: FilePath -> Artefact ()
@@ -251,15 +213,15 @@ writeTests _testDir = do
     --     hsDecls
     --     moduleBaseName
     --     testDir
-    throwPure_TODO 22 "Test generation  integrated into the artefact API"
+    throwPure_TODO 22 "Test generation integrated into the artefact API"
 
 {-------------------------------------------------------------------------------
   Helpers
 -------------------------------------------------------------------------------}
 
-write :: FileOverwritePolicy -> String -> FileLocation -> String -> ArtefactM ()
+write :: FileOverwritePolicy -> String -> FileLocation -> String -> DelayedIOM ()
 write pol what loc str =
-    delay $ WriteFile $ FileDescription what loc pol (TextContent str)
+    delay $ WriteToFile $ FileDescription what loc pol (TextContent str)
 
 writeByCategory ::
      FileOverwritePolicy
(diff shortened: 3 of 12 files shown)


hs-bindgen/src-internal/HsBindgen.hs
```diff
@@ -171,22 +171,30 @@ writeTests testDir = do
...
       traceWith tracer $ RunArtefactWriteFile what path
       liftIO $ do
-        createDirectoryIfMissing True $ takeDirectory path
-        writeFile path str
+        -- Check if file exists and policy forbids overwriting
+        fileExists <- doesFileExist path
+        case fop of
+          ProtectExistingFiles
+            | fileExists ->
+            throwIO (FileAlreadyExistsException path)
```
dschrempf: I think this should be an Error trace instead (but others may disagree). ArtefactM has access to the tracer, so this should not be a problem.

hs-bindgen/src-internal/HsBindgen.hs
```diff
@@ -171,22 +171,30 @@ writeTests testDir = do
...
+          _ -> do
+            createDirectoryIfMissing True $ takeDirectory path
+            writeFile path str
 
 writeByCategory ::
      String
   -> FilePath
+  -> FileOverwritePolicy
```
dschrempf: The order of arguments is different, compared to `write`, could we have the same order? The same is true for `genBindingSpec`.

hs-bindgen/src-internal/HsBindgen/BindingSpec/Gen.hs
```diff
@@ -46,12 +49,22 @@ import HsBindgen.Language.Haskell qualified as Hs
...
-genBindingSpec hsModuleName path getMainHeaders omitTypes =
-      BindingSpec.writeFile path
+genBindingSpec hsModuleName path fop getMainHeaders omitTypes =
+      (\ubs -> do
+        -- Check if file exists and policy forbids overwriting
+        fileExists <- doesFileExist path
+        case fop of
+          ProtectExistingFiles
+            | fileExists ->
+            throwIO (FileAlreadyExistsException path)
```
dschrempf: Also here, this may be an Erro trace (@edsko, what do you think?).

hs-bindgen/src-internal/HsBindgen/Config.hs
```diff
@@ -46,7 +48,8 @@ data Config_ path = Config {
   , programSlicing  :: ProgramSlicing
 
     -- * Backend
-  , haddockPathStyle :: PathStyle
+  , haddockPathStyle    :: PathStyle
+  , fileOverwritePolicy :: FileOverwritePolicy
```
dschrempf: This should probably not be part of the common configuration, because it is unused by TH and Literate mode.

hs-bindgen/src-internal/HsBindgen/Config/Internal.hs
```diff
@@ -87,9 +94,16 @@ data FrontendConfig = FrontendConfig {
 data BackendConfig = BackendConfig {
       backendTranslationOpts :: TranslationOpts
     , backendHaddockConfig   :: HaddockConfig
+    , backendFileOverwrite   :: FileOverwritePolicy
     }
   deriving stock (Show, Eq, Generic)
-  deriving anyclass Default
+
+instance Default BackendConfig where
```
dschrempf: Why did you change the `deriving` clause?

hs-bindgen/src-internal/HsBindgen/Config/Internal.hs
```diff
@@ -101,3 +115,28 @@ checkBackendConfig tracer backendConfig =
...
+  deriving (Show, Eq)
+
+instance Default FileOverwritePolicy where
+  def = ProtectExistingFiles
+
+data FileAlreadyExistsException =
+  FileAlreadyExistsException FilePath
+  deriving Show
+
+instance Exception FileAlreadyExistsException where
```
dschrempf: As stated above, I think this should be an error trace instead; but you must have made a conscious decision, so I am curious about your reasoning!

hs-bindgen/src-internal/HsBindgen.hs
edsko: He is using an exception because he wanted to terminate the program as soon as we failed to write one file, and we are currently doing the same for the guard on creating the directory. However, I have asked Armando to look at an alternative approach where we collect all files we want to write first, check if _any_ of them exists, and only when none do, then write them all; I hope is that this will be a better experience for the user, avoiding ending up with a bunch of new files and old files. If that works out, an error trace would indeed make sense.

hs-bindgen/src-internal/HsBindgen/Config/Internal.hs
bolt12: This seems to be an artefact of some previous approach I had, thanks for spotting it.

hs-bindgen/src-internal/HsBindgen/Config.hs
bolt12: I see, that's a good point. I will see to it

bolt12:
I have successfully implemented the solution where if there are any files already present, then the command is going to fail without written anything! Reviews are welcome!

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -140,6 +168,21 @@ runArtefacts
       FinalModuleSafe     -> liftIO backendFinalModuleSafe
       FinalModuleUnsafe   -> liftIO backendFinalModuleUnsafe
       FinalModules        -> liftIO backendFinalModules
+      -- File writes.
+      PendingFileWrite what path content -> do
```
dschrempf: This is a nice approach! I actually think we could generalize this so that also directory creation (`--create-output-dirs`) is a "pending file system operation" (or so?).

I think like this we can really ensure that we do not touch the file system when there is an error [trace].

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -140,6 +168,21 @@ runArtefacts
...
+          (ProtectExistingFiles, _)
+            | fileExists ->
+              throwError (FileAlreadyExistsException path)
+          (_, TextContent c) ->
+            liftIO $ do
+              traceWith tracer $ RunArtefactWriteFile what path
+              createDirectoryIfMissing True $ takeDirectory path
+              writeFile path c
+          (_, BindingSpecContent ubs) ->
+            liftIO $ BindingSpec.writeFile path ubs
```
dschrempf: Sorry, that's a nit: Could we maybe move the implementation into a where clause or so? I really like how each branch has only one line, at the moment.

hs-bindgen/src-internal/HsBindgen/Artefact.hs
dschrempf: (Not saying this should be part of the PR).

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -93,6 +101,23 @@ data ArtefactEnv = ArtefactEnv {
...
+--
+-- > createDirectoryIfMissing True $ takeDirectory path
+-- > writeFile path str
+--
+-- If it's BindingSpecContent then, depending on the FileOverwritePolicy, then
+-- this code is going to run:
+--
+-- > BindingSpec.writeFile path ubs
+--
+data WriteContent
```
dschrempf: Maybe just `Content`, or `FileContent`?

hs-bindgen/src-internal/HsBindgen.hs
```diff
@@ -171,32 +169,27 @@ writeTests testDir = do
...
-write :: String -> Maybe FilePath -> String -> ArtefactM ()
-write _    Nothing     str = liftIO $ putStrLn str
-write what (Just path) str = do
-      tracer <- artefactTracer <$> ask
-      traceWith tracer $ RunArtefactWriteFile what path
-      liftIO $ do
-        createDirectoryIfMissing True $ takeDirectory path
-        writeFile path str
+write :: String -> Maybe FilePath -> String -> Artefact ()
+write _    Nothing str     = liftIO $ putStrLn str
```
dschrempf: Sorry for the nit, I think we  should either align all arguments or none.

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -140,6 +168,21 @@ runArtefacts
       FinalModuleSafe     -> liftIO backendFinalModuleSafe
       FinalModuleUnsafe   -> liftIO backendFinalModuleUnsafe
       FinalModules        -> liftIO backendFinalModules
+      -- File writes.
+      PendingFileWrite what path content -> do
+        fileExists <- liftIO $ doesFileExist path
+        case (fop, content) of
+          (ProtectExistingFiles, _)
+            | fileExists ->
+              throwError (FileAlreadyExistsException path)
```
dschrempf: I think this should be an error trace. (Bear with me, I might be wrong; but I think we check for Error traces after each artefact, so we should be fine).

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -140,6 +168,21 @@ runArtefacts
...
+          (_, TextContent c) ->
```
dschrempf: I am unsure if this is having the desired effect: Write all files, or none.

The `PendinfFileWrite` artefacts will still be executed one after the other, and if one fails in the middle, the earlier ones will have happened already. Is that right, or am I missing something?

I think we could have a "file system operation" storage in the artefact monad. Only at the end of the execution, when no Error trace has happened, these file system operations are executed one after the other.

dschrempf (commented):
Thanks! Sorry for my nitty reviews, I have become fond of the Artefact module...

I think the "pending write" idea is a great one, I am just unsure if it has the desired effect (write all files, or none) in the way it is implemented at the moment.

I think we must have a "file system operation" storage in the artefact monad. Only at the end of the execution, when no Error trace has happened, these file system operations are executed one after the other.

hs-bindgen/src-internal/HsBindgen/Artefact.hs
bolt12: This is already an error trace. I modified Trace exception to include this special case!

hs-bindgen/src-internal/HsBindgen/Artefact.hs
bolt12: Yes, I think you are right and I misunderstood what the code was doing!

bolt12: Yes, I think you are right and I misunderstood what the code was doing!

hs-bindgen/src-internal/HsBindgen/Artefact.hs
bolt12: Yeah I can do it as part of this PR, just need to polish the idea and actually make it work! 😛

bolt12: Done!

hs-bindgen/src-internal/HsBindgen/TH/Internal.hs
```diff
@@ -228,6 +228,10 @@ toBindgenConfigTH packageRoot config = do
         bindgenConfig =
           toBindgenConfig
             (toFilePath packageRoot <$> config)
+            DoNotCreateDirStructure
```
dschrempf: Since you define conservative defaults anyway, we could use the defaults here, and move the comments to the definitions of the Default instances?

hs-bindgen/src-internal/HsBindgen/Config/Internal.hs
```diff
@@ -87,6 +91,8 @@ data FrontendConfig = FrontendConfig {
 data BackendConfig = BackendConfig {
       backendTranslationConfig :: TranslationConfig
     , backendHaddockConfig     :: HaddockConfig
+    , backendOutputDirPolicy   :: OutputDirPolicy
```
dschrempf: I think it is OK to have the policies in the backend configuration. On the other hand, then they have to be provided even when we do not create output files (e.g., TH mode).

We could also provide them with the respective artefacts, but then we have to do so for every artefact involving file output. What do you think?

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -89,12 +95,35 @@ instance MonadIO Artefact where
...
     }
 
+-- | A file system action to be executed
+data FileSystemAction =
+      CreateDir FilePath
+    | WriteFile String FilePath FileContent
+
+-- | Content to be written to a file
+--
+-- If it's TextContent then, depending on the FileOverwritePolicy, then this
```
dschrempf: Comment needs to be fixed.

hs-bindgen/src-internal/HsBindgen/Config/Internal.hs
dschrempf: (This would have the advantage that we do not have to provide values to `toBindgenCOnfig`).

hs-bindgen/src-internal/HsBindgen.hs
```diff
@@ -177,25 +233,20 @@ writeTests testDir = do
...
 writeByCategory ::
      String
   -> FilePath
   -> BaseModuleName
   -> ByCategory String
-  -> ArtefactM ()
+  -> Artefact ()
 writeByCategory what hsOutputDir moduleBaseName =
-    mapM_ (uncurry writeCategory) . Map.toList . unByCategory
+    Foldable.foldl' (>>) (pure ()) . map (uncurry writeCategory) . Map.toList . unByCategory
```
dschrempf: Is there a reason why you have removed `mapM_`? I think it is nicer than a fold.

hs-bindgen/app/HsBindgen/Cli/ToolSupport/Literate.hs
```diff
@@ -62,6 +64,9 @@ parseOpts = do
 
     input  <- strArgument $ metavar "IN"
     output <- strArgument $ metavar "OUT"
+
+    outputDirPolicy <- parseOutputDirPolicy
```
dschrempf: I think this should not be a command line option for literate mode. I am unsure how literate mode works in detail, but I think GHC uses a new tmp dir, so if this is overwriting something, it may be a bug. I would try using the conservative options here.

hs-bindgen/app/HsBindgen/Cli/Info/UseDeclGraph.hs
```diff
@@ -48,6 +50,8 @@ parseOpts =
       <*> parseBaseModuleName
       <*> optional parseOutput'
       <*> parseInputs
+      <*> parseOutputDirPolicy
```
dschrempf: I think we should also go over these client commands (`IncludeGraph`, `UseDeclGraph`, `Frontend` etc.). Do they actually use the command line options at all, or is it better if we just use default values (see literate mode below). Of course, this could be a separate ticket and PR, just saying.

I just want to avoid that we offer command line options for things we don't even use in the command.

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -103,21 +132,24 @@ data ArtefactEnv = ArtefactEnv {
...
 runArtefacts
   tracer
   tracerStateRef
   BootArtefact{..}
   FrontendArtefact{..}
   BackendArtefact{..}
-  artefact = runReaderT (runExceptT (runArtefact artefact)) env
+  artefact = do
+    (result, _, actions) <- runRWST (runExceptT (runArtefact artefact)) env ()
+    return (result, actions)
```
dschrempf: Is there a reason why you return the actions and not perform them here?

I have the feeling they should be performed here. This also makes sense in that the comment at `FileContent` describes what the file operations do, but the code is not in the module but in `HsBindgen.hs`.

Also, I mentioned this already but I think we didn't discuss it properly yet: Why do we use an exception instead of an error trace? To my knowledge, we should avoid panics if possible and think about `hs-bindgen` as if it were a compiler.

The `Bind` operation of `Artefact` checks for error traces, and we could use that machinery and just emit an error trace.

hs-bindgen/src-internal/HsBindgen.hs
```diff
@@ -89,6 +97,56 @@ hsBindgen
...
+                 | Just _ <- mbp -> pure mbp
+                 | otherwise -> do
+                   fileExists <- doesFileExist path
+                   pure $ if fileExists
+                             then Just path
+                             else mbp
+               _ -> pure mbp
+           ) Nothing actions
+  case outputDirPolicy of
+    DoNotCreateDirStructure
```
dschrempf: I thought much less about this than you, and there may be subtleties involved I
do not see; but I have two remarks:

- Could we do this in reverse order? Only check all dirs and paths if we configured to protect them, respectively.

- Doesn't the right fold always traverse the complete list of actions, even if it finds a conflicting file early on? I think it is enough to stop at the first existing directory.

hs-bindgen/src-internal/HsBindgen/Config/Internal.hs
bolt12: hmm I took them out of the backend config because of TH, and I agree that it isnt obvious what's the best approach. So I'd say let's keep it like this unless we need to add any more flags like these 2

hs-bindgen/src-internal/HsBindgen.hs
bolt12: I think this is some rebase artefact I don't remember changing this

hs-bindgen/app/HsBindgen/Cli/ToolSupport/Literate.hs
bolt12: Okay, thank you!

hs-bindgen/app/HsBindgen/Cli/Info/UseDeclGraph.hs
bolt12: I didn't really think it through when doing this, was just type checking to make the compiler happy. But I think you are right, probably the conservative values would be better here

hs-bindgen/src-internal/HsBindgen.hs
bolt12: 1. Yeah we could, no real reason why I decided to do this :)

bolt12: 2. Yes, however it will ignore all others if one is found. I can write an explicit recursion function that does what you suggest

PR #1324, opened 2025-11-25
Many minor fixes to the `hs-bindgen` manual

### LINKED ISSUE #1100: Work through of `hs-bindgen` manual
The `hs-bindgen` manual keeps diverging from the implementation.

We should skim through the manual, and pick up all issues we have to fix.

For example,
- [ ] the `Invocation` section is too general.

dschrempf:
This PR contains a lot of minor changes to the manual and closes #1100.

Sections that I didn't read in detail:
- [x] 03-Enums.md
- [x] 04-Unions.md (DONE; fixed person structure, see comment below)
- [x] 05-ZeroCopy.md
- [ ] 06-Functions.md (see comment below)
- [ ] 07-Globals.md (see comment below)
- [x] 08-Macros.md

I have only skimmed through the function and global sections since they are very long and discuss things others know more about than I do. Should we create separate tickets for those, or leave them as they are?

### DIFF
--- a/manual/LowLevel/Translation/06-Functions.md
+++ b/manual/LowLevel/Translation/06-Functions.md
@@ -6,38 +6,65 @@ This chapter discusses how `hs-bindgen` generates bindings for C functions.
 
 ## Safe vs unsafe foreign imports
 
-When importing a C function, GHC allows us to choose between two calling
-conventions: `safe` and `unsafe`. The distinction is important:
+When importing a C function, we can choose between two calling conventions:
+`safe` and `unsafe`. The [Haskell 2010 Language Report
+states](https://www.haskell.org/onlinereport/haskell2010/haskellch8.html):
 
-* **Safe** foreign imports may call back into Haskell code. They are more
-  expensive.
-* **Unsafe** foreign imports may not call back into Haskell. They are faster
-  because they avoid this overhead, but using `unsafe` incorrectly can lead to
-  undefined behavior.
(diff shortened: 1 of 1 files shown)


dev/ci.md
```diff
@@ -1,3 +1,5 @@
+# Continues development, continuous integration
```
dschrempf: typo

dschrempf:
I have a question wrt `Unions.md`. 

We discuss the structure `person`

```c
struct person {
  char* name;
  int occupation_tag;

  union occupation {
    // .. as before ..
  } occupation;
};
```

and its corresponding bindings

```haskell
data Person = Person
  { person_name           :: Ptr CChar
  , person_occupation_tag :: CInt
  , person_occupation     :: Occupation
  }
```

However, this structure _does not exist_ in the actual C file. Instead, we use a
pointer to `person`, but do not provide a definition:

```c
typedef union occupation {
  struct student {
    char* university;
    int year;
  } student;

  struct employee {
    char* company;
    struct person* supervisor;
    int salary;
  } employee;
} occupation;
```

The relevant generated binding is empty!

```haskell
{-| __C declaration:__ @person@
    __defined at:__ @manual_examples.h:77:12@
    __exported by:__ @manual_examples.h@
-}
data Person
```

Thoughts?

edsko:
Hmm, that feels like it's just an omission. Without the definition of `person`, it's not at all clear how to use the `union`; we need that `tag`.

Issue #1298, opened 2025-11-18
Check current target against target mentioned in prescriptive binding spec


TravisCardwell:
We currently ignore the `target` in prescriptive binding specs entirely. _If_ it is present, we should verify that it matches the current target (making it optional is part of https://github.com/well-typed/hs-bindgen/issues/1524); a mismatch should trace an error message.

~~A option with CLI flag `--ignore-target` should be added.  When specified the above traces should use level `Notice` instead.~~

TravisCardwell:
We decided to *not* implement `--ignore-target` (at this time).

edsko:
Obsolete: we removed `target` altogether (#1547; cc @TravisCardwell ).

Issue #1232, opened 2025-10-28
Add target to binding specifications


TravisCardwell:
Some binding specifications may be specific to a target, but it is also possible for some binding specifications to work across multiple targets.  Recording the target in generated binding specifications would be helpful, but we need to decide how/if that information is used.

Here are some ideas:

* Support target-specific and target-agnostic binding specifications

    For this, we could add an *optional* top-level `target`:

    ```yaml
    version:
      hs_bindgen: 0.1.0
      binding_specification: '1.0'
    target: x86_64-pc-linux
    ctypes:
      ...
    hstypes:
      ...
    ```

    If an external or prescriptive binding specification specifies a target, it must match the build target.  When merging two binding specifications, targets must match exactly when both are specified.

* Make all binding specifications target-specific

    For this, the same YAML is used, but `target` is required.

    All external and prescriptive binding specifications must match the build target.  When merging two binding specifications, targets must match exactly.

* Make the target informational only

    Perhaps it would be good to put the `target` in an object that indicates that it is informational.  It is of course optional.

    ```yaml
    version:
      hs_bindgen: 0.1.0
      binding_specification: '1.0'
    meta:
      target: x86_64-pc-linux
    ctypes:
      ...
    hstypes:
      ...
    ```

    All external and presriptive binding specifications are used regardless of any target specified.  When merging two binding specifications, we should probable decide between `First` and `Last` behavior.

In the future, we would like to make it "easy" to generate bindings for large projects that have dependencies.  We have talked about using Cabal for dependency resolution.  For a given dependency, perhaps we can check if there is a prescriptive binding specification that should be used to generate the bindings and binding specification for the build target.  To make this work when there are differences across targets, perhaps the target should *not* be informational.

Perhaps each project can specify a *list* of prescriptive binding specifications.  For the first option, we could first check for a binding specification that specifies the build target and then check for a binding specification that does not specify a target if one is not found.  For the second option, we could just check for a binding specification that specifies the build target.

edsko:
Discussing this with @TravisCardwell .

1. We will make `target` required, but allow value `any`, to emphasize these are not platform specific:

```yaml
target: any
```

[...]

PR #1330, opened 2025-11-27
Refactor `Target`, add to binding specifications

TravisCardwell:
This PR refactors how we deal with targets.  We previously had a fixed set of supported but `Target`s that was used with an arbitrary `TargetEnv`.  The `TargetEnv` type is removed, and significant environments are now distinguished in `Target`.  `Target` is therefore now a sum type that enumerates the complete set of supported platforms.

Function `targetTriple` associates each `Target` with a *canonical* target triple string.  Function `parseTargetTriple` parses these strings, only accepting the exact canonical target triples.  Users *must* use canonical target triple strings in command-line arguments and binding specifications.

Currently, the supported targets are as follows:

| `Target`                     | `targetTriple`           |
| :---                         | :---                     |
| `Target_Linux_GNU_X86_64`    | `x86_64-pc-linux-gnu`    |
| `Target_Linux_GNU_X86`       | `i386-pc-linux-gnu`      |
| `Target_Linux_GNU_AArch64`   | `aarch64-pc-linux-gnu`   |
| `Target_Linux_Musl_X86_64`   | `x86_64-pc-linux-musl`   |
| `Target_Linux_Musl_AArch64`  | `aarch64-pc-linux-musl`  |
| `Target_Windows_MSVC_X86_64` | `x86_64-pc-windows-msvc` |
| `Target_Windows_GNU_X86_64`  | `x86_64-pc-windows-gnu`  |
| `Target_Darwin_X86_64`       | `x86_64-apple-darwin`    |
| `Target_Darwin_AArch64`      | `aarch64-apple-darwin`   |

New CLI command `hs-bindgen-cli info targets` lists the supported targets (canonical target triples).

We have been treating targets as something that only concerns cross-compilation.  The (host) target may have a significant effect on how bindings are generated, however, even when not doing cross-compilation.  With this PR, we *always* determine the current target.  We previously only checked that targets specified by users are supported, but we generated bindings using the host target even if it is not supported.  Now, we check the host target when users do not specify one.

Function `parseTargetTripleLenient` is only used to parse the host target triple as reported by LLVM/Clang, translating it to one of the supported targets.  Target triples are a mess, so this parser is *not* expected to be perfect.  In cases where the reported target triple cannot be parsed, users should configure one of the canonical target triples.  Unit tests are added for this function, including all of the target triples that I have observed on Windows and macOS.

Now, the Clang arguments and current target are determined together.  First, the Clang arguments are determined using the `ClangConfig`.  If the target is specified, then those Clang arguments and the target are used.  If the target is not specified, however, those Clang arguments are used to call `libclang` and get the current target triple, and the lenient parser is used to attempt to translate the current target triple to a supported target.  If it can be translated to a supported target, we then reconstruct Clang arguments using that target.  Subsequent calls therefore use one of our canonical target triples.

Support for the `BINDGEN_EXTRA_CLANG_ARGS_<TARGET>` environment variable is removed.  Please see the explanation in [this comment](https://github.com/well-typed/hs-bindgen/issues/1232#issuecomment-3584567613).

This implementation throws an `UnableToDetermineTargetException` exception if the call to `libclang` fails or the target triple returned by `libclang` does not translate to a supported target.  This is done so that we do not continue with a default/inappropriate or unsupported target.

Binding specifications now specify a target, either `any` or a canonical target triple for one of the supported targets, as described in the issues.  Here is the current status:

* [x] Generated binding specifications specify the current target
    * [ ] In the future, we may want to specify `any` in the case that the generated bindings are not target-specific. #1334
* [x] Prescriptive binding specifications with a target that matches the current target are used.
* [x] Prescriptive binding specifications with a target that does not match the current target result in an error.  The error is not immediate, so an empty prescriptive binding specification is used.
* [x] Prescriptive binding specifications with an `any` target are used.
    * [ ] In a future PR, we need to trace an error if any of the generated bindings are target specific in this case. #1298
* [x] External binding specifications with a target that matches the current target are used.
* [x] External binding specifications with a target that does not match the current target result in an error.  Such an external binding specification is ignored (not merged).
* [x] External binding specifications with an `any` target are used.
* ~~The `--ignore-target` option is not added yet.  This should be done in a future PR.~~

Golden tests for binding specifications are refactored, with tests for targets added.  Some `TODO` comments are added for tests that cannot be done yet.

### DIFF
--- a/hs-bindgen/app/HsBindgen/Cli/Info/Targets.hs
+++ b/hs-bindgen/app/HsBindgen/Cli/Info/Targets.hs
@@ -10,6 +10,7 @@ module HsBindgen.Cli.Info.Targets (
   , exec
   ) where
 
+import Data.List qualified as List
 import Options.Applicative hiding (info)
 
 import HsBindgen.Config.ClangArgs
@@ -26,4 +27,35 @@ info = progDesc "List supported targets"
 -------------------------------------------------------------------------------}
 
 exec :: IO ()
-exec = mapM_ (putStrLn . targetTriple) [minBound..]
+exec = mapM_ putStrLn $
+      formatRow tableHeader
+    : List.replicate (sum lens - 3) '-'
+    : map formatRow tableRows
+  where
+    tableHeader :: [String]
+    tableHeader = ["Target", "Processor", "OS", "C library"]
+
+    targetRow :: Target -> [String]
+    targetRow t = targetTriple t : case t of
+      Target_Linux_GNU_X86_64    -> ["Intel/AMD 64-bit", "Linux",   "glibc"]
+      Target_Linux_GNU_AArch64   -> ["ARM 64-bit",       "Linux",   "glibc"]
+      Target_Linux_Musl_X86_64   -> ["Intel/AMD 64-bit", "Linux",   "Musl"]
+      Target_Linux_Musl_AArch64  -> ["ARM 64-bit",       "Linux",   "Musl"]
+      Target_Windows_MSVC_X86_64 -> ["Intel/AMD 64-bit", "Windows", "MSVC"]
+      Target_Windows_GNU_X86_64  -> ["Intel/AMD 64-bit", "Windows", "glibc"]
+      Target_Darwin_X86_64       -> ["Intel/AMD 64-bit", "macOS",   ""]
+      Target_Darwin_AArch64      -> ["Apple/ARM 64-bit", "macOS",   ""]
+
+    tableRows :: [[String]]
+    tableRows = map targetRow [minBound..]
+
+    lens :: [Int]
+    lens =
+      (+ 3) . maximum . map length <$> List.transpose (tableHeader : tableRows)
+
+    formatRow :: [String] -> String
+    formatRow ss = List.dropWhileEnd (== ' ') $
+      concatMap (uncurry formatCol) (zip ss lens)
+
+    formatCol :: String -> Int -> String
+    formatCol s n = take n $ s ++ List.repeat ' '


TravisCardwell:
This PR really needs to be tested across multiple platforms.  The PR tests all run on Linux, however.  The Windows and macOS tests only run on merge.

Just thinking about possibilities:

* Perhaps we can configure merge workflows with `workflow_dispatch` so that we can manually run them before merging when desired?

* Perhaps testing on Windows and macOS is higher priority than building the examples?  What if we test Windows and macOS on PR and (just) build the examples on merge?

TravisCardwell:
Some things that we definitely need to discuss in the review:

* The set of supported targets - selection strategy
* The `parseTargetTripleLenient` implementation - too lenient or too strict?

TravisCardwell:
I added some more information to the `hs-bindgen-cli info targets` command:

```
Target                   Processor          OS        C library
---------------------------------------------------------------
x86_64-pc-linux-gnu      Intel/AMD 64-bit   Linux     glibc
aarch64-pc-linux-gnu     ARM 64-bit         Linux     glibc
x86_64-pc-linux-musl     Intel/AMD 64-bit   Linux     Musl
aarch64-pc-linux-musl    ARM 64-bit         Linux     Musl
x86_64-pc-windows-msvc   Intel/AMD 64-bit   Windows   MSVC
x86_64-pc-windows-gnu    Intel/AMD 64-bit   Windows   glibc
x86_64-apple-darwin      Intel/AMD 64-bit   macOS
aarch64-apple-darwin     Apple/ARM 64-bit   macOS
```

I have not been able to test on macOS yet.  I bet that it uses an Apple C library, but I am not sure.  I will create issues about investigating the following possible targets:

* ~~Windows Musl~~ : Research indicates that this is *not* something people normally do.
* Darwin GNU
* Darwin Musl

Issue #1267, opened 2025-11-08
Simplify `Id`


edsko:
The evolution of `Id` throughout the frontend passes is

```hs
type Id Parse                    = C.PrelimDeclId
type Id ConstructTranslationUnit = C.PrelimDeclId
type Id HandleMacros             = C.PrelimDeclId
type Id NameAnon                 = C.DeclId
type Id ResolveBindingSpecs      = C.DeclId
type Id Select                   = C.DeclId
type Id HandleTypedefs           = C.DeclId
type Id MangleNames              = (C.NamePair, C.NameOrigin)
```

This is somewhat suboptimal, for a few reasons:

### 1. `MangleNames`

It's not clear from this definition that `MangleNames` actually just _adds_ information; `(C.NamePair, C.NameOrigin)` is isomorphic to `(C.DeclId, Hs.Identifier)`, which would make that a lot more obvious.
 
### 2. `NamePair` is re-introduced in multiple places

In addition to the use of `NamePair` in `Id`, we also have

```hs
type FieldName    Parse                    = C.Name
type FieldName    ConstructTranslationUnit = C.Name
type FieldName    HandleMacros             = C.Name
type FieldName    NameAnon                 = C.Name
type FieldName    ResolveBindingSpecs      = C.Name
type FieldName    Select                   = C.Name
type FieldName    HandleTypedefs           = C.Name
type FieldName    MangleNames              = C.NamePair

type ArgumentName Parse                    = Maybe C.Name
type ArgumentName ConstructTranslationUnit = Maybe C.Name
type ArgumentName HandleMacros             = Maybe C.Name
type ArgumentName NameAnon                 = Maybe C.Name
type ArgumentName ResolveBindingSpecs      = Maybe C.Name
type ArgumentName Select                   = Maybe C.Name
type ArgumentName HandleTypedefs           = Maybe C.Name
type ArgumentName MangleNames              = Maybe C.NamePair
```

And we'd need yet another case for macros (#1269). It would be much better if this was done _once_; _everywhere_ there is a `Name`, no matter whether it's a declaration, a field name, an argument name, or a macro name, after name mangling that should be a `NamePair`.

### 3. Complexity

We have a proliferation of name types; they all serve a purpose, but if we can find a way to simplify this _without_ losing the distinctions that we want to make, that would be preferable.

## Proposed solution

I think the way forward is as follows:

[...]

PR #1336, opened 2025-11-28
Add `haskellId` to `DeclId`

edsko:
Prior to this commit, we had

```hs
type Id Parse                    = C.PrelimDeclId
type Id ConstructTranslationUnit = C.PrelimDeclId
type Id HandleMacros             = C.PrelimDeclId
type Id NameAnon                 = C.DeclId NameAnon
type Id ResolveBindingSpecs      = C.DeclId ResolveBindingSpecs
type Id Select                   = C.DeclId Select
type Id HandleTypedefs           = C.DeclId HandleTypedefs
type Id MangleNames              = (C.NamePair, C.NameOrigin)
```

Now this is

```hs
type Id Parse                    = C.PrelimDeclId
type Id ConstructTranslationUnit = C.PrelimDeclId
type Id HandleMacros             = C.PrelimDeclId
type Id NameAnon                 = C.DeclId NameAnon
type Id ResolveBindingSpecs      = C.DeclId ResolveBindingSpecs
type Id Select                   = C.DeclId Select
type Id HandleTypedefs           = C.DeclId HandleTypedefs
type Id MangleNames              = C.DeclId MangleNames
```

The important change is that `DeclId` now has a `haskellId` field:

```hs
data DeclId (p :: Pass) =
    DeclIdNamed   (NamedDeclId   p)
  | DeclIdBuiltin (BuiltinDeclId p)

data NamedDeclId (p :: Pass) = NamedDeclId{
      name      :: Name
    , origin    :: NameOrigin
    , haskellId :: HaskellId p
    }

data BuiltinDeclId (p :: Pass) = BuiltinDeclId{
      name      :: Name
    , haskellId :: HaskellId p
    }
```

where `HaskellId` is a new type family in `IsPass`.

We use `HaskellId` also in `CommentRef`:

```hs
data CommentRef p = CommentRef C.Name (Maybe (HaskellId p))
```

We do not use `Id p` here anymore, in order to pave the way for _always_ including qualifiers (ordinary/struct/..) in `DeclId` (which we do not a priori know for cross-references in comments).

We no longer use `NamePair` at all for declartions. This gets us one step closer to the vision in https://github.com/well-typed/hs-bindgen/issues/1267.

### DIFF
--- a/manual/hs/manual/app/Manual/Functions/FirstOrder.hs
+++ b/manual/hs/manual/app/Manual/Functions/FirstOrder.hs
@@ -5,7 +5,6 @@ module Manual.Functions.FirstOrder (examples) where
 import Control.Monad ((<=<))
 import Foreign as F
 import Foreign.C (withCString)
-import Foreign.C qualified as FC
 import System.IO.Unsafe
 
 import HsBindgen.Runtime.FunPtr
@@ -24,19 +23,6 @@ import FunctionPointers.Safe qualified as FunPtr
 
 hashSafe :: String -> Int
 hashSafe s = fromIntegral $ unsafePerformIO $ withCString s hash
-
-{-------------------------------------------------------------------------------
-  Function pointer instances
--------------------------------------------------------------------------------}
-
-foreign import ccall "dynamic" mkApply1Fun ::
-     F.FunPtr (F.FunPtr FunPtr.Int2int -> FC.CInt -> IO FC.CInt)
-  -> F.FunPtr FunPtr.Int2int
-  -> FC.CInt
-  -> IO FC.CInt
-
-instance FromFunPtr (F.FunPtr FunPtr.Int2int -> FC.CInt -> IO FC.CInt) where
-  fromFunPtr = mkApply1Fun
 
 {-------------------------------------------------------------------------------
   Examples
--- a/hs-bindgen/fixtures/macros/macro_in_fundecl/Example.hs
+++ b/hs-bindgen/fixtures/macros/macro_in_fundecl/Example.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CApiFFI #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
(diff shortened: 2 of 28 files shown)


edsko:
I don't know where the hash changes are coming from, but looking into it, I noticed that we're actually hashing something in a suboptimal way. I will address that first, and also include some comments in the generate code whenever we use a hash, and then go from there.

edsko:
After rebasing this on previous improvements to hash handling (#1345, #1348, #1349, #1350, #1352, #1353) the unexpected hash changes are gone from this PR. There are still a few fixture changes; this is because `isDefinedInCurrentModule` was previously missing a few changes. Specifically, when looking for "domain specific" or "locally defined" (i.e., not orphan) types, it was missing macros 

```c
#define I int
I (*bar4(long x))(short) { return 0; }
```

as well as domain specific types inside function pointers (higher order functions)

```c
extern int (* const apply1_nopointer_res (void)) (int2int, int);
```

We were therefore missing some `ToFunPtr`/`FromFunPtr` instances, which are now added (cc @bolt12 ).

manual/hs/manual/app/Manual/Functions/FirstOrder.hs
```diff
@@ -25,19 +24,6 @@ import FunctionPointers.Safe qualified as FunPtr
...
--------------------------------------------------------------------------------}
-
-foreign import ccall "dynamic" mkApply1Fun ::
-     F.FunPtr (F.FunPtr FunPtr.Int2int -> FC.CInt -> IO FC.CInt)
-  -> F.FunPtr FunPtr.Int2int
-  -> FC.CInt
-  -> IO FC.CInt
-
-instance FromFunPtr (F.FunPtr FunPtr.Int2int -> FC.CInt -> IO FC.CInt) where
-  fromFunPtr = mkApply1Fun
```
edsko, line 39: This change is due to the fact that we were missing `FromFunPtr` instance generation (see https://github.com/well-typed/hs-bindgen/pull/1336#issuecomment-3591712838). Kind of satisfying that this results in a duplicate instance error, and that this instance can now simply be deleted.

@jorisdral I _believe_ you added this instance; do you think we need to update any docs anywhere?

Issue #1357, opened 2025-12-01
Haskell representation: allow name overrides for records and newtypes


TravisCardwell:
For records, this would look like

```yaml
hstypes:
- hsname: CTm
  representation:
    record:
      constructor: ...
      fields:
      - cTm_sec
      - cTm_min
      ...
```

It will _not_ be possible override _single_ field names; if a list of fields is specified, _all_ fields must be specified.

For newtypes, the syntax is identical; the list of fields will simply be a singleton.

This ticket is _not_ about _choosing_ a representation; a prescriptive binding spec specify `record` or `newtype` _must_ match what we have actually generated; making this _configurable_ is the scope of other tickets (#308, #1531).

Note: Squashing is not handled through representation at all, but by mapping multiple C types to the same Haskell type (#1436).

TravisCardwell:
Resolving this issue will resolve #308.

TravisCardwell:
We discussed one possible use case of a `type` alias.  In some C code, an `enum` is used to define constant values but an integral type is used elsewhere.  Such (ill-typed) usage is transparent in C because of automatic coercion.  Example:

```c
enum category { ... };

struct widget {
  ...
  int category;
  ...
};
```

We discussed specifying an `alias` Haskell representation for `enum category`, so that the Haskell type would be something like `type Category = CInt`, allowing it to be used directly with `widget.category`.

Some notes while I write this comment:

* We should still generate the patterns for the `enum` constants, I guess, perhaps implementing a special case for `alias` representations for C `enum` types.

* The C compiler chooses the underlying representation of the `enum`, and the type may differ across platforms.  This is transparent in C due to automatic coercion, but we probably need to be careful about this when using a Haskell `alias`.  For example, the underlying type for `enum category` may be `CUInt`.

jorisdral:
Note that for type aliases like `type Category = CInt` it's important that we import `CInt` and its constructor so that `foreign import`s compile (see issue #1282). So maybe a type alias also needs an `ffitype`?

Issue #1356, opened 2025-11-30
Implement C representation


TravisCardwell:
Binding specifications allow specification of the C type representation.

Example:

```yaml
version:
  hs_bindgen: 0.1.0
  binding_specification: '1.0'
target: any
hsmodule: HsBindgen.Runtime.Prelude
ctypes:
- headers:
  - stdio.h
  - wchar.h
  cname: FILE
  hsname: CFile
  representation: opaque
```

We need to implement the functionality for this:

* [ ] Prescriptive binding specifications
    * [ ] Check that each Haskell type is referenced by exactly one C type with representation `default` or `opaque`
    * [ ] `representation: default`
        * [ ] Check that we generate a type and constructor, tracing an error if we cannot
    * [ ] `representation: opaque`
        * [ ] Generate a type with no constructor
    * [ ] `representation: alias`
        * [ ] Determine how to "track aliases"

* [ ] External binding specifications
    * [x] Specify C type representation in `stdlib`
    * [ ] Can ignore C type representation in external binding specifications?

* [ ] Generated binding specifications
    * [ ] Output the correct C representation
        * [ ] Be sure to check that `struct foo;` is opaque

Note that C type representation is different from Haskell type representation (#308).

TravisCardwell:
Resolving this issue will resolve #809.

edsko:
This is obsolete; we can close it once #1529 is done ; marking as part of `release-0.1-alpha1` because #1529 is.

TravisCardwell:
#1529 is resolved in #1545.

PR #1358, opened 2025-12-01
Add `ctypes` and `hstypes` representation syntax

TravisCardwell:
Add `ctypes` and `hstypes` representation syntax

This PR adds the binding specification syntax for both C type representation and Haskell type representation, using the syntax we decided in #799.

**C type representation**

* `default`: The C declaration corresponds to both a type and a constructor.
* `opaque`: The C declaration corresponds to a type only, so it may only be used via a reference.
* `alias`: The C type should be considered an alias of a different C type.

Implementation issue: #1356

**Haskell type representation**

* `record`: A type is generated using `data`.  A binding specification may specify a list of field names.
* `newtype`: A type is generated using `newtype`.
* `alias`: A type is generated using `type`.

Implementation issue: #1357

**Question**

My understanding is that the `record` Haskell representation corresponds to any type defined with `data`, both those with constructors and those without.  I think that this might be a bit confusing, as many developers think of a "record type" as a type with fields (created using `data` *or* `newtype`), which excludes types without constructors.  Perhaps it would be preferable to used `data`, `newtype`, and `type` instead?

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/Stdlib.hs
+++ b/hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/Stdlib.hs
@@ -57,7 +57,7 @@ bindingSpec = BindingSpec.BindingSpec{..}
     integralTypes :: [(CTypeKV, HsTypeKV)]
     integralTypes =
       let aux (t, hsIdentifier) =
-            mkType t hsIdentifier cD hsN intI ["inttypes.h", "stdint.h"]
+            mkTypeN t hsIdentifier cD intI ["inttypes.h", "stdint.h"]
       in  map aux [
               ("int8_t",         "Int8")
             , ("int16_t",        "Int16")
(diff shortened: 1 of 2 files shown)


TravisCardwell:
Going through my notes, I spotted something that might be related to the `newtype` Haskell representation.  At one point, we talked about being able to override a type, "for ill-typed programs `enum`/`int`."  Is this something that we still want to do?  Perhaps a type override would be specified for a `newtype` Haskell representation?

If we implement this, I guess `Storable` instance methods will need to perform conversion.  In cases where the override size differs, we need to decide what to do about overflow and underflow.

EDIT: This could apply to `struct` fields as well.  Example:

```c
enum category { ... };

struct widget {
  ...
  int category;
  ...
};
```

If the program always uses `enum category` values in `widget.category`, then a user may want to override the type of the `widget_category` field for Haskell type `Widget`, to use `Category` instead of `CInt`.  If `Category` wraps `CInt`, then Haskell conversion is also coercion, but `Category` could wrap `CUInt`, in which case the types do not match exactly.

TravisCardwell:
We decided to *not* support overriding types, at least at this time.

TravisCardwell:
As we discussed yesterday, we will keep the current Haskell representation names but add an `opaque` representation.

* A `record` representation translates to a `data` type with a constructor.
* An `opaque` representation translates to a `data` type with no constructor.

The other representations are unchanged:

* A `newtype` representation translates to a `newtype` type (with a constructor).
* An `alias` representation translates to a `type` type.

Note that an `opaque` Haskell type is generally used opaque C types.  It is used with `Ptr`, as it cannot have a `Storable` instance.

TravisCardwell:
As we discussed yesterday, we will allow specification of a `newtype` field name, using the same `fields` syntax used for `record` specifications.

Since we always give generated `newtype` declarations named fields, I am representing the type as follows:

```haskell
-- | Haskell newtype representation
data HsNewtypeRep = HsNewtypeRep {
      -- | Field name
      hsNewtypeRepField :: Maybe Hs.Identifier
    }
```

In a prescriptive binding specification, a `Nothing` value just means that a name is not specified.  In this case, `hs-bindgen` determines the name.

External binding specifications that are generated by `hs-bindgen` will always specify the name.  External binding specifications for types not generated by `hs-bindgen`, however, may not have a named field.  `Nothing` is used in this case, even though the semantics differ.  Note that the same semantics difference exists for `record` fields for types not generated by `hs-bindgen` that lack field names.

TravisCardwell:
As we discussed yesterday, we will also allow specification of `record` and `newtype` types to specify the constructor name.

TravisCardwell:
@edsko The discussed changes have been made.

PR #1359, opened 2025-12-01
Fix the `compile-fixtures.sh` script

jorisdral:
The script was expecting fixtures to have `.pp` as part of the file extension, but that was recently changed in PR #1341. The script's fixture detection is now updated.

To prevent this type of bug in the future, the script now has a `KNOWN_FIXTURES_COUNT` constant variable that records how many fixtures the script expects to find. If the count of found fixtures does not match the count of known fixtures, then the script throws an error. `KNOWN_FIXTURES_COUNT` should from now on be updated if new fixtures are added or removed. The text that is printed to the console in case of a count mismatch also hints that this variable should be updated.

### DIFF
--- a/scripts/ci/compile-fixtures.sh
+++ b/scripts/ci/compile-fixtures.sh
@@ -6,7 +6,7 @@ usage() {
     cat <<EOF
 Usage: $(basename "$0") [OPTIONS]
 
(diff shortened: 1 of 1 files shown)


jorisdral:
Merging this because it is a relatively simple fix

PR #1364, opened 2025-12-02
Add an `absurd` in `adjustFunctionTypesToPointers`

jorisdral:


jorisdral:
Merging this because it has no functional consequences

PR #1367, opened 2025-12-02
Generate single line comments

### LINKED ISSUE #1346: Render title-only haddock comments as single lines
Currently we render these as

```hs
{-| blah

-}
```

it would be nicer if we rendered these as

```hs
-- | blah
```

instead.

bolt12:
This closes #1346


edsko:
@bolt12 I think something isn't _quite_ right, since no fixtures are changed. An example of a fixture with single line comments is `functions/callbacks/Example.hs`.

bolt12:
Turns out I had to consider the case where there was no title but there was only one metadata entry

PR #1369, opened 2025-12-03
Suggested changes to overwrite files and output directory policies

dschrempf:
- Use `StateT` instead of writer, like so we can access the actions before exiting the monad stack.
- Stay in `ExceptT` so we can use `throwError` instead of the implicit `Either` early return.
- Separate "check adherence to policy" from "execute file action".

Edit:
- [x] I am still fixing the tracer; somehow the safe tracer is not used anymore.


dschrempf:
Merging this into Armandos branch, we will discuss using the other PR.

Issue #1379, opened 2025-12-06
Think about binding specifications for auxiliary definitions


edsko:
(detailed to be added)

edsko:
If we do this, we might need to invent a syntax for auxiliary definitions (`DeclAux`; see #1404); something like `aux: Foo` perhaps, to refer to the auxiliary type for `Foo` (these aux types don't have names of their own).

TravisCardwell:
`DeclAux` is recursive.  Perhaps something like this could work?

```yaml
# aux for foo
- cname: aux foo

# aux for aux for foo
- cname: aux (aux foo)

# aux for aux for aux for foo
- cname: aux (aux (aux foo))
```

Only the `DeclAux` constructor recurses, and there is no branching, so the parenthesis are technically unnecessary...

```yaml
# aux for foo
- cname: aux foo

# aux for aux for foo
- cname: aux aux foo

# aux for aux for aux for foo
- cname: aux aux aux foo
```

Note that we decided against using an `anon` prefix for anonymous types and instead chose syntax like `struct @foo`.  We could use similar syntax of auxiliary types, but I think that a prefix would be less confusing.

I guess we do not need to distinguish between different kinds of auxiliary types?

TravisCardwell:
The tentative syntax in the code is `aux: Foo`, but we should probably avoid using `:` in the syntax because it does not work well with YAML.

edsko:
We decided to go a slightly different route here. We will _not_ have `DeclAux` at all, reserving `DeclId` for declarations that actually appear in the C source. Instead, auxiliary declarations will be introduced in the Haskell translation phase (possibly assisted by earlier annotations). This means we have no way of referring to them directly in binding specs, but that's okay; auxiliary are anyway not really independent from their "parent". If we do want to make certain aspects of auxiliary types configurable, then we could do something like

```yaml
  ctypes:
    - cname: foo
      header: ..
      hsname: Foo
  hstypes:
    - hsname: Foo
      auxiliary:
      - kind: AuxFunPtrPayload
        name: Foo_MyAuxSuffix
   - hsname: Foo_MyAuxSuffix
     instances: ...
```

TravisCardwell:
Note that the current implementation results in two `ctypes` with the same `cname` that reference different `hsnames`.

[...]

Issue #844, opened 2025-07-01
Be more explicit about specifications for anonymous declarations


edsko:
Suppose we have

```c
typedef struct {
  // .. fields
} foo;

struct foo {
  // .. fields
};
```

and suppose furthermore that we make it possible for binding specs to say that the `typedef foo` should _not_ be squashed (#799). When this happens, with the existing infrastructure a binding specification for `struct foo` would become ambiguous. Perhaps @TravisCardwell was right all along and we should have a specific syntax for referring to anonymous declarations. Proposed syntax:

```yaml
types:
   - header: ..
     cname: ..
    ..
   - header: ..
     anonymous: ..
    ..
```

TravisCardwell:
@edsko and I discussed this.  We would like to also be able to reference anonymous types when slicing, so we think that it is preferable to use a prefix after all.  We would be able to use the same syntax in pretty-printed output.

Note that the prefix must not result in valid C names.  Generated names should stand out.

Note that the syntax is strictly for anonymous types.  It is not used for renamed types, as binding spec resolution and slicing is done before types are renamed.

We would like to use a `$` prefix, mirroring GHC syntax.  Selection is done using regular expressions, however, and `$` is a special character that could cause problems as well as confusion.

We also considered an `anon:` prefix, but it does not stand out as much as `$`.  The colon is YAML syntax, so such names would need to be quoted in binding specification files.

TravisCardwell:
We decided to use an `@` prefix.  This character is not currently used by YAML, but it is reserved for future use and therefore must be quoted.  Example:

```yaml
cname: '@S1_c'
```

edsko:
@TravisCardwell I'm going to close this as done. Let me know if that is not correct.

PR #1381, opened 2025-12-06
Obsolete `QualDeclId`

edsko:


hs-bindgen/fixtures/attributes/attributes/bindingspec.yaml
```diff
@@ -5,7 +5,7 @@ target: x86_64-pc-linux-musl
 hsmodule: Example
 ctypes:
 - headers: attributes/attributes.h
-  cname: struct __sFILE
+  cname: FILE
```
edsko, line 8: @TravisCardwell In a case like this where we squash a typedef around a struct, and then use the name of the typedef _for_ the struct, the generated binding spec now refers to the typedef instead of the struct. I think this is okay (in fact, better).

hs-bindgen/fixtures/attributes/attributes/Example.hs
```diff
@@ -157,7 +157,9 @@ instance ( TyEq ty ((HsBindgen.Runtime.HasCField.CFieldType Bar) "bar_i")
   getField =
     HsBindgen.Runtime.HasCField.ptrToCField (Data.Proxy.Proxy @"bar_i")
 
-{-| __defined at:__ @attributes\/attributes.h:22:9@
+{-| __C declaration:__ @baz@
+
+    __defined at:__ @attributes\/attributes.h:22:9@
```
edsko, line 162: This name was initially omitted from the documentation because the declaration was anonymous. However, we now refer to the enclosing typedef instead, which I think is fine (better, in fact).

hs-bindgen/fixtures/documentation/doxygen_docs/bindingspec.yaml
```diff
@@ -40,9 +37,6 @@ ctypes:
 - headers: documentation/doxygen_docs.h
   cname: processor_fn_t
   hsname: Processor_fn_t
-- headers: documentation/doxygen_docs.h
-  cname: processor_fn_t_Deref
-  hsname: Processor_fn_t_Deref
```
edsko, line 45: @TravisCardwell This is an auxiliary type that we generate, with no corresponding C type at all. I've opened https://github.com/well-typed/hs-bindgen/issues/1379 to think about cases like this.

hs-bindgen/fixtures/documentation/doxygen_docs/bindingspec.yaml
```diff
@@ -4,6 +4,9 @@ version:
 target: x86_64-pc-linux-musl
 hsmodule: Example
 ctypes:
+- headers: documentation/doxygen_docs.h
+  cname: struct @data_union_t_as_parts
+  hsname: Data_union_t_as_parts
```
edsko, line 9: @TravisCardwell As a small step towards https://github.com/well-typed/hs-bindgen/issues/844, I've changed `getCQualName` in `HsBindgen.BindingSpec.Gen` to insert these `@` signs. We may well want to do this differently, but at least we're being a bit more explicit now (and it was natural to insert them now, the logic for creating these keys is now quite different, and no longer depends on aliases at all).

hs-bindgen/fixtures/edge-cases/flam/bindingspec.yaml
```diff
@@ -4,15 +4,15 @@ version:
 target: x86_64-pc-linux-musl
 hsmodule: Example
 ctypes:
+- headers: edge-cases/flam.h
+  cname: struct @foo_bar
+  hsname: Foo_bar
```
edsko, line 9: @TravisCardwell Note that we are now also inserting the `struct` prefix correctly.

hs-bindgen/fixtures/macros/macro_in_fundecl_vs_typedef/Example/FunPtr.hs
```diff
@@ -98,7 +98,7 @@ $(HsBindgen.Runtime.Prelude.addCSource (HsBindgen.Runtime.Prelude.unlines
   , "/* test_macrosmacro_in_fundecl_vs_typ_Example_get_struct_name3_ptr */"
   , "__attribute__ ((const))"
   , "void (*hs_bindgen_680e8a5d673ce9c1 (void)) ("
-  , "  struct struct4 *arg1,"
+  , "  struct4 *arg1,"
```
edsko, line 101: In C we now prefer the original reference to the typedef.

hs-bindgen/fixtures/types/enums/nested_enums/bindingspec.yaml
```diff
@@ -4,6 +4,9 @@ version:
 target: x86_64-pc-linux-musl
 hsmodule: Example
 ctypes:
+- headers: types/enums/nested_enums.h
+  cname: enum @exB_fieldB1
```
edsko, line 8: Anonymous enum.

hs-bindgen/src-internal/HsBindgen/BindingSpec/Gen.hs
```diff
@@ -240,23 +234,35 @@ genBindingSpec'
...
-            C.QualName cName cNameKind
-          C.NameOriginGenerated{} ->
-            let cName' = fromMaybe cName (listToMaybe (C.declAliases declInfo))
-            in  C.QualName cName' C.NameKindOrdinary
-          C.NameOriginRenamedFrom fromCName ->
-            C.QualName fromCName cNameKind
-      C.DeclIdBuiltin builtin ->
-        C.QualName builtin.name cNameKind
+getCQualName :: C.FinalDeclId -> Maybe C.QualName
+getCQualName declId =
```
edsko, line 238: @TravisCardwell These are the changes I mentioned above.

TravisCardwell:
Great!

The previous implementation of the `@` syntax was blocked because we were not handling "aliases" appropriately yet.  For example, the manual `vector` tests were failing because we generated binding specifications for the anonymous `struct @vector` type but did not recognize `vector` as an alias for that type, so uses as an external binding specification did not have external references to `vector`.

It sounds like this PR completely changes how we handle `typedef` aliases for anonymous types.  When a `typedef` is used to provide a name for an anonymous `struct`, like `typedef struct { ... } vector`, the record specifications (`representation: record`) should now be specified using the `typedef` name (`vector`), *not* the `struct` name (`struct @vector`).  I confirmed this behavior in `main`.

TravisCardwell:
In preparation for our discussion, I dumped the declarations after every frontend pass when generating bindings for `vector.h` and analyzed them.

1. **`Parse`** - `PrelimDeclId` is now always qualified: it includes the `NameKind`.  We get `ParseResultSuccess` values for both the anonymous `struct` and the `typedef`.

    ```haskell
    PrelimDeclIdAnon
      (AnonId "manual/c/vector.h:3:9")
      (NameKindTagged TagKindStruct)
    ```

    ```haskell
    PrelimDeclIdNamed "vector" NameKindOrdinary
    ```

2. **`ConstructTranslationUnit`** - We construct `Decl` values for both the anonymous `struct` and the `typedef`.

3. **`HandleMacros`** - No macros are involved.  The `ReparseNotNeeded` annotations are removed.

4. **`NameAnon`** - The anonymous `struct` and `typedef` declarations are renamed, respectively, as follows.  Both are assigned `DeclIdNamed "vector"`.  The `name` is therefore *not* a unique identifier, but the `DeclIndex`, `UseDeclGraph`, and `DeclUseGraph` all use the `PrelimDeclId` (still available via `origDeclId`) for queries.

    ```haskell
    DeclId
      { name = DeclIdNamed "vector"
      , nameKind = NameKindTagged TagKindStruct
      , origDeclId = OrigDeclId
          ( PrelimDeclIdAnon (AnonId "manual/c/vector.h:3:9") (NameKindTagged TagKindStruct)
          )
      , haskellId = ()
      }
    ```

    ```haskell
    DeclId
      { name = DeclIdNamed "vector"
      , nameKind = NameKindOrdinary
      , origDeclId = OrigDeclId (PrelimDeclIdNamed "vector" NameKindOrdinary)
      , haskellId = ()
      }
    ```

5. **`ResolveBindingSpecs`** - No binding specifications are involved.  Annotations are added to indicate that there are neither C nor Haskell specifications.

6. **`Select`** - Both the anonymous `struct` and `typedef` declarations are selected.  There are no changes to the declarations.

7. **`HandleTypedefs`** - The declarations are squashed: the `typedef` declaration is removed, and the `DeclId` for the anonymous `struct` declaration is updated to the following.

    ```haskell
    DeclId
      { name = DeclIdNamed "vector"
      , nameKind = NameKindTagged TagKindStruct
      , origDeclId = OrigDeclId (PrelimDeclIdNamed "vector" NameKindOrdinary)
      , haskellId = ()
      }
    ```

    The different `NameKind`s indicate that the declaration was squashed.  In addition to automatic squashing of anonymous declarations within `typedef` declarations, however, users are able to squash other declarations (using `representation: alias` for the C `typedef` declaration), and we need to retain information to distinguish these cases for generating binding specifications.  The above `DeclId` does *not* distinguish between `typedef struct { ... } vector` and `typedef struct vector { ... } vector`.

8. **`MangleNames`** - Haskell names are assigned to the type, constructor, and fields.

TravisCardwell:
We have discussed the idea of annotating declarations instead of removing them from the list.  Examples:

* When a declaration is omitted via a prescriptive binding specification, we currently remove it from the list.  Instead, we could annotate the declaration as omitted.
* When a declaration is externally bound, we currently remove it from the list.  Instead, we could annotate the declaration as external.
* When a declaration is not selected, we currently remove it from the list.  Instead, we could annotate the declaration as not selected.
* When a declaration for a `typedef` is squashed, we currently remove it from the list.  Instead, we could annotate the declaration as squashed.

The idea is that we would retain all of the information about all of the ("parsed") declarations.  We would need to look at the annotation to determine which declarations (still) need to be processed at each pass/stage, checking for errors as appropriate.  We would only generate code for the declarations that omitted, externally bound, etc.

Binding specifications will be generated from the list of frontend declarations, and a `squashed` declaration would retain the C specification in cases where it is specified by a prescriptive binding specification.

The current `DeclId` design is sufficient if we do this.

TravisCardwell:
The final `vector` `DeclId` is as follows:

```haskell
DeclId
  { name = DeclIdNamed "vector"
  , nameKind = NameKindTagged TagKindStruct
  , origDeclId = OrigDeclId (PrelimDeclIdNamed "vector" NameKindOrdinary)
  , haskellId = "Vector"
  }
```

We can only use the `OrigDeclId` when generating tests.  (`vector` is valid, while `struct vector` is not.)

TravisCardwell:
I would like to see an example of an anonymous declaration that is *not* in a `typedef`.

```c
struct rect {
  struct { int x; int y; } tl;
  struct { int x; int y; } br;
};
```

Here are the `DeclId`s after all passes in the frontend:

```
$ cabal run hs-bindgen-cli -- internal frontend --unique-id ex -I . ex.h
```

```haskell
DeclId
  { name = DeclIdNamed "rect_tl"
  , nameKind = NameKindTagged TagKindStruct
  , origDeclId = OrigDeclId (PrelimDeclIdAnon (AnonId "./ex.h:2:3") (NameKindTagged TagKindStruct))
  , haskellId = "Rect_tl"
  }
```

```haskell
DeclId
  { name = DeclIdNamed "rect_br"
  , nameKind = NameKindTagged TagKindStruct
  , origDeclId = OrigDeclId (PrelimDeclIdAnon (AnonId "./ex.h:3:3") (NameKindTagged TagKindStruct))
  , haskellId = "Rect_br"
  }
```

```haskell
DeclId
  { name = DeclIdNamed "rect"
  , nameKind = NameKindTagged TagKindStruct
  , origDeclId = OrigDeclId (PrelimDeclIdNamed "rect" (NameKindTagged TagKindStruct))
  , haskellId = "Rect"
  }
```

This looks good.  The `OrigDeclId`s for the anonymous `struct` declarations are `PrelimDeclIdAnon`s.

TravisCardwell:
I am unable to find how/where we are adding the `@` syntax.  I could not find it in the commit diffs, and it is quite difficult to search for.  I is not in the places where I expect it to be.

So, to investigate, I created an example that needs the `@` syntax.  In creating bindings for the following header, the goal is to give a better Haskell name to the anonymous `struct`.

```c
struct person {
  char *name;
  struct { int year; int month; int day; } dob;
};
```

The following prescriptive binding specification should match the name that we assign to the anonymous `struct` and use Haskell name `Date` instead.

```yaml
version:
  hs_bindgen: 0.1.0
  binding_specification: '1.0'
target: x86_64-pc-linux-gnu
hsmodule: Person
ctypes:
  - headers: ex.h
    cname: struct @person_dob
    hsname: Date
```

The following runs the preprocessor with that prescriptive binding specification, and we get an error.

```
    --unique-id ex \
    --hs-output-dir ex-src \
    --module Person \
    --prescriptive-binding-spec ex.yaml \
    -I . \
    ex.h
[Error  ] [HsBindgen] [resolve-binding-specs] Binding specification for type not used: 'struct @person_dob'
An error happened (see above)
```

When resolving binding specification, we parse `C.QualName`s from the YAML/JSON file.  We are currently not parsing the `@` syntax specially, so `struct @person_dob` is parsed to the following.

```haskell
QualName
  { qualNameName = "@person_dob"
  , qualNameKind = NameKindTagged TagKindStruct
  }
```

In `ResolveBindingSpecs`, we perform a lookup to match a declaration with a C type in the prescriptive binding specification.  The `DeclId` is as follows.

```haskell
DeclId
  { name = DeclIdNamed "person_dob"
  , nameKind = NameKindTagged TagKindStruct
  , origDeclId = OrigDeclId (PrelimDeclIdAnon (AnonId "./ex.h:3:3") (NameKindTagged TagKindStruct))
  , haskellId = ()
  }
```

We construct a `C.QualName` using `C.declQualName`, which is implemented using `C.declIdQualName`.

```haskell
declIdQualName :: DeclId p -> QualName
declIdQualName declId = QualName (declIdName declId) declId.nameKind
```

This is not taking into account that the declaration is anonymous, so we get the following.

```haskell
QualName
  { qualNameName = "person_dob"
  , qualNameKind = NameKindTagged TagKindStruct
  }
```

This does not match.

In my original/blocked anonymous syntax PR, `C.QualName` was given an additional `QualNameAnon` constructor, separating the syntax from the semantics.  Note that `QualNameAnon` has a `TagKind` (*not* a `NameKind`).

```haskell
-- | C name, qualified to distinguish kinds and anonymous names
data QualName =
    -- | Valid C name
    --
    -- This is the parsed representation of a @libclang@ C spelling.
    QualName {
        qualNameName :: Name
      , qualNameKind :: NameKind
      }

    -- | Generated name for an anonymous C type
    --
    -- Syntax: @struct \@foo@ references an anonymous C @struct@ type with
    -- generated name @foo@.
  | QualNameAnon {
        qualNameAnonGeneratedName :: Name
      , qualNameAnonTagKind       :: TagKind
      }
```

hs-bindgen/src-internal/HsBindgen/BindingSpec/Gen.hs
```diff
@@ -240,23 +234,35 @@ genBindingSpec'
...
+          -- * Anonymous declarations inside typedefs are given a name, but their
+          --   /original/ declaration is set to be the typedef instead.
+          --
+          -- However, anonymous structs inside other structs or unions will
+          -- remain anonymous. For these we instead use the name assigned by
+          -- the @NameAnon@ pass.
+          --
+          -- TODO <https://github.com/well-typed/hs-bindgen/issues/844>
+          -- This is WIP.
+          C.QualName ("@" <> declIdName declId) kind
```
edsko, line 263: This isn't _quite_ right I think. We should be able to rename declarations further, if needed for other reasons. Ideally we'd have the `DeclId NameAnon` name here in the `origDeclId`, so that we have the name assigned by `NameAnon` always available (which is meant not to be configurable precisely so that we can refer to it).

PR #1386, opened 2025-12-08
Parse typedefs only once

### LINKED ISSUE #1283: Parse `typedef`s only once
At the moment, when parsing a `typedef`, we traverse recursive `typedef`s until we find the bottom type.

When parsing functions, for example, we also parse return types and arguments. Now, if a funtion return type is a `typedef`, we parse the `typedef` again, recursively, until we find the bottom type.

We should parse each `typedef` only once.

bolt12:
The `HandleTypedefs` pass was redundantly traversing typedef chains at every use site.

Add a type cache to `TypedefAnalysis` that memoizes resolved types. Thread it through `handleUseSites` using the `State` monad. On first traversal of a typedef, cache the result; subsequent uses hit the cache instead of re-traversing the chain.

This fixes #1283

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type/Monad.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type/Monad.hs
@@ -6,6 +6,9 @@
 module HsBindgen.Frontend.Pass.Parse.Type.Monad (
     ParseType -- opaque
   , run
+    -- * Caching API
+  , lookupCache
+  , insertCache
     -- * Errors
   , ParseTypeExceptionInContext(..)
   , ParseTypeException(..)
@@ -28,6 +31,7 @@ import Clang.LowLevel.Core
 import HsBindgen.Frontend.AST.Internal qualified as C
 import HsBindgen.Frontend.Pass.Parse.IsPass
 import HsBindgen.Imports
+import HsBindgen.Language.C qualified as C
 
 {-------------------------------------------------------------------------------
   Definition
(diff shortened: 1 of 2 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type/Monad.hs
```diff
@@ -36,24 +39,32 @@ import HsBindgen.Util.Tracer
...
     , Applicative
     , Monad
     , MonadIO
+    , MonadState (Map CXType (C.Type p))
     )
 
-instance MonadError ParseTypeException ParseType where
-  throwError err   = Wrap $ throwIO err
-  f `catchError` h = Wrap $ catch (unwrap f) (unwrap . h)
+instance MonadError ParseTypeException (ParseType p) where
```
bolt12: Move the definition of ParseTypeException to IsPass.hs and we can avoid the `p` type parameter

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type/Monad.hs
```diff
@@ -36,24 +39,32 @@ import HsBindgen.Util.Tracer
...
-      unwrap :: IO a
+newtype ParseType p a = Wrap {
+      unwrap :: StateT (Map CXType (C.Type p)) IO a
     }
   deriving newtype (
       Functor
     , Applicative
     , Monad
     , MonadIO
+    , MonadState (Map CXType (C.Type p))
```
bolt12: Don't derive it and offer an API for caching instead

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type/Monad.hs
```diff
@@ -36,24 +39,32 @@ import HsBindgen.Util.Tracer
...
+
+  The ParseType monad includes state for caching typedef resolution to avoid
+  redundant type parsing during typedef chain resolution. It is parametrized
+  over the phase type to avoid circular dependencies.
 -------------------------------------------------------------------------------}
 
-newtype ParseType a = Wrap {
-      unwrap :: IO a
+newtype ParseType p a = Wrap {
+      unwrap :: StateT (Map CXType (C.Type p)) IO a
```
bolt12: Do not use CXType as the map key, use PrelimDeclId instead.

edsko (approved):

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type/Monad.hs
edsko: This is now done.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type/Monad.hs
edsko: But this still needs to be done.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type/Monad.hs
edsko: And this too.

PR #1368, opened 2025-12-03
Add `hs-bindgen-cli info builtin-macros`

TravisCardwell:
We are able to print the names of builtin macros, but unfortunately the definitions are not available via the `libclang` API.

PR #1388, opened 2025-12-09
Nix: Update Flake lock file

dschrempf:
In response to https://github.com/well-typed/hs-bindgen/pull/1368#event-21381071228


dschrempf:
Putting this on the queue because it only touches the Nix lock file.

Issue #1387, opened 2025-12-09
Next steps for identifier refactoring and binding spec updates


edsko:
Next steps:

- [x] Introduce `C.DeclName` and `C.ScopedName` in `HsBindgen.Language.C`, replacing `C.Name` in `HsBindgen.Frontend.Naming` (which will ultimateley go entirely)
- [x] Introduce `DeclIdAnon` as sketched below
- [ ] Update binding spec parser/pretty-printer to deal correctly with the `@` syntax (though we could extend this also to some other passes: pretty-printing in error messages, potentially Haddock documentation, potentially selection if we were to re-introduce selection of anonymous declarations). 
- [x] ~Move `HandleTypedefs` to before `ResolveBindingSpecs`. We decided _not_ to make squashing configurable, to avoid catch-22s between binding spec resolution and squashing.~

Some observations on  #1381

* The `origDeclId` guarantees uniqueness, no matter what internal transformations we apply.
* The information suggested in https://github.com/well-typed/hs-bindgen/pull/1381#issuecomment-3629489153 for removed declarations I _think_ can quite conveniently be stored in the `DeclIndex`. It already has types for each kind of entry (omitted, external, etc.); it would be easy to add additional information there if needed. This can be used for example to inclue `omit` entries in the generated binding spec.
* https://github.com/well-typed/hs-bindgen/pull/1381#issuecomment-3629499172 : yes, that's the _purpose_ of the `origDeclId`: to indicate how we should refer to this declaration _in C_.

> I am unable to find how/where we are adding the @ syntax. 

I even marked the very spot for you in the PR review :roll_eyes: 

with `PrelimDeclId` gone, `DeclId NameAnon` would look something like this (and would be the source and target for binding specs...)?

```hs
data DeclId p = DeclId{
      name :: DeclIdName p 
    , nameKind :: NameKind
    , ..
    }

data DeclIdName p =    
    DeclIdNamed C.Name
    -- | The name assigned in the NameAnon pass 
    --
    -- This is precisely the @\@@ syntax.
  | DeclidAnon (AnonAssignedName p) 
  | DeclIdBuiltin C.Name
```

[...]

Issue #1160, opened 2025-10-01
Support external binding specifications for built-in typedefs


edsko:
In addition to built-in macros (#1087), there exist also (a handful of?) builtin typedefs. Unlike built-in macros, these don't have a definition at all, they are baked into the compiler. Users _may_ want to provide external binding specifications for these.

Marking this as a low-priority enhancement for now, since these are very rare and very low-level.

PR #1391, opened 2025-12-09
Builtins are not a different kind of _name_

edsko:
Built-ins cannot have corresponding declaration (that's what makes them _built-in_ after all), so they cannot appear in the list of decls; built-in _macros_ are a bit of an exception to this, maybe, as they _do_ appear, but without a definition, so we can just skip over them.

For _use sites_, built-ins would not be another name (that is, yet another `TypeRef`), but instead would need to be special-cased, something like

```hs
data Type = ... | TypeBuiltin BuiltinType

data BuiltinType = BuiltinVaList | ...
```

However, we currently don't in fact support any built-in types at all.

### DIFF
--- a/hs-bindgen/app/HsBindgen/Cli/Info/BuiltinMacros.hs
+++ b/hs-bindgen/app/HsBindgen/Cli/Info/BuiltinMacros.hs
@@ -116,10 +116,11 @@ getBuiltinMacroNames tracer clangArgs =
       ClangInputMemory "hs-bindgen-builtins.h" ""
 
     visit :: Fold IO C.Name
-    visit = simpleFold $ \curr ->
-      C.getPrelimDeclId curr C.NameKindOrdinary >>= \case
-        C.PrelimDeclIdBuiltin name _kind -> foldContinueWith name
-        _otherwise                       -> foldBreak
+    visit = simpleFold $ \curr -> do
+      mBuiltin <- C.checkIsBuiltin curr
+      case mBuiltin of
+        Just name -> foldContinueWith (C.Name name)
+        Nothing   -> foldBreak
 
 -- | Get stringified definitions of the specified macros
 --
@@ -154,10 +155,9 @@ getMacros tracer clangArgs names =
     parseName = fmap C.Name . Text.stripPrefix "BUILTIN_X_" . C.getName
 
     visit :: Fold IO (C.Name, String)
-    visit = simpleFold $ \curr ->
+    visit = simpleFold $ \curr -> do
       C.getPrelimDeclId curr C.NameKindOrdinary >>= \case
         C.PrelimDeclIdAnon{}           -> foldContinue
-        C.PrelimDeclIdBuiltin{}        -> foldContinue
         C.PrelimDeclIdNamed name _kind ->
           (fromSimpleEnum <$> clang_getCursorKind curr) >>= \case
             Right CXCursor_VarDecl -> case parseName name of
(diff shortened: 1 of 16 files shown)


edsko:
@TravisCardwell could I get your eyes on this? This PR was motivated by a comment of yours that stuck in my head: you said that `DeclNameBuiltin` (when we were discussing https://github.com/well-typed/hs-bindgen/issues/1387) was a bit of an exception. That made me realize that our treatment of built-ins was actually kind of wrong; hence this PR.

Which is now making me wonder if we should keep the `PrelimDeclId`/`Declid` distinction as it exists now; it avoids awkward circularity between passes and the AST, it avoids awkward possibility of recursion. It also means that in `DeclId` we don't need a separate `DeclIdName` type at all anymore, avoiding the awkward similarity between the yet-to-be-introduced `DeclName`  type and `DeclIdName`.

If this all makes sense to you, feel free to merge this PR; and then perhaps you could take a look at introducing `DeclName` and `ScopedName` in `HsBindgen.Language.C`, deleting `Name` from `Naming`, and then I can take it from there again?

TravisCardwell:
There are six kinds of builtin macros:

* Integer values
    * Example: `#define __STDC__ 1`
* Float values
    * Example: `#define __DBL_DENORM_MIN__ 4.9406564584124654e-324`
* String values
    * Example: `#define __VERSION__ "Clang 21.1.6"`
* Types
    * Example: `#define __UINT_LEAST8_TYPE__ unsigned char`
* Functions
    * Example: `#define __INT64_C(c) c##L`
* Raw syntax
    * Example: `#define __INT64_C_SUFFIX__ L`

Only the types may be seen in use sites.  Example:

```c
struct foo {
    __UINT_LEAST8_TYPE__ a;
    __UINT_LEAST8_TYPE__ b;
};
```

Currently, we simply generate bindings with the underlying type.  IMHO, this behavior is fine.

```haskell
data Foo = Foo
  { foo_a :: FC.CUChar
  , foo_b :: FC.CUChar
  }
```

If we want to add support for macros to `hs-bindgen`, however, perhaps the goal would be to allow users to generate `newtype`s?

```haskell
newtype UINT_LEAST8_TYPE = UINT_LEAST8_TYPE
  { unUINT_LEAST8_TYPE :: FC.CUChar
  }

data Foo = Foo
  { foo_a :: UINT_LEAST8_TYPE
  , foo_b :: UINT_LEAST8_TYPE
  }
```

That would require knowing the underlying type.  It is *possible*, if we use "stringification" like in the `hs-bindgen-cli info builtin-macros` utility and parse the resulting string.  We could do this for all builtin macros and create `C.Decl`s for those where types are successfully parsed.  We would not need a separate `TypeBuiltin` constructor to do this.

Thoughts?

TravisCardwell:
Regarding names of builtin macros, I think that the name of a builtin macro is no different from the name of an ordinary macro.  I am not sure what we would gain from distinguishing builtin macros at the name level.  There may be good reasons to distinguish them at a declaration level, for documentation if nothing else.

On the topic of builtin macro names, I should point out that function macros are tricky.  The spelling for builtin macro `__INT64_C(c)` is just `__INT64_C`, and we are not able to inspect the tokens for builtin macros.  I do not think that this will be an issue because we will not use such macros.

TravisCardwell:
I have yet to get an example of a builtin `typedef` (#1160) that I can investigate.  I do not know how to enumerate them, and I have not seen any documentation.  (The priority of that issue is low, so I have not dug into it.)

I suspect that the situation is similar, however: there may be good reasons to distinguish them at a declaration level but not a name level.

TravisCardwell (approved):
Looks good, and I like how it simplifies things a bit.

TravisCardwell:
> If this all makes sense to you, feel free to merge this PR; and then perhaps you could take a look at introducing `DeclName` and `ScopedName` in `HsBindgen.Language.C`, deleting `Name` from `Naming`, and then I can take it from there again?

:+1: I posted my status in a [comment](https://github.com/well-typed/hs-bindgen/issues/1387#issuecomment-3635013591) in #1387.

edsko:
Opened https://github.com/well-typed/hs-bindgen/issues/1394 about supporting built-in macro types.

PR #1393, opened 2025-12-09
Improve error messages when tests unexpectedly fail or succeed

dschrempf:
The new messages are, for example

```
Expected 'hs-bindgen' to succeed, but received an error: BindgenErrorReported AnErrorHappened

Expected 'hs-bindgen' to fail, but it succeeded with the following list of declarations:
fromList []
```


dschrempf:
Putting on the queue since this is a tiny change that only affects the test suite output.

Issue #1375, opened 2025-12-04
Never rename C declarations


edsko:
Currently we rename C declarations, for example, struct `foo` inside typedef `foo` we rename to `foo_Deref`; but this is quite weird: it's not the C declaration that changes, but rather it's _Haskell_ name. We do this only because `MangleNames` currently sits very late in the frontend pipeline (and was originally not even part of the frontend at all). Now that we're streamlining all this as part of https://github.com/well-typed/hs-bindgen/issues/1267 , we can fix this; this would then obsolete the need for a `NameOrigin`.

edsko:
Going to close this as `wontfix`, as being able to manipulate declarations prior to generating Haskell types is actually quite useful; for example, we squash typedefs, introduce auxiliary definitions, etc. We could do all that differently, perhaps, but I'm not sure it's worth spending the time. The new approach for `DeclId`, and in particular the new `origDeclId` field, introduced in https://github.com/well-typed/hs-bindgen/pull/1381 already improves the situation quite significantly.

edsko:
Going to reopen this after all, but kind of tentatively, keeping it open as an option. See also https://github.com/well-typed/hs-bindgen/issues/1387 .

edsko:
@TravisCardwell Let's keep this in mind as a potential alternative, somewhere, somehow. It would be quite nice if `origDeclId` would not be needed at all.

edsko:
Some thoughts here:

```c
typedef struct { int x; int y; } point;
```

We have a few choices of how to refer to this struct in say binding specifications:

1. We can refer to it as `point`; that is, _using the name of the typedef_. This is what we do as of https://github.com/well-typed/hs-bindgen/pull/1381 , but it means we effectively "rename": we assign the name of the typedef to the struct (this happens in `HandleTypedefs`).
2. We could refer to it as `@point`, that is, the name _assigned_ to the anonymous struct (in `NameAnon`). Unfortunately, this is possible anymore as of https://github.com/well-typed/hs-bindgen/pull/1292, where we have given up on trying to always accurately detect anonymous structs, which `libclang` basically makes impossible.
3. We could refer to it as `struct point`, using the name assigned to the struct by `libclang >= 16`. This is "effectively" using the original name (there is https://github.com/well-typed/hs-bindgen/issues/1307 to worry about, but that's hopefully relatively minor).

[...]

PR #1395, opened 2025-12-10
Introduce `DeclName` and `ScopedName`

edsko:
Opening this draft PR to track the work that @TravisCardwell and I are doing on this.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/AST/Internal.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/AST/Internal.hs
@@ -41,7 +41,7 @@ module HsBindgen.Frontend.AST.Internal (
     -- * Show
   , ValidPass
     -- * Helper functions
-  , declQualName
+  , declName
   ) where
 
 import Prelude hiding (Enum)
@@ -332,7 +332,7 @@ newtype Comment p =
 -- between namespaces (i.e., @struct foo@ is simply referred to as @foo@). In
 -- 'MangleNames' we will /search/ for a matching name and set the @HaskellId@
 -- accordingly, so that we can generate an approprate reference in the Haddocks.
-data CommentRef p = CommentRef C.Name (Maybe (HaskellId p))
+data CommentRef p = CommentRef Text (Maybe (HaskellId p))
 
 {-------------------------------------------------------------------------------
   Macros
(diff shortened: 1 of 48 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/AST/Internal.hs
```diff
@@ -411,7 +411,7 @@ data Type p =
 -- Invariant: the 'Hs.Identifier' must satisfy the rules for legal Haskell
 -- names, for its intended use (constructor, variable, ..).
 data NamePair = NamePair {
-      nameC       :: C.Name
+      nameC       :: C.ScopedName
```
edsko, line 417: We should explicitly point out that `NamePair` is only used for `ScopedName`s (this is true as of https://github.com/well-typed/hs-bindgen/pull/1336).

hs-bindgen/src-internal/HsBindgen/Frontend/Naming.hs
```diff
@@ -320,8 +188,7 @@ checkIsBuiltin curr = do
 -- All declarations have names after renaming in the @NameAnon@ pass.  This type
```
edsko, line 157: Stray comment

edsko:
@TravisCardwell There were still a _lot_ of changes required to push this through :) But I think it's done now. NOTE: I've rebased (force-pushed) this branch to insert a tiny commit at the start ("Simple utility for experimenting with language-c"), but I have _not_ changed your code. Instead, I have pushed another commit on top with my changes, so that it's easier for you to see what changed and what has not. 

If this all looks OK to you, could you merge your commit and mine so that we don't end up with commit points on `main` that break the build, and then merge?

Then in the morning let's talk about whether we can make #1375 a reality, I think we're inching closer, and it would be so nice not to have to worry about this "decl id" vs "orig decl id" thing anymore; especially functions like `resolveExtBinding` that take both are confusing.

TravisCardwell:
Indeed, this change cuts deep.

In reviewing the changes, I noticed remaining references to `QualName` within the names of variables and a function.  They were all trivial to fix.  I pushed this commit so that the `diff` is available here, before squashing.  Note that the branch is also rebased on the current `main`.

I squashed the commits, annotating the squashed commit so that it is recognized as having multiple authors.

PR #1399, opened 2025-12-10
Introduce `FunctionName` and use `UniqueSymbol` in `ForeignImportDecl`

dschrempf:
This is a prerequisite for #1319 and introduces a distinction between exported and internal function names.

(PR 1 in our internal communications).

Like so, we can ensure that we only rename "exposed" functions (and, in particular, functions without internal use sites).

#### Summary of a meeting we had, clarifying nomeclature

We have to be careful with names:
- Original C function
- C wrapper (beware, we use the word `wrapper` only in this case); has a globally unique name (now called `UserlandCAPIWrapper`)
- Foreign import of C wrapper (uses the same name as the C wrapper).
  The foreign import should _not_ have a Haddock.

We must export a Haskell function with the same name as the C function, and with the same signature.

__We distinguish__: Does the C wrapper have the same signature as the original C function?

If the function is "not fancy", the C wrapper has the same signature.
- Create an alias `aliasOrig` to the foreign import of the C wrapper. The alias has the same name as the original C function and no suffix.

  Haddocks should state that the `aliasOrig` corresponds to the original C function

If the function is "fancy", the C wrapper does NOT have the same signature. We need another function on the Haskell side. We also want to have access to the foreign import.
- Create an alias `aliasCWrapper` to the foreign import of the C wrapper. The alias has a different type than the original C function. Use the suffix `_wrapper` (may later be moved into a separate module so we avoid the suffix altogether).

  Haddocks should state that we provide `aliasCWrapper` as a convenience function if the user has some of the arguments already in memory.

- Create a Haskell function `restoreOrigSignature` with the same name as the original C function, and the same signature.

  Haddocks should state that the `restoreOrigSignature` corresponds to the original C function

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST.hs
@@ -76,6 +76,7 @@ import HsBindgen.Backend.Hs.AST.Type
 import HsBindgen.Backend.Hs.CallConv
 import HsBindgen.Backend.Hs.Haddock.Documentation qualified as HsDoc
 import HsBindgen.Backend.Hs.Origin qualified as Origin
+import HsBindgen.Backend.SHs.AST (FunctionName)
 import HsBindgen.Backend.SHs.AST qualified as SHs
 import HsBindgen.Backend.UniqueSymbol (UniqueSymbol)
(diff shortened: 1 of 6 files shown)


dschrempf:
Will also close #1400

PR #1402, opened 2025-12-11
Improve Select pass trace message

dschrempf:
Trace constructor name: `TransitiveDependencyOfDeclarationUnselectable`

Pretty printed trace:

```
[Warning] [HsBindgen] [select] Could not select declaration:
  'f' at "./hs-bindgen/examples/golden/program-analysis/selection_bad.h:6:6" (direct select predicate match) because a transitive dependency is unavailable:
    'size_t' at "./hs-bindgen/examples/golden/program-analysis/selection_bad_size_t.h:2:15": not selected
```

However, I do think we have a problem with nomenclature here: "Unselectable" vs "unavailable". Internally, we call declarations that we can not select "unselectable". But some of them are "unselectable" to us (i.e., `hs-bindgen`) because a user didn't select them in the first place. That contradicts them being "unselectable", if that makes sense :stuck_out_tongue:.

EDIT:
We decided to not export `Unselectable`, and clearly state that the selectability is "from the perspective of `hs-bindgen`" (and not the user).

Further, we split the messages up into one saying the transitive dependency is unusable, and another one saying that the transitive dependency is not selected.

#### New messages

```
[Warning] [HsBindgen] [select] Could not select declaration:
  'f' at "./hs-bindgen/examples/golden/program-analysis/selection_bad.h:6:6" (direct select predicate match) because a transitive dependency is not selected:
    'size_t' at "./hs-bindgen/examples/golden/program-analysis/selection_bad_size_t.h:2:15"
    Consider adjusting the select predicate
```

Transitive dependency unusable:

```
[Warning] [HsBindgen] [select] Could not select declaration:
  'struct DependOnFailByValue' at "./hs-bindgen/examples/golden/program-analysis/selection_fail.h:16:8" (direct select predicate match) because a transitive dependency is unusable:
    'struct Fail' at "./hs-bindgen/examples/golden/program-analysis/selection_fail.h:6:8": parse failed
```

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
@@ -44,6 +44,20 @@ type DeclId = C.PrelimDeclId
 -- Declaration itself.
 type Decl = C.Decl Select
 
+
+-- | Internal data type! This data type 'Unselectable' refers to declarations
+-- that are not selectable _from the perspective of `hs-bindgen`_; and, in
+-- particular, not from the perspective of the user (they can change the select
+-- predicate).
+data Unselectable =
+    -- | We (i.e., `hs-bindgen`) can not select a declaration selected because
+    --   it or one of its dependencies is unusable.
+    UnselectableBecauseUnusable Unusable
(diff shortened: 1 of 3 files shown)


dschrempf:
I quote: "No space left on device" lol?

dschrempf:
Trying again.

Issue #1307, opened 2025-11-19
Deal with `clang`-generated name collisions


edsko:
Given

```c
struct foo { char a; char b; };
typedef struct { double x; double y; } foo;
```

`clang-16` and up assign the name `foo` to the anonymous struct (older versions return an empty string), resulting in name clashes when we construct the `DeclIndex` (which will then throw out _both_ `struct foo`s, and then selection will throw out the typedef also due to missing transitive dependency). This is a bit frustrating, because in `HandleTypedefs` we will then _squash_ the second `struct foo`, and the name clash would disappear again. 

If we theoretically would squash the typedef as we process it, we would avoid the issue. In practice, this is hard to do: we use the `UseDecl` graph to find out which structs have exactly one use site, and we don't have the `UseDecl` graph until parsing is complete. Just to be explicit, during parsing, instead of

```c
typedef struct { double x; double y; } foo;
```

we will see

```c
typedef struct foo { double x; double y; } foo;
```

so it's not _a priori_ clear that this struct cannot have other use sites (_if_ we could reliably detect anonymous structs we could do better, but we can't; that's precisely the problem). See also detailed discussion in https://github.com/well-typed/hs-bindgen/pull/1292.

edsko:
Going to make this as a low priority enhancement, because I think this example will be rare in practice. Not calling this a `bug`, because we're not generating _invalid_ code; we're merely missing some declarations.

edsko:
Going to close this as `wontfix`: supporting this use case affects the simplification of the identifiers quite significantly, so it has a high cost, for very little benefit. Note that @dschrempf 's work on collision detection means that we at least detect the situation, report an appropriate warning, and just omit the declarations.

PR #1404, opened 2025-12-11
New `AssignAnonIds` pass

edsko:
@TravisCardwell I got quite a way with this, but didn't _quite_ finish. Status: 

* There is now a new pass `AssignAnonIds`, which is ultimately going to replace `NameAnon` (but hasn't yet)
* The usage analysis of anon IDs is done, as well as the logic for assigning names
* The logic for _applying_ this is partly done: we update def sites, but not yet use sites, so this would be the first port of call for you. 
* After that we need to update the rest of the pipeline to replace `NameAnon` by `AssignAnonIds`, this is not yet done at all (that is, not even started).

Note that this is so far all untested code, but hey, it's Haskell, and it type-checks, so it must be right?


edsko:
This is a precursor to https://github.com/well-typed/hs-bindgen/issues/1375#issuecomment-3640839932 ; the point is that after this PR, the `DeclIndex` can use normal `DeclId`, rather than `PrelimDeclId`.

TravisCardwell:
The goal of `AssignAnonIds` is to assign names to anonymous types *before* we construct the `DeclIndex`, `UseDeclGraph`, and `DeclUseGraph` in the `ConstructTranslationUnit` pass.

Current passes:

1. `Parse`
2. `ConstructTranslationUnit`
3. `HandleMacros`
4. `NameAnon`
5. ...

Target passes:

1. `Parse`
2. `AssignAnonIds`
3. `ConstructTranslationUnit`
4. `HandleMacros`
5. ...

We must name uses of anonymous types in macros, however.  In `AssignAnonIds`, macros are still unparsed; they are represented using a list of tokens.  The `HandleMacros` pass now just registers failed macros in the `DeclIndex`, though, so perhaps we could handle macros earlier and pass the list of failed macros to the `ConstructTranslationUnit` pass...

Passes that resolve this issue:

1. `Parse`
2. `HandleMacros`
3. `AssignAnonIds`
4. `ConstructTranslationUnit`
5. ...

EDIT: Oh, I just realized that the `ConstructTranslationUnit` pass **sorts**, and `HandleMacros` relies on that!  :thinking:

TravisCardwell:
There was a `TODO` in `UpdateUseSites`:

```haskell
-- TODO: We should think about whether or not this case fail. In principle
-- the fact that we have a use site means that the anonymous type is used;
-- however, we also don't assign a name when an anonymous type is /unusable/,
-- rather than unused (for example, if they appear in a function sig).
```

If there is an unusable anonymous type, such as one in a function signature, we must process it when updating use sites.  We cannot leave such a type as is, so we have no choice but to fail so that the declaration is dropped, right?

I renamed the existing trace message to `AssignAnonIdsSkippedDecl` because it is specific to declarations (with a `DeclInfo` argument), and I added a new `AssignAnonIdsSkippedUse` trace message for this case.  We already warn about this, so the log level is also `Debug`.

I implemented the `UpdateUseSites` instances.  I think that we will likely move the `HandleMacros` pass as suggested in my previous comment, so I went ahead and implemented the instances for `C.CheckedMacro` and `C.CheckedMacroType`, and I stubbed the case as follows.

```haskell
C.DeclMacro macro ->
  -- TODO https://github.com/well-typed/hs-bindgen/pull/1404#issuecomment-3644126724
  undefined
  --C.DeclMacro    <$> updateUseSites chosenNames macro
```

If we indeed go this route, we need to delete the `TODO` and `undefined`, as well as uncomment that last line.  We will also need to update pass names to reflect the new ordering.

If we go a different route, we should not forget to remove those instances.

TravisCardwell:
Discussing with @edsko, we can just keep the *unparsed* macro, without any use updates.  I changed the type to `type MacroBody    AssignAnonIds = UnparsedMacro` and implemented this.

TravisCardwell:
The `ParseResult` type uses types appropriate for the `Parse` pass.  In particular, `ParseSuccess` contains a `C.PrelimDeclId` and a `C.Decl Parse`.  Some refactoring is required to make `ParseSuccess` work with `C.PrelimDeclId` before assigning names and `C.DeclId` after assigning names.

Combining the passes (#1408) will allow us to keep `C.PrelimDeclId` internal to the `Parse` pass.  The output of the `Parse pass` and all other passes will use `C.DeclName`, allowing us to remove `Id` from `IsPass`.

TravisCardwell:
While orienting myself, I will document our design decisions.

We decided to name anonymous declarations *before* the `ConstructTranslationUnit` pass so that the `DeclIndex`, `UseDeclGraph`, and `DeclUseGraph` can all use named identifiers as keys.  A new `AssignAnonIds` pass implements this, and the old `NameAnon` pass is removed.  The passes are now as follows.

1. `Parse`
2. `AssignAnonIds`
3. `ConstructTranslationUnit`
4. `HandleMacros`
5. `ResolveBindingSpecs`
6. `Select`
7. `HandleTypedefs`
8. `MangleNames`

The `DeclIndex`, `UseDeclGraph`, and `DeclUseGraph` now use `C.DeclId AssignAnonIds` as keys.  Here are the relevant types, where the `C` types are still straightforward and `HaskellId AssignAnonIds` is `()` until the `MangleNames` pass.

```haskell
data DeclId (p :: Pass) = DeclId{
      name       :: C.DeclName
    , origDeclId :: OrigDeclId
    , haskellId  :: HaskellId p
    }

data OrigDeclId =
    OrigDeclId PrelimDeclId
  | AuxForDecl PrelimDeclId

data PrelimDeclId =
    PrelimDeclIdNamed C.DeclName
  | PrelimDeclIdAnon AnonId C.NameKind

newtype AnonId = AnonId SingleLoc
```

We still have to know the `PrelimDeclId` (in addition to assigned names) to query the `DeclIndex`, `UseDeclGraph`, and `DeclUseGraph`, which is not as clean as it was before.  The benefit is that traversal of a graph provides `DeclId`s directly in results, so we do not have to deal with `PrelimDeclId`s *directly* in later passes.  They are still used, but we do not have to think too much about them.

Note that the `PrelimDeclId` is used to determine if a `DeclId` has an anonymous origin.  A binding specification with a `cname` of `struct @foo` matches a declaration with a `name` of `DeclName "foo" (NameKindTagged TagKindStruct)` and a `origDeclId` of `OrigDeclId PrelimDeclIdAnon{}` (where we assume that the `C.NameKind` is consistent).

We considered combining the `Parse`, `AssignAnonIds`, and `ConstructTranslationUnit` passes into single, initial `Parse` pass.  The motivation was to use a single `Id` type across the passes, without any type parameters.  That does not work well, as we use the parameter for the `HaskellId`, and we cannot really encapsulate type `PrelimDeclId` within the `Parse` pass because it is used in `DeclId`.

With the `Id` type parameter, we have types like the following.

| #    | Pass                       | `Id p`                              |
| ---: | :---                       | :---                                |
| 1    | `Parse`                    | `C.PrelimDeclId`                    |
| 2    | `AssignAnonIds`            | `C.DeclId AssignAnonIds`            |
| 3    | `ConstructTranslationUnit` | `C.DeclId ConstructTranslationUnit` |
| 4    | `HandleMacros`             | `C.DeclId HandleMacros`             |
| 5    | `ResolveBindingSpecs`      | `C.DeclId ResolveBindingSpecs`      |
| 6    | `Select`                   | `C.DeclId Select`                   |
| 7    | `HandleTypedefs`           | `C.DeclId HandleTypedefs`           |
| 8    | `MangleNames`              | `C.DeclId MangleNames`              |

The `ResolveBindingSpecs` pass processes IDs of type `C.DeclId HandleMacros` and must coerce them to type `C.DeclId AssignAnonIds` for use with the `DeclIndex`, `UseDeclGraph`, and `DeclUseGraph` (as well as coerce them to type `C.DeclId ResolveBindingSpecs` to pass on to the next pass).  We did not yet have `CoercePass` instances for such "reverse" coercion, so I added some.  Is this an appropriate approach, or is there a different one that is preferred?

I refactored the binding specification modules.  The change of unique ID from `C.PrelimDeclId` to `C.DeclId AssignAnonIds` is straightforward.  A more significant change is handling of anonymous types.  `DeclId p` cannot be used because we cannot parse them from strings (in YAML) since they contain source locations.  Some tentative code in `HsBindgen.BindingSpec.Gen` implemented the `@` syntax, but it was not in a good place because that syntax needs to be used in other places as well.  I tentatively added a `C.NameAnon` type (in `HsBindgen.Frontend.Naming`) that wraps `Text` and represents C name (spellings) with `@` syntax when a name is assigned by `hs-bindgen` to an anonymous type.  Function `declIdNameAnon :: DeclId p -> Maybe NameAnon` is still a `Maybe` type because we need the `Nothing` case for auxiliary declarations, and there are now more `panicPure` cases that should never be reached because auxiliary declarations are not created until the `NameTypedefs` pass.

I am committing the binding specification changes as a separate (WIP) commit.

TravisCardwell:
I continued refactoring, and now the library and CLI builds.  Tests still need to be refactored.  `AttachedParseMsg` has been removed; investigations indicate that this type was not simply renamed for moved to a different module.  We need to refactor the tests to work with whatever new API is implemented.

TravisCardwell:
(Rebased)

TravisCardwell:
I refactored the tests.  They now build!  Now to work on them.

```
145 out of 822 tests failed
```

Note that I replaced `expectFromPrelimDeclId` with `expectFromDeclId`, and I went ahead and used `C.declNameText` to render the C name (`struct foo` instead of just `foo`).  That likely accounts for some failures; those will be trivial to fix.  (IMHO, who really should use `@` syntax in traces and tests.)

edsko:
Mostly orthogonal, but since I happen to see it: the `NameKind` of `DeclOpaque` is probably no longer necessary now that everything is always qualified.

edsko:
Note to self: are we dealing with

```c
typedef struct { int x; int y; } points[10];
```

?

edsko:
Another thought: now that `AssignAnonIds` no longer uses the `DeclUseGraph`, do we still need edge labels in the `DeclUseGraph` or `UseDeclGraph`?

edsko:
Another thought: I _think_ the current represntation in `Analyis/AnonUsage` would actually _miss_ the case in https://github.com/well-typed/hs-bindgen/issues/1418.

TravisCardwell:
> Another thought: now that `AssignAnonIds` no longer uses the `DeclUseGraph`, do we still need edge labels in the `DeclUseGraph` or `UseDeclGraph`?

Do we use edge information in trace messages?  If nothing else, edges are nice as documentation when dumping a graph.

TravisCardwell:
All binding spec fixture changes were just insignificant reordering, due to using a different type as the key of the `Map` in the data type.

TravisCardwell:
There was one remaining `DeclOpaque` issue, probably caused by merging.  With that fixed, all tests pass!  I rebased, so the branch is current.

edsko:
Removing `origDeclId`:

* We need it for uniqueness in the graphs. There is the pesky problem of https://github.com/well-typed/hs-bindgen/issues/1307 , which we decided to ignore. The only side note here is that we need to ensure that when we construct the translation unit, we construct the `DeclIndex` _before_ anything else, because that is where we do conflict detection. 

* We still need to know if something was originally anonymous or not. Proposal:

```hs
data DeclId = 
    -- | Named declaration
    --
    -- The name is as it appears in the C source (or appears to be that way,
    -- due to clang assigned names for anonymous declarations).
    DeclNamed DeclName

    -- | Anonymous declaration
    --
    -- Declaration without a name in the C source, such as an anonymous struct
    -- inside another struct.
    -- 
    -- We do /NOT/ record the original anon ID here, because that is a source
    -- location, which is impossible to construct in many places (for example,
    -- when parsing @struct \@foo@ in binding specs).
  | DeclAnon DeclName

    -- | Auxiliary datatype in support /of/ a C declaration, but that does not
    -- itself exist in the C source at all (this is what distinguishes these from
    -- 'DeclAnon': they /do/ appear in the code, just don't have a name).
    --
    -- We record the declaration that this is in support of. We do /NOT/
    -- give this a name of its own ('DeclId' names are names as they appear
    -- in the C source, but these do not exist in C).
    --
    -- TODO <https://github.com/well-typed/hs-bindgen/issues/1379>
    -- If we add support for auxiliary types in binding specs, we need to
    -- invent a syntax for these; it might be something as simple as
    -- @aux: Foo@ to denote the auxiliary type for @Foo@ (that is,
    -- we specify /this/ 'DeclId', but not the 'DeclName' assigned /to/
    -- the auxiliary type).
  | DeclAux DeclId
```

then

```hs
type family Id Parse         = PrelimDeclId
type family Id AssignAnonIds = DeclId
type family Id ...           = DeclId
type family MangleNames      = (DeclId, Hs.Identifier)
```

We'll make these changes as part of this PR.

edsko:
The change from `_Deref` to `_Aux` we'll do in a separate PR, to have an easier time checking the fixtures.

TravisCardwell:
The C names rendered in Haddock documentation got messed up, so that even `struct foo` was rendered as just `foo` (with no anonymous types or squashing involved).  This is even in `main`!  The fixture diffs confirm that this is fixed now.

TravisCardwell:
I worked on tests a bit first, accepting the correct fixtures, with have the following changes:

* Haddock documentation C names are now correct.  (See previous comment.)
* We changed the way we refer to anonymous declaration in squashed `typedef`s.  `typedef struct { ... } foo` is now referred to as `struct foo`, *not* `foo`.

Note that there were some changes to the generated C code in test `macros/macro_in_fundecl_vs_typedef`.  The changes are (still) correct and are more consistent with the C source.

Status: 42 out of 836 tests failed

TravisCardwell:
Haddock documentation for auxiliary declarations document C declarations that do not exist.

Example:

```c
typedef void (*f1)(int x, int y);
```

The Haddock documentation includes:

```
__C declaration:__ @f1_Deref@
```

Tests with this issue:

* [ ] `documentation/doxygen_docs`
* [ ] `edge-cases/distilled_lib_1`
* [ ] `functions/callbacks`
* [ ] `functions/circular_dependency_fun`
* [ ] `macros/reparse`
* [ ] `types/typedefs/multi_level_function_pointer`
* [ ] `types/typedefs/typedefs.h`

Perhaps we should change this?

On a related note, Haddock documentation for anonymous declarations also document C declarations that do not exist, but in this case `libclang` does not disclose if they exist or not.

hs-bindgen/fixtures/documentation/doxygen_docs/Example.hs
```diff
@@ -220,6 +220,8 @@ pattern COLOR_BLUE = Color_enum 2
 
 {-| Auxiliary type used by 'Event_callback_t'
 
+__C declaration:__ @event_callback_t@
```
edsko, line 228: This is a "true" change in the docs we generate (we previously did not mention C declarations for auxiliary types), but I think one that is acceptable.

hs-bindgen/fixtures/documentation/doxygen_docs/Example.hs
```diff
@@ -606,6 +608,8 @@ pattern STATUS_ERROR = Status_code_t (-99)
 
   Allows access to high and low parts separately
 
+__C declaration:__ @struct \@data_union_t_as_parts@
```
edsko, line 618: This too is a true change, but a good one: we now record the anonymous C declaration, using the same syntax as in binding specs. I verified that this way of escaping the `@` sign  works.

hs-bindgen/fixtures/documentation/doxygen_docs/bindingspec.yaml
```diff
@@ -77,6 +83,9 @@ hstypes:
   - Ord
   - Show
   - Storable
+- hsname: Event_callback_t_Deref
```
edsko: This auxiliary type does not belong in the binding specs. That will be resolved when we generate binding specs from the C declarations rather than the Haskell ones. (cc @TravisCardwell )

hs-bindgen/fixtures/edge-cases/flam/bindingspec.yaml
```diff
@@ -4,15 +4,15 @@ version:
...
-  hsname: Foo_bar
 - headers: edge-cases/flam.h
   cname: struct diff
   hsname: Diff
 - headers: edge-cases/flam.h
   cname: struct foo
   hsname: Foo
+- headers: edge-cases/flam.h
+  cname: struct @foo_bar
+  hsname: Foo_bar
```
edsko, line 15: I'm a bit puzzled by this reordering, but I guess it doesn't matter.

hs-bindgen/fixtures/macros/macro_in_fundecl_vs_typedef/Example/FunPtr.hs
```diff
@@ -98,7 +98,7 @@ $(HsBindgen.Runtime.Prelude.addCSource (HsBindgen.Runtime.Prelude.unlines
   , "/* test_macrosmacro_in_fundecl_vs_typ_Example_get_struct_name3_ptr */"
   , "__attribute__ ((const))"
   , "void (*hs_bindgen_680e8a5d673ce9c1 (void)) ("
-  , "  struct4 *arg1,"
+  , "  struct struct4 *arg1,"
```
edsko, line 101: Since we do not change `DeclId` at all anymore, we now mirror struct vs typedef usage in C also in the wrappers that we generate. I think that's fine.

hs-bindgen/fixtures/documentation/doxygen_docs/Example.hs
TravisCardwell: This one is documenting the C declaration that the auxiliary type is for.  I agree that it is acceptable. :+1:

hs-bindgen/fixtures/documentation/doxygen_docs/Example.hs
TravisCardwell: Great!  I like it!  :+1:

hs-bindgen/fixtures/edge-cases/flam/bindingspec.yaml
TravisCardwell: We recently decided to order C types by source location, but I have not implemented it yet.  It should not be too difficult...  I guess we should use the same traversal (order) of the include graph, which would make the order the same as the order of `[C.Decl]` in the internal AST.

hs-bindgen/fixtures/macros/macro_in_fundecl_vs_typedef/Example/FunPtr.hs
TravisCardwell: I noticed this one as well, and think it is better that we follow the C. :+1:

hs-bindgen/fixtures/edge-cases/flam/bindingspec.yaml
TravisCardwell: Reviewing fixture changes now, I confirm that the ordering is very peculiar!

TravisCardwell:
I added a commit that generates C type specs for squashed `typedef`s.  I did not write details in the commit message since the commit will likely be squashed.

We need this to work now because the `vector` test uses external bindings and needs it.  This will be implemented for real as part of the C representation (#1356) in a separate PR, *after* we refactor to generate binding specifications in the frontend (#1448), but the current "workaround" just adds the minimum specification required and only works because we have not yet implemented squashing specification.  I am currently in "hack mode," unsure of the final design, so I made minimal possible changes for now.

In the `DeclIndex`, I added a `UsableSquashed SourcePath Hs.Identifier` constructor.  We pattern match before `UsableSquashed` is even used, which is a bit awkward, but we already do so with `UsableExternal`.  Function `registerSquashedDeclarations` simply replaces any `UsableSuccess` entries, like is done when registering omitted declarations.  Function `getSquashed` gets the squashed entries, like is down to get the omitted entries.

The lookup of the `Hs.Identifier` is done in `MangleNames` using the `NameMap`, which requires the `C.DeclId` of the wrapped type.  Both the `SourcePath` and the `C.DeclId` of the wrapped type are in the `DeclUseGraph`, so I added these to the `Squash` constructor in the `TypedefAnalysis`.

When updating the `DeclIndex` in `MangleNames`, it is possible for a squashed type to not be found in the `NameMap`.  I do not think that selection is reflected in the `DeclIndex` (yet?), so this is not always an error.  `MangleNames` is after `Select`, so a squashed entry may refer to a declaration that is not selected.  I ignore this case.

We do not generate binding specifications in the frontend yet, so I pass the information about squashed types to `genBindingSpec` in the same way that information about omitted types is passed.  Cached `frontendSquashedTypes` are added to the `FrontendArtefact` and are requested by `writeBindingSpec`.

Note that `genBindingSpec` still uses a `Map`, so the order is quite strange.  This will be resolved when we order by source location).

All fixture changes look good!

Here is a minimal example:

```c
typedef struct { double x, y; } pt;
```

We previously generated the following binding specification:

```yaml
version:
  hs_bindgen: 0.1.0
  binding_specification: '1.0'
target: x86_64-pc-linux-gnu
hsmodule: Generated
ctypes:
- headers: ex.h
  cname: struct pt
  hsname: Pt
hstypes:
- hsname: Pt
  instances:
  - Eq
  - Show
  - Storable
```

With this change, we generate the following binding specification:

```yaml
version:
  hs_bindgen: 0.1.0
  binding_specification: '1.0'
target: x86_64-pc-linux-gnu
hsmodule: Generated
ctypes:
- headers: ex.h
  cname: pt
  hsname: Pt
- headers: ex.h
  cname: struct pt
  hsname: Pt
hstypes:
- hsname: Pt
  instances:
  - Eq
  - Show
  - Storable
```

TravisCardwell:
Yesterday, I added code so that C type specifications are generated for squashed `typedef`s, but when talking about it with @edsko, we realized that `MangleNames` is not choosing names of squashed types correctly when specified using C type specifications.

Using `typedef struct foo { int x, y; } bar;`, where this is the only use of `struct foo` in the C headers, as an example:

* If a name is specified for neither `struct foo` nor `bar`:
    * `struct foo` is translated to Haskell type `Bar`, since the name of the `typedef` is used.
    * `bar` is not translated to a Haskell type, as it is squashed (considered an alias of `struct foo`).
    * Generated binding specifications for `struct foo` and `bar` should both specify Haskell name `Bar`.

* If a name is specified for `struct foo` but not `bar`:

    ```yaml
    ctypes:
      - headers: ex.h
        cname: struct foo
        hsname: Hoge
    ```

    * `struct foo` is translated to Haskell type `Hoge`, since the name is specified.
    * `bar` is not translated to a Haskell type, as it is squashed (considered an alias of `struct foo`).
    * Generated binding specifications for `struct foo` and `bar` should both specify Haskell name `Hoge`.

* If a name not specified for `struct foo` but is specified for `bar`:

    ```yaml
    ctypes:
      - headers: ex.h
        cname: bar
        hsname: Piyo
    ```

    * `struct foo` is translated to Haskell type `Piyo`, since the name of the `typedef` is used, and that name is specified.
    * `bar` is not translated to a Haskell type, as it is squashed (considered an alias of `struct foo`).
    * Generated binding specifications for `struct foo` and `bar` should both specify Haskell name `Piyo`.

* If the same name is specified for both `struct foo` and `bar`:

    ```yaml
    ctypes:
      - headers: ex.h
        cname: struct foo
        hsname: Hoge
      - headers: ex.h
        cname: bar
        hsname: Hoge
    ```

    * `struct foo` is translated to Haskell type `Hoge`, since the name is specified.
    * `bar` is not translated to a Haskell type, as it is squashed (considered an alias of `struct foo`).
    * Generated binding specifications for `struct foo` and `bar` both specify Haskell name `Hoge`, like in the prescriptive binding specification above.

* If different names are specified for `struct foo` and `bar`:

    ```yaml
    ctypes:
      - headers: ex.h
        cname: struct foo
        hsname: Hoge
      - headers: ex.h
        cname: bar
        hsname: Piyo
    ```

    * `struct foo` is translated to Haskell type `Hoge`, since the name is specified.
    * `bar` is translated to Haskell `newtype` `Piyo`, since the name is specified.  Since the Haskell name is not the same as that for `struct foo`, it is *not* an alias.
    * Generated binding specifications for `struct foo` and `bar` specify the different names, like in the prescriptive binding specification above.
    * This case is not implemented yet.  The binding specification syntax is implemented, but we have not yet implemented customization of squashing.

As a quick solution for the current PR, I proposed creating a local map of specified names, and we agreed to move forward with this solution for now.  Previously, `nameForDecl` checked for specified names as the list of declarations are traversed.  Now, we build a `NameMap` of specified names first, passing that map to `nameForDecl`.  It is used to check for specified names in the general case as well as for both the wrapped and squashed declarations.  In particular, no recursion is needed.  We can refactor this in the future, as the code stabilized.

----

With this implemented, I tested the above cases.

If a name is specified for neither `struct foo` nor `bar`:

```haskell
data Bar = ...
```

<details><summary>Generated binding specification</summary>

```yaml
version:
  hs_bindgen: 0.1.0
  binding_specification: '1.0'
target: x86_64-pc-linux-gnu
hsmodule: Generated
ctypes:
- headers: ex.h
  cname: bar
  hsname: Bar
- headers: ex.h
  cname: struct foo
  hsname: Bar
hstypes:
- hsname: Bar
  instances:
  - Eq
  - Show
  - Storable
```

</details>

If a name is specified for `struct foo` but not `bar`:

```haskell
data Hoge = ...
```

<details><summary>Generated binding specification</summary>

```yaml
version:
  hs_bindgen: 0.1.0
  binding_specification: '1.0'
target: x86_64-pc-linux-gnu
hsmodule: Generated
ctypes:
- headers: ex.h
  cname: bar
  hsname: Hoge
- headers: ex.h
  cname: struct foo
  hsname: Hoge
hstypes:
- hsname: Hoge
  instances:
  - Eq
  - Show
  - Storable
```

</details>

If a name not specified for `struct foo` but is specified for `bar`:

```haskell
data Piyo = ...
```

<details><summary>Generated binding specification</summary>

```yaml
version:
  hs_bindgen: 0.1.0
  binding_specification: '1.0'
target: x86_64-pc-linux-gnu
hsmodule: Generated
ctypes:
- headers: ex.h
  cname: bar
  hsname: Piyo
- headers: ex.h
  cname: struct foo
  hsname: Piyo
hstypes:
- hsname: Piyo
  instances:
  - Eq
  - Show
  - Storable
```

</details>

If the same name is specified for both `struct foo` and `bar`:

```haskell
data Hoge = ...
```

<details><summary>Generated binding specification</summary>

```yaml
version:
  hs_bindgen: 0.1.0
  binding_specification: '1.0'
target: x86_64-pc-linux-gnu
hsmodule: Generated
ctypes:
- headers: ex.h
  cname: bar
  hsname: Hoge
- headers: ex.h
  cname: struct foo
  hsname: Hoge
hstypes:
- hsname: Hoge
  instances:
  - Eq
  - Show
  - Storable
```

</details>

I added the following tests for the rename cases:

* `binding-specs/bs_pre_rename_squash_both`
* `binding-specs/bs_pre_rename_squash_struct`
* `binding-specs/bs_pre_rename_squash_typedef`

TravisCardwell:
I added a commit that prevents specifications from being generated for auxiliary types.  Discussing with @edsko yesterday, we decided to add a new `Newtype` origin to distinguish auxiliary types.  Note that this is a temporary solution.  After we refactor to do more in the frontend, only using the backend for code generation, we will generate binding specifications from the list of `C.Decl`s.

I added an `Aux C.Typedef` constructor to the `Newtype` origin.  In `Hs.Translation`, I added a `(C.Typedef -> Origin.Newtype)` constructor parameter to `typedefDecs`, allowing the caller to specify which constructor to use. In `genBindingSpec`, this origin is checked, and no specifications are generated for `Aux` `Newtype`s.

hs-bindgen/fixtures/types/complex/vector_test/bindingspec.yaml
```diff
@@ -7,6 +7,9 @@ ctypes:
 - headers: types/complex/vector_test.h
   cname: vector
   hsname: Vector
+- headers: types/complex/vector_test.h
+  cname: struct vector
+  hsname: Vector
```
edsko, line 12: @TravisCardwell I guess this is the critical change to the manual? :)

PR #1411, opened 2025-12-12
`Backend.Hs`: add a smart constructor for newtype declarations

jorisdral:
This is in preparation for #1282

I've also introduced a dedicate state type that currently only holds the `InstanceMap`, but I'm planning on adding additional state

EDIT: I also added a `HsM` monad type that we now use in the `Hs` translation step

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
@@ -3,7 +3,7 @@ module HsBindgen.Backend.Hs.Translation (
     generateDeclarations
   ) where
 
-import Control.Monad.State qualified as State
+import Control.Monad.State qualified as State hiding (MonadState)
 import Data.List qualified as List
 import Data.List.NonEmpty qualified as NonEmpty
 import Data.Map.Strict qualified as Map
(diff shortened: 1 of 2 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/Newtype.hs
```diff
@@ -0,0 +1,52 @@
...
+import HsBindgen.Backend.Hs.Origin qualified as Origin
+import HsBindgen.Backend.Hs.Translation.Instances qualified as Hs
+import HsBindgen.Backend.Hs.Translation.State (TranslationState)
+import HsBindgen.Backend.Hs.Translation.State qualified as State
+import HsBindgen.Imports
+import HsBindgen.Language.Haskell qualified as Hs
+
+-- | Smart constructor for creating a 'Hs.Newtype'
+newtypeDec ::
+     State.MonadState TranslationState m
```
jorisdral, line 19: Maybe we should use a dedicated monad instead, e.g., `HsM`?

edsko (approved):

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/Newtype.hs
jorisdral: Done

Issue #1363, opened 2025-12-02
We lose `FunPtr` information when binding specifications are involved


jorisdral:
In the Haskell FFI, we have a distinction between "ordinary" pointers and *function* pointers. `hs-bindgen` has to figure out for each C pointer whether it should translate to an ordinary pointer or a function pointer, and generate bindings using the former or the latter. When binding specifications are involved, this seems to go wrong in some cases.

I vaguely recall that I had diagnosed this before when I was working on support for function pointers, but I can't find an issue for it now, so here's the gist:

### Example

```c
typedef int A (int);
typedef A B;
int foo (int x (int));
int bar (A x);
int baz (B x);
```

We first generate bindings and a binding spec for just the `A` typedef. Then we do the same for the `B` typedef, using the `A` binding spec as an input. Finally, we generate bindings for the `foo`, `bar`, and `baz` functions using the `A` and `B` binding specs as inputs.

```bash
cabal run hs-bindgen-cli -- \
    preprocess \
    --hs-output-dir . \
    --module A \
    --select-by-decl-name "A" \
    --gen-binding-spec A.bindingspec.yaml \
    foo.h

cabal run hs-bindgen-cli -- \
    preprocess \
    --hs-output-dir . \
    --module B \
    --select-by-decl-name "B" \
    --gen-binding-spec B.bindingspec.yaml \
    --external-binding-spec A.bindingspec.yaml \
    foo.h

cabal run hs-bindgen-cli -- \
    preprocess \
    --hs-output-dir . \
    --module C \
    --external-binding-spec A.bindingspec.yaml \
    --external-binding-spec B.bindingspec.yaml \
    foo.h
```

The resulting modules look roughly like this:

```hs
module A where

newtype A = A
  { un_A :: CInt -> IO CInt
  }
  deriving newtype (HasBaseForeignType)

instance ToFunPtr A where
instance FromFunPtr A where

instance ( TyEq ty ((CFieldType A) "un_A")
instance HasCField A "un_A" where
```

```hs
module B where

import A qualified

newtype B = B
  { un_B :: A.A
  }
  deriving newtype (HasBaseForeignType)

instance ( TyEq ty ((CFieldType B) "un_B")
         ) => HasField "un_B" (Ptr B) (Ptr ty) where
instance HasCField B "un_B" where
```

```hs
module C where

import A qualified
import B qualified

foo ::
     FunPtr (CInt -> IO CInt)
  -> IO CInt

bar ::
[...]

PR #1416, opened 2025-12-15
Add `ffitype` to binding specs

jorisdral:
This is part of #1282

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/BindingSpec/Gen.hs
+++ b/hs-bindgen/src-internal/HsBindgen/BindingSpec/Gen.hs
@@ -15,6 +15,7 @@ import Data.ByteString (ByteString)
 import Data.List.NonEmpty qualified as NonEmpty
 import Data.Map.Strict qualified as Map
 import Data.Set qualified as Set
+import Data.Vec.Lazy qualified as Vec
 
 import Clang.HighLevel.Types
 import Clang.Paths
@@ -164,8 +165,15 @@ genBindingSpec'
                 cTypeSpecIdentifier = Just hsIdentifier
               , cTypeSpecRep        = Nothing  -- TODO implement
               }
+            hsRecordRep = BindingSpec.HsRecordRep {
+                hsRecordRepConstructor = Just $ Hs.Identifier $ Hs.getName $ Hs.structConstr hsStruct
+              , hsRecordRepFields = Just [
+                    Hs.Identifier $ Hs.getName $ Hs.fieldName field
+                  | field <- Vec.toList $ Hs.structFields hsStruct
+                  ]
+              }
             hsTypeSpec = BindingSpec.HsTypeSpec {
-                hsTypeSpecRep = Nothing -- TODO implement
+                hsTypeSpecRep = Just $ BindingSpec.HsTypeRepRecord hsRecordRep
               , hsTypeSpecInstances =
                   mkInstSpecs
                     ( maybe Map.empty BindingSpec.hsTypeSpecInstances
@@ -218,7 +226,7 @@ genBindingSpec'
             , hsNewtypeRepFFIType = Hs.newtypeFFIType hsNewtype
             }
           hsTypeSpec = BindingSpec.HsTypeSpec {
-              hsTypeSpecRep = Just $ BindingSpec.HsTypeRepNewtype $ hsNewtypeRep
+              hsTypeSpecRep = Just $ BindingSpec.HsTypeRepNewtype hsNewtypeRep
             , hsTypeSpecInstances =
                 mkInstSpecs
                   ( maybe Map.empty BindingSpec.hsTypeSpecInstances
--- a/hs-bindgen/examples/golden/program-analysis/program_slicing_simple.h
+++ b/hs-bindgen/examples/golden/program-analysis/program_slicing_simple.h
@@ -4,3 +4,5 @@ struct foo {
   uint64_t sixty_four;
   uint32_t thirty_two;
 };
+
+int bar (uint64_t x, uint32_t y);
--- a/hs-bindgen/examples/golden/program-analysis/program_slicing_simple.yaml
+++ b/hs-bindgen/examples/golden/program-analysis/program_slicing_simple.yaml
@@ -9,6 +9,9 @@ ctypes:
   hsname: Int64
 hstypes:
 - hsname: Int64
+  representation:
+    newtype:
+      ffitype: Int64
   instances:
   - Bits
   - Bounded
--- a/hs-bindgen/fixtures/arrays/array/bindingspec.yaml
+++ b/hs-bindgen/fixtures/arrays/array/bindingspec.yaml
@@ -24,6 +24,12 @@ ctypes:
   hsname: Tripletlist
 hstypes:
 - hsname: Example
+  representation:
+    record:
+      constructor: Example
+      fields:
+      - example_triple
+      - example_sudoku
   instances:
   - Eq
   - Show
--- a/hs-bindgen/fixtures/attributes/attributes/bindingspec.yaml
+++ b/hs-bindgen/fixtures/attributes/attributes/bindingspec.yaml
@@ -21,26 +21,57 @@ ctypes:
   hsname: Qux
 hstypes:
 - hsname: Bar
+  representation:
+    record:
+      constructor: Bar
+      fields:
+      - bar_c
+      - bar_i
   instances:
   - Eq
   - Show
   - Storable
 - hsname: Baz
+  representation:
+    record:
+      constructor: Baz
+      fields:
+      - baz_c
+      - baz_i
   instances:
   - Eq
   - Show
   - Storable
 - hsname: FILE
+  representation:
+    record:
+      constructor: FILE
+      fields:
+      - fILE__r
+      - fILE__w
+      - fILE__close
   instances:
   - Eq
   - Show
   - Storable
 - hsname: Foo
+  representation:
+    record:
+      constructor: Foo
+      fields:
+      - foo_c
+      - foo_i
   instances:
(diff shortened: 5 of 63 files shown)


TravisCardwell:
As this PR is marked as not ready for review, I did not review it.  I took a peek to see *where* the new stuff is put, though.

* `FFIType` is defined in new module `HsBindgen.BindingSpec.Private.FFIType`.  Something defined here should never change across future versions of binding specification.  (That is what `HsBindgen.BindingSpec.Private.Common` is for, so perhaps it would be better to define this type there.)  Is this intended?  (If it is possible that we may change this type in future versions of binding specifications, to change the syntax or fix a bug, it should instead be defined in `HsBindgen.BindingSpec.Private.V1`.)

* In a binding specification, it is added to the `HsTypeSpec`.  This means that it may be specified for *any* type.  We can of course trace errors when it is used inappropriately.  Is this intended?  (I think the initial proposal was to restrict it to `newtype` representations.  FWIW, I am fine with either.  The current design has more error cases but it easier to write.)

jorisdral:
> As this PR is marked as not ready for review, I did not review it. I took a peek to see _where_ the new stuff is put, though.

Whoops, I forgot to remove that text from the PR description. It should be ready to review

>  * `FFIType` is defined in new module `HsBindgen.BindingSpec.Private.FFIType`.  Something defined here should never change across future versions of binding specification.  (That is what `HsBindgen.BindingSpec.Private.Common` is for, so perhaps it would be better to define this type there.)  Is this intended?  (If it is possible that we may change this type in future versions of binding specifications, to change the syntax or fix a bug, it should instead be defined in `HsBindgen.BindingSpec.Private.V1`.)

Ah, okay, I was not aware of that design. I can put it in `V1` instead

>  * In a binding specification, it is added to the `HsTypeSpec`.  This means that it may be specified for _any_ type.  We can of course trace errors when it is used inappropriately.  Is this intended?  (I think the initial proposal was to restrict it to `newtype` representations.  FWIW, I am fine with either.  The current design has more error cases but it easier to write.)

The new `HsTypeRep` field is not implemented everywhere yet (some TODOs here and there), so I thought I'd add `FFIType` to the `HsTypeSpec` instead. We could move it into `HsTypeRep` later. And yes, we might want to use it only for newtype representations, but I'm also thinking it might be nice to keep `FFIType` separate so that it can be used more generally to carry information about types across binding specification boundaries. It's just the beginning of an idea however

EDIT: for example, I've been thinking of how to properly fix https://github.com/well-typed/hs-bindgen/issues/1363. The ffitype field might help there. That's also one reason why I think the ffitype might have to be a required field, so that we can always see whether the underlying type of an external reference is a function, or a foreign type, or something else

TravisCardwell:
> I'm also thinking it might be nice to keep `FFIType` separate so that it can be used more generally to carry information about types across binding specification boundaries. It's just the beginning of an idea however

`ffitype` will be useful for more than just `newtype`s?  `HsTypeSpec` is indeed the appropriate place in that case.

> That's also one reason why I think the ffitype might have to be a required field, so that we can always see whether the underlying type of an external reference is a function, or a foreign type, or something else

We use the same data type and file format for two different purposes, so the data type and file format must support both of them.  `ffitype` must be optional in prescriptive binding specifications, and it is required in external binding specifications.  It must therefore be optional in the data type and file format, and we can check that `ffitype`s are specified when resolving external binding specifications.

In other words, the requirement must be implemented in `Frontend`, not `BindingSpec`.  One option is to implement the requirement in the `resolveExtBinding` function in the `ResolveBindingSpecs` pass.  `hsname` is also optional in the data type and file format but required in external binding specifications, and this function checks that the `hsname` is specified (and traces an error otherwise).  We can add a similar check that the Haskell type is also specified and has a `ffitype` (and trace an error otherwise).  Perhaps it would be easiest to make `extHsSpec :: HsTypeSpec` (no `Maybe`) as well as add a new field for the `FFIType` (no `Maybe`).

hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/FFIType.hs
```diff
@@ -0,0 +1,38 @@
...
+
+import HsBindgen.Runtime.BaseForeignType
+
+data FFIType =
+    Function
+  | Array
+  | Data
+  | Basic BasicForeignType
+  | Builtin BuiltinForeignType
+  deriving stock (Show, Eq, Ord, Generic, Read)
```
jorisdral:
```suggestion
data FFIType =
    Basic BasicForeignType
  | Builtin BuiltinForeignType
  deriving stock (Show, Eq, Ord, Generic, Read)
```

hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/FFIType.hs
```diff
@@ -0,0 +1,38 @@
...
+
+instance Aeson.FromJSON FFIType where
+  parseJSON = Aeson.withText "FFIType" $ \t ->
+    let s = Text.unpack t
+    in  case readMaybe s of
+          Just bft -> return bft
+          Nothing   -> Aeson.parseFail $ "unknown FFI type: " ++ s
+
+instance Aeson.ToJSON FFIType where
+  toJSON = Aeson.String . Text.pack . show
```
jorisdral: Remove `Basic` and `Builtin` from the rendering: just the Haskell type

jorisdral:
@TravisCardwell I'm going to clean up this PR a bit after a review I had with @edsko, I'll let you know once it can be reviewed

jorisdral:
@TravisCardwell this should now be ready for you to take a look, if you want to.

 Some changes:
* I moved `ffitype` into the `HsNewtypeRep` datatype, so that it is only used for newtypes
* `FFIType` now only has the `Basic` and `Builtin` constructors

So for now, `ffitype` is just for newtypes. Also, it's not required to be specified. If `Nothing :: Maybe FFIType`, then it is assumed that the underlying type isn't an FFI type. @edsko and I talked about it and that should probably be fine for now

hs-bindgen/src-internal/HsBindgen/BindingSpec/Gen.hs
```diff
@@ -211,8 +211,13 @@ genBindingSpec'
...
             , cTypeSpecRep        = Nothing  -- TODO implement
             }
+          hsNewtypeRep = BindingSpec.HsNewtypeRep {
+              hsNewtypeRepConstructor = Just $ Hs.Identifier $ Hs.getName $ Hs.newtypeConstr hsNewtype
+            , hsNewtypeRepField = Just $ Hs.Identifier $ Hs.getName $ Hs.fieldName $ Hs.newtypeField hsNewtype
+            , hsNewtypeRepFFIType = Hs.newtypeFFIType hsNewtype
+            }
           hsTypeSpec = BindingSpec.HsTypeSpec {
-              hsTypeSpecRep = Nothing -- TODO implement
+              hsTypeSpecRep = Just $ BindingSpec.HsTypeRepNewtype $ hsNewtypeRep
```
TravisCardwell: I see that I forgot to implement the `hsNewtypeRep` in binding spec generation.  Sorry about that, and thank you for fixing it!

hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/Common.hs
```diff
@@ -265,6 +271,44 @@ instance (Aeson.ToJSON o, Aeson.ToJSON r) => Aeson.ToJSON (AOmittable o r) where
 -- | Aeson representation of 'Omittable' with same omitted and required type
 type AOmittable' a = AOmittable a a
 
+{-------------------------------------------------------------------------------
+  FFI type
+-------------------------------------------------------------------------------}
+
+data FFIType =
```
TravisCardwell: Defining `FFIType` in the `Common` module means that it can never change across binding specification versions.  There is no chance that `BasicForeignType` and `BuiltinForeignType` will ever need to be changed?

The Aeson instance are implemented using `Read` and `Show`.  That is fine if these types will indeed never change.  If it is possible that may be updated, however, the rendering and parsing must be implemented in a different way.

TravisCardwell (approved):
Oh, in an external binding specification, even a `newtype` may not have an `ffitype`.  Looking at the fixtures, I see that `newtype`s around `struct`s, `union`s, and constant arrays are examples.  Makes sense!

Looks good!

hs-bindgen/src-internal/HsBindgen/BindingSpec/Gen.hs
jorisdral: No worries -- it's also missing for structs, and the `cTypeSpecRep` fields are also missing, but I didn't want to touch anything that wasn't strictly necessary. We could open a separate PR for those?

hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/Common.hs
jorisdral: I think `BuiltinForeignType` might change in the future. `BasicForeignType` is determined by the Haskell 2010 language report, so that seems unlikely to change. I'll move `FFIType` out of the `Common` module

hs-bindgen/src-internal/HsBindgen/BindingSpec/Gen.hs
TravisCardwell: Oh, the `struct` case reminds me why those are `TODO`s.  We will "soon" change the implementation so that binding specifications are generated from `[C.Decl]` and `DeclIndex`.  Identifier refactoring came up, but I will get back to binding specs when that is done.

We do not need new issues, as this is already listed in #1356 and #1357.  These should be done *after* the above refactoring.

PR #1417, opened 2025-12-15
Fix the constructor import issue (1)

### LINKED ISSUE #1282: FFI Constructor Import Issue

jorisdral:
Resolves #1282

This PR is stacked on top of PR #1416, so that one should be merged first

### DIFF
--- a/.github/workflows/compile-fixtures.yml
+++ b/.github/workflows/compile-fixtures.yml
@@ -5,11 +5,7 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
   merge_group:
-    branches:
-      - main
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -5,11 +5,7 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
   merge_group:
-    branches:
-      - main
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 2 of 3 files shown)


manual/c/ffitype.h
jorisdral, line 1: Specify how these are split into binding specs and Haskell modules

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
```diff
@@ -1,56 +1,240 @@
...
   -> C.DeclName
   -> CallConv
   -> Origin.ForeignImport
   -> Maybe HsDoc.Comment
   -> Safety
   -> [Hs.Decl]
-foreignImportDecs name resultType parameters origName callConv origin comment safety =
-    [ Hs.DeclForeignImport foreignImportDecl ]
-    -- TODO: prevent the "newtype constructor not in scope" bug. See issue #1282.
+foreignImportDecs transState name resultType parameters origName callConv origin comment safety
```
jorisdral, line 48: What about structs, unions, etc. If an ffi type does not exist, generate a wrapper?

jorisdral:
Moving this back to draft because some investigation is needed

jorisdral:
This PR is rather out of date, so I decided to open a new one rather than updating this one. So, superseded by #1562

PR #1426, opened 2025-12-16
Generate Prim instance for structs, unions and enums

bolt12:


hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST.hs
```diff
@@ -374,6 +389,123 @@ data PokeCField ctx =
...
+--
+-- * Storable: @peekByteOff ptr byteOffset@
+-- * Prim: @indexByteArray# arr# (numFields# *# i# +# fieldPos#)@
+--
+-- <https://hackage.haskell.org/package/primitive/docs/Data-Primitive-Types.html#t:Prim>
+--
+type PrimInstance :: Star
+data PrimInstance = PrimInstance
+    { primSizeOf           :: Int#  -- ^ Total size in bytes
+    , primAlignment        :: Int#  -- ^ Alignment requirement
```
bolt12: No need to work with unboxed ints

hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST.hs
```diff
@@ -374,6 +389,123 @@ data PokeCField ctx =
...
+      (Idx ctx)   -- ^ Element index variable
+      Int#         -- ^ Field position (0-based)
+      Int         -- ^ Total number of fields
+  | -- | Index from Addr#: calls field type's indexOffAddr#
+    IndexOffAddrField
+      HsType      -- ^ Field type
+      (Idx ctx)   -- ^ Address variable
+      (Idx ctx)   -- ^ Element index variable
+      Int#         -- ^ Field position (0-based)
+      Int         -- ^ Total number of fields
```
bolt12: Better to split these types

hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST.hs
```diff
@@ -374,6 +389,123 @@ data PokeCField ctx =
...
+      [(HsType, Int)] -- ^ Fields: (type, position)
+      Int             -- ^ Total number of fields
+    -- | Read from Addr# with state threading
+  | ReadOffAddrFields
+      (Idx ctx)       -- ^ Address variable
+      (Idx ctx)       -- ^ Element index variable
+      (Idx ctx)       -- ^ State variable
+      [(HsType, Int)] -- ^ Fields: (type, position)
+      Int             -- ^ Total number of fields
+  deriving (Generic, Show)
```
bolt12: Better to split these types

hs-bindgen-runtime/src/HsBindgen/Runtime/SizedByteArray.hs
```diff
@@ -20,3 +26,72 @@ instance (GHC.KnownNat n, GHC.KnownNat m) => Storable (SizedByteArray n m) where
...
+            (# s3#, frozen# #) -> (# s3#, SizedByteArray (ByteArray frozen#) #)
+    where
+      size# = case fromIntegral (GHC.natVal (Proxy @n)) of I# s -> s
+      offset# = i# Exts.*# size#
+
+  writeOffAddr# addr# i# (SizedByteArray (ByteArray src#)) s# =
+    Exts.copyByteArrayToAddr# src# 0# (Exts.plusAddr# addr# offset#) size# s#
+    where
+      size# = case fromIntegral (GHC.natVal (Proxy @n)) of I# s -> s
+      offset# = i# Exts.*# size#
```
edsko, line 97: That was some tricky code! :-o

hs-bindgen-runtime/test/Test/HsBindgen/Runtime/SizedByteArray.hs
```diff
@@ -0,0 +1,225 @@
...
+    toBytes (PA.indexPrimArray arr 1) @?= [0x55, 0x66, 0x77, 0x88]
+    toBytes (PA.indexPrimArray arr 2) @?= [0xAA, 0xBB, 0xCC, 0xDD]
+
+prop_readByteArray_roundtrip :: [Word8] -> Property
+prop_readByteArray_roundtrip bytes = ioProperty $ do
+    let input = mkSizedByteArray @8 @8 (take 8 $ bytes ++ repeat 0)
+
+    -- Create a mutable array, write the value, read it back
+    marr <- BA.newByteArray 8
+    BA.writeByteArray marr 0 input
```
edsko: I'm happy you added tests for all this! Since you've gone through all that trouble anyway, and since these are already _properties_, not just unit tests, can we test with indices other than 0?

manual/hs/manual/manual.cabal
```diff
@@ -52,9 +52,11 @@ executable run-manual
   -- Internal.
   build-depends:
     , c-expr-runtime
+    , ghc-prim
     , hs-bindgen-runtime
     , hs-game
     , hs-vector
+    , primitive
```
edsko, line 59: It would be great if we could actually use the `Prim` instance for a struct in the manual code; one example would suffice; perhaps it would be nice if we would actually show a _use case_; for example, a `PrimArray` of structs. If we do that, it would also be nice if we could (briefly) show that in the actual manual text.

hs-bindgen/src-internal/HsBindgen/Backend/Extensions.hs
```diff
@@ -116,10 +117,19 @@ exprExtensions = \case
     EUnusedLam body -> exprExtensions body
     ECase x alts -> mconcat $
         exprExtensions x
-      : [ exprExtensions body
-        | SAlt _con _add _hints body <- alts
-        ]
+      :  [ exprExtensions body
+         | SAlt _con _add _hints body <- alts
```
edsko: Nit: rather than doing three separate pattern matches on `alta`, I guess it would be nicer to do a single list, and then with a `case` statement decide which extensions we need.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -284,6 +285,116 @@ structDecs opts haddockConfig info struct spec fields = do
                                   }
                           }
 
+        primDecl :: [Hs.Decl]
```
edsko, line 290: Let's move all `Prim` related stuff to a new module `Hs/Translation/Prim`. We're slowly but surely trying to reduce the massive size of this `Translation` module.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -284,6 +285,116 @@ structDecs opts haddockConfig info struct spec fields = do
...
+                                  , Hs.primReadByteArray = Hs.Lambda "arr"
+                                                         $ Hs.Lambda "i"
+                                                         $ Hs.Lambda "s"
+                                                         $ Hs.ReadByteArrayFields readFieldData
+                                  -- writeByteArray# :: MutableByteArray# s -> Int# -> a -> State# s -> State# s
+                                  , Hs.primWriteByteArray = Hs.Lambda "arr"
+                                                          $ Hs.Lambda "i"
+                                                          $ Hs.Lambda "struct"
+                                                          $ Hs.Lambda "s"
+                                                          $ Hs.makeElimStruct (IS IZ) hsStruct
```
edsko: Having thought about it again, I think this initial `s` parameter here, although _bit_ unfortunate, is acceptable: it emphasizes that `write` and `read`, unlike `index`, are in IO.

hs-bindgen/src-internal/HsBindgen/Backend/HsModule/Translation.hs
```diff
@@ -287,10 +287,17 @@ resolveExprImports = \case
     EUnusedLam body -> resolveExprImports body
     ECase x alts -> mconcat $
         resolveExprImports x
-      : [ resolveExprImports body
-        | SAlt _con _add _hints body <- alts
-        ]
+      :  [ resolveExprImports body
+         | SAlt _con _add _hints body <- alts
```
edsko: Similar nit here regarding the multiple matches.

hs-bindgen/src-internal/HsBindgen/Backend/SHs/AST.hs
```diff
@@ -294,7 +308,12 @@ pattern EInt i <- EIntegral (fromInteger -> i) (Just HsPrimInt)
 
 -- | Case alternatives
 data SAlt ctx where
-    SAlt :: Hs.Name Hs.NsConstr -> Add n ctx ctx' -> Vec n NameHint -> SExpr ctx' -> SAlt ctx
+    SAlt
```
edsko, line 320: These could do with some comments, and maybe a simple example. Especially `SAltNoConstr` confused me initially.

hs-bindgen/src-internal/HsBindgen/Backend/SHs/Translation.hs
```diff
@@ -287,6 +293,338 @@ translateStorableInstance struct Hs.StorableInstance{..} mbComment = do
       , instanceComment = mbComment
       }
 
+translatePrimInstance ::
```
edsko: Let's move all of this to `SHs/Translation/Prim` as well. I don't think we've started splitting the `Hs` to `SHs` translation just yet, but we might as well.

hs-bindgen/src-internal/HsBindgen/Backend/TH/Translation.hs
```diff
@@ -549,15 +576,40 @@ mkExpr env = \case
...
+                              -- guaranteed by the type.
+                              (xs, env') <- newNames env (AS AZ) hints
+                              case xs of
+                                []    -> panicPure "impossible happened"
+                                [v]   -> TH.match
+                                           (TH.varP v)
+                                           (TH.normalB $ mkExpr env' b)
+                                           []
+                                (_:_) -> panicPure "impossible happened"
+                         | SAltNoConstr hints b <- alts
```
edsko: ... and here too.

edsko (approved):
Nice work, that was _not_ an easy task! :scream: Left some minor review remarks, feel free to merge once they are addressed.

jorisdral:
Drive-by comment: if we implement test-suite generation, then we could test these generated `Prim` instances with something like https://hackage.haskell.org/package/quickcheck-classes-0.6.5.0/docs/Test-QuickCheck-Classes.html#v:primLaws

edsko:
Oh! Could we use that to test our instance for `SizedByteArray`?

jorisdral:
> Oh! Could we use that to test our instance for `SizedByteArray`?

Yes, I think we could -- we'd have to think a little about `Arbitrary` instances for `SizedByteArray`, but then you could just run the tests once that is set up

jorisdral:
FYI, I wrote some utility code for running these types of tests in `lsm-tree`: https://github.com/IntersectMBO/lsm-tree/blob/0fdb12ba5405ef1c413fb9aba70a129302d599dd/lsm-tree/test/Test/Util/QC.hs#L2

edsko:
> Yes, I think we could -- we'd have to think a little about `Arbitrary` instances for `SizedByteArray`, but then you could just run the tests once that is set up

I think for that it would suffice to just hand-pick a few select values for size (and possibly alignment), no need to get fancier than that I _think_.

bolt12:
I like that idea!

bolt12:
I'll move @jorisdral 's idea into a new issue, so that this PR can go in a little bit more easily!

edsko:
> I'll move @jorisdral 's idea into a new issue, so that this PR can go in a little bit more easily!

That's fine.

edsko (approved):

edsko:
Discussed this briefly, we decided to undo the change to `HsPrimType`, so that unboxed ints are _not_ considered primitive. This will limit the impact of this PR.

edsko:
@bolt12 You removed the wrong PR from the queue, I think (#1435). `HsPrimType` still needs changing back.

Issue #1439, opened 2025-12-18
"assignment of read-only location" errors when passing `const`-qualified structs/unions by pointer


jorisdral:
Some types like structs and unions can't be passed by value from Haskell to C and vice versa, so normally we'd pass around those types of values inside pointers. When the result of a function is a `const`-qualified struct or union, then we hit a compiler error. 

For example, if we have this C code:

```c
struct TyStruct { int x; };
const struct TyStruct foo (void);
```

Then currently we would generate this Haskell code:

```hs
$(HsBindgen.Runtime.Prelude.addCSource (HsBindgen.Runtime.Prelude.unlines
  [ "#include <binding_specifications.h>"
  , "void hs_bindgen_4c7688c596dcfdcd ("
  , "  struct TyStruct const *arg1"
  , ")"
  , "{"
  , "  *arg1 = foo();"
  , "}"
  ]))

foreign import ccall safe "hs_bindgen_4c7688c596dcfdcd" hs_bindgen_4c7688c596dcfdcd ::
     HsBindgen.Runtime.ConstPtr.ConstPtr TyStruct
  -> IO ()

foo_wrapper ::
     HsBindgen.Runtime.ConstPtr.ConstPtr TyStruct
  -> IO ()
foo_wrapper = hs_bindgen_4c7688c596dcfdcd

foo ::
     IO TyStruct
foo =
  HsBindgen.Runtime.CAPI.allocaAndPeekConst (\z0 ->
                                               hs_bindgen_4c7688c596dcfdcd z0)
```

And when we try to compile this we get this error:

```c
/tmp/ghc467809_0/ghc_14.c: In function ‘hs_bindgen_4c7688c596dcfdcd’:

/tmp/ghc467809_0/ghc_14.c:6:9: error:
     error: assignment of read-only location ‘*arg1’
        6 |   *arg1 = foo();
          |         ^
  |
6 |   *arg1 = foo();
  |         ^

<no location info>: error:
    `gcc' failed in phase `C Compiler'. (Exit code: 1)
```

`arg1` points to a constant struct, which can't be modified, so `*arg1 = foo()` will fail because it tries to overwrite the structs. `arg1 = &foo()` won't work either, because the struct is passed by value by (on the stack?) by `foo`, and so it has no address.

jorisdral:
[...]

PR #1440, opened 2025-12-18
Use `ConstPtr` for all function result types passed through pointers

jorisdral:
We had switched to using `ConstPtr` in the function arguments of bindings, but
we were not properly handling yet the case where result types (like structs and
unions) are communicated through a pointer. We add a new `allocaAndPeekConst`
function and we use this instead of `allocaAndPeek` whenever we are dealing with
a pointer to a const struct or union.


jorisdral:
I've opened #1439 for the bug I found in this PR

jorisdral:
After some careful thought, I do not think the fix in this PR is correct. I've opened issue #1490 instead.

PR #1443, opened 2025-12-19
Update `index-state`, add a mechanism for resetting caches

jorisdral:
It's good to update the `index-state` once in a while so that we don't fall too far behind. 

I've also added a mechanism for resetting caches, because cache sizes can grow over time and it's nice to start with a fresh cabal store every once in a while.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "nixpkgs-lib": "nixpkgs-lib"
       },
       "locked": {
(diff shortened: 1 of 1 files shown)


dschrempf:
We should merge this. May I?

dschrempf:
I rebased and also took the opportunity to update the Nix Flake lock file.

PR #1444, opened 2025-12-19
Add Invocation documentation to the manual

### LINKED ISSUE #944: Manual: invocation (TH, preprocessor)
There is already a skeleton in place (https://github.com/well-typed/hs-bindgen/blob/main/manual/LowLevel/Invocation.md), but this should describe how to use `hs-bindgen` in both TH and in preprocessor mode.

bolt12:
This PR sits on top of #1324 and closes #944

### DIFF
--- a/examples/literate-example/README.md
+++ b/examples/literate-example/README.md
@@ -66,7 +66,7 @@ literate-example/
 
 3. When `cabal build` is invoked:
    - Cabal detects the `.lhs` file
-   - Invokes `hs-bindgen-cli tool-support literate app/SimpleBindings.lhs <output>`
+   - Invokes `hs-bindgen-cli tool-support literate app/SimpleBindings.lhs app/SimpleBindings.hs -I c-lib --module=SimpleBindings --unique-id org.hs-bindgen.literate-example --enable-program-slicing simple.h`
    - The preprocessor reads the configuration from the file
    - The preprocessor generates bindings (equivalent to `hs-bindgen-cli preprocess`)
    - The preprocessor writes generated code to the output path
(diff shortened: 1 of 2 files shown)


examples/preprocessor-example/preprocessor-example.cabal
```diff
@@ -0,0 +1,30 @@
...
+author:        Edsko de Vries, Armando Santos
+maintainer:    edsko@well-typed.com
+build-type:    Simple
+
+common lang
+  ghc-options:      -Wall
+  default-language: GHC2021
+  build-depends:    base >=4.18 && <5
+
+executable preprocessor-example
```
edsko: This isn't a _preprocessor_ example but a _literate_ example right? Usually we use the term "preprocessor" to refer to the CLI. (I haven't read the rest of the PR, just scanned it out of curiosity).

edsko:
@dschrempf Could I ask you to review this PR, since you've been looking at the manual much more than me recently? @bolt12 Once Dominik approves feel free to merge.

examples/literate-example/README.md
```diff
@@ -0,0 +1,108 @@
...
+   ]
+   ```
+
+   These are the same arguments you would pass to `hs-bindgen-cli preprocess`.
+
+   **The file can contain anything** — it's just passed to the preprocessor, which reads it and generates Haskell code.
+
+3. When `cabal build` is invoked:
+   - Cabal detects the `.lhs` file
+   - Invokes `hs-bindgen-cli tool-support literate app/SimpleBindings.lhs <output>`
```
dschrempf: Before `<output>` it also adds all flags defined in the `.lhs` file. If we write out the full command, it should contain these flags.

PR #1450, opened 2025-12-22
Refactor C standard

TravisCardwell:
We decided to always get the C standard from `libclang`.  Users may specify a Clang standard via Clang arguments.

Clang standards specify *both* the C standard and use of GNU extensions. New type `ClangCStandard` is the product of `CStandard` and `Gnu`. While these types should generally be specified together, but it is convenient to case match them separately.  See the `libclang` commit for details.

This commit removes `cStandard` and `gnu` from `ClangArgsConfig`. `Boot` is refactored to get the C standard using `libclang`.  Various refactoring is done to work with these changes.

The `--standard` and `--gnu` options are removed from the CLI.

The C standard types are removed from the `TH` interface.  New module `HsBindgen.Clang.CStandard` re-exports `Clang.CStandard`, so we can make the module public or re-export types in the `TH` interface if desired, but it is no longer configurable.

The `c-expr-runtime` test is changed to use C17, since the Clang argument for it is not dependent on the LLVM/Clang version.

The `hs-bindgen` tests use `c23` or `c2x` depending on the LLVM/Clang version.  This will soon be refactored!

### DIFF
--- a/c-expr-runtime/test/Main.hs
+++ b/c-expr-runtime/test/Main.hs
@@ -11,8 +11,6 @@ module Main where
 -- base
 import Control.Arrow
   ( first )
-import Control.Exception
-  ( throwIO )
 import Data.List
   ( isPrefixOf )
 import Data.String
@@ -108,9 +106,8 @@ main = do
       , "The test-suite will use whichever C header files it finds on your system."
       ]
 
-  stdClangArg <- either throwIO return $
-    Clang.getStdClangArg Clang.C23 Clang.DisableGnu
-  let platform = hostPlatform
+  let stdClangArg = "-std=c17"  -- C23 arg depends on libclang version
+      platform = hostPlatform
       clangArgs = Clang.ClangArgs $ stdClangArg :
         case platformOS hostPlatform of
           Windows -> ["-target", "x86_64-pc-windows"]
--- a/cabal.project.base
+++ b/cabal.project.base
@@ -5,4 +5,4 @@ index-state: 2025-12-19T00:00:00Z
 source-repository-package
     type: git
     location: https://github.com/well-typed/libclang
-    tag: 363bc3593b18670b497c36f29d11ad49df9eea4d
+    tag: 27f20c2e320f779314175ee3e82072fec61b8391
--- a/examples/blocktest/generate-and-run.sh
+++ b/examples/blocktest/generate-and-run.sh
@@ -29,7 +29,7 @@ cabal run --project-dir="${PROJECT_ROOT}" -- hs-bindgen-cli \
     --hs-output-dir hs-project/generated \
     --unique-id blocksdemo.well-typed.com \
     --module Iterator \
-    --standard c23 \
+    --clang-option '-std=c2x' \
     --enable-blocks \
     iterator.h
 
--- a/examples/libpcap/generate.sh
+++ b/examples/libpcap/generate.sh
@@ -19,7 +19,7 @@ module_flags=(
 libclang_flags=(
     # Enable GNU extensions. GNU extensions define types, such as `u_int`, which
     # `libpcap` requires.
-    --gnu
+    --clang-option '-std=gnu17'
 )
 
 # `hs-bindgen` parses the C header files using `libclang` and interprets/reifies
--- a/flake.lock
+++ b/flake.lock
@@ -21,17 +21,17 @@
     "libclang-bindings-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1768172012,
-        "narHash": "sha256-HgrqgUdJN1A7bZF5oIchicMo9dKDDJmJImHynH5hZfM=",
+        "lastModified": 1768982090,
+        "narHash": "sha256-EgM9ZJvbM0+SQ8f9P27Rnx4dY8Q1fWtGVXsRl5f0Vfk=",
         "owner": "well-typed",
         "repo": "libclang",
-        "rev": "363bc3593b18670b497c36f29d11ad49df9eea4d",
+        "rev": "27f20c2e320f779314175ee3e82072fec61b8391",
         "type": "github"
       },
       "original": {
         "owner": "well-typed",
         "repo": "libclang",
-        "rev": "363bc3593b18670b497c36f29d11ad49df9eea4d",
+        "rev": "27f20c2e320f779314175ee3e82072fec61b8391",
         "type": "github"
       }
     },
--- a/flake.nix
+++ b/flake.nix
@@ -5,7 +5,7 @@
     flake-parts.url = "github:hercules-ci/flake-parts";
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
     libclang-bindings-src = {
-      url = "github:well-typed/libclang?rev=363bc3593b18670b497c36f29d11ad49df9eea4d";
+      url = "github:well-typed/libclang?rev=27f20c2e320f779314175ee3e82072fec61b8391";
       flake = false;
     };
   };
--- a/hs-bindgen/app/HsBindgen/App.hs
+++ b/hs-bindgen/app/HsBindgen/App.hs
@@ -33,7 +33,6 @@ module HsBindgen.App (
 import Data.Char qualified as Char
 import Data.Default (Default (..))
 import Data.Either (partitionEithers)
-import Data.List qualified as List
 import Data.Maybe (catMaybes)
 import Options.Applicative
 import Options.Applicative.Extra (helperWith)
@@ -247,8 +246,6 @@ parseClangArgsConfig = do
     -- ApplicativeDo to be able to reorder arguments for --help, and to use
     -- record construction (i.e., to avoid bool or string/path blindness)
     -- instead of positional one.
-    cStandard        <- parseCStandard
-    gnu              <- parseGnu
     enableBlocks     <- parseEnableBlocks
     builtinIncDir    <- parseBuiltinIncDirConfig
     extraIncludeDirs <- many parseIncludeDir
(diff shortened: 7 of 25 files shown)


TravisCardwell:
CI is testing with LLVM 15, and I needed to update some examples/tests to configure the version (using Clang arguments) correctly.

The most problematic is the `blocktest` example.  I think it requires C23 and does not work with C17, so the Clang argument depends on the version of LLVM being used.  I started writing some Bash, but then I realized that the test already hard-codes `c2x` in the `Makefile`, so I just used that.  I tested configuring `c2x` with LLVM/Clang 21.1.6, and it works fine and does not even show deprecation warnings.  (I am unable to test `blocktest` locally because I do not have the dependencies.)

TravisCardwell:
Failures on Windows!  I will investigate another day.

TravisCardwell:
I realized what is wrong.

When using a version of LLVM/Clang that targets MSVC by default and not explicitly specifying a C standard, LLVM is *not* in standards mode.  It does not use any of the standards; it uses "Microsoft C."

To support Microsoft C, I think we need to recognize it in the `libclang-bindings` and `hs-bindgen` implementations...

The current types are isomorphic to this:

```haskell
data CStandard      = C89 | C95 | C99 | C11 | C17 | C23
data Gnu            = DisableGnu | EnableGnu
data ClangCStandard = ClangCStandard CStandard Gnu
```

(This represents a superset of the standards supported by LLVM/Clang on Linux, as LLVM does not support `C95` with GNU extensions.  This is not a problem because we always detect the C standard, so a `C95` standard will always be coupled with `DisableGnu`.)

This does not support Microsoft C.  We could add Microsoft C support by changing to types like the following:

```haskell
data CStandard = C89 | C95 | C99 | C11 | C17 | C23
data Gnu       = DisableGnu | EnableGnu
data ClangC    = ClangCStandard CStandard Gnu | ClangCMicrosoft
```

If we want to do this, here are the relevant macros for detecting Microsoft C:

```c
#define _MSC_BUILD 1
#define _MSC_EXTENSIONS 1
#define _MSC_FULL_VER 195035717
#define _MSC_VER 1950
#define _MSVC_EXECUTION_CHARACTER_SET 65001
```

Thoughts?

dschrempf:
That is interesting. I like your suggestion of data types; it is reflecting the status-quo. Do we still parse any of these values (e.g., from the command line)? If so, we also need to change the data structures in these use sites.

TravisCardwell:
We currently do not parse any of these types.  Users may configure the C standard using Clang arguments (such as `--clang-option=-std=gnu23`), and `hs-bindgen` always gets the C standard from `libclang` (by interpreting builtin macros).

When I talked with @edsko yesterday, he said that he would not mind if we add the `--standard` option back, as a convenience only, if desired.  (It would be a convenience for specifying the `-std` Clang argument, and we would still get the C standard from `libclang`.)  This came up because we have an example that previously used `--standard c23` in a Bash script.  Having to specify `c23` or `c2x` depending on the version of LLVM/Clang using Bash would be annoying, but I found that specifying `c2x` works fine even in the latest LLVM/Clang release, with no deprecation warnings!  So, we are good for now.

I mention this because I think that if we do ever want to parse C standards, we should use the exact same strings that LLVM/Clang uses, such as `c23` and `gnu23`.  There is no such string for Microsoft C, however, so we would need to decide what to do about that.

Another possible future need for parsing would be adding the C standard to binding specifications, for the same reasons that `target` was added.  That is something that is pending discussion.

edsko:
I think perhaps we are trying to be too precise. The key point is that bindgen _mostly doesn't care_ what C standard is in use; it's only in a few places where we need to know (is `bool` a primitive or not is the only thing that comes to mind right now). Given that we now anyway do not _tell_ `clang` what version to use, but instead _ask_ it what version it uses, I think we should distinguish only what we need to distinguish.

edsko:
Perhaps we should regard the Clang version as something abstract/opaque; a module in `HsBindgen` that we can ask questions to ("should we support `bool`"?) but don't export any _value_ for.

TravisCardwell:
Feature predicates sound useful.  Perhaps they would be implemented based on the queried C standard information?

Microsoft C is not standard because they pick and choose which parts of the various standards they feel like supporting.  Perhaps some features depend on the Microsoft C version (`_MSC_VER`)...  In that case, perhaps the types should include it.

```haskell
data CStandard = C89 | C95 | C99 | C11 | C17 | C23
data Gnu       = DisableGnu | EnableGnu
data ClangC    = ClangCStandard CStandard Gnu | ClangCMicrosoft Int
```

edsko:
> Feature predicates sound useful. Perhaps they would be implemented based on the queried C standard information?

My suggestion was not necessarily a completely different approach (yes, we'd still figure to the version, as you are doing), but merely try to be less precise how we call that version; giving us just enough info to decide the things that _we_ need to decide. But if that's not in fact any easier, than never mind :)

edsko:
Note to self: this is not yet ready; we'll continue this work after the break. Depends on https://github.com/well-typed/libclang/pull/20, which has already been merged; _if_ the `libclang` main commit needs to be bumped for `hs-bindgen` in the meantime, might need to temporarily revert that.

TravisCardwell:
I reverted the `libclang` commit in https://github.com/well-typed/libclang/pull/24 so that #1499 can be merged.

It will be trivial to cherry-pick the commit and add a new one to distinguish Microsoft C when working on this again.

TravisCardwell:
This branch needed rebasing to work with the `OverloadedRecordDot` refactoring made over the holidays.  The target removal (#1555) done on Friday changed the Clang arguments code, which is an even more significant change.  I went ahead and rebased it this morning, before it becomes more difficult to work with.  Since this changes the configuration API (and CLI), it is probably good to get this in for the alpha anyway.

Support for Microsoft C is added.  It includes the Microsoft C version number, which is needed to implement feature predicates.

```haskell
data ClangCStandard =
    ClangCStandard CStandard Gnu
  | ClangCMicrosoft MicrosoftCVersion

data CStandard = C89 | C95 | C99 | C11 | C17 | C23

data Gnu = DisableGnu | EnableGnu

type MicrosoftCVersion = Integer
```

This branch currently uses a temporary commit to test.  If approved:

1. Merge https://github.com/well-typed/libclang/pull/25
2. Remove temporary commit
3. Update `libclang` dependency to new `main` commit

TravisCardwell:
It looks like GHC cannot target MSVC.  Ben talked about it as recently as last summer [here](https://discourse.haskell.org/t/ghc-9-14-1-alpha1-released/12786/16).

While it is possible for users to generate bindings using MSVC headers, perhaps it does not make sense because GHC will use `glibc`.  I do not think that the way we install LLVM using `KyleMayes/install-llvm-action` on Windows is valid.  We should always/only use the LLVM/Clang that is installed alongside GHC.  When building the manual or examples, we should only use that LLVM/Clang, *not* GCC.

Note that the version of LLVM/Clang is determined by the version of GHC being using.  It cannot be a variable in the test matrix.  I confirmed that GHC 9.14.1 uses LLVM/Clang 20.1.7, so the version is not constant across the versions of GHC that we support.

How should we implement this branch?  Perhaps we should *keep* the `ClangCMicrosoft` constructor in `libclang` so that we can accurately detect use of MSVC in `hs-bindgen`.  When we detect use of MSVC in `hs-bindgen`, perhaps we can have an error that explains that the wrong LLVM/Clang installation is being used (not just the wrong target).

Thoughts?

edsko:
Since this is a confusing topic, it's useful to be precise here:

First, "generating code for a target" means:

1. a choice of machine architecture (x86, ARM, ..); this will equal the host machine, unless we are cross-compiling
2. a choice of OS (file formats, etc.)
3. a choice of standard library implementation (a set of headers)

For some targets (like `msvc`) there is also a strange interaction between the choice of target and the C "standard" supported (quotes very much intentional).

Suppose we are generating bindings for some library `Lib`, which defines a struct `A`. There are three C compilers involved:

1. The C compiler used to compile `Lib` when it was installed on the system. 
2. The C compiler used by `hs-bindgen` (that is, `clang`) to query properties of the compiled `Lib`
3. The C compiler used by `ghc` to compile the little snippets of C code (wrappers etc.) that are generated by `hs-bindgen`

The most critical pairing here, from the perspective of `hs-bindgen`, is between compiler 1 and compiler 2.

In principle compiler 3 is not that important. No compatibility constraints exist here: no data structures are defined, and the functions are just simple functions. Of course basic ABI compatibility must be present (such as parameter passing: when we call `Lib`, the argument passing convention (register/stack/..) expected by it must match the code that ghc/the C compiler generates for these snippets); but I'd expect that that is fine always).

However, objects compiled targeting say  `x86_64-pc-windows-msvc` cannot be combined into the same executable with objects targeting say `x86_64-pc-windows-gnu`: you get linker errors. For one, multiple C standard libraries would result in duplicate symbols, and there may well be host of other problems too (even different file formats). 

[Currently GHC cannot target MSVC](https://discourse.haskell.org/t/ghc-9-14-1-alpha1-released/12786/16): that is, compiler 3 cannot be `clang` targeting `x86_64-pc-windows-msvc`. Until that is fixed, there is not much point in supporting `msvc` in `hs-bindgen` (compiler 2). This isn't really a problem with `hs-bindgen`, however, or even a problem with `hs-bindgen` generating incompatible code, or making the wrong decisions (that's between compiler 1 and compiler  instead). 

We therefore decided to "support" MSVC for now. All that means is that we can detect Microsoft C version numbers, and can do the relevant feature detection. Right now the only feature detection we do anyway is "is `bool` supported", which has an easy answer for MSVC: no. If in the future feature detection would become difficult for MSVC, then we could decide to simply not support it; not because we'd generate incorrect code ,but because it would anyway not be usable due to current limitations in GHC

PR #1451, opened 2025-12-22
Update libclang script

dschrempf:
The script
- Checks if the Cabal and Nix revisions of the libclang bindings are equal.
- If yes, it goes ahead and updates the revisions to the latest ones on the git repository (HEAD).
- If Nix is installed and Flakes are enabled, it updates the lock of `libclang-bindings-src`. If not, it warns the user that this should be done.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -31,6 +31,7 @@
       "original": {
         "owner": "well-typed",
         "repo": "libclang",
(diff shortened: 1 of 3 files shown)


dschrempf:
I will merge this since it only contains a script, and does not even change lock file targets.

PR #1454, opened 2025-12-23
Do not export UniqueSymbol constructor

### LINKED ISSUE #1398: Ensure that locally unique symbols are unique across a module
```haskell
data UniqueSymbol = UniqueSymbol{
      unique :: String
    , source :: String
    }
locallyUnique :: String -> UniqueSymbol
locallyUnique source = UniqueSymbol{
      unique = "hs_bindgen_" ++ hashString source
    , source
    }
```

`locallyUnique` is essentially hashing an underlying string (source or unhashed
value).

We should ensure uniqueness of `locallyUnique` across a module. This is
especially relevant when we combine multiple binding categories in Template
Haskell and literate modes.

dschrempf:
Force usage of `globallyUnique` or `locallyUnique`.

I have also audited that use-sites of these functions all have different arguments; that is, we do not generate euqal unique symbols.

Closes #1398.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/UniqueSymbol.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/UniqueSymbol.hs
@@ -3,7 +3,7 @@
 -- Intended for unqualified import.
 module HsBindgen.Backend.UniqueSymbol (
     -- * Generating unique names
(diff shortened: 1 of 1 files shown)


dschrempf:
Putting this on the queue since it is a minor change.

PR #1461, opened 2025-12-24
Refactor binding spec encode/decode API

TravisCardwell:
This PR does the following:

* Simplify the `readFile` API
    * Removes the functions that we do not use
* Refactor the `encode` API
    * Always encode to strict `ByteString`, to unify the API
    * Uses a `Format` argument, to make it easier to work with
* Remove the `writeFile` API
    * Was not exposed in the public API
* Fixed private API usage
    * Fixed in `Hs.Translation` by using re-exported types from public API
    * Fixed in `writeBindingSpec` by getting rendered file content
    * Removed binding spec usage from `DelayedIO`
* Ordered `ctypes` by source location in generated specs
    * Note: will refactor to create common functionality after the break

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclIndex.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclIndex.hs
@@ -76,7 +76,7 @@ data Usable =
       -- information required to match the select predicate also to external
       -- declarations.
     | UsableExternal
-    | UsableSquashed SourcePath Hs.Identifier
+    | UsableSquashed SingleLoc Hs.Identifier
     deriving stock (Show, Generic)
 
 -- | Unusable declaration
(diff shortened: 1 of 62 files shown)


edsko:
Opened https://github.com/well-typed/hs-bindgen/issues/1462 to track the refactoring.

edsko:
> Always encode to strict ByteString, to unify the API

Always using `lazy` would maybe be preferable? (There is no need for this to be a contiguous block of memory.) But these files are small enough that it doesn't really matter, so fine as-is.

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclIndex.hs
```diff
@@ -95,7 +95,7 @@ data Unusable =
       -- information required to match the select predicate also to omitted
       -- declarations.
       -- | Omitted by prescriptive binding specifications
-    | UnusableOmitted           SourcePath
+    | UnusableOmitted           SingleLoc
```
edsko, line 98: @dschrempf ^ omitted types now have location information.

PR #1466, opened 2025-12-25
Share `Type` between internal and external AST

edsko:


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ResolveBindingSpecs/IsPass.hs
```diff
@@ -64,6 +64,12 @@ data ResolvedExtBinding = ResolvedExtBinding{
     }
   deriving stock (Show, Eq, Ord, Generic)
 
+extDeclIdPair :: ResolvedExtBinding -> C.DeclIdPair
+extDeclIdPair ext = C.DeclIdPair{
+      cName  = ext.extCDeclId
+    , hsName = ext.extHsRef.extRefIdentifier
```
edsko, line 70: @TravisCardwell This is perhaps a bit iffy, extracting `extRefIdentifier` here and disgarding the module. It's not essential code, it's just to make some code more uniform (the introduction of `extBindingId` means that the pretty printer does not need to be pass specific), but it's still somewhat dubious. That said, I wonder if `Hs.Identifier` perhaps should _always_ have a module attached to it...? :thinking:

edsko: https://github.com/well-typed/hs-bindgen/issues/1467

PR #1469, opened 2025-12-25
Drop external AST

edsko:
Also moved `CheckedMacro` and co to `HandleMacros` (from `AST.Decl`), it's not really part of the syntax of C declarations.


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ResolveBindingSpecs/IsPass.hs
```diff
@@ -49,6 +48,23 @@ instance IsPass ResolveBindingSpecs where
...
+-- Although we have interpreted /part/ of this binding specification during
+-- name mangling, we leave the /full/ binding specification in the AST, because
+-- we need it when we  /generate/ the output binding specification.
+--
+-- TODO: This is not quite right: we should distinguish between binding
+-- specifications for different classes of things (declarations of types,
+-- functions, etc.). When we do, we should not associate them with the top-level
+-- 'Decl' but instead with specific 'DeclKind's. When we change this, this will
+-- have consequences for "HsBindgen.Language.Haskell.Origin" also.
+data PrescriptiveDeclSpec = PrescriptiveDeclSpec {
```
edsko, line 62: @TravisCardwell I hope this name makes sense; I _think_ it does, but would be good if you could confirm :)

TravisCardwell: This name sounds good to me.

PR #1494, opened 2026-01-07
Add underlying type to external binding spec references

### LINKED ISSUE #1363: We lose `FunPtr` information when binding specifications are involved

jorisdral:
Resolves #1363

For the example in #1363, we were previously generating:

```hs
bar :: Ptr A -> IO CInt
baz :: Ptr B -> IO CInt
```

We are now generating:

```hs
bar :: FunPtr A -> IO CInt
baz :: FunPtr B -> IO CInt
```

This is also visible in the `binding_spec/fun_arg/function_pointer` golden test that I've added. 

I've also added golden tests for other types of "special" function arguments: arrays, enums, structs, and unions. These types were exhibiting bugs of a similar nature to the function pointer bug. All these bugs were related to not being able to properly classify the types of function arguments without an underlying type.

I've also opened #1495: it would make it easier to write these types of tests in the future, and it would allow us to compile the fixtures

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
@@ -129,13 +129,15 @@ scanAllFunctionPointerTypes = foldMap $ \decl ->
       fp@(C.TypePointers _n (C.TypeFun args res)) ->
            Set.singleton fp
         <> foldMap scanTypeForFunctionPointers (res : args)
-      C.TypePointers _ t               -> scanTypeForFunctionPointers t
-      C.TypeIncompleteArray  t         -> scanTypeForFunctionPointers t
-      C.TypeConstArray _ t             -> scanTypeForFunctionPointers t
-      C.TypeBlock t                    -> scanTypeForFunctionPointers t
-      C.TypeQual _ t                   -> scanTypeForFunctionPointers t
-      C.TypeTypedef (C.TypedefRef _ t) -> scanTypeForFunctionPointers t
-      _                                -> Set.empty
+      C.TypePointers _ t        -> scanTypeForFunctionPointers t
+      C.TypeIncompleteArray  t  -> scanTypeForFunctionPointers t
+      C.TypeConstArray _ t      -> scanTypeForFunctionPointers t
+      C.TypeBlock t             -> scanTypeForFunctionPointers t
+      C.TypeQual _ t            -> scanTypeForFunctionPointers t
+      -- TODO: do we have to scane the underlying type too? If so, elaborate in
+      -- a comment why we do this.
+      C.TypeTypedef (C.Ref _ t) -> scanTypeForFunctionPointers t
+      _                         -> Set.empty
 
 -- | Check if a type is defined in the current module
 isDefinedInCurrentModule :: DeclIndex -> C.Type Final -> Bool
@@ -223,7 +225,7 @@ structDecs opts haddockConfig info struct spec fields = do
     pure decls
   where
     structName :: Hs.Name Hs.NsTypeConstr
-    structName = Hs.unsafeHsIdHsName info.id.hsName
+    structName = Hs.unsafeHsIdHsName info.id.unsafeHsName
 
     structFields :: Vec n Hs.Field
     structFields = flip Vec.map fields $ \field -> Hs.Field {
@@ -418,7 +420,7 @@ opaqueDecs haddockConfig info spec = do
     return [decl]
   where
     name :: Hs.Name Hs.NsTypeConstr
-    name = Hs.unsafeHsIdHsName info.id.hsName
+    name = Hs.unsafeHsIdHsName info.id.unsafeHsName
 
     decl :: Hs.Decl
     decl = Hs.DeclEmpty Hs.EmptyData {
@@ -451,7 +453,7 @@ unionDecs haddockConfig info union spec = do
           newtypeOrigin newtypeComment candidateInsts knownInsts
       where
         newtypeName :: Hs.Name Hs.NsTypeConstr
-        newtypeName = Hs.unsafeHsIdHsName info.id.hsName
+        newtypeName = Hs.unsafeHsIdHsName info.id.unsafeHsName
 
         newtypeConstr :: Hs.Name Hs.NsConstr
         newtypeConstr = union.names.constr
@@ -652,7 +654,7 @@ enumDecs opts haddockConfig info enum spec = do
           newtypeOrigin newtypeComment candidateInsts knownInsts
       where
         newtypeName :: Hs.Name Hs.NsTypeConstr
-        newtypeName = Hs.unsafeHsIdHsName info.id.hsName
+        newtypeName = Hs.unsafeHsIdHsName info.id.unsafeHsName
 
         newtypeConstr :: Hs.Name Hs.NsConstr
         newtypeConstr = enum.names.constr
@@ -808,7 +810,7 @@ typedefDecs opts haddockConfig info mkNewtypeOrigin typedef spec = do
           newtypeOrigin newtypeComment candidateInsts knownInsts
       where
         newtypeName :: Hs.Name Hs.NsTypeConstr
-        newtypeName = Hs.unsafeHsIdHsName info.id.hsName
+        newtypeName = Hs.unsafeHsIdHsName info.id.unsafeHsName
 
         newtypeConstr :: Hs.Name Hs.NsConstr
         newtypeConstr = typedef.names.constr
@@ -991,8 +993,10 @@ typedefFunPtrDecs opts haddockConfig origInfo n (args, res) origNames origSpec =
 
     -- TODO <https://github.com/well-typed/hs-bindgen/issues/1379>
     -- The name of this auxiliary type should be configurable.
-    auxHsName :: Hs.Identifier
-    auxHsName = origInfo.id.hsName <> "_Aux"
+    auxDeclIdPair :: DeclIdPair
+    auxDeclIdPair =
+        -- Still refer to the /original/ C decl...?
+        renameHsName (<> "_Aux") origInfo.id
 
     auxInfo :: C.DeclInfo Final
     auxInfo = C.DeclInfo {
@@ -1000,11 +1004,7 @@ typedefFunPtrDecs opts haddockConfig origInfo n (args, res) origNames origSpec =
         , headerInfo   = origInfo.headerInfo
         , comment      = Just auxComment
         , availability = C.Available
-        , id           = DeclIdPair{
-                             -- Still refer to the /original/ C decl...?
-                             cName  = origInfo.id.cName
-                           , hsName = auxHsName
-                           }
+        , id           = auxDeclIdPair
         }
 
     auxComment :: C.Comment Final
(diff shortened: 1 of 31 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -129,13 +129,14 @@ scanAllFunctionPointerTypes = foldMap $ \decl ->
...
-      C.TypeQual _ t                   -> scanTypeForFunctionPointers t
-      C.TypeTypedef (C.TypedefRef _ t) -> scanTypeForFunctionPointers t
-      _                                -> Set.empty
+      C.TypePointers _ t        -> scanTypeForFunctionPointers t
+      C.TypeIncompleteArray  t  -> scanTypeForFunctionPointers t
+      C.TypeConstArray _ t      -> scanTypeForFunctionPointers t
+      C.TypeBlock t             -> scanTypeForFunctionPointers t
+      C.TypeQual _ t            -> scanTypeForFunctionPointers t
+      -- TODO: remove ref case
+      C.TypeTypedef (C.Ref _ t) -> scanTypeForFunctionPointers t
```
jorisdral: Not sure if this case should be removed -- we should write a comment about why we're going into the underlying type regardless

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Type.hs
```diff
@@ -62,18 +62,20 @@ data TypeF tag p =
   | TypeComplex C.PrimType
 
     -- | Reference to named type other than a typedef
+    --
+    -- TODO: enum should get annotations with underlying types just like
+    -- typedefs, so that we can erase the enum types and replace them with
+    -- their underlying type.
```
jorisdral, line 68: We should open an issue for this. My thinking is that enums are just "newtypes", so we should give them the same treatment as typedefs

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Type.hs
```diff
@@ -62,18 +62,20 @@ data TypeF tag p =
...
+    --
+    -- TODO: macros should get annotations with underlying types just like
+    -- typedefs, so that we can erase the macro types and replace them with
+    -- their underlying type. See issue #1200.
   | TypeRef (Id p)
 
     -- | Reference to typedef
     --
-    -- TODO: TypedefRefF for erased should be void, and reuse 'TypeRef'.
-    --
```
jorisdral, line 70: The first part of the TODO was already done. I'm not sure what the intent of the second part is. Maybe we should elaborate in the comment?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames.hs
```diff
@@ -224,7 +223,7 @@ mangleDeclId declId = WrapM $ do
     nm <- asks (.nameMap)
     case Map.lookup declId nm of
       Just hsName -> return $ DeclIdPair declId hsName
-      Nothing     -> panicPure $ "Missing declaration: " <> show declId
+      Nothing     -> return $ DeclIdPair declId "do not use"
```
jorisdral: I'm not sure what to do in this case. Because we now apply the mangling on the underlying types of external binding references, we might hit declaration identifiers that we haven't pregenerated a fixed name for, and so we have to fill in *something*. Should we just generate a Haskell identifier that is precisely the C name? Or should we pregenerate the fixed names for these underlying types too?

jorisdral: @edsko @TravisCardwell thoughts?

edsko: I guess there is no easy answer here; we could introduce a lot of type-level machinery to make it clear that these underlying types may not have a Haskell name, but that seems a lot of work for little benefit.  Perhaps `return . panicPure` would be cleanest, then if we do accidentally use that Haskell name we'd at least panic at _that_ point; however, that would make the AST unprintable, which is also unpleasant. 

Maybe `DeclIdPair` should have `Maybe Hs.Identifier` instead, and then at use sites we use a function that panics if that is pure...?

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Type.hs
edsko: I think the second part was Edsko being confused.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames.hs
jorisdral: Good to note: I do not think we will ever use this Haskell name anywhere, but it is not statically enforced. 

One thing I pondered on was remove names from the underlying type AST altogether, but then we'll have two slightly different `Type` ASTs that we have to juggle...

edsko: Yeah, I'm realizing more and more that this is an exercise in deciding what is worth enforcing at the type level and what is it not.

jorisdral: > Maybe `DeclIdPair` should have `Maybe Hs.Identifier` instead, and then at use sites we use a function that panics if that is pure...?

Right, "defer" the panic to some pointer later in time where we have to decide whether we need the name or not. I think that could work

jorisdral: I've changed it to `Maybe Hs.Identifier` and introduced a panic whenever it is assumed to be a `Just` when it is a `Nothing`

jorisdral: And now it's an ADT with a callstack and a reason

hs-bindgen/src-internal/HsBindgen/Frontend/Naming.hs
```diff
@@ -87,14 +98,82 @@ instance PrettyForTrace DeclId where
...
+
+optNoIdentifier :: HasCallStack => Reason -> OptIdentifier
+optNoIdentifier = NoIdentifier callStack
+
+renameOptIdentifier :: (Hs.Identifier -> Hs.Identifier) -> OptIdentifier -> OptIdentifier
+renameOptIdentifier f = \case
+    NoIdentifier cstack reason -> NoIdentifier cstack reason
+    Identifier x -> Identifier (f x)
+
+-- |A Haskell identifier is not available.
```
jorisdral:
```suggestion
-- | A Haskell identifier is not available.
```

hs-bindgen/src-internal/HsBindgen/Frontend/Naming.hs
```diff
@@ -87,14 +98,82 @@ instance PrettyForTrace DeclId where
...
+      case hsName of
+        NoIdentifier cstack reason ->
+          panicPure $ concat
+            [ prettyReason reason, "\n"
+            , "Call stack:\n"
+            , prettyCallStack cstack
+            ]
+        Identifier x -> x
+
+data OptIdentifier = Identifier Hs.Identifier | NoIdentifier CallStack Reason
```
jorisdral: Possibly rename this to
```hs
data AssignedIdentifier = AssignedIdentifier Hs.Identifier | NoAssignedIdentifier CallStack Reason
```

jorisdral: And a comment about why it could be `NoAssignedIdentifier` -- and point to the `MangleNames` pass

edsko (approved):
Glad that some of the fog is lifting.

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Type.hs
jorisdral: Resolved in #1509

PR #1496, opened 2026-01-08
Fix alignment of types in TH

### LINKED ISSUE #1422: Fix alignment of types of foreign imports in TH fixtures
We noticed that the types of the foreign imports in the Template Haskell fixutres (`th.txt` files) are aligned in a weird way:

```haskell
foreign import ccall safe "hs_bindgen_1c811bfb80de8f77" hs_bindgen_1c811bfb80de8f77 :: CDouble ->
                                                                                       CDouble
```

I briefly tried to change that, but the functions are quite complicated. I think we use `TH.ppr_dec`, which is a library function? Armando knows his way around there.

We want the output to look like

```haskell
foreign import ccall safe "hs_bindgen_1c811bfb80de8f77" hs_bindgen_1c811bfb80de8f77 ::
     FC.CDouble
  -> FC.CDouble
```

bolt12:
This closes #1422

### DIFF
--- a/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden/Check/TH.hs
+++ b/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden/Check/TH.hs
@@ -213,7 +213,7 @@ prettyWithDocumentationMap isTop docMap dec =
             formatDecDoc docMap dec
       TH.$$ thCompatInstanceD docMap overlap cxt typ decs
 
-    -- The only reason foreign import pretty printer is overwritten here is 
+    -- The only reason foreign import pretty printer is overwritten here is
     -- for our benefit of making the type signatures easier to read. It is
     -- not what the default TH pretty printer would print.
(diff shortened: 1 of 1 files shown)


hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden/Check/TH.hs
```diff
@@ -213,6 +213,17 @@ prettyWithDocumentationMap isTop docMap dec =
...
+    TH.ForeignD (TH.ImportF callconv safety impent name typ) ->
+            formatDecDoc docMap dec
+      TH.$$ TH.hsep [ TH.text "foreign import"
+                    , pprCallconv callconv
+                    , pprSafety safety
+                    , TH.text (show impent)
+                    , TH.ppr name
+                    , TH.text "::"
+                    ]
+      TH.$$ TH.nest nestDepth (thCompatPprType typ)
```
bolt12, line 228: add a comment describing why we overwrite the default renderer

PR #1499, opened 2026-01-08
Fix output locations format to avoid mangling STDIO

### LINKED ISSUE #1270: Literate mode mangles `STDIO`
We have observed source locations strings for anonymous declarations with error test in them.  For example, we see `/usr/include/bits/types/struct_timespec.h:7:0: error: 1` instead of `/usr/include/bits/types/struct_timespec.h:7:1`.  This only happens in some situations, notably when running `cabal run test-pp`.

While I do not fully understand why this is happening yet, debugging indicates that the way that Cabal runs the literate preprocessor causes `STDIO` to be mangled.

bolt12:
Fix #1270 

I confirmed with @TravisCardwell reproducible that the issue is fixed! Thank you!

### DIFF
--- a/cabal.project.base
+++ b/cabal.project.base
@@ -5,4 +5,4 @@ index-state: 2025-12-19T00:00:00Z
 source-repository-package
     type: git
     location: https://github.com/well-typed/libclang
-    tag: b8305bd9c3e8e8f63d6f415553388be7f233ef45
+    tag: 363bc3593b18670b497c36f29d11ad49df9eea4d
--- a/flake.lock
+++ b/flake.lock
@@ -21,17 +21,17 @@
     "libclang-bindings-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1765293798,
-        "narHash": "sha256-fZvyZo5emD1W5LwrIg2zTUYE9StRh3xF8TcuXjzKIAk=",
+        "lastModified": 1768172012,
+        "narHash": "sha256-HgrqgUdJN1A7bZF5oIchicMo9dKDDJmJImHynH5hZfM=",
         "owner": "well-typed",
         "repo": "libclang",
-        "rev": "b8305bd9c3e8e8f63d6f415553388be7f233ef45",
+        "rev": "363bc3593b18670b497c36f29d11ad49df9eea4d",
         "type": "github"
       },
       "original": {
         "owner": "well-typed",
         "repo": "libclang",
-        "rev": "b8305bd9c3e8e8f63d6f415553388be7f233ef45",
+        "rev": "363bc3593b18670b497c36f29d11ad49df9eea4d",
         "type": "github"
       }
     },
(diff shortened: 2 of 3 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/HsModule/Render.hs
```diff
@@ -224,11 +224,9 @@ instance Pretty CommentKind where
...
 prettyHashIncludeArgLoc :: C.HeaderInfo -> SingleLoc -> CtxDoc
 prettyHashIncludeArgLoc info loc = PP.string $
-    List.intercalate ":" [
-        escapePaths info.includeArg.path
-      , show (singleLocLine loc)
-      , show (singleLocColumn loc)
-      ]
+    -- Use space instead of first colon to avoid GHC literate preprocessor mangling
+    escapePaths info.includeArg.path ++ " "
+      ++ show (singleLocLine loc) ++ ":" ++ show (singleLocColumn loc)
```
bolt12, line 229: This also needs a similar change in libclang in the function `prettySingleLoc`!

TravisCardwell:
This PR did not update the `source-repository-package` for the `libclang` repo, so it did not include the source location change made there and therefore still had the issue.

To test, I added the following to `test_01.h`:

```c
struct mangle_test {
  struct { long double ld; } foo;
};
```

The issue was reproduced when building (`cabal build all`):

```
[Warning] [HsBindgen] [select-parse-delayed] 'struct mangle_test'  at  "examples/test_01.h:415:0: error:
    8"
      Could not select declaration:
        Parse failure:
          No bindings generated: Unsupported long double
```

We cannot simply update the `source-repository-package`, however, because the `libclang` repo has a commit for #1450, which is not ready yet.  We attempted to always get the C standard being used from `libclang`, but this does not work on Windows when using MSVC targets because Microsoft does not follow standards.  We need to distinguish the non-standard Microsoft C case.  I reverted the problematic `libclang` commit so that we can update the `source-repository-package`.  That commit can be cherry-picked when we work on #1450, adding a new commit to distinguish Microsoft C.

With the `source-repository-package` updated, the issue is indeed resolved:

```
Template Haskell error: [Warning] [HsBindgen] [select-parse-delayed] 'struct mangle_test'  at  "./examples/test_01.h 415:8"
  Could not select declaration:
    Parse failure:
      No bindings generated: Unsupported long double
```

I see that this PR was not merged just because of an intermittent CI issue.  I will go ahead and merge now.

TravisCardwell:
CI for macOS is still failing.

```
 Warning: cannot determine version of /opt/homebrew/bin/git :
""
dyld[85653]: Symbol not found: _libiconv
  Referenced from: <E1ACFC9D-3501-35D5-A81E-9C76DD619092> /opt/homebrew/Cellar/git/2.52.0_1/bin/git
  Expected in:     <B27D3B05-B1FB-3728-B61A-FAA07DF2FCFD> /usr/lib/libiconv.2.dylib
/Users/runner/work/_temp/d1aedc74-e808-4fb3-b939-b2365cf056ae.sh: line 1: 85644 Abort trap: 6           cabal build all --dry-run
Error: Process completed with exit code 134.
```

Perhaps this was triggered by an update to the runner image and is not an intermittent issue.

PR #1500, opened 2026-01-08
Add category selection with single-module generation

### LINKED ISSUE #1449: Add command line options for binding category selection to literate mode
Since literate mode, like TH, is constrained to producing just a single file, it could be rather useful to have command line options to choose between safe, unsafe, or pointers, or indeed a combination thereof, though if a combination is desired, renaming must be done; probably a command line option for adding a prefix or a suffix would suffice.

bolt12:
This fixes #1449

### DIFF
--- a/examples/c-qrcode/hs-project/c-qrcode.cabal
+++ b/examples/c-qrcode/hs-project/c-qrcode.cabal
@@ -15,6 +15,7 @@ library
   import:             warnings
   exposed-modules:
     QRCodeGenerator.Generated
+    QRCodeGenerator.Generated.FunPtr
     QRCodeGenerator.Generated.Safe
     QRCodeGenerator.Generated.Unsafe
 
--- a/examples/botan/hs-project/botan.cabal
+++ b/examples/botan/hs-project/botan.cabal
@@ -32,6 +32,7 @@ library
   exposed-modules: Botan
   other-modules:
     Generated.Botan
+    Generated.Botan.FunPtr
     Generated.Botan.Safe
     Generated.Botan.Unsafe
 
--- a/examples/c-minisat/hs-project/c-minisat.cabal
+++ b/examples/c-minisat/hs-project/c-minisat.cabal
@@ -15,6 +15,7 @@ library
   import:           warnings
   exposed-modules:
     Generated.Minisat
+    Generated.Minisat.FunPtr
     Generated.Minisat.Global
     Generated.Minisat.Safe
     Generated.Minisat.Unsafe


bolt12:
I have only implemented single category option without the chance of combining. I feel like we need to discuss the renaming approach @edsko

hs-bindgen/app/HsBindgen/App.hs
```diff
@@ -174,6 +217,35 @@ parseConfig = Config
...
+parseCategoryOptions = CategoryOptions <$> many parseCategory
+  where
+    parseCategory :: Parser CategorySelection
+    parseCategory = option readCategory $ mconcat [
+          long "category"
+        , metavar "CHOICE"
+        , help $ concat [
+              "Binding category to generate (safe|unsafe|pointer). "
+            , "Generates a single module file combining the selected category with types and globals. "
+            , "Can be specified multiple times for combinations (future). "
```
dschrempf: We could alread support multiple combinations by using a "category + suffix" approach (where the suffix can be the empty string at the risk of the user).

hs-bindgen/test/pp/Test/PP/TestSafe.lhs
dschrempf, line 1: Instead of naming these TestN, we could rename them to `TestSafe.lhs` etc?

hs-bindgen/src-internal/HsBindgen.hs
```diff
@@ -157,6 +158,32 @@ writeBindings fileOverwritePolicy path = do
     bindings <- getBindings
     write fileOverwritePolicy "bindings" (UserSpecified path) bindings
 
+-- | Write bindings to a directory (single module combining all categories).
+--
+-- Unlike 'writeBindings', this writes to a directory and automatically
+-- constructs the file path from the module name, similar to
+-- 'writeBindingsMultiple' but generating only one file.
+writeBindingsSingleToDir ::
```
dschrempf, line 167: I wonder why this function is necessary. Should we maybe replace `writeBindings`? Is `writeBindings` used after your changes?

hs-bindgen/app/HsBindgen/Cli/Preprocess.hs
```diff
@@ -64,30 +66,49 @@ parseOpts =
...
+    --
+    -- When no category is specified (default), use multi-module mode for
+    -- backwards compatibility.
     artefact :: Artefact ()
     artefact = do
-        writeBindingsMultiple
-          opts.fileOverwritePolicy
-          opts.outputDirPolicy
-          opts.hsOutputDir
+        if usesSingleModule opts.categoryOptions
```
dschrempf: I have the feeling that we should we move this logic into the artefact, and maybe have one entry function
```haskell
     FileOverwritePolicy
  -> OutputDirPolicy
  -> FilePath
  -> CategoryOptions(MaybeIncludingFilePath?)
  -> Artefact ()
```

hs-bindgen/app/HsBindgen/Cli/Preprocess.hs
```diff
@@ -64,30 +66,49 @@ parseOpts =
 -------------------------------------------------------------------------------}
 
 exec :: GlobalOpts -> Opts -> IO ()
-exec global opts =
+exec global opts = do
+    -- Build category choice from options
+    categoryChoice <- case buildCategoryChoice opts.categoryOptions of
```
dschrempf: The only reason why this throws an error is because we do not support something yet? If so, I suggest supporting it right away (using the category+suffix), or having a type that doesn't allow the error (e.g., Maybe Category).

dschrempf: (Or is this because we are "parsing" the category strings?)

dschrempf (commented):
The task this PR is solving popped up on my board so I had a look! It will be great when we can choose categories on the command line, I am looking forward to it!

hs-bindgen/app/HsBindgen/App.hs
bolt12: I see, I wasn't aware of this! Great!

hs-bindgen/test/pp/Test/PP/TestSafe.lhs
bolt12: Yeah!

hs-bindgen/app/HsBindgen/Cli/Preprocess.hs
bolt12: Yeah, that's a good suggestion!

hs-bindgen/app/HsBindgen/Cli/Preprocess.hs
bolt12: You are right, is just because we might either combine or get nothing at all

hs-bindgen/app/HsBindgen/App.hs
```diff
@@ -50,6 +61,63 @@ import HsBindgen.Frontend.RootHeader (UncheckedHashIncludeArg)
...
+        , cGlobal = IncludeTermCategory def
+        }
+
+    -- Build a choice that includes multiple categories with renaming
+    buildCombinedChoice :: [CategorySelection] -> ByCategory Choice
+    buildCombinedChoice selections = ByCategory {
+          cType   = IncludeTypeCategory
+        , cSafe   = includeWithSuffix Safe   "_safe"
+        , cUnsafe = includeWithSuffix Unsafe "_unsafe"
+        , cFunPtr = includeWithSuffix Pointer "_pointer"
```
bolt12: Make the user choose which suffix it wants

hs-bindgen/app/HsBindgen/App.hs
```diff
@@ -174,6 +242,35 @@ parseConfig = Config
...
+    parseCategory = option readCategory $ mconcat [
+          long "category"
+        , metavar "CHOICE"
+        , help $ concat [
+              "Binding category to generate (safe|unsafe|pointer). "
+            , "Generates a single module file combining the selected categories with types and globals. "
+            , "Can be specified multiple times to combine categories (functions suffixed with category name). "
+            , "Default: safe (literate mode) or all categories in separate modules (preprocessor mode)"
+            ]
+        ]
```
bolt12: Change the way we do this:

either specify:
`--file-per-category` - should be enabled by default

or specify:
`--single-file` which then enables the user to further specify:

`--safe <suffix>`     - should only generate one file with safe bindings and add the specified suffix
`--unsafe <suffix>`   - should only generate one file with unsafe bindings and add the specified suffix
`--pointer <suffix>`  - should only generate one file with pointer bindings and add the specified suffix
`--global`
`--types`

examples/c-qrcode/hs-project/c-qrcode.cabal
```diff
@@ -15,7 +15,6 @@ library
   import:             warnings
   exposed-modules:
     QRCodeGenerator.Generated
-    QRCodeGenerator.Generated.FunPtr
```
bolt12, line 18: double check if this is needed

hs-bindgen/app/HsBindgen/App/Output.hs
```diff
@@ -0,0 +1,147 @@
...
+parseOutputMode defMode =
+  asum [
+    (flag' FilePerModule $ mconcat [
+        long "file-per-module"
+      , help "Generate one file per binding category (default)"
+      ])
+  , (flag' SingleFile $ mconcat [
+        long "single-file"
+      , help "Generate a single module file"
+      ]) <*> parseSingleFileCategory
```
bolt12: We should be able to combine the categories

hs-bindgen/app/HsBindgen/Cli/Preprocess.hs
```diff
@@ -64,7 +68,15 @@ parseOpts =
...
 exec :: GlobalOpts -> Opts -> IO ()
-exec global opts =
+exec global opts = do
+    let bindgenConfig :: BindgenConfig
+        bindgenConfig =
+            toBindgenConfig
+              opts.config
+              opts.uniqueId
+              opts.baseModuleName
+              (buildCategoryChoice opts.outputOptions)
```
bolt12: keep it in the `where`

hs-bindgen/test/pp/Test/PP/TestPointer.lhs
bolt12, line 1: Add a test that uses more than one category

hs-bindgen/app/HsBindgen/Cli/Preprocess.hs
```diff
@@ -74,20 +86,13 @@ exec global opts =
   where
     artefact :: Artefact ()
     artefact = do
-        writeBindingsMultiple
+        writeBindingsToDir
```
bolt12: Move the if then else to here.

hs-bindgen/src-internal/HsBindgen/Backend/Category.hs
```diff
@@ -182,5 +183,15 @@ useUnsafeCategory = ByCategory {
...
 
+-- | Use 'CType', 'CFunPtr', and 'CGlobal'; do not rename declarations.
+useFunPtrCategory :: ByCategory Choice
+useFunPtrCategory = ByCategory {
+      cType   = IncludeTypeCategory
+    , cSafe   = ExcludeCategory
+    , cUnsafe = ExcludeCategory
+    , cFunPtr = IncludeTermCategory def
+    , cGlobal = IncludeTermCategory def
+    }
```
bolt12, line 210: not needed after all!

PR #1505, opened 2026-01-12
Refactor Aeson representation data family

TravisCardwell:
The terminology was not accurate.  We implement a translation between a type and its Aeson representation, but all the types and Aeson representations are for the binding specification version defined in the module (`V1`).  When a version depends on a previous version, the previous version will be imported qualified.  Within a module, there should be no reference to the version (aside from the module name), and all references to previous versions should be via qualifications.

The data family is therefore renamed to `ARep`, short for "Aeson representation."

Type class `PureV1Spec` is used when there is an isomorphism between a type and its Aeson representation, which is orthogonal to purity.  It is renamed to `ARepIso`.  Type class `ARepKV` is added to handle cases where the Aeson representation encodes a key, where the type of the key is defined in an associated data family (`ARepK`).

Such type classes should be used in all version modules, so they are now defined in the `Common` module and are parameterized by the data family. In some cases, type applications are necessary (`fromARep @ARep`), but this is rare.  If a future version has too many of these, helpers may be used (such as `fromARep' = fromARep @ARep`).

All of the auxiliary types are now defined as instances of a data family, either `ARep` or `ARepK`.  We still have separate functions to convert between a type and its Aeson representation when extra context is required (such as the path for use in errors) or returned (such as lists of errors).  These functions are moved to the sections of the relevant types.

The Aeson representation data family, type aliases, and helper functions are *not* exported yet.  When we implement `V2`, we can add the needed exports.  I expect that it would be convenient to export all of them; it is a private module after all.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/Common.hs
+++ b/hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/Common.hs
@@ -14,6 +14,10 @@ module HsBindgen.BindingSpec.Private.Common (
   , BindingSpecResolveMsg(..)
   , BindingSpecMergeMsg(..)
   , BindingSpecMsg(..)
+    -- * Aeson representation
+    -- $ARep
+  , ARepIso(..)
+  , ARepKV(..)
     -- * Omittable
   , Omittable(..)
   , AOmittable(..)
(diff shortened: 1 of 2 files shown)


TravisCardwell:
On Friday, @edsko told me to go ahead and merge this when ready.  I will try, but I expect to hit the macOS issue that we are seeing in #1499.

PR #1506, opened 2026-01-12
Work around `macos-latest` GitHub runner issues

dschrempf:
Work in progress.

In particular, the LLVM install messes around with include directories, and the then `git` can not find `libiconv`.

This PR does not fix the issue, instead works around by bumping the LLVM version used so the libraries align again.


dschrempf:
Fixed by #1507

Issue #1424, opened 2025-12-16
Support anonymous enums


edsko:
Given

```c
enum {
  A,
  B
};
```

we should generate constants `A` and `B` in the enums underlying type (rather than generate a new type).

edsko:
We should probably only do this once the identifier refactoring is complete.

bolt12:
This means that we should not create the newtype for the top level enum

jorisdral:
> This means that we should not create the newtype for the top level enum

If you don't generate the newtype but generate constants in the underlying type, then you'd have to go through all types to replace all references to the enum by the underlying type as well

edsko:
There cannot be any references to the enum, as its anonymous :)

jorisdral:
> There cannot be any references to the enum, as its anonymous :)

I was about to correct myself but you were first :P

bolt12:
Same 😄

edsko:
Some people are just a bit faster than others 🐰 🐢

PR #1509, opened 2026-01-12
Add underlying types to enum references

jorisdral:


edsko:
@jorisdral @bolt12 There might be some overlap between this and #1424.

jorisdral:
~> @jorisdral @bolt12 There might be some overlap between this and #1424.~

~Right, that's true. This PR is about modifying "refences to enum types" to include the underlying type. "Enum declarations" already have that information. If there is a need to replace enum references by the underlying type (see https://github.com/well-typed/hs-bindgen/issues/1424#issuecomment-3743090291), then maybe this PR is useful for you @bolt12 ?~

EDIT: never mind, anonymous enums can not be referred to. In that case, then I think this PR probably does not affect #1424, but I might be overlooking something

jorisdral:
Okay, I've actually found a use for the changes in this code! So I'm going to get this PR ready for review

EDIT: so the use I've found for this is that I want to erase enum references to fix the constructor import issue in #1282. C enums are translated to Haskell newtypes, which can exhibit the constructor import issue if we do not replace the enum by its "ffi type" in foreign imports

hs-bindgen/src-internal/HsBindgen/Frontend/LanguageC/PartialAST/FromLanC.hs
```diff
@@ -181,7 +181,10 @@ instance Apply (LanC.CTypeSpecifier a) PartialType where
       LanC.CEnumType (LanC.CEnum mTag mDef _attrs _a) _a' -> \partial -> do
         name <- checkNotAnon mTag C.TagKindEnum
         checkNoDef "enum definition" mDef
-        notFun (typeRef name) $ partial
+        typeEnv <- getReparseEnv
+        case Map.lookup name.text typeEnv of
+          Nothing -> notFun (typeRef name) $ partial
+          Just typ -> notFun typ partial
```
jorisdral, line 187: Unsure about this code, because I do not fully understand how the reparse algorithm works. Need to look into this more

jorisdral: FYI: I shouldn't use `typeRef` because it creates a `C.TypeRef` type, when we should be creating a `C.TypeEnum`

jorisdral: Okay I think I figured it out -- I indeed do a lookup here, but i have to add the enum reference types to the `ReparseEnv` as well

jorisdral:
Now ready for review. Note that the change is not really visible in the fixtures now, but it will be when solving #1282

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Type.hs
```diff
@@ -312,6 +340,7 @@ instance IsPass p => ValidTypeTag Canonical p where
   type instance TypeQualifierF     Canonical p = Void
   type instance TypeExtBindingRefF Canonical p = Void
   type instance TypeMacroRefF      Canonical p = Void
+  type instance TypeEnumRefF       Canonical p = EnumRef p
```
jorisdral:
```suggestion
  type instance TypeEnumRefF       Canonical p = Void
```

PR #1512, opened 2026-01-13
Add underlying types to macro type references

jorisdral:


jorisdral:
> Let's rename the `ref` field of `Ref` to `name`, to avoid awkward `ref.ref.foo`.

Will do

Issue #1533, opened 2026-01-14
Handle name mangler failures in `Select` (by using the `DeclIndex`)


dschrempf:
This is a placeholder for the second part of https://github.com/well-typed/hs-bindgen/issues/1373

- The first part will be that we only warn about collisions in `MangleNames` (and leave the Errors for other failures in place).
- This, second part, is about handling name mangler failures in the `Select` pass, by recording them in the `DeclIndex`. This also requires that we switch the order of the `MangleNames` and `Select` passes.

dschrempf:
Assigning same labels as https://github.com/well-typed/hs-bindgen/issues/1373

Issue #1373, opened 2025-12-03
Bug: not handling declaration name collisions


TravisCardwell:
Working on something else, a fixture change for [`hs-bindgen/examples/golden/declarations/redeclaration.h`](https://github.com/well-typed/hs-bindgen/blob/fd83626704209e73a9c3b424181d81e5c0178e91/hs-bindgen/examples/golden/declarations/redeclaration.h) highlighted an issue.  The relevant C declarations are the following:

```c
union y;
union Y { int m; int o; };
union Y;
```

C declarations `union y` and `union Y` are distinct types in C that are both translated to Haskell type `Y`.  `union y` is opaque, while `union Y` is not.

The [`TestCase`](https://github.com/well-typed/hs-bindgen/blob/fd83626704209e73a9c3b424181d81e5c0178e91/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden.hs#L462-L472) handles some traces for other declarations, but not `union y` or `union Y`.

Running `preprocess` confirms this:

```
$ mkdir -p ex-src
$ cabal run hs-bindgen-cli -- preprocess \
    --unique-id ex \
    --hs-output-dir ex-src \
    -I hs-bindgen/examples/golden/declarations \
    redeclaration.h
[Notice ] [HsBindgen] [select-parse-delayed] During parse:
  'x' at "hs-bindgen/examples/golden/declarations/redeclaration.h:11:5":
    Bindings may result in duplicate symbols; consider using 'static' or 'extern'; or if that is not an option, consider attributing hidden visibility to the symbol
[Warning] [HsBindgen] [select-parse-delayed] Could not select declaration:
  Parse failure:
    'n' at "hs-bindgen/examples/golden/declarations/redeclaration.h:15:12":
      No bindings generated: unsupported storage class simpleEnum CX_SC_Static
```

Running `frontend` confirms that there are two declarations that have Haskell name `Y`:

```
$ cabal run hs-bindgen-cli -- internal frontend \
    --unique-id ex \
    -I hs-bindgen/examples/golden/declarations \
    redeclaration.h \
  | friendly
...
  ,Decl {
      declInfo = DeclInfo {
        declLoc = "hs-bindgen/examples/golden/declarations/redeclaration.h:29:7"
      , declId = DeclIdNamed (NamedDeclId {name = "y", origin = NameOriginInSource, haskellId = "Y"})
      , declAliases = []
      , declHeaderInfo = Just (
          HeaderInfo {
[...]

PR #1542, opened 2026-01-14
Detect and warn about Haskell name collisions

dschrempf:
Part 1 of https://github.com/well-typed/hs-bindgen/issues/1533

Related: https://github.com/well-typed/hs-bindgen/issues/1373


dschrempf:
@edsko I am inclined to put this on the queue without another review, because we have discussed quite a lot. Let me know what you think!

PR #1550, opened 2026-01-15
Clarify usage and requirements of `SizedByteArray`

### LINKED ISSUE #1442: (Re)consider alignment requires for `SizedByteArray`
In particular, even _if_ the start of the `ByteArray` would be aligned, that would still not be the case for values at indices other than `0` (in the `Prim` instances); if this is really important, we should add some invariants regarding the size also (padding would need to be added so that `offset + i * size` would still be properly aligned).

edsko:
Closes #1442.

### DIFF
--- a/hs-bindgen-runtime/hs-bindgen-runtime.cabal
+++ b/hs-bindgen-runtime/hs-bindgen-runtime.cabal
@@ -81,7 +81,7 @@ library
 
   build-tool-depends: hsc2hs:hsc2hs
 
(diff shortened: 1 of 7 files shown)


edsko:
@joris Could I ask for your review here? This is a response to #1442 . The only thing to review is the new documentation I added to `SizedByteArray`, to see if that makes sense to you.

jorisdral (approved):

PR #1551, opened 2026-01-15
Handle name mangler failures

dschrempf:
I think this is ready for review.

This changes the order of the MangleNames and Select passes!
- Add support for MangleNamesFailures to DeclIndex and register MangleNames
  failures in DeclIndex
- Remove unused MangleNames trace (MangleNamesMissingIdentifier)
- Improve representation of squashed declarations in DeclIndex

Minor fixes:
- Improve suggestion in trace message ("enable program slicing")

### DIFF
--- a/hs-bindgen/examples/golden/declarations/select_scoping.h
+++ b/hs-bindgen/examples/golden/declarations/select_scoping.h
@@ -21,8 +21,7 @@ struct ParsedUnselectable {
 
 // This declaration is parsed _and_ selected. The dependency is parsed but
 // unselectable because another transitivie dependency is unusable.
-// `ParsedAndSelected4` is squashed, and so we do not get direct trace messages
-// in the select pass (but a squash notice).
+// `ParsedAndSelected4` is squashed.
 typedef struct ParsedUnselectable ParsedAndSelected4;
 
 // This declaration is parsed _and_ selected. However, a transitive dependency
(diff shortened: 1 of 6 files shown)


edsko (changes requested):
Let's change selection so that the selection predicate is applied to the squash source, rather than the squash target. Detailed discussion to follow soon, will post a link.

dschrempf:
I have revamped the select pass, generating select messages from the declaration index
- This has the advantage that we now also get messages for selection roots that are not in the list of declaration attached to the unit.
- It has the disadvantage that we have to test for selection and transitive selectability twice (declaration index and list of declarations). We could cache this computation.

Further, we now apply the selection predicate to the squash source. Before this change, the location was obtained from the squash target only. I have changed this behaior. We always matched against the declaration name of the squash source (not the target, as I incorrectly claimed in private communications, and to which Edsko refers to above).

However, the squash target also has to be selected (and here we do not match against the squash source).

For example, if we have
```c
typedef struct foo {...} bar;
```
we need to select `foo` and `bar`. (By default, this is not be a problem, because we match all declarations in the main header; but if a user selected `bar` only, the example above would not be translated, I believe. We would emit appropriate trace messages though.)

dschrempf:
(I am putting this on the queue since we discussed this about 2 hours ago).

Issue #1547, opened 2026-01-15
Remove `Target`


edsko:
The `Target` triple in `hs-bindgen` was introduced for two reasons, which we discuss in turn.

## The target triple could direct generation

We thought we needed to make different decisions based on the architecture; we have since moved away from this entirely, and just depend on what `libclang` tells us. Nothing in `hs-bindgen` makes any decisions based on the choice of `Target` anymore.

If we _do_ need to make decisions based on the platform (for example, depending on [endianness](https://github.com/well-typed/hs-bindgen/issues/1546)), it would be better to query clang, either through its API or through the builtin macros.

## We record the target triple in binding specs

The sole reason for this is to verify that a given set of bindings, as described by an external binding specification, is suitable for a choice of target platform.

This seems plausible at first, but it actually doesn't really solve the problem very well at all:

* The target triple lists the choice of C library implementation (glibc, musl), but actually that does not influence the bindings in a meaningful way at all (our manually written bindings for standard types are platform independent).
* Conversely, if library B depends on library A, we have no way of knowing at all what kind of CPP library A uses to make different decisions; the target triple listed in the binding spec for library A would be insufficient to know whether those bindings are suitable for a given target platform.

As we have said before: **hs-bindgen should be considered to be part of the build process**. Generated bindings really are only valid for a given choice of platform _and all dependencies_. Users may choose of course to distribute generated bindings, but then _they_ must know what exactly those bindings depend on, and act accordingly. This is no different really from distributing any other kind of build artefacts (like binaries).

PR #1555, opened 2026-01-15
Remove target

TravisCardwell:
See #1547 for details.


TravisCardwell:
This is not tested on Windows or macOS until it enters the merge queue.  I do not expect problems, however.

TravisCardwell:
While discussing yesterday, I did a quick test where I changes the `hs-bindgen` tests to use `x86_64-pc-linux-gnu` (instead of `x86_64-pc-linux-musl`) and ran the tests to see what failures occur.  With the target removed, I tried this again.

Changes:

* I changed the target from `x86_64-pc-linux-musl` to `x86_64-pc-linux-gnu`.
* I changed the extra include directories from the vendored `musl` include directory to the compiler builtin directory for the version of LLVM/Clang that I am testing with.
* I changed the `testGetTargetTriple` test to check for the GNU triple.
* I disabled the `test_programAnalysis_program_slicing_simple` test, because the `glibc` implements things differently than `musl`.

Differences:

* The TH dependent files of course differ.
* Function `erf` (defined in `functions/simple_func.h`) is pure when using `musl` but is in `IO` when using `glibc`.

With those fixtures accepted, all tests pass.

(This is just to check for differences; I am *not* committing these changes.)

PR #1556, opened 2026-01-16
Refactor to use a common `ARep` data family

TravisCardwell:
This PR refactors the binding specification code to define an `ARep` data family in `Common`.  It is parameterized by a version, so that all versions can use it.

The version modules are designed to be imported qualified, and qualifications are used to specify versions.  I therefore implemented this using an open kind `ARepV`, implemented exactly like we implement `Pass`.  Each version module should define Aeson representation version `V`.  Helper functions `fromARep'` and `toARep'` are specialized to `V` (the current version), to minimize syntactical noise.  Other versions should be referenced via qualification; for example the `V2` module can use `V1.V` to refer to the `V1` Aeson representation version.

(The current version, such as `V1` in the `V1` module, should only appear in the module name.  Specifying it in types, functions, etc. is a code smell, IMHO.)

This design does *not* use quantified instances.  `V1` is the base, and subsequent versions will build on top of it.  (If we completely break backwards compatibility at some point, that new version will be a new base.)

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/Common.hs
+++ b/hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/Common.hs
@@ -15,7 +15,8 @@ module HsBindgen.BindingSpec.Private.Common (
   , BindingSpecMergeMsg(..)
   , BindingSpecMsg(..)
     -- * Aeson representation
-    -- $ARep
+  , ARepV
+  , ARep
   , ARepIso(..)
   , ARepKV(..)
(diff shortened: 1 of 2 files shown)


edsko (approved):
Let's not call all of these versions `V`. Design philosophy aside, this results in unpleasant haddocks; something like

```hs
module V2 where

upgradeVersionFoo :: ARep V1.V Foo -> ARep V Foo
```

would render as `ARep V Foo -> ARep V Foo`, which is just confusing.

PR #1558, opened 2026-01-16
CI: Add macOS and Windows jobs to PR test matrix

### LINKED ISSUE #1557: Re-enable one Windows and one OSX test for PRs
Currently we have to wait until merge to see if there are any problems on those platforms, which is sometimes a bit annoying.

bolt12:
Previously, macOS and Windows CI tests only ran on push to main, making it difficult to catch platform-specific issues early during PR review. This commit adds one macOS and one Windows job to the PR test matrix.

Fixes #1557

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -33,6 +33,8 @@ jobs:
       run: |
         { echo 'MATRIX_COMBINATIONS<<EOF'
           echo '{"runner": "ubuntu-latest", "ghc": "9.4", "llvm": "15"},'
+          echo '{"runner": "macos-latest"  , "ghc": "9.4", "llvm": "14"},'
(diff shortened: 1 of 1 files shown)


dschrempf (approved):
Instead of running the same combination of `(OS, GHC, LLVM)` twice, we could take the chance, and run different versions of GHC or LLVM in the PR and the merge queue.

edsko (approved):
Let's go with this for now; It would be nicer if we would test with a _single_ ghc and llvm version only, so that any differences in test outcome can _only_ be due to difference in platform, but that would either mean testing with llvm-14 across all platforms, which is suboptimal because llvm-14 lacks some features we kind of need, or usin llvm-15 across all platforms, which is currently difficult to do on Windows. We can revise again later.

PR #1559, opened 2026-01-19
Fail early after libclang errors

### LINKED ISSUE #1548: Fail early after libclang errors
When we run `preprocess` on a non-existent file, we get

```
[Error  ] [Libclang ] [clang] unable to resolve #include <nonexistingfile.h>
[Warning] [HsBindgen] [select] No declarations matched the select predicate
[Notice ] [HsBindgen] [artefact-no-bindings-multiple] All binding categories with base module name "Generated" are empty
```

Similarly, when a C file has clang errors, we get

```
[Error  ] [Libclang ] [clang] example.h:26:32: error: some error
[Warning] [HsBindgen] [select] No declarations matched the select predicate
[Notice ] [HsBindgen] [artefact-no-bindings-multiple] All binding categories with base module name "Generated" are empty
```

In both cases we should exit early.

dschrempf:
Fixes #1548

### DIFF
--- a/hs-bindgen/app/hs-bindgen-cli.hs
+++ b/hs-bindgen/app/hs-bindgen-cli.hs
@@ -57,6 +57,7 @@ execCliParser = do
             , clangArgsFooter
             , parsePredicateFooter
             , selectSliceFooter
+            , exitCodeFooter
             ]
         ]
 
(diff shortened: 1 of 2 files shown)


dschrempf:
Note to myself: Documentation about error codes still missing (Invocation, command line help). [DONE]

Issue #1438, opened 2025-12-18
Use argument names from C AST instead of Haskell AST for documentation translation


dschrempf:
At the moment, the frontend provides us with a `C.Decl`, which contains a `C.Comment`.

For functions, we also have a list of function parameters and their names (if available). We then go ahead and re-parse the Doxygen comment (`C.Comment`), extracting information about the parameters, translating it to `Hs.Comments` (main comment, parameter-specific comments etc.).

We construct (simplified)

```haskell
data Function = Function {
    ...
    , parameters :: [FunctionParameter]
    ...
    }
data FunctionParameter = FunctionParameter
  { name    :: Maybe (Hs.Name Hs.NsVar)
  , type    :: ClosedType
  , comment :: Maybe HsDoc.Comment
  }
```

That is, we store the parameter names (if present), and their translated Doxygen comments.

Observations:
- Reparsing the Doxygen comment with the information obtained from the function declaration is similar to a "Pass" in the Frontend.
- That is, we could use a pass-like architecture to "enrich" our declarations with Haddocks (this would remove some Haddock-related code complexity).
- We do not need the parameter names anymore after reparsing the Doxygen comment; we could through them away (that would require a "Pass p"-like type-safe structure similar to the one we have in the frontend).

edsko:
I definitely agree that the translation code is too monolithic; and we have in fact started  the process to break it up into separate, much smaller, modules. I'm not sure separate _passes_ are required though; in particular, translating Doxygen to Haddock in one pass and then the code in another feels unnecessary: after all, the translated comments don't _inform_ the subsequent pass in any way.

Having `name` in `FunctionParameter` agree does seem unnecessary; it would be better if comment generation would use the names as present in the _C_ AST. That said, I could imagine the parameter names being useful to improve the readability of any Haskell wrappers we generate. 

So I propose the following: 

- [x] Temporarily remove `name` from `FunctionParameter`
- [x] Fix documentation generation. The resulting fixtures should be unchanged  
- [x] Reintroduce `name` again, in order to make it possible to improve the readability of the function wrappers we generate (for which I have created a new, low priority, ticket: https://github.com/well-typed/hs-bindgen/issues/1541).

PR #1560, opened 2026-01-19
Use argument names from C AST instead of Haskell AST

bolt12:
This also fixed a bug where when `data` was used as the C parameter name, the mangler would change it to `data'`.


bolt12:
This closes #1438

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/Function.hs
```diff
@@ -145,11 +145,10 @@ functionDecs safety opts haddockConfig moduleName info origCFun _spec =
     foreignImportParams :: [Hs.FunctionParameter]
     foreignImportParams = [
            Hs.FunctionParameter{
-               name    = fmap (Hs.unsafeHsIdHsName . (.hsName)) mbName
-             , typ     = Type.inContext Type.FunArg (toPrimitiveType (classifyArgPassingMethod ty))
+               typ     = Type.inContext Type.FunArg (toPrimitiveType (classifyArgPassingMethod ty))
              , comment = Nothing
              }
-        | (mbName, ty) <- origCFun.args
+        | (_mbName, ty) <- origCFun.args
```
bolt12: Return the list of pairs (name, funParam)

PR #1562, opened 2026-01-20
Fix the constructor import issue (2)

### LINKED ISSUE #1282: FFI Constructor Import Issue

jorisdral:
Resolves #1282

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
@@ -1,8 +1,22 @@
 -- | Generate Haskell foreign imports (using the 'HasBaseForeignType' class)
 module HsBindgen.Backend.Hs.Translation.ForeignImport (
-    foreignImportDec
-  , hasBaseForeignTypeDecs
+    FunName (..)
+  , FunParam (..)
+  , FunRes (..)
+  , foreignImportDec
+  , foreignImportWrapperDec
+  , foreignImportDynamicDec
   ) where
+
+import Data.Function
+import Data.Map.Strict qualified as Map
+import Data.Maybe (isJust)
+import Data.Set qualified as Set
+import DeBruijn (Idx (IZ))
+import Optics.Core
+import Text.Printf (printf)
+
+import HsBindgen.Runtime.BaseForeignType qualified as BFT
 
 import HsBindgen.Backend.Hs.AST qualified as Hs
 import HsBindgen.Backend.Hs.AST.Type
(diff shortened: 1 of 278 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
```diff
@@ -1,59 +1,440 @@
...
-foreignImportDec name result parameters origName callConv origin safety =
-    Hs.DeclForeignImport foreignImportDecl
+  -> [Hs.Decl]
+foreignImportDec transState name params res origName callConv origin safety
+  | willForeignImportTypecheck transState funType
+  = [ Hs.DeclForeignImport foreignImportDecl
+    , Hs.DeclFunction funDecl
+    ]
+  | otherwise
+  = panicPure "foreignImportDec: generated binding would not compile"
```
jorisdral, line 71: It's currently a panic regardless of the reason why type checking would fail. There are a few reasons why type checking would fail:

1. `hs-bindgen` has a bug, where it tries to generate a foreign import for an invalid type (such as a Haskell `data` type)
2. In an external binding spec, an `ffitype` was erroneously omitted
3. In an external binding spec, a `HasBaseForeignType` instance was erroneously omitted

Reason 1 should be a panic, reasons 2 and 3 should be caught in the fronted, so that we can emit a warning message and omit the generated bindings.

And we should take the same approach for functions `foreignImportWrapperDec` and `foreignImportDynamicDec`

If we decide to make this change, I'd prefer to do it in a follow-up PR

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
```diff
@@ -1,59 +1,440 @@
...
+unsafeToBase ntMap ty = case toBase ntMap ty of
+    Nothing ->
+      panicPure $ printf "Type does not have base foreign type: %s" (show ty)
+    Just ty' ->
+      ty'
+
+toBase :: NewtypeMap -> HsType -> Maybe HsType
+toBase ntMap ty =
+      fmap fromBaseForeignType
+    $ toBaseForeignType ntMap ty
```
jorisdral, line 300: Maybe `ForeignImportDecl`, `ForeignImportWrapper`, and `ForeignImportDynamic` should use `BFT.BaseForeignType` rather than `HsType`? It would require some (boring) plumbing, but the upside would be that we enforce statically that a foreign import declaration only uses base foreign types 

If we decide to make this change, I'd prefer to do it in a follow-up PR

examples/constructor-import-issue/c/constructor_import_issue.h
```diff
@@ -0,0 +1,11 @@
+typedef char A;
+
+typedef A B;
+
+extern B ex_global;
+
+extern B ex_func (B x);
```
jorisdral:
```
foreign import "ex_func" ... :: B -> IO B
```

```
foreign import "ex_func" hs_bindgen_29812748_base :: CChar -> IO CChar

hs_bindgen_29812748  :: B -> IO B
hs_bindgen_29812748 = fromBaseForeignType hs_bindgen_29812748_base

ex_func :: B -> IO B
ex_func = hs_bindgen_29812748
```

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
```diff
@@ -1,59 +1,440 @@
...
         , origName   = origName
         , callConv   = callConv
         , origin     = origin
-        , comment    = mbUniqueSymbol
+        , comment    = fiComment
         , safety     = safety
         }
+      where
+
+    fiName = Hs.InternalName (name.uniqSymbol & #unique %~ (<> "_base"))
```
jorisdral, line 90: Add a comment that this name is also unique because it is created from a unique name + suffix

examples/constructor-import-issue/c/constructor_import_issue.h
jorisdral, line 1: We should explain with comments which declarations live in which module/external binding spec

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
```diff
@@ -1,59 +1,440 @@
...
   -> Safety
-  -> Hs.Decl
-foreignImportDec name result parameters origName callConv origin safety =
-    Hs.DeclForeignImport foreignImportDecl
+  -> [Hs.Decl]
+foreignImportDec transState name params res origName callConv origin safety
+  | willForeignImportTypecheck transState funType
+  = [ Hs.DeclForeignImport foreignImportDecl
+    , Hs.DeclFunction funDecl
+    ]
```
jorisdral, line 69: This generates two declarations even if the type signatures are exactly the same. Maybe we should run a simplifier pass to "squash" the two when applicable? Let's open a ticket for this

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
jorisdral: We should open a ticket for the follow-up work

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
jorisdral: We're probably not going to do this now, so I won't open an issue, but I'll leave a note here for possible improvements in the future

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
jorisdral: Created an issue: #1566

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
jorisdral: Opened an issue: #1567

examples/constructor-import-issue/c/constructor_import_issue.h
jorisdral: Done

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
jorisdral: Done

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
jorisdral: Done

PR #1563, opened 2026-01-20
Add `PtrConst` replacement for `ConstPtr`

### LINKED ISSUE #1537: Add a custom `PtrConst` replacement for `ConstPtr`
We will define 

```hs
#if MIN_VERSION_base(...)
type PtrConstr a = ConstrPtr a
#else
newtype PtrConstr a = PtrConst (Ptr a) -- constructor not exported
#endif
```

And then we additionally offer an API for working with this:

```hs
peek :: PtrConst a -> IO a
unsafeToPtr :: PtrConst a -> Ptr a
```

(but not `poke`).

This is then also consistent with the nomenclature in https://github.com/well-typed/hs-bindgen/issues/1536.

jorisdral:
Resolves #1537

### DIFF
--- a/examples/c-qrcode/generate-and-run.sh
+++ b/examples/c-qrcode/generate-and-run.sh
@@ -32,7 +32,7 @@ cabal run hs-bindgen-cli -- \
     "$SCRIPT_DIR/QR-Code-generator/c/qrcodegen.h"
 
 echo "# "
-echo "# Creating cabal.project.local"
+echo "# Updating cabal.project.local"
 echo "# "
 
 LINE=$(
--- a/examples/c-rogueutil/generate-and-run.sh
+++ b/examples/c-rogueutil/generate-and-run.sh
@@ -32,16 +32,20 @@ cabal run hs-bindgen-cli -- \
     "$SCRIPT_DIR/rogueutil/rogueutil.h"
 
 echo "# "
-echo "# Creating cabal.project.local"
+echo "# Updating cabal.project.local"
 echo "# "
 
-cat > "$SCRIPT_DIR/hs-project/cabal.project.local" <<EOF
+LINE=$(
+    cat <<-EOF
 package c-rogueutil
     extra-include-dirs:
         $SCRIPT_DIR/rogueutil
     extra-lib-dirs:
         $SCRIPT_DIR/rogueutil
 EOF
+)
+grep -qxF "$LINE" "$SCRIPT_DIR/hs-project/cabal.project.local" || echo "$LINE" >>"$SCRIPT_DIR/hs-project/cabal.project.local"
+cat "$SCRIPT_DIR/hs-project/cabal.project.local"
 
 echo "# "
 echo "# Done!"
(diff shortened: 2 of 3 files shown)


hs-bindgen-runtime/src/HsBindgen/Runtime/PtrConst.hs
```diff
@@ -0,0 +1,98 @@
...
+--
+-- A @'PtrConst' a@ is a pointer to a type @a@ with a C @const@ qualifier. For
+-- instance, the Haskell type @PtrConst CInt@ is equivalent to the C type @const
+-- int*@. @const@-qualified contents of a pointer should not be modified, but
+-- reading the contents is okay.
+type PtrConst :: Type -> Type
+type role PtrConst phantom
+newtype PtrConst a = PtrConst { un :: Ptr a }
+    deriving stock (Eq, Ord)
+    deriving newtype Storable
```
jorisdral, line 39: This deviates from the description of #1537 in that it is unconditionally a `newtype`, rather than a type synonym on older base versions. 

I think it is better to have consistency with an unconditional newtype. There are `unsafeToConstPtr` and `unsafeFromConstPtr` escape hatches that one could use if they want a `ConstPtr` instead

(It also made the implementation slightly easier)

hs-bindgen-runtime/src/HsBindgen/Runtime/Marshal.hs
```diff
@@ -346,6 +346,7 @@ instance ReadRaw    (Ptr a)
 instance WriteRaw   (Ptr a)
 
 -- TODO Foreign.C.ConstPtr from base-4.18.0.0 ?
+-- TODO PtrConst from hs-bindgen-runtime ?
```
jorisdral: Should we do this now?

hs-bindgen-runtime/src/HsBindgen/Runtime/PtrConst.hs
```diff
@@ -0,0 +1,98 @@
+{-# LANGUAGE RoleAnnotations #-}
+{-# LANGUAGE StandaloneKindSignatures #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module HsBindgen.Runtime.PtrConst (
+    -- TODO: remove the 'PtrConst' constructor when FFI types are implemented.
+    -- See issue #1282.
```
jorisdral: Shouldn't forget to do this in #1562

jorisdral: #1562 is merged now, so I should make the change here instead

jorisdral: Actually, I can only make this change in #1573. I tried to make the change now but I get compiler errors

hs-bindgen-runtime/src/HsBindgen/Runtime/PtrConst.hs
jorisdral: Made it a type synonym after all

Issue #1455, opened 2025-12-23
Users need to import packages other than `hs-bindgen-runtime`


dschrempf:
Pulling along the [`hs-bindgen` Nix tutorial](https://github.com/well-typed/hs-bindgen-tutorial-nix), I noticed that we now have to add build dependencies `primitive` and `ghc-prim`.

I think the declarations we use from there should be re-exported from `hs-bindgen-runtime` (but there may be complications I am not aware of).

I am leaving this ticket here, so we don't forget to discuss this.

Relevant error:
```
       > src/Generated/Pcap.hs:26:1: error: [GHC-87110]
       >     Could not load module ‘Data.Primitive.Types’.
       >     It is a member of the hidden package ‘primitive-0.9.1.0’.
       >     Perhaps you need to add ‘primitive’ to the build-depends in your .cabal file.
       >     Use -v to see a list of the files searched for.
       >    |
       > 26 | import qualified Data.Primitive.Types
       >    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       >
       > src/Generated/Pcap.hs:40:1: error: [GHC-87110]
       >     Could not load module ‘GHC.Prim’.
       >     It is a member of the hidden package ‘ghc-prim-0.12.0’.
       >     Perhaps you need to add ‘ghc-prim’ to the build-depends in your .cabal file.
       >     Use -v to see a list of the files searched for.
       >    |
       > 40 | import GHC.Prim ((*#), (+#))
```

edsko:
I think this will be resolved by https://github.com/well-typed/hs-bindgen/issues/1627 .

PR #1568, opened 2026-01-20
Do not import `GHC.Prim` from `ghc-prim` in our bindings

jorisdral:
It has a buggy interaction with `-Wunused-packages`. Instead, we import `GHC.Exts` from `base`.

See https://gitlab.haskell.org/ghc/ghc/-/issues/23212 for a description of the bug

### DIFF
--- a/hs-bindgen/fixtures/attributes/attributes/Example.hs
+++ b/hs-bindgen/fixtures/attributes/attributes/Example.hs
@@ -20,7 +20,7 @@ import qualified GHC.Ptr as Ptr
 import qualified GHC.Records
 import qualified HsBindgen.Runtime.HasCField
 import Data.Void (Void)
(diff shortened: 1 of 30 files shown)


edsko (approved):
I guess this would be somewhat obsoleted by https://github.com/well-typed/hs-bindgen/issues/1455 but fine for now.

jorisdral:
> I guess this would be somewhat obsoleted by #1455 but fine for now.

It might not be -- I think both `'(GHC.Exts.+#)` or `'(GHC.Base.+#)` are resolved to `GHC.Prim` by template Haskell because `(+#)` is re-exported from `GHC.Exts` and `GHC.Base`. So I'm not sure if re-exporting from `hs-bindgen-runtime` would fix the issue, but we should check of course

PR #1569, opened 2026-01-20
Derive Bitfield for newtypes

### LINKED ISSUE #1289: Derive `Bitfield` instances for `newtype`s generated for `typedef`s
If we generate bindings for: 
```c
struct myStruct {
  int x : 2;
};
```
And then compile the bindings, everything works.

If instead we generate bindings for:
```c
typedef int myInt;
struct myStruct {
  myInt y : 2;
};
```

Then we get a GHC error that roughly says:
```hs
    • No instance for ‘HsBindgen.Runtime.Bitfield.Bitfield MyInt’
        arising from a use of ‘HsBindgen.Runtime.HasCField.peekCBitfield’
```

It should be simple to derive a `Bitfield` instance because `MyInt` is just a newtype around `CInt`.

Alternatively, we can disallow bit-field types like `myInt`. The [C reference section on bit-fields](https://en.cppreference.com/w/c/language/bit_field.html) seems pretty restrictive about what types are allowed as bit-fields, but implementations like `gcc` and `clang` seem to be more flexible.

bolt12:
Closes #1289

### DIFF
--- a/scripts/ci/compile-fixtures.sh
+++ b/scripts/ci/compile-fixtures.sh
@@ -76,7 +76,7 @@ KNOWN_EMPTY=(
 #
 # This number is used for sanity checks. Make sure to update this number when
 # new fixtures are added or old ones are removed.
-KNOWN_FIXTURES_COUNT=136
+KNOWN_FIXTURES_COUNT=137
 
 # Default options
 JOBS=4


TravisCardwell:
I am currently working on the new `ResolveInstances` pass, so I am looking at this PR so that I can update my implementation/plans accordingly.

My understanding is that primitive integral types should have `Bitfield` instances.  Looking at the code:

* We do not provide instances for `short` types.  Intentional?
* We do not provide instances for `long long` types.  Intentional?

Compilers generally allow C99 fixed-size integral types to also be used in bit-fields.  If we want to be able to derive `Bitfield` instances for `newtype`s, perhaps we need to add instances for all of the Haskell types that we translate those to:

* `Int8`, `Int16`, `Int32`, `Int64`
* `Word8`, `Word16`, `Word32`, `Word64`
* `CIntMax`, `CUIntMax`
* `CIntPtr`, `CUIntPtr` ?

Regarding instance derivation, I guess this uses the `newtype` strategy: we derive an instance (using `GeneralizedNewtypeDeriving`) only if the underlying type has an instance.  We do this by default, even when not explicitly requested, and no other strategies are valid.  Is this accurate?

bolt12:
@TravisCardwell, I have had to add some BitField instances for primitive types, but I am unsure if that's the way to go, it feels harmful, so I went with it. I am not sure if it was intentional or not, probably @edsko will tell me when reviewing this.

Regarding the deriving strategy, yes, I think the strategy we want is newtype deriving and not any other.

hs-bindgen-runtime/src/HsBindgen/Runtime/Bitfield.hs
```diff
@@ -51,14 +51,18 @@ class Bitfield a where
     -- For example this may extend the sign for signed types.
     extend :: Word64 -> Int -> a
 
+instance Bitfield CUShort where extend = unsignedExtend
```
bolt12: Add instances for all primitive data type (see Travis' list)

hs-bindgen/fixtures/binding-specs/bs_pre_name_type/bindingspec.yaml
```diff
@@ -15,6 +15,7 @@ hstypes:
       - un_MySym
       ffitype: CChar
   instances:
+  - Bitfield
```
edsko, line 18: @TravisCardwell This is confusing to me; notice how new `Bitfield` class is included in the external binding spec, even though @bolt12 did not add `Bitfield` to the `instances` field. Perhaps I am confusing two things, but perhaps what we discussed this morning wasn't quite right...? :thinking: 

(@bolt12 This is independent of your PR, feel free to ignore!)

TravisCardwell: * `Bitfield` is added to the `integralInsts` in `HsBindgen.Backend.Hs.Translation.Instances`.
* It is also added to the `candidateInsts` for `typedef`s in `HsBindgen.Backend.Hs.Translation`.
* `bitfieldDecl` generates an instance if the resolved instance includes the class.
* An instance is therefore generated for any `newtype` with an integral underlying type.
* This example has an integral underlying type, and the instance is correctly generated.
* What we discussed this morning is correct.  Some classes are not implemented fully/correctly.  @bolt12 implemented this correctly, but *others* are still inconsistent.

Sorry for the quick/terse response (without links); I am currently not working :bow:

edsko: I see. Then I don't think I really understand what `instances` is used for.

Also, stop responding when you're not working! :grimacing:

edsko: But thanks :)

PR #1570, opened 2026-01-20
Improve interface with flexible array members

dschrempf:
This is still a WIP, but when it is ready for review, it would be nice to get your feedback @jorisdral and @TravisCardwell.

In particular,
- @jorisdral could you check the `HsWithFlexibleArrayMember` clause of `toFFIType` (I guess this extends to `toBaseForeignType`)
- @TravisCardwell could you check the `HsWithFlexibleArrayMember` clause of `getInstances` (also related to #648, and #1572)

(I will ping you again when the PR is ready).

TODOs:
- [x] Separate module for translating `struct`s
- [x] Separate functions for translating structs with and without FLAMs

### DIFF
--- a/manual/hs/manual/app/Manual/Types/Structs.hs
+++ b/manual/hs/manual/app/Manual/Types/Structs.hs
@@ -12,8 +12,9 @@ import Foreign.C qualified as FC
 import System.IO.Unsafe
 
 import HsBindgen.Runtime.ConstPtr
-import HsBindgen.Runtime.FlexibleArrayMember qualified as FLAM
+import HsBindgen.Runtime.FlexibleArrayMember qualified as Flam
 import HsBindgen.Runtime.IncompleteArray qualified as IA
+import HsBindgen.Runtime.Marshal (ReadRaw (..))
 
 import Manual.Tools
 import Structs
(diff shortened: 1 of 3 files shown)


TravisCardwell:
> check the `HsWithFlexibleArrayMember` clause of `getInstances`

`Set.empty` looks good.  A type alias does not have any instances.  Also, `WithFlexibleArrayMember` is not a type class (https://github.com/well-typed/hs-bindgen/issues/1572#issuecomment-3775036090).

BTW, I am currently working on the `ResolveInstances` pass, and `getInstances` will be removed (likely after this has merged).

edsko:
@TravisCardwell @dschrempf 

> Set.empty looks good. A type alias does not have any instances. Also, WithFlexibleArrayMember is not a type class (https://github.com/well-typed/hs-bindgen/issues/1572#issuecomment-3775036090).

I think this confusion might have been resolved already, but just in case it's not: `Set.empty` is _not_ the right answer. If we have something like

```c
struct pascal {
  int length;
  char[] elems;
}

struct outer {
  struct pascal foo;
};
```

then returning the empty set for `WithFlexibleArrayMember` would mean that `outer` cannot have any instances either. It's a stop-gap; it's not the correct answer.

Or am I misunderstanding something?

dschrempf:
I think you are right, Edsko. I left a TODO comment referencing #1572, so we can solve this apprpriately!

Also Joris confirmed that `WithFlexibleArrayMember` has no FFI types.

PR #1573, opened 2026-01-20
Remove the `ffitype` field from binding specs

### LINKED ISSUE #1534: Delete `ffitype` from binding specs
We realized that it's not required after all, since we can look at the underlying C type (#1494).

jorisdral:
Resolves #1534


hs-bindgen/fixtures/binding-specs/fun_arg/typedef/enum/bindingspec.yaml
```diff
@@ -50,7 +48,6 @@ hstypes:
       constructor: E
       fields:
       - un_E
-      ffitype: CInt
```
jorisdral, line 53: Turns out this `ffitype` was wrong, should have been a `CUInt`

hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST/Type.hs
```diff
@@ -63,14 +63,11 @@ data HsPrimType =
...
     | HsPrimCDouble
-
-  -- Int8 Int16 Int32 Int64
-  -- Word8 Word16 Word32 Word64
   deriving stock (Eq, Ord, Generic, Show)
 
 data HsType =
     HsPrimType HsPrimType
-  | HsTypRef (Hs.Name Hs.NsTypeConstr)
+  | HsTypRef (Hs.Name Hs.NsTypeConstr) (Maybe HsType)
```
jorisdral: Should add a comment what the `Maybe` represents

hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST/Type.hs
```diff
@@ -79,11 +76,10 @@ data HsType =
   | HsConstPtr HsType
   | HsIO HsType
   | HsFun HsType HsType
-  | HsExtBinding Hs.ExtRef BindingSpec.CTypeSpec BindingSpec.HsTypeSpec
+  | HsExtBinding Hs.ExtRef BindingSpec.CTypeSpec BindingSpec.HsTypeSpec HsType
```
jorisdral: Should add a comment what the `Maybe` represents

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/Type.hs
```diff
@@ -79,141 +75,71 @@ inContext ctx = go ctx
...
+      Just "uint8_t"  -> prim Hs.HsPrimWord8
+      Just "uint16_t" -> prim Hs.HsPrimWord16
+      Just "uint32_t" -> prim Hs.HsPrimWord32
+      Just "uint64_t" -> prim Hs.HsPrimWord64
+      _ -> Nothing
+  | otherwise
+  = Nothing
+  where
+    prim :: Hs.HsPrimType -> Maybe Hs.HsType
+    prim = Just . Hs.HsPrimType
```
jorisdral: This mapping is incomplete, but sufficient for the tests to run succesfully. Going to make it complete later

jorisdral: I'm undecided about where this mapping should live...

jorisdral: Are there situations where `size_t` is not a `ResolvedExtBinding`? Or situations where `size_t` is a custom C declaration, not from the standard library?

TravisCardwell: > Are there situations where `size_t` is not a `ResolvedExtBinding`?

Yes.  When using `hs-bindgen` with the `stdlib` external bindings disabled, one can generate bindings that include needed parts of the standard library, using program slicing.  The implementation for `size_t` would therefore be included (CPP-expanded for the target).

> Or situations where `size_t` is a custom C declaration, not from the standard library?

I can imagine such situations when working with embedded systems, but perhaps it is unlikely on any platform that GHC can run on.

TravisCardwell: I am uneasy about such name-based translation.  Why do we need to special-case things like this?  We are unable to determine the primitive types from the underlying types?

edsko: This PR is a draft so I'm a bit hesitant to comment at this stage, but agreed with this observation. _If_ we really need name-based translation, then that's kind of what the whole binding specs infra does (match against header etc).

jorisdral: This name-based translation was used just as a Proof of Concept -- I think the translation should happen not based only on the name, but with proper resolution of the name + header etc, so using the binding specs infra in some way

> We are unable to determine the primitive types from the underlying types?

The true underlying type for `size_t` is something like, say, `CUInt` on my system, but I'd prefer to map it `CSize` because

* `CSize` is what typically use in Haskell for `size_t`
* References to `CSize` are platform-independent, even if the implementation of `CSize` depends on the platform 
* `CUInt` is *not* platform-independent

What do you think?

jorisdral: Actually, we're probably getting rid of this mapping altogether. See https://github.com/well-typed/hs-bindgen/issues/1534#issuecomment-3777543811

jorisdral: Mapping is removed

PR #1585, opened 2026-01-22
Support simultaneous declaration of multiple anon decls

### LINKED ISSUE #1430: Support simultaneous declaration of "multiple" anon decls
Consider

```c
typedef struct { int s1_field; } s1a, s1b;

typedef struct { int s2_field; } s2a, *s2b;

typedef struct { int s3_field; } *s3a, *s3b;

void f(s2a x, s2b y) {
        *y = x;
}
```

Note that the assignment in `f` is valid: `s2a` and `s2b` refer to the _same_ "anonymous" declaration. So this is another case of "multiple" uses of an anon decl (see also #1418).

bolt12:
This closes #1430

### DIFF
--- a/hs-bindgen/fixtures/edge-cases/anon_multiple_typedefs/th.txt
+++ b/hs-bindgen/fixtures/edge-cases/anon_multiple_typedefs/th.txt
@@ -345,7 +345,7 @@ test :: Point2a -> Point2b -> IO Unit
 -}
 test = \x_0 -> \y_1 -> with x_0 (\x_2 -> hs_bindgen_c97a0d4458699ad7 x_2 y_1)
 -- __unique:__ @test_edgecasesanon_multiple_typed_Example_Unsafe_test@
-foreign import ccall safe "hs_bindgen_f90e97b8d269be4e" hs_bindgen_f90e97b8d269be4e_base ::
+foreign import ccall unsafe "hs_bindgen_f90e97b8d269be4e" hs_bindgen_f90e97b8d269be4e_base ::
     Ptr Void
  -> Ptr Void
  -> IO Unit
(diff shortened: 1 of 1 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/AnonUsage.hs
```diff
@@ -94,6 +94,33 @@ resolveConflicts anonId new old =
...
+        -- We choose the first typedef (in source order).
+        old
+      (TypedefDirect _, TypedefIndirect _) ->
+        -- Example:
+        --
+        -- > typedef struct { int x; int y; } point2a, *point2b;
+        --
+        -- IMPORTANT: This case only occurs in Clang < 16. In Clang >= 16,
+        -- the indirect typedef does not reference the anonymous struct at
+        -- all, so there is no conflict to resolve.
```
bolt12: This can happen for Typedef, Typedef too, in Clang > 16 that case is not a problem

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/AnonUsage.hs
```diff
@@ -94,6 +94,33 @@ resolveConflicts anonId new old =
...
+        --
+        -- We choose the first typedef (in source order).
+        old
+      (TypedefIndirect _, TypedefIndirect _) ->
+        -- Example:
+        --
+        -- > typedef struct { int x; int y; } *point3a, *point3b;
+        --
+        -- We choose the first typedef (in source order).
+        old
```
bolt12: This is the only case where Clang > 16 can find a conflict

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/AnonUsage.hs
```diff
@@ -94,6 +94,33 @@ resolveConflicts anonId new old =
...
+        --
+        -- > typedef struct { int x; int y; } point2a, *point2b;
+        --
+        -- IMPORTANT: This case only occurs in Clang < 16. In Clang >= 16,
+        -- the indirect typedef does not reference the anonymous struct at
+        -- all, so there is no conflict to resolve.
+        old
+      (TypedefIndirect _, TypedefDirect _) ->
+        -- Mirror of the above case (for when indirect comes first in source
+        -- order, though this is unlikely in practice).
```
bolt12: This is critical that we do this because Clang >= 16 will use the first direct declaration to name the anonymous structure

Issue #976, opened 2025-08-06
Instance documentation fails for multi-parameter type classes in Template Haskell


bolt12:
In #959 we can't attach documentation on top of TH generated instances due to a GHC bug (See https://gitlab.haskell.org/ghc/ghc/-/issues/26276).

The workaround for now is to add the documentation to the first declaration of the instance.

edsko:
Marking this as low priority as it's currently blocked on upstream, and the impact on the quality of the generated docs is quite small.

PR #1584, opened 2026-01-22
Improve Guasi: Documentation functions use 'Comment' (no `Maybe Comment`)

dschrempf:
https://github.com/well-typed/hs-bindgen/pull/1586 includes the commit in this PR, closing this one.

PR #1586, opened 2026-01-22
Avoid ambiguous usage errors in TH-mode

dschrempf:
When attaching documentation to declarations with ambiguous names (e.g., `reverse`), `withDecDoc` fails because it tries to attach the documentation to the declaration defined elsewhere.

Now we avoid `withDecDoc` altogether and use a custom `putDocName` function that attaches documentation to declarations with the provided name _in the current module_.

This PR also fixes a bug, where we created `safe` foreign import functions in TH mode independent of the actual safety specified.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/TH/Translation.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/TH/Translation.hs
@@ -728,7 +728,7 @@ mkDecl = \case
               | field <- record.fields
               ]
 
-        traverse_ (uncurry putDocNameM) docs
+        traverse_ (uncurry putLocalDocM) docs
 
         decl <-
           TH.dataD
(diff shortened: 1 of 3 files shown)


dschrempf:
This PR incldues the commit in https://github.com/well-typed/hs-bindgen/pull/1584, so closing the other one.

dschrempf:
I think we could use a similar procedure to fix https://github.com/well-typed/hs-bindgen/issues/976 (documentation of instances).

edsko:
Since @dschrempf  might not be around today, I'll take over this PR.

edsko:
I have reported the issue upstream: https://gitlab.haskell.org/ghc/ghc/-/issues/26817 .

edsko:
Opened https://github.com/well-typed/hs-bindgen/issues/1587 to track the fact that our TH fixtures contain duplicate symbols.

edsko:
> I think we could use a similar procedure to fix https://github.com/well-typed/hs-bindgen/issues/976 (documentation of instances).

I don't _think_ it's related? But would be great if it was true of course :)

PR #1594, opened 2026-01-26
Make it possible to keep a type opaque

TravisCardwell:
This is implemented by specifying `representation: emptydata` in the Haskell type specification.  This is a significant change to a type that must be taken into account in all passes (from `ResolveBindingSpecs`).  This involves a number of design decisions.  I describe the implemented design below, but we can discuss options and select alternatives if desired.

We need to ensure that a specified representation is compatible with the underlying kind.  I implemented this in `ResolveBindingSpecs`, adding `TODO`s for implementing validation of other representations.  For the `emptydata` representation, the current implementation accepts only the following underlying kinds:

* `struct`
* `union`
* `enum`
* `typedef`
* opaque type
* type macro

If a user specifies `emptydata` representation for a different kind, such as a function, a warning is omitted and that `emptydata` representation is removed for subsequent processing.

Making a type opaque affects selection.  In particular, types that would be selected via program slicing should no longer be selected if they are only used in a type that is made opaque.  Selection handles dependencies using the graphs, so I implemented this by removing edges in the graphs, like how we make selection work with external bindings.  For external bindings, edges *to* the vertices are removed.  For types made opaque, the edges *from* the vertices are removed.

We also need to decide what to do about anonymous declarations defined in a type that is made opaque, as discussed in [this comment](https://github.com/well-typed/hs-bindgen/issues/809#issuecomment-3798235154).  The current implementation does *not* translate such (direct) anonymous declarations, as such types cannot be used elsewhere.  (Note that this is easy to implement, while translating such anonymous types would be more involved.)

We currently still create `Hs` declarations in the backend.  The current implementation does this for _any_ declaration that specifies `emptydata` (without further validation: the validation is already done in `ResolveBindingSpecs`).


TravisCardwell:
I added some tests for *nested* types within a type that is made opaque:

```c
// When struct oaa is translated to an opaque type, neither anonymous struct
// should be translated.
struct oaa {
  struct {
    struct {
      int e, f;
    } c, d;
  } a, b;
};

// When struct oan is translated to an opaque type, the anonymous struct should
// not be translated but struct named should be translated because it may still
// be used even when struct oan is opaque.
struct oan {
  struct {
    struct named {
      int e, f;
    } c, d;
  } a, b;
};
```

TravisCardwell:
The `fun_arg` tests organized tests differently than the existing tests.  I added a commit that reorganizes the other tests to match.

edsko (approved):
Let's see if we can replace the `Decl..` by `DeclOpaque`, so that the modifications of the use-decl graph are "justified".

TravisCardwell:
Changing the code to mutate the declaration itself, changing the `kind` to `DeclOpaque`, works fine.

Issue #1519, opened 2026-01-13
Avoid referring to "renaming" Haskell declarations (internally and in documentation)


edsko:
Prior to the Great Identifier Simplification, we implemented certain features as _renaming_ C declarations:

* Name clashes resulting from typedefs would be handled by renaming the struct wrapped in the typedef (in the no-longer-existent `HandleTypedefs`) 
* Prescriptive binding specs could be applied to the C declaration

We don't do this anymore: the C declaration has a fixed identifier, and is never changed; naming decisions like the ones mentioned above still exist, of course, but they don't _rename_ anything; instead, they merely change how we _name_ the corresponding Haskell declarations. "Renaming" in this context doesn't really make sense, because we're _generating_ those Haskell declarations; there is nothing to "rename". 

We should grep for this, both internally in the code and (more importantly) in documentation.

edsko:
(cc also @TravisCardwell )

TravisCardwell:
I renamed the binding specification tests, in #1522.

PR #1595, opened 2026-01-27
Use "assign name" instead of "rename" in the name mangler

dschrempf:
Fixes https://github.com/well-typed/hs-bindgen/issues/1519

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
@@ -1250,3 +1250,20 @@ anonEnumConstantDecs haddockConfig info anonEnumConstant =
           , comment = mkHaddocks haddockConfig info patSynName
           }
   in  [typeSigDecl]
+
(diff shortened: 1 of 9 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -1250,3 +1250,19 @@ anonEnumConstantDecs haddockConfig info anonEnumConstant =
           , comment = mkHaddocks haddockConfig info patSynName
           }
   in  [typeSigDecl]
+
+{-------------------------------------------------------------------------------
+  Internal helpers
+-------------------------------------------------------------------------------}
+
+-- TODO D.
+renameHsName :: (Hs.Identifier -> Hs.Identifier) -> DeclIdPair -> DeclIdPair
```
dschrempf, line 1260: How should we name this function?

dschrempf: (Refer to name mangler ticket).

PR #1614, opened 2026-01-28
Support top-level anonymous structs

### LINKED ISSUE #1429: Support top-level anonymous structs
I'm not sure quite how useful this is, I suspect not very useful at all. But in principle a header could define

```c
struct { int x; int y; } a;
```

The strange thing is that this is a non-static non-extern global, which is a strange beast already, but it's _possible_. 

I _think_ this is the _only_ case we'd have to consider; in particular, I _think_ an `extern` declaration such as

```c
extern struct { int x; int y; } a;
```

is not usable, because if another C file has the "corresponding" declaration, it would have to repeat the anonymous struct and those two types would be incompatible.

Marking this as a low priority enhancement for now, until we have a specific use case.

bolt12:
This closes #1429

### DIFF
--- a/hs-bindgen/fixtures/globals/globals/Example/Global.hs
+++ b/hs-bindgen/fixtures/globals/globals/Example/Global.hs
@@ -24,6 +24,10 @@ module Example.Global
     , Example.Global.streamBinary
     , Example.Global.streamBinary_len
     , Example.Global.some_global_struct
+    , Example.Global.anonPoint
+    , Example.Global.anonPair
+    , Example.Global.anonEnum
+    , Example.Global.anonEnumCoords
     )
   where
 
--- a/hs-bindgen/fixtures/globals/globals/Example.hs
+++ b/hs-bindgen/fixtures/globals/globals/Example.hs
@@ -20,6 +20,15 @@ module Example
     , Example.Version_t(..)
     , Example.Struct1_t(..)
     , Example.Struct2_t(..)
+    , Example.AnonPoint(..)
+    , Example.AnonPair(..)
+    , Example.AnonEnum(..)
+    , pattern Example.VAL_A
+    , pattern Example.VAL_B
+    , Example.AnonEnumCoords(..)
+    , pattern Example.X
+    , pattern Example.Y
+    , pattern Example.Z
     )
   where
 


hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/AnonUsage.hs
```diff
@@ -64,6 +64,16 @@ data Context =
...
+
+    -- | Anonymous declaration used as the type of a global variable
+    --
+    -- E.g.
+    --
+    -- > struct { int x; int y; } a;
+    --
+    -- In this case, we use the name of the global variable as the name of
+    -- the struct.
+  | TopLevelStruct (C.DeclInfo Parse)
```
bolt12:
```suggestion
  | GlobalVar (C.DeclInfo Parse)
```

hs-bindgen/examples/golden/globals/globals.h
```diff
@@ -433,3 +433,9 @@ struct2_t some_global_struct = {
         },
      }
   };
+
+// Basic anonymous struct global
+struct { int x; int y; } anonPoint;
+
+// Anonymous struct with initializer
+struct { int a; int b; } anonPair = { 1, 2 };
```
bolt12, line 442: Add test case for `enum`

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/AssignAnonIds/ChooseNames.hs
```diff
@@ -100,6 +102,25 @@ chooseNames (AnonUsageAnalysis usageAnalysis) =
...
+    --
+    -- For example, given:
+    --
+    -- > struct { int x; int y; } a;
+    --
+    -- the struct is named "a".
+    --
+    -- Similar to 'nameForTypedefDirect', we set @isAnon@ to @False@ because
+    -- the struct effectively takes on the identity of the global variable name.
+    nameForTopLevelStruct :: AnonId -> DeclId -> DeclId
```
bolt12: change the function name here

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -704,11 +704,11 @@ varDecl info = \curr -> do
                    || (isTentative && declCls == DefinitionUnavailable)
 
         pure $ (fails ++) $
-          if not (null anonDecls) then [
+          if not (null anonDecls) && not isDefn then [
               parseFail info.id info.loc $
                 ParseUnexpectedAnonInExtern
             ]
```
bolt12: We can remove this condition.

bolt12: Otherwise, check linkage (line 668)

bolt12:
We found a bug on `varDecl` and the test was panicking but we weren't seeing it. After debugging we pin pointed the issue. We couldn't have found it if we didn't test for anon global enums. See here what the fix was: [`ef50029` (this PR)](https://github.com/well-typed/hs-bindgen/pull/1614/commits/ef50029d6d0360925580108658cc5fdecaefc6c5#diff-99799ef54c69cbceb68c63d2608ba344c79030d6078e73d89488a9f80f368004R743)

edsko:
So there are currently two problems with this, which actually require a bit of thought. Given

```c
enum { ... } someGlobalVar1;
```

We currently try to generate a wrapper to get a pointer to that global var:

```c
???? hs_bindgen_get_someGlobalVar1() {
  return &someGlobalVar1;
}
```

but what type should this return? The `enum` is anonymous; we cannot refer to it!

Secondly, for

```c
struct { 
  int x; 
  int y;
} someGlobalVar2;
```

we have a similar problem. In this case, the struct cannot really be passed around at all; perhaps we should instead generate wrappers that get us pointers to the _fields_ rather than the struct...?

Let's put this on hold until at least after the alpha release.

hs-bindgen/fixtures/globals/globals/Example/Global.hs
```diff
@@ -133,6 +133,30 @@ $(HsBindgen.Runtime.Internal.CAPI.addCSource (HsBindgen.Runtime.Internal.CAPI.un
   , "{"
   , "  return &some_global_struct;"
   , "}"
+  , "/* test_globalsglobals_Example_get_anonPoint */"
+  , "__attribute__ ((const))"
+  , "struct anonPoint *hs_bindgen_6a74c096c69eaf13 (void)"
```
bolt12: Fix this to generate a void* pointer

scripts/ci/compile-fixtures.sh
```diff
@@ -251,6 +251,7 @@ compile_fixture() {
         -optc -std=gnu2x \
         -optc -Wno-deprecated-declarations \
         -optc -Wno-attributes \
+        -optc -Wno-incompatible-pointer-types \
```
bolt12: remove this

hs-bindgen/examples/golden/globals/globals.h
```diff
@@ -433,3 +433,15 @@ struct2_t some_global_struct = {
...
+
+// Basic anonymous enum global
+enum { VAL_A = 0, VAL_B = 1 } anonEnum;
```
bolt12, line 445: Fix this to generate a C wrapper that returns the underlying enum type, conditionally on wether it is anon or not.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -704,11 +704,7 @@ varDecl info = \curr -> do
                    || (isTentative && declCls == DefinitionUnavailable)
 
         pure $ (fails ++) $
-          if not (null anonDecls) then [
-              parseFail info.id info.loc $
-                ParseUnexpectedAnonInExtern
-            ]
```
bolt12, line 730: Re-introduce this condition and also take into consideration the new `IsExtern` information to not parse extern anon declarations at all.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -973,6 +973,21 @@ addressStubDecs uniqueId haddockConfig moduleName sizeofs info ty runnerNameSpec
...
+    -- since @struct anonName@ \/ @enum anonName@ would be invalid C.
+    cStubType :: C.Type Final
+    cStubType = case ty of
+        C.TypeRef ref
+          | ref.cName.isAnon
+          -> C.TypePointers 1 C.TypeVoid
+        C.TypeEnum ref
+          | ref.name.cName.isAnon
+          -> C.TypePointers 1 ref.underlying
+        _ -> stubType
```
edsko, line 992: Nice, glad to see this was so simple!

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclIndex.hs
```diff
@@ -328,6 +328,8 @@ fromParseResults results = flip execState empty $ mapM_ aux results
         case (a, b) of
           (C.DeclMacro macroA, C.DeclMacro macroB) ->
             sameMacro macroA macroB
+          (C.DeclGlobal tyA, C.DeclGlobal tyB) ->
+            tyA == tyB
```
edsko: I think this was probably a remnant from the isExtern thing, this change can probably go?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/AssignAnonIds.hs
```diff
@@ -176,7 +176,7 @@ instance UpdateUseSites C.DeclKind where
       C.DeclEnum     x         -> C.DeclEnum     <$> updateUseSites x
       C.DeclAnonEnumConstant x -> C.DeclAnonEnumConstant <$> updateUseSites x
       C.DeclFunction x         -> C.DeclFunction <$> updateUseSites x
-      C.DeclGlobal   x         -> C.DeclGlobal   <$> updateUseSites x
+      C.DeclGlobal       x     -> C.DeclGlobal     <$> updateUseSites x
```
edsko: Unnecessary change

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames.hs
```diff
@@ -517,7 +517,7 @@ instance MangleInDecl C.DeclKind where
       C.DeclAnonEnumConstant x -> C.DeclAnonEnumConstant <$> mangleInDecl info x
       C.DeclFunction x         -> C.DeclFunction         <$> mangleInDecl info x
       C.DeclMacro    x         -> C.DeclMacro            <$> mangleInDecl info x
-      C.DeclGlobal   x         -> C.DeclGlobal           <$> mangle            x
+      C.DeclGlobal       x     -> C.DeclGlobal           <$> mangle            x
```
edsko: Same

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -724,34 +707,34 @@ varDecl info = \curr -> do
...
+                  (map parseSucceed (anonDecls ++ otherDecls) ++) $
+                    singleton $ parseSucceedWith msgs (mkDecl $ C.DeclGlobal typ)
+                VarConst IsExtern
+                  | not (null anonDecls) ->
+                    [ parseFail info.id info.loc ParseUnexpectedAnonInExtern ]
+                VarConst _ ->
+                  (map parseSucceed (anonDecls ++ otherDecls) ++) $
+                    singleton $ parseSucceedWith msgs (mkDecl $ C.DeclGlobal typ)
+                VarThreadLocal ->
+                  map parseSucceed (anonDecls ++ otherDecls)
```
edsko: I'm inclined to suggest that for VarThreadLocal and VarUnsupported we also omit any kind of nested declarations.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -724,34 +707,34 @@ varDecl info = \curr -> do
...
-                    ParseUnknownStorageClass storage
+              potentialDuplicate = [
+                  ParsePotentialDuplicateSymbol (visibility == PublicVisibility)
+                | isDefn && linkage == ExternalLinkage
                 ]
+              msgs = nonPublicVisibility ++ potentialDuplicate
+           in case cls of
+                VarGlobal IsExtern
+                  | not (null anonDecls) ->
+                    [ parseFail info.id info.loc ParseUnexpectedAnonInExtern ]
```
edsko, line 739: Looking at this I'm wondering if the `VarConst` vs `VarGlobal` still serves a purpose?

bolt12: Try combining/merging these two constructors.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
bolt12: Return a singleton with a parse failure

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
bolt12: Worked!

edsko: Nice :)

PR #1618, opened 2026-01-28
Standard library external binding specifications import from LibC

dschrempf:
I leave the re-export of `LibC` in `Prelude` for now, because otherwise, TH complains.


dschrempf:
I am closing this PR, as this approach does not work.

`mkName` fails when the module names don't match exactly (even when the definitions of the symbols are the same).

edsko:
Perhaps you closed this a bit too soon; let's first figure out exactly what goes wrong.

dschrempf:
Reopening; we accept that we have to import `HsBindgen.Runtime.LibC` when standard library external binding specifications are used.

dschrempf:
I reopened but that doesn't work, haha, funny.

dschrempf:
Trying again.

Issue #1502, opened 2026-01-09
Restructure `hs-bindgen-runtime`


dschrempf:
We need to restructure the `HsBindgen.Runtime.*` modules to improve user experience.

In particular, we have to clarify
- which parts are user-facing (if any);
- which parts are binding-facing;
- why do we have a `Prelude`, and what is in the `Prelude`;
- how we structure/cut the modules.

#### Goals (if possible)

- Generating bindings in Template Haskell mode should only require the import of one module (e.g., `HsBindgen.Runtime.Prelude`).
- Using the generated bindings may require import of appropriate modules, but we strive to provide a convenience module re-exporting all required definitions. There are, however, definitions that need to be imported qualified (e.g., `FLAM` module).
- Users should only need to import the `hs-bindgen-runtime` package.

#### Misc

`HsBindgen.Runtime.LibC`:
- [x] Rename `LibC` to `LibC.Auxiliary`
- [x] In `LibC`: Re-export by name all standard library binding specifications definitions defined in in base and `LibC.Auxiliary`
- [x] Re-export `LibC` by module from `HsBindgen.Runtime.Prelude`

#### Ideas

- We could also split the module structure into user-facing, and internal (i.e., binding-facing) parts.

#### Candidates for user-facing modules
- CEnum interface
- HasCField interface

edsko:
This we should definitely do prior to `release-0.1`, but not necessarily for `release-0.1-alpha1`.

dschrempf:
I am looking into this.

To me it is not entirely clear how we want to cut the runtime modules.

We have `HsBindgen.Runtime.Prelude`, which exports many declarations required by
the generated bindings. Therefore, we call the `hs-bindgen` runtime prelude
_binding-facing_. The `hs-bindgen` runtime prelude is also required when using
`hs-bindgen` in Template Haskell mode.

Then, we have many modules exposing functionality that is required when _using_
the bindings; we call these _user-facing_. For example, in the modules we
discuss in the manual, we import the following runtime modules:

```haskell
import HsBindgen.Runtime.CEnum
import HsBindgen.Runtime.ConstPtr
import HsBindgen.Runtime.ConstantArray
import HsBindgen.Runtime.FlexibleArrayMember
import HsBindgen.Runtime.FunPtr
import HsBindgen.Runtime.HasCField
import HsBindgen.Runtime.IncompleteArray
import HsBindgen.Runtime.Marshal
import HsBindgen.Runtime.Ptr
```

[...]

PR #1593, opened 2026-01-26
Make array types mutable

jorisdral:
Resolves #1536

PR #1628, opened 2026-01-29
Runtime: User-facing prelude

dschrempf:


dschrempf:
Ping joris on opaque re-export

edsko:
Note to self: we cannot close #1502 until this PR has been updated (or a separate PR) for both `PtrConst` (https://github.com/well-typed/hs-bindgen/pull/1563) and the updated arrays (https://github.com/well-typed/hs-bindgen/pull/1593).

@jorisdral FYI, most of the user-facing modules are now intended to be used qualified (I think this works really well with how to set up the new `PtrConst` module).

dschrempf:
And now I know what I wanted to talk about with @jorisdral ;-).

We are re-exporting the opaque IncompleteArray and ConstantArray types. (But both of you said these modules will change significantly, so maybe this is not a concern).

edsko (approved):

edsko:
@jorisdral I agreed with @dschrempf that he'd merge this once CI goes green. Your array stuff will need rebasing, but I think the conflcits shouldn't be too bad, and hopefully the new Prelude structure is now very clear so slotting in the array stuff into it shouldn't be too hard. Let me know if anything is unclear.

PR #1630, opened 2026-01-29
Introduce cross compilation example

### LINKED ISSUE #738: Experiment with and document the cross-compilation workflow
We should do this in both preprocessor and TH mode. 

Related: 

* #342 (the original issue is based on incorrect assumptions, but there is still some valuable pointers in there)
* #732 
* #734

bolt12:
This closes #738


examples/cross-compilation/c-src/Makefile
bolt12, line 1: Use clang instead of gcc (should be a cross compiling version)

examples/cross-compilation/generate-and-run.sh
```diff
@@ -0,0 +1,433 @@
+#!/usr/bin/env bash
+#
+# generate-and-run.sh - Cross-compilation workflow for hs-bindgen
+#
+# Demonstrates the full cross-compilation pipeline:
+#   Phase 1: Build C libraries (native + cross-compiled)
+#   Phase 2: Generate Haskell bindings (hs-bindgen --target=<triple>)
+#   Phase 3: Build and run native Haskell executable
+#   Phase 4: Cross-compile and run under QEMU (iserv + cross-GHC)
```
bolt12: Simplify to just run one example (arm64)

examples/cross-compilation/generate-and-run.sh
```diff
@@ -0,0 +1,433 @@
...
+#
+# How it works:
+#   1. GHC encounters a TH splice
+#   2. GHC sends the splice to iserv via pipe (using the GHCi.Message protocol)
+#   3. iserv (running under QEMU) evaluates the splice
+#   4. iserv sends the result back to GHC via pipe
+#   5. GHC incorporates the result into the compiled output
+#
+# Why build from source: Nix's cross-GHC (pkgsCross.*.buildPackages.ghc) does
+# not ship an iserv binary. However, the ghci package is available in the
```
bolt12: If we build GHC from source we get _build/stage1/bin/ghc-iserv-dyn binary, so maybe it is on the nix derivation build path after all

examples/cross-compilation/generate-and-run.sh
```diff
@@ -0,0 +1,433 @@
...
+#   module Main (main) where
+#   import GHCi.Server (defaultServer)
+#   main :: IO ()
+#   main = defaultServer
+#
+# We compile with two additional GHC flags to configure the RTS:
+#
+#   -fkeep-cafs: Prevents the RTS from garbage-collecting CAFs (Constant
+#     Applicative Forms -- top-level thunks). Without this flag, the RTS would
+#     GC those results, causing crashes from dangling pointers.
```
bolt12: This is weird and should look more deeply into this.

bolt12: better explanation, ideally from the GHC manual, something we can trust :D

examples/cross-compilation/generate-and-run.sh
```diff
@@ -0,0 +1,433 @@
...
+  extra-lib-dirs: $LIB_DIR
+
+package *
+  ghc-options: -fexternal-interpreter -pgmi $iserv_wrapper
+CABAL_EOF
+
+    if "$CABAL_PATH" build "$EXE_NAME" \
+        --with-compiler="$GHC_PATH" \
+        --with-ghc-pkg="$ghc_pkg_path" \
+        ${hsc2hs_path:+--with-hsc2hs="$hsc2hs_path"} \
```
bolt12: We probably do not need this, but we could add an hsc file to our example to 0demonstrate that if we had hsc files how we would do it (i.e. pass hsc2hs path)

examples/cross-compilation/generate-and-run.sh
bolt12, line 1: Simplify this script to just run under aarch64.

manual/LowLevel/Usage/08-CrossCompilation.md
```diff
@@ -1,3 +1,439 @@
...
+  foo.h
+```
+
+> [!NOTE]
+> When cross-compiling on NixOS/Nix, the `BINDGEN_EXTRA_CLANG_ARGS` environment
+> variable set by the Nix dev shell contains host-specific include paths that
+> conflict with the target. Similarly, `BINDGEN_BUILTIN_INCLUDE_DIR` controls
+> where `hs-bindgen` looks for Clang's built-in headers (like `stddef.h`); the
+> Nix dev shell sets it to prevent heuristic searching, but those paths are
+> host-specific. Clear both when cross-compiling.
```
edsko: This is specific to the nix flake we provide i guess? we should clarify that.

manual/LowLevel/Usage/08-CrossCompilation.md
```diff
@@ -1,3 +1,439 @@
...
+On other systems, you can:
+
+1. **Use a cross-compilation toolchain**: Install a cross-compiler (e.g.,
+   `aarch64-linux-gnu-gcc`) which typically includes target headers.
+
+2. **Download target sysroot**: Obtain a sysroot containing headers for the
+   target platform.
+
+3. **Use Docker/containers**: Run `hs-bindgen` in a container for the target
+   architecture.
```
edsko: Not really cross compiling in that case I guess..?

manual/LowLevel/Usage/08-CrossCompilation.md
```diff
@@ -1,3 +1,439 @@
...
+> ```cabal
+> package *
+>   ghc-options: -fexternal-interpreter -pgmi /path/to/iserv-wrapper.sh
+> ```
+
+#### Why build iserv from source
+
+Nix's cross-GHC (`pkgsCross.*.buildPackages.ghc`) does not ship an iserv binary
+(`ghc-iserv` or `ghc-iserv-dyn`). However, the `ghci` package is available in
+the cross-GHC's package database and provides
```
edsko: This section might need updating.

manual/LowLevel/Usage/08-CrossCompilation.md
```diff
@@ -1,3 +1,439 @@
...
+main :: IO ()
+main = defaultServer
+```
+
+We compile with two additional GHC flags to configure the RTS:
+
+- **`-fkeep-cafs`**: prevents the RTS from garbage-collecting CAFs (Constant
+  Applicative Forms -- top-level thunks like `x = expensiveComputation`) after
+  evaluation. Without this flag, the RTS would GC those results, causing
+  crashes from dangling pointers.
```
edsko: If we figure this out for the docs of the script, we should update it here also.

manual/LowLevel/Usage/08-CrossCompilation.md
```diff
@@ -1,3 +1,439 @@
...
+
```
edsko: Perhaps better yet: perhaps we should avoid the duplication between the manual and the script, and just refer to the manual where possible?

manual/LowLevel/Usage/08-CrossCompilation.md
```diff
@@ -1,3 +1,439 @@
...
+### hsc2hs and cross-compilation
+
+[`hsc2hs`](https://downloads.haskell.org/ghc/latest/docs/users_guide/utils.html#writing-haskell-interfaces-to-c-code-hsc2hs)
+is a GHC preprocessor that processes `.hsc` files into `.hs` files.  It handles
+directives like `#{size struct foo}`, `#{alignment struct foo}`, and
+`#{peek struct foo, bar}` by compiling a small C program that prints the values,
+running it, and substituting the results into the Haskell source.
+
+`hs-bindgen-runtime` uses `.hsc` files, so `hsc2hs` is needed during the build.
```
edsko, line 360: ah, here is why we need hsc2hs.

examples/cross-compilation/generate-and-run.sh
bolt12: It is not

bolt12:
Closed in favour of #1243 which already cherry picks this PR's commits.

PR #1653, opened 2026-02-05
Minor cleanup in preparation for the alpha release

### LINKED ISSUE #1648: Unnecessary "unique identifier too long" notice
We used to include these identifiers as-is in the names of our generated wrappers, but now we merely use it to produce a hash, so this warning is no longer needed.

edsko:
* Drop unnecessary "unique ID too long" notice
  
  This was necessary previously, because we included that ID as-is in our C identifiers; however, now it is only used to determine the hash.

* Improve `hs-bindgen-runtime` package description

* Bump `libclang` bindings to https://github.com/well-typed/libclang/pull/27

* Bump `index-state`

* Support ghc-9.14, relax bounds

* Minor improvements to literate mode

Closes #1648


cabal.project.base
```diff
@@ -1,8 +1,49 @@
...
+allow-newer: vec-0.5.1:base
+
+--
+-- Temporary package bounds overrides for QuickCheck 2.17
+--
+
+allow-newer: aeson-2.2.3.0:QuickCheck
+allow-newer: fin-0.3.2:QuickCheck
+allow-newer: skew-list-0.1:QuickCheck
+allow-newer: vec-0.5.1:QuickCheck
```
jorisdral, line 49: We can make this conditional on the GHC version

```
if(impl(ghc >=9.14))
  allow-newer: aeson
```

c-expr-dsl/c-expr-dsl.cabal
```diff
@@ -8,11 +8,18 @@ maintainer:      Well-Typed LLP
...
-tested-with:     GHC ==9.2 || ==9.4 || ==9.6 || ==9.8 || ==9.10 || ==9.12
 synopsis:        DSL for the language support by c-expr-runtime
+tested-with:
+  GHC ==9.2.8
+   || ==9.4.8
+   || ==9.6.7
+   || ==9.8.4
+   || ==9.10.3
+   || ==9.12.2
+   || ==9.14.1
```
jorisdral, line 19: I typically don't bother with the minor version, mainly because it's more likely that someone forgets to update this field than it is that the minor version affects anything (but I could be wrong)

edsko: Hm, I can see where you are coming from, but I _think_ I still prefer the full version :) But I could be convinced here I think.

cabal.project.base
edsko: https://github.com/well-typed/hs-bindgen/issues/1665

PR #1661, opened 2026-02-05
Bump libclang-bindings commit in nix

bolt12:


dschrempf:
For reference, there is a script that updates the libclang revision in both, `cabal.project.base` and `flake.nix`.

PR #1663, opened 2026-02-05
Avoid frontend binding spec errors

TravisCardwell:
This PR downgrades some errors to warnings:

* `ResolveBindingSpecsModuleMismatch` messages are traced when a prescriptive binding specification does not match the module that is being generated.  This is unlikely to result in problems.

* `ResolveBindingSpecsNoHsTypeSpec` messages are traced when an external binding specification is missing Haskell type specifications for a type.  This is usually an issue because we need the instances for the type.  Types with no instances should still specify the Haskell type, even if there are no instances.

* `ResolveBindingSpecsTypeNotUsed` messages are traced when a prescriptive binding specification specifies a type that is not used.  This is useful to quickly discover mistakes/typos in type names.

* The `ResolveBindingSpecsExtHsRefNoIdentifier` trace is for a check that we must do in the `ResolveBindingSpecs` pass.  We cannot create an external binding without a Haskell identifier.  We could check for this case in boot and trace an error, but I think we should just downgrade the trace to a warning.

This PR also removes `ResolveBindingSpecsNoHsTypeRep`:

* This error was needed when we included the `FFIType` in binding specifications.  Now that we no longer do that, this error can be removed.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ResolveBindingSpecs/IsPass.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ResolveBindingSpecs/IsPass.hs
@@ -149,7 +149,7 @@ instance PrettyForTrace ResolveBindingSpecsMsg where
 instance IsTrace Level ResolveBindingSpecsMsg where
   getDefaultLogLevel = \case
     ResolveBindingSpecsModuleMismatch{}       -> Warning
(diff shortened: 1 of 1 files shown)


edsko:
Regarding `ResolveBindingSpecsExtHsRefNoIdentifier`, it means that if we have a external binding spec with an unusable entry, we warn about that entry and then move on without it (just like we throw out unusable C declarations).

Issue #1668, opened 2026-02-06
Refactor instance generation for more guidance


edsko:
I tried adding support for `Generic`; the first step was easy enough, adding the appropriate entry to `SupportedInstances`:

```hs
struct = Map.fromList [
    ..
  , mkDef Generic Independent Stock []
  ]
```

However, it was far from clear after that how to proceed. The code simply compiled, but no `Generic` instance was produced. It's not terribly surprising that this didn't work without further action, but it would be much better if adding an instance like this would result in a panic somewhere; "don't know to to generate `Generic`". There are declarations such as `optDecls` in `Structure.hs`, but that only gets passed "candidate" instances, and apparently only `Dependent` instances are candidates (why?).

When we move instance resolution to the frontend, we should think about how we can make this process better.

TravisCardwell:
This issue is essentially a sub-task of #1447.  The implementation in the frontend will be *much* better than the current implementation in the backend, which mixes instance resolution with translation.

`Dependent` is used when an instance depends on instances of dependencies.  When the strategy is `Newtype` or `Stock`, just adding an entry to `SupportedInstances` is enough to make instance resolution consider the instance and add a deriving declaration when appropriate.  We also have to maintain the lists of instances provided by the primitive and `stdlib` types, though.  As I mentioned before, the current design makes maintaining this difficult, but it will be considerably improved with #1447.

`Independent` is used when an instance does *not* depend on instances of dependencies.  You may note that none of the supported instances marked as `Independent` have a default strategy that can be automatically derived.  They all require specific logic to possibly inspect the declaration to determine if an instance should be generated/derived.  Most of these generate instance declarations, not just deriving declarations.  These are currently implemented during translation.  With #1447, we will be able to resolve which instances should be generated/derived separately.  That is what the `HsBindgen.Instances.Instance` type is mean to be used for.  (I will likely rename it to `ResolvedInstance`, btw.)  It is currently unused, patiently waiting for our implementation to catch up with our design.

> There are declarations such as `optDecls` in `Structure.hs`, but that only gets passed "candidate" instances, and apparently only `Dependent` instances are candidates (why?).

That `optDecls` is the list of deriving instance declarations computed just based on what instances the dependencies of the type have.  All `Independent` entries, as well as entries with strategy `HsBindgen`, require special logic and therefore cannot be computed in `optDecls`.  To illustrate the need for `Dependence`, consider `Storable` instances for `struct` declarations, which are `Dependent` but cannot be derived.

PR #1669, opened 2026-02-06
Derive `Generic` for all generated types

edsko:
@TravisCardwell This is quite a bad hack (although _somewhat_ justifiable); I couldn't figure out how to do it better in time for the alpha release. I've opened https://github.com/well-typed/hs-bindgen/issues/1668 for us to consider how we could make this easier, but meanwhile you might want to refactor this to make it bit better with the rest.

### DIFF
--- a/hs-bindgen/fixtures/arrays/array/Example.hs
+++ b/hs-bindgen/fixtures/arrays/array/Example.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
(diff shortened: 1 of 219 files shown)


TravisCardwell:
~This hack derives `Generic` instances for all records translated from `struct` declarations?  If a field of the record does not have a `Generic` instance, perhaps the generated code is invalid?  For example, an external binding specification may specify use of a Haskell type that is not generated by `hs-bindgen` and does not have a `Generic` instance.~

Ah, I was not thinking correctly.  Dependencies do not need `Generic` instances to simply derive `Generic`.  It is the recursive use generics in applications of `Generic` that result in the `Generic` errors that are often seen.

PR #1673, opened 2026-02-06
Make sure we can parse `short int`

edsko:
See comments in `FromLanC.hs`.


hs-bindgen/src-internal/HsBindgen/Frontend/LanguageC/PartialAST/FromLanC.hs
```diff
@@ -111,6 +111,14 @@ withSign f = \case
...
         $ (if unknown.isConst then C.TypeQual C.QualConst else id)
         $ C.TypePrim $ f unknown.sign
+    alreadyKnown@(PartialKnown (KnownType _typ)) ->
+      -- In an example such as
+      --
+      -- > short int
+      --
+      -- a fully constructed type will already be present when we see the @int@
+      -- part; in this case, we just stick with the type we have already have.
+      return alreadyKnown
```
edsko, line 121: This isn't quite right: `long double` would now be parsed incorrectly.

PR #1676, opened 2026-02-06
Simplify `HasField` instances by expanding `CFieldType`/`CBitfieldType`

### LINKED ISSUE #1667: HasField instances should expand `CFieldType`
We currently generate instances such as

```hs
instance ( TyEq ty (CFieldType Driver) "version"
         ) => HasField "version" (Ptr Driver) (Ptr ty) where (..)
```

this makes the instance needlessly (I think) difficult to read: the type equality constraint is needed, unfortunately, but it would be nicer if the instance would make it immediately obvious what that `ty` actually expands to, rather than requiring readers to manually do the lookup of `CFieldType` (by looking at the `HasCField` instance).

jorisdral:
Resolves #1667 

Where we were previously generating instance headers like:

```hs
instance ( TyEq ty ((HsBindgen.Runtime.HasCField.CFieldType Triplet) "unwrapTriplet")
         ) => GHC.Records.HasField "unwrapTriplet" (Ptr.Ptr Triplet) (Ptr.Ptr ty) where
```

Which uses the following `HasCField` instance:

```hs
instance HsBindgen.Runtime.HasCField.HasCField Triplet "unwrapTriplet" where
  type CFieldType Triplet "unwrapTriplet" =
    (HsBindgen.Runtime.ConstantArray.ConstantArray 3) FC.CInt
```

We now generate instance headers where `CFieldType` (or `CBitfieldType`) are expanded:

```hs
instance GHC.Records.HasField "unwrapTriplet" (Ptr.Ptr Triplet) (Ptr.Ptr ((HsBindgen.Runtime.ConstantArray.ConstantArray 3) FC.CInt)) where
```

### DIFF
--- a/hs-bindgen/fixtures/arrays/array/Example.hs
+++ b/hs-bindgen/fixtures/arrays/array/Example.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE DerivingVia #-}
-{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE MagicHash #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
@@ -10,7 +9,6 @@
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
-{-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 module Example where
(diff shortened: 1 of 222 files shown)


hs-bindgen/fixtures/arrays/array/Example.hs
```diff
@@ -46,8 +43,7 @@ newtype Triplet = Triplet
     , F.Storable
     )
 
-instance ( TyEq ty ((HsBindgen.Runtime.HasCField.CFieldType Triplet) "unwrapTriplet")
```
jorisdral, line 49: TODO: remove `TyEq` if possible

jorisdral: It looks like I can't remove it because it is also used in macro types

PR #1678, opened 2026-02-09
Reflect `Generic` in generated binding specs

TravisCardwell:
The hack in #1669 derives instances, but it does not reflect those instances in generated binding specifications.  This commit removes the hack and implements (backend) `Generic` instance resolution.

Note that the hack derived `Generic` instances for `union` types, and this implementation does as well.  This is not a problem, but I mention it in case it was unintentional.

There are two kinds of changes to the fixtures:

* `Generic` is now derived along with the other (`stock`) instances:

```diff
-  deriving stock (GHC.Generics.Generic)
-  deriving stock (Eq, Ord, Read, Show)
+  deriving stock (GHC.Generics.Generic, Eq, Ord, Read, Show)
```

* `Generic` instances are now reflected in generated binding specifications.

I confirmed that there are no other changes.


TravisCardwell:
(Rebased)

Issue #1520, opened 2026-01-13
Introduce `_Aux` types for function pointers through other-than-pointer indirection layers


edsko:
Given

```c
typedef void (*fun1)();
typedef void (**fun2)();
typedef void (* const fun3)();
typedef void (*fun4[2])();
```

we generate `_Aux` types for `fun1` and `fun2`, but not for `fun3` or `fun4`: we look specifically for a number of pointers around a function:

https://github.com/well-typed/hs-bindgen/blob/b05894b60ea5f55e455944b36585ebbd6c39150f/hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs#L169-L175

but don't allow for other kinds of indirection.

This is the same bug as we had for `HandleTypedefs`, where we needed a similar generalization (that was done in #1404).

edsko:
Note that C does not allow pointers-to-constant functions; I have no idea why not (https://en.cppreference.com/w/c/language/const.html explicitly mentions that "If a function type is declared with the const type qualifier (through the use of typedef), the behavior is undefined."); I started thing about this ticket because I wondered if we would also miss introducing an `_Aux` type of we had a pointer-to-const-function, but it turns out such a thing cannot be defined!

edsko:
Similar bug: https://github.com/well-typed/hs-bindgen/issues/1445 . That's now three places where we have a similar bug; I wonder if we can do something to avoid these 🤔

jorisdral:
> Similar bug: [#1445](https://github.com/well-typed/hs-bindgen/issues/1445) . That's now three places where we have a similar bug; I wonder if we can do something to avoid these 🤔

I think one of the ingredients for preventing these types of bugs/misses is writing complete case expressions without wildcards for C types, forcing us to consider all cases. 

Helper functions can maybe also do the tedious bits for us. But then it's also the case that the many functions on C types that we have each want slightly different results. It's hard to write a single "fold" over C types that fits all use cases

PR #1684, opened 2026-02-09
Avoid generating unnecessary ToFunPtr and FromFunPtr instances

### LINKED ISSUE #1650: Avoid generating unnecessary `FromFunPtr`/`ToFunPtr` instances
Given

```c
struct Driver;
typedef int (*RunDriver)(struct Driver* self);

struct Driver {
  RunDriver run;
};
```

we currently generate

```hs
newtype RunDriver_Aux = RunDriver_Aux
  { unwrapRunDriver_Aux :: (Ptr.Ptr Driver) -> IO FC.CInt
  }

newtype RunDriver = RunDriver (FunPtr RunDriver Aux))

instance HsBindgen.Runtime.FunPtr.FromFunPtr RunDriver_Aux where
  fromFunPtr = ..

instance HsBindgen.Runtime.FunPtr.FromFunPtr ((Ptr.Ptr Driver) -> IO FC.CInt) where
  fromFunPtr = ...
```

The first `FromFunPtr` instance here comes from the instances we provide for typedefs around function pointers, the second from scanning all uses of function types everywhere. However, that "everywhere" should _skip_ typedefs around function pointers, otherwise we're essentially dealing with the same thing twice. 

This isn't a huge deal, because if users use the wrong instance, they will get a `FunPtr (Ptr Driver -> IO CInt)` rather than a `FunPtr RunDriver_Aux`, which is not actually usable in any API call, so the "extra" instance is unlikely to cause trouble.

bolt12:
This closes #1650

### DIFF
--- a/hs-bindgen/fixtures/functions/typedef_funptr/Example.hs
+++ b/hs-bindgen/fixtures/functions/typedef_funptr/Example.hs
@@ -3,7 +3,6 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE DerivingVia #-}
-{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE MagicHash #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
@@ -11,7 +10,6 @@
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
-{-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 module Example where
@@ -29,7 +27,6 @@ import qualified HsBindgen.Runtime.Internal.HasFFIType
 import qualified HsBindgen.Runtime.Marshal
 import qualified Prelude as P
 import Data.Void (Void)
-import HsBindgen.Runtime.Internal.TypeEquality (TyEq)
 import Prelude ((<*>), Eq, IO, Int, Ord, Show, pure)
 
 {-| Auxiliary type used by 'RunDriver'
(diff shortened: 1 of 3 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -137,9 +139,14 @@ scanAllFunctionPointerTypes = foldMap $ \decl ->
...
-      -- a comment why we do this.
-      C.TypeTypedef (C.Ref _ t) -> scanTypeForFunctionPointers t
+
+      -- Don't scan through typedef function pointers; these are handled by
+      -- typedefFunPtrDecs (see generateDecs). Scanning would create duplicate
+      -- ToFunPtr/FromFunPtr instances. We still scan other typedefs to find
+      -- nested function pointers.
+      C.TypeTypedef (C.Ref _ t) -> case t of
+        C.TypePointers _ (C.TypeFun _ _) -> Set.empty
+        _                                -> scanTypeForFunctionPointers t
```
bolt12, line 150: Check this issue: https://github.com/well-typed/hs-bindgen/issues/1520, when we address it we should make sure we update this code. Add a reference (comment) here mentioning the issue.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -137,9 +139,14 @@ scanAllFunctionPointerTypes = foldMap $ \decl ->
...
       C.TypeBlock t             -> scanTypeForFunctionPointers t
       C.TypeQual _ t            -> scanTypeForFunctionPointers t
-      -- TODO: do we have to scane the underlying type too? If so, elaborate in
-      -- a comment why we do this.
-      C.TypeTypedef (C.Ref _ t) -> scanTypeForFunctionPointers t
+
+      -- Don't scan through typedef function pointers; these are handled by
+      -- typedefFunPtrDecs (see generateDecs). Scanning would create duplicate
+      -- ToFunPtr/FromFunPtr instances. We still scan other typedefs to find
+      -- nested function pointers.
```
bolt12, line 146:
```suggestion
      -- nested function pointers.
      -- When #1520 is addressed this could should be updated.
```

PR #1692, opened 2026-02-11
Document how to use hs-bindgen wo shared libraries

### LINKED ISSUE #873: Document how to use `hs-bindgen` without shared libraries
We should probably also document how to use `hs-bindgen` without shared libraries.  I think that it is a common case for users to want to compile in their own project C code, so that the only shared libraries used are those of dependencies.  Perhaps this should be done in a separate PR, however, which adds such an example.

Perhaps it would be nice to be able to build the manual with shared libraries or without them.  The `.cabal` files differ, but we could use a flag.  In my quick test, I simply created symbolic links to the `c` directory, but we could refactor the manual so that C source is stored within the relevant Haskell package directory.

<details><summary><code>ldd run-manual</code></summary>

```
  linux-vdso.so.1 (0x00007f535151d000)
  libm.so.6 => /usr/lib/libm.so.6 (0x00007f53513eb000)
  libgmp.so.10 => /usr/lib/libgmp.so.10 (0x00007f5351345000)
  libc.so.6 => /usr/lib/libc.so.6 (0x00007f5351155000)
  /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f535151f000)
```

</details>

If we refactor the manual so that C source is stored within the relevant Haskell package directory, the `extra-include-dirs` and `extra-lib-dirs` configuration in `cabal.project.local` files would also be come unnecessary.  A drawback is that more paths would need to be added to `LD_LIBRARY_PATH` (Linux), `DYLD_LIBRARY_PATH` (macOS), and `PATH` (Windows) to execute the resulting executable as we currently do it.  Alternatively, we could collect the executable and shared libraries into a build directory.  We would not need to set `PATH` at all on Windows.  I am not sure about macOS.  We would still need `LD_LIBRARY_PATH` in Linux:

```
$ env LD_LIBRARY_PATH=build build/run-manual
```

_Originally posted by @TravisCardwell in https://github.com/well-typed/hs-bindgen/issues/869#issuecomment-3042949166_

bolt12:
Closes #873

### DIFF
--- a/manual/LowLevel/Usage/01-Invocation.md
+++ b/manual/LowLevel/Usage/01-Invocation.md
@@ -599,7 +599,7 @@ MacOS.
     buildable: false
   ```
 
-## Using `hs-bindgen` with bundled C source files (without shared libraries)
+## Using `hs-bindgen` with bundled C source files
 
 All examples in the preceding sections assume that the C library you are
 binding to is built separately and linked as a shared library. However, when


manual/LowLevel/Usage/01-Invocation.md
```diff
@@ -598,3 +598,60 @@ MacOS.
   if !(os(linux) && arch(x86_64))
     buildable: false
   ```
+
+## Using `hs-bindgen` with bundled C source files (without shared libraries)
```
bolt12:
```suggestion
## Using `hs-bindgen` with bundled C source files
```

Issue #1689, opened 2026-02-10
The standard libary external binding specifications are platform-specific


dschrempf:
In particular, the definitions

```
            , ("int_least16_t",  "Int16")
            , ("uint_fast16_t",  "Word16")

            , ("fexcept_t", "CFexceptT")
```

mismatch the actual C types used in our `musl-include` headers:

```c
typedef signed int      int32_t;
typedef int32_t int_fast16_t;
typedef unsigned int    uint32_t;
typedef uint32_t uint_fast16_t;

typedef unsigned short fexcept_t;
```

To obtain platform-independent standartd library external binding specifications, we need indirections similar to `CSize` in `base`:

```haskell
newtype CSize = CSize WordXX -- platform-specific; requires external tooling (autoconf?)!
```

```haskell
newtype IntFast16T = InstFast16T IntXX -- platform-specific (see above)
newtype UIntFast16T = UIntFast16T IntXX -- platform-specific (see above)
```

(In theory, we need this for all types of variable size that we have in our standard library external binding specifications; it is just that we observe mismatches in `musl-include` for the three types mentioned above.)

When solving this ticket, please also check the TODO items in the golden test `standard_library_external_binding_specs.h`.

edsko:
@TravisCardwell Let's discuss this. I feel like the best way forward might be to simply disable these types from our standard binding spec for release 0.1, and then introduce the necessary CPP for 0.2. But if you think it's not too much work, then we can also do the CPP now already.

edsko:
Marking this as a high priority bug because we're generating code that does not compile.

jorisdral:
FYI we'd need an `autoconf` script to determine the sizes of C types. This is what `ghc-internal` also does for types like `size_t`/`CSize`. See https://gitlab.haskell.org/ghc/ghc/-/blob/master/libraries/ghc-internal/configure.ac?ref_type=heads#L261

TravisCardwell:
This is my mistake.  Sorry for overlooking that!

Thankfully, we do not need to use `configure` or CPP for this because we are already using `hsc2hs`.  I implemented the new types in #1695.

TravisCardwell:
Oh no...  Implementing this using `hsc2hs` (or `configure` and CPP) does not work.

The underlying types differ in different standard library implementations.  In other words, these types are not just platform-dependent, they are libc-dependent.  The standard library used is a runtime choice (by the user), so **we cannot implement such types in `hs-bindgen-runtime`**.

[...]

PR #1695, opened 2026-02-11
Remove `C[U]Int(Le|F)ast` types

TravisCardwell:
~~This PR adds Haskell types corresponding to `int_least8_t` ..., `uint_least_8_t` ..., `int_fast8_t` ..., and `uint_fast8_t` ... types to `LibC` so that they use the correct underlying types per platform.~~

That did not work.  See https://github.com/well-typed/hs-bindgen/issues/1689#issuecomment-3887853443

This PR now removes those types from `stdlib`.

### DIFF
--- a/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden.hs
+++ b/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden.hs
@@ -796,6 +796,8 @@ noHandleMacrosTraces = multiTracePredicate ([] :: [String]) (\case
 -- This test sets the parse predicate to parse all declarations, which results
 -- in use of 'HsBindgen.Runtime.LibC.CBool'.  This test also works without
 -- setting the parse predicate, but it results in use of 'FC.CBool' instead.
+-- This configuration should be removed when
+-- https://github.com/well-typed/hs-bindgen/issues/1627 is resolved.
 test_bindingSpecs_stdlib_instances :: TestCase
 test_bindingSpecs_stdlib_instances =
     defaultTest "binding-specs/stdlib/instances"
@@ -806,6 +808,8 @@ test_bindingSpecs_stdlib_instances =
 -- This test sets the parse predicate to parse all declarations, which results
 -- in use of 'HsBindgen.Runtime.LibC.CBool'.  This test also works without
 -- setting the parse predicate, but it results in use of 'FC.CBool' instead.
+-- This configuration should be removed when
+-- https://github.com/well-typed/hs-bindgen/issues/1627 is resolved.
 test_bindingSpecs_stdlib_return_values :: TestCase
 test_bindingSpecs_stdlib_return_values =
     defaultTest "binding-specs/stdlib/return_values"


TravisCardwell:
What's up with [this](https://github.com/well-typed/hs-bindgen/blob/311d5d8d2e42195fedfec601c4b18d06cf556765/hs-bindgen/examples/golden/binding-specs/standard_library_external_binding_specs.h#L3)?

```c
/* LAST UPDATE: February 10, 2026; revision 73a232ec01989e9ac38549aaf3829e7544f2085e */
```

It reminds me of CVS/Subversion days!  These days, such metadata is managed by the RCS, Git in our case.  Do we need this?  Am I missing something?

Perhaps generating and including such information in *generated* files could be helpful, but this is not a generated file.

TravisCardwell:
The new `binding-specs/standard_library_external_binding_specs` test does not use naming consistent with existing tests, and it is not next to the existing standard library instances test, so I am refactoring a bit.

I think that such tests better fit under `types`, but I do not really care, so I am moving my test under `binding-specs` for consistency.  We should use directories for organization, not prefixes of header filenames, however.  Here are the new names:

* `binding-specs/stdlib/instances`
* `binding-specs/stdlib/return_values`

In doing this, I noticed some unfortunate behavior.  These tests should not need custom selection predicates, because external binding specifications may be used with types that are not parsed.  The new test explicitly parses all declarations, however:

```haskell
    defaultTest "binding-specs/stdlib/return_values"
      & #onFrontend .~ (\cfg -> cfg
          & #parsePredicate .~ BTrue
          )
```

With that custom selection predicate, `HsBindgen.Runtime.LibC.CBool` is used.  Without it, `FC.CBool` is used.  Perhaps we should change our handling of (non-stdlib) booleans to (also) use `HsBindgen.Runtime.LibC.CBool` (instead of `FC.CBool` directly)?

edsko (approved):

dschrempf:
Hi @TravisCardwell thanks for looking into this.

With respect to the test location, I just chose the folder that sounded like it was supposed to contain tests about binding specifications. Thanks for creating a sub-folder, that makes things easier.

I don't exactly remember now why I had to add the parse-all parse predicate. I think it was unrelated to the `CBool` issue. Also, you mentione selection predicates in your post, but there are no custom selection predicates involved in this test. Could you clarify?

The include statement `#include <stdbool.h>` changed `FC.CBool` to the type we define in our runtime.

edsko:
https://github.com/well-typed/hs-bindgen/issues/1696 for the parse predicate strangeness.

TravisCardwell:
> I don't exactly remember now why I had to add the parse-all parse predicate. I think it was unrelated to the `CBool` issue. Also, you mentione selection predicates in your post, but there are no custom selection predicates involved in this test. Could you clarify?

When using the following, we import `CBool` from `hs-bindgen-runtime`:

```haskell
defaultTest "binding-specs/stdlib/return_values"
  & #onFrontend .~ (\cfg -> cfg
      & #parsePredicate .~ BTrue
      )
```

Without the parse predicate, we import `CBool` directly from `base`:

```haskell
defaultTest "binding-specs/stdlib/return_values"
```

Indeed, use of `#include <stdbool.h>` is also a factor, as is the C standard.  I think we should always import types from `hs-bindgen-runtime` when they are available to avoid such insignificant changes of imports.

PR #1704, opened 2026-02-13
Add `FunctionArg` to C AST

jorisdral:
Function arguments used to be represented by tuples, but this is somewhat unwieldy, so we add a new datatype for function arguments instead.


jorisdral:
I'm going to merge this because it is only a minor refactoring that I think is uncontroversial

PR #1707, opened 2026-02-16
Add post-qualified-imports flag

### LINKED ISSUE #1641: Support post-qualified imports in generated code
We're currently generating prepositive imports

```hs
import qualified Data.Proxy
```

In projects where `-Wprepositive-qualified-module` is enabled this will lead to ghc warnings (it is useful to enable this warning, if for no reason other than to have HLS generated post-qualified imports). We should have a command line flag to switch to generating post-qualified imports instead.

bolt12:
This closes #1641


hs-bindgen/src-internal/HsBindgen/Config/Prelims.hs
```diff
@@ -78,6 +84,41 @@ data FieldNamingStrategy =
...
+{-------------------------------------------------------------------------------
+  Module render config
+-------------------------------------------------------------------------------}
+
+-- | Configuration for rendering Haskell modules.
+--
+-- Bundles settings that affect the surface syntax of generated modules but
+-- do not change semantics.
+data ModuleRenderConfig = ModuleRenderConfig {
+      fieldNamingStrategy :: FieldNamingStrategy
```
bolt12: Maybe this field should not be here, but keep the ModuleRenderConfig type

bolt12:
@dschrempf would you be so kind as to take a look at this PR since it adds a new configuration flag?

dschrempf (commented):
Hi Armando! I was going through the changes, and I think we have an inconsistent state (or design) here. I think part of the changes already happened in earlier PRs.

In detail, we provide configuration options twice:
- to the `hsBindgen` call, and
- when registering artifacts (e.g., `getBindings`).

This is apparent, for example, in `Preprocess.hs`. The "field naming strategy" is provided as part of the `BindgenConfig` (`FrontendConfig` _and_ `BackendConfig` (why both?, so in total we actually provide it three times)) to the function `hsBindgen`, and to the artifact `writeBindingsSingleToDir`.

```haskell
hsBindgen :: ... -> BindgenConfig -> ... -> Artefact a -> IO a
writeBindingsSingleToDir :: FieldNamingStrategy -> ... -> Artefact ()
```

This seems wrong.

I could see this is correct if we say that users might want to call `writeBindingsSingleDir` with different `FieldNamingStrategy`ies _in one `hsBindgen` call_.

Instead, configuration options provided to `hsBindgen` that are also required by artifacts should be available as `BootArtefact`, `FrontendArtefact`, or `BackendArtefact`.

In fact, I see that `fieldNamingStrategy` is a record field of `BackendArtefact`, so the state is even more confusing.

The new `ModuleRenderConfig` is following this inconsistent pattern. It is part of `BackendConfig`, and also provided when registering artifacts. In this case, I could imagine that one wants to run `hsBindgen` once, but creates modules with different `ModuleRenderConfig`s.

I am not saying we should go this or that route, my point is we should be consistent! I also may have missed details that you thought about and that influenced your design choice. Let me know!

dschrempf:
(See new issue: https://github.com/well-typed/hs-bindgen/issues/1750), feel free to merge this PR! Thanks @bolt12.

scripts/ci/compile-fixtures.sh
```diff
@@ -96,7 +96,7 @@ set_known_werror_unclean
 #
 # This number is used for sanity checks. Make sure to update this number when
 # new fixtures are added or old ones are removed.
-KNOWN_FIXTURES_COUNT=159
+KNOWN_FIXTURES_COUNT=160
```
bolt12:
```suggestion
KNOWN_FIXTURES_COUNT=161
```

PR #1709, opened 2026-02-16
Add Non-Portability manual section and example

### LINKED ISSUE #893: Manual: Discuss that generated binding specs may be system dependent
When we run `hs-bingden` on

```c
#ifdef ..
..
#else
..
#endif
```

then the generated binding spec may look _entirely_ different depending on how that CPP gets resolved. Worse, when generating bindings against some libraries, the headers of those libraries may have been generated by `configure` scripts, with nothing in the header itself indicating that anything is specific to the local machine. It is therefore impossible (in general anyway) for `hs-bindgen` to _know_ when this happens.

We should add a section to the manual that highlights this, gives an examples, and suggests ways of dealing with it.

Related to #860.

bolt12:
This closes #893


edsko (approved):
Very nice! Manual section looks great, and good that we now have a nice example of that CPP flag mismatch.

Issue #1447, opened 2025-12-20
Introduce instance resolution pass in the frontend


edsko:
A good starting place might be is to look for use of the `TranslationConfig`, which specifies which instances we currently want to generate by default, as well as the `TranslationState`, which is used to track our conclusions (`instanceMap`, specifically).

jorisdral:
In #2091, I am adding `HasField` instances for union fields but not every union field can be given a `HasField` instance: it can only be given an instance if the field type is storable. The `TypeClass` datatype doesn't allow us to specify that "for some fields" there is a `HasField` instance. We should take this into account when implementing instance resolution in the frontend

PR #1712, opened 2026-02-17
Frontend pass: adjust function arguments that are array type

jorisdral:
Builds on top of #1711

That is, if we have this C code in a header:

```c
typedef char T [];
void foo (T xs);
```

Then in the parsed C AST each array type in a function argument position will be adjusted to a pointer to the array element type.

```c
void foo (char * xs);
```

In our Haskell bindings, we generated code for the adjusted AST like so:

```Haskell
class IsArray a where
  type Elem a :: Type

newtype T = T (IncompleteArray CChar)
  deriving newtype IsArray
foo :: Ptr (Elem T) -> IO ()
```

`Elem` is a new associated type in the `IsArray` class, which as the name suggests maps an array type to its array *element* type. For example, `Elem T = CChar`. We use `Elem` instead of `CChar` directly because `CChar` can be an implementation detail of `T`.

Constrast this new approach with the previous approach that we were taking. In that previous approach, we would adjust the type of `foo` to:

```c
void foo (T * xs);
```

This is largely equivalent, but the problem is here that converting `T *` to `T` (or vice versa) requires a pointer dereference or a cast in other cases, while converting `T` to `char *` (or vice versa) does not need either

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -4,11 +4,7 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
   merge_group:
-    branches:
-      - main
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/compile-fixtures.yml
+++ b/.github/workflows/compile-fixtures.yml
@@ -5,11 +5,7 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
   merge_group:
-    branches:
-      - main
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -5,11 +5,7 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
   merge_group:
-    branches:
-      - main
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}


hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/Instances.hs
```diff
@@ -283,3 +286,44 @@ getHsPrimTypeInsts = \case
...
+        HsFun{} -> False
+        HsExtBinding _ref _cTypeSpec hsTypeSpec _ ->
+          Inst.IsArray `Set.member` hsTypeSpecInsts hsTypeSpec
+        HsByteArray{} -> False
+        HsSizedByteArray{} -> False
+        HsBlock{} -> False
+        HsComplexType{} -> False
+        HsStrLit{} -> False
+        HsWithFlam{} -> False
+        HsEquivStorable{} -> False
```
jorisdral: @TravisCardwell `IsArray` is a new class that I added in this PR. There are two instances for the array types in the runtime:

```hs 
class IsArray a where 
  type Elem a :: Type
  ...
instance Storable a => IsArray (IncompleteArray a) where ...
instance Storable a => IsArray (ConstantArray n a) where ...
```

I want to newtype-derive instances in generated bindings. For example:

```hs
newtype A = A (ConstantArray CInt)
  deriving newtype IsArray
newtype B = A
  deriving newtype IsArray
```

This seems incompatible with `getInstances` because `getInstances` checks that a type and *all* its dependencies have an instance of the same class. Here I have a superclass constraint for a different class. How do I resolve this?

And should I maybe define `IsArray` like so?

```hs
class Storable (Elem a) => IsArray a where ...
```

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/Newtype.hs
```diff
@@ -44,7 +45,18 @@ newtypeDec name constr field orig comment candidateInsts knownInsts = do
...
+            , resolvedIsArrayInst
+            ]
+
+        resolvedIsArrayInst :: Set Inst.TypeClass
+        resolvedIsArrayInst
+          | Inst.IsArray `Set.member` candidateInsts
+          , Hs.hasInstance_IsArray transState.instanceMap (Just name) field.typ
+          = Set.singleton Inst.IsArray
+          | otherwise
+          = Set.empty
```
jorisdral: @TravisCardwell see also this ad-hoc call to `hasInstance_IsArray`

hs-bindgen/src-internal/HsBindgen/Instances.hs
```diff
@@ -225,6 +226,9 @@ instance Default SupportedInstances where
           , mkDef HasFFIType Dependent   Newtype   []
           , mkDef HasField   Independent HsBindgen []
           , mkDef Integral   Dependent   Newtype   []
+            -- TODO: IsArray instances are not dependent in the same way that
+            -- for example Storable instances are.
+          , mkDef IsArray    Dependent   Newtype   []
```
jorisdral: @TravisCardwell and is this right?

.github/workflows/haskell.yml
```diff
@@ -4,11 +4,7 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
   merge_group:
-    branches:
-      - main
```
TravisCardwell, line 11: Is removal of specification of PR and MR branches in the workflows on purpose?

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/Instances.hs
TravisCardwell: The example in the first comment:

```c
typedef char T [];
void foo (T xs);
```

```haskell
newtype T = T (IncompleteArray CChar)
  ...

foo :: Ptr (Elem T) -> IO ()
```

We need to determine if we can (`newtype`) derive an `IsArray` instance for `T`.

We can (`newtype`) derive an `IsArray` instance for `T` when the wrapped type (`IncompleteArray CChar`) has an `IsArray` instance.  Instance resolution is currently implemented using `Hs.Name NsTypeConstr` as the key, so it cannot handle parameterized types like this.

If we hard-code the logic of those two `IsArray` instances, then we can (`newtype`) derive an `IsArray` instance for `T ~ IncompleteArray CChar` when the `IncompleteArray` element type (`CChar`) has a `Storable` instance.  This is something that we can check, but it is indeed not the general `newtype` deriving logic, so it needs to be done specially.

In the `hasInstance_IsArray` function quoted above, perhaps the `HsIncompleteArray{} -> True` case is incorrect?  We must check that the `IncompleteArray` element type has a `Storable` instance, using `getInstances`, no?  This is also applies to the `HsConstArray{} -> True` case.

TravisCardwell: > And should I maybe define `IsArray` like so?

```
class Storable (Elem a) => IsArray a where ...
```

I do not this that this would help.

TravisCardwell: We will (at long last) add constraint generation when we implement instance resolution in the frontend (#1447).  Such special logic would need to be implemented there as well, but I do not think that doing so would pose a significant problem.  Perhaps we can implement "rewrite rules" to make the implementation less ad hoc.  IMHO, doing so in the backend is not worth it since we will soon rewrite it anyway.

hs-bindgen/src-internal/HsBindgen/Instances.hs
TravisCardwell: No, I do not think that this is right.  For general `newtype` deriving, we check that the wrapped type has an instance.  In the above example, the wrapped types is a `IncompleteArray CChar`, and the case in `getInstances` is as follows:

```haskell
HsIncompleteArray hsType' ->
  -- constrain by Array item type in next step
  aux (acc /\ arrayInsts) $ hsType' : hsTypes
```

The accumulator is intersected with `{Eq, Show}`, and then the recursive call prepends the wrapped type to the (empty) list of other dependencies.  That intersection already rules out `IsArray` instances, and we should *not* be checking for an `IsArray` instance of the element type, so this does not work for two reasons.

To use special logic, we need to use the `HsBindgen` strategy.  (The special logic can still generate a deriving declaration.)  Since the dependency is not the general type of dependency for `newtype` deriving, we should probably make it `Independent` and document the special case in a comment.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/Newtype.hs
TravisCardwell: In `resolvedIsArrayInst`, we check the following:

```haskell
| Inst.IsArray `Set.member` candidateInsts
```

If we use `Independent`, then `IsArray` will never be a member of `candidateInsts`.

hs-bindgen/src-internal/HsBindgen/Instances.hs
TravisCardwell: Chatting with @edsko, I realized that we *also* need to derive `IsArray` instances for types like `newtype Foo = Foo T`, so perhaps we need to use `Dependent` but *also* have special-case handling of `ConstArray`.

If we move the `Storable` constraint to the method, then it should be a lot easier to implement, using `Dependent`.

edsko:
I propose that we change the definition of the class to avoid some of the instance resolution difficulties. Rather than

```hs
class IsArray a where
  type Element a :: Type

  withPtrFirstElem :: a -> (Ptr (Element a) -> IO r) -> IO r

instance Storable a => IsArray (ConstArray n    a) where ..
instance Storable a => IsArray (IncompleteArray a) where ..
```

define instead

```hs
class IsArray a where
  type Element a :: Type

  withPtrFirstElem :: Storable a => a -> (Ptr (Element a) -> IO r) -> IO r

instance IsArray (ConstArray n    a) where ..
instance IsArray (IncompleteArray a) where ..
```

Moving the constraint to the _class_ does not help (they'd still appear as constraints on the instances), but moving it to the method _does_ help, and this makes semantic sense: you can get a pointer to the elements only of those elements are `Storable`. It's also strictly more general; there might be other methods in the class that don't depend on `Storable`, for example

```hs
class IsArray a where
  type Element a :: Type

  toList :: a -> [Element a]
  withPtrFirstElem :: Storable a => a -> (Ptr (Element a) -> IO r) -> IO r
```

.github/workflows/haskell.yml
jorisdral: this was on purpose -- I stack my PRs sometimes, in which case it is useful if CI is run. And `branches: main` is superfluous because the merge queue only applys to the default branch, which is `main`

jorisdral:
@edsko @TravisCardwell thanks for the input!

> I propose that we change the definition of the class to avoid some of the instance resolution difficulties. Rather than
> ...

I've applied this change. ~~We still need the special-cased instance resolution for now, because `getInstances` won't resolve to the right instances.~~ I've now incorporated a special case for `IsArray` in `getInstances`, instead of the separate `hasInstance_isArray` function we used to have. I've opened #1739 to track that we should revisit the special handling once #1447 is resolved

edsko:
@TravisCardwell are you happy with this PR as is, in the knowledge that instance resolution will change anyway? Or should something still be corrected?

.github/workflows/haskell.yml
TravisCardwell: Thanks for the confirmation!  I did the same in the new Windows workflows.

TravisCardwell (approved):
The `IsArray` instance implementation looks good!  It is easy to understand, and the `TODO` comments are clear.

PR #1733, opened 2026-02-19
Changelogs for globals and `hs-bindgen-runtime` refactors

dschrempf:


dschrempf:
@edsko Maybe this can even be merged without the CI.

hs-bindgen-test-runtime/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Revision history for hs-bindgen-test-runtime
 
-## 0.1.0 -- YYYY-mm-dd
+## ?.?.? -- YYYY-mm-dd
+
+* BREAKING: classify internal modules as such.
```
edsko: I guess this is _technically_ "breaking", but in practice, hopefully, this wouldn't affect to many people right? Or have we changed definitions that people might actually use?

hs-bindgen-test-runtime/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
...
+* BREAKING: add an internal prelude, re-exporting all definitions required by
```
edsko: This is not a breaking change right? I's a new module?

hs-bindgen/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 * BREAKING: occurrences of the `CFieldType`/`CBitfieldType` type families in
   class instance heads are now replaced by their definition.
+* MINOR: re-export all global definitions used by `hs-bindgen` generated code
+  from `hs-bindgen-runtime`. This may affect required imports when `hs-bindgen`
+  is used in Template Haskell mode.
```
edsko: I think it would be useful to give hints about how people should update their code (here and elsewhere, we should make a habit of that).

hs-bindgen-test-runtime/CHANGELOG.md
jorisdral: IMO it's nicer for users to err on the safe side. That is, mark something as breaking rather than non-breaking

hs-bindgen-test-runtime/CHANGELOG.md
jorisdral, line 1: AFAIK `hs-bindgen-test-runtime` was not part of the alpha release, so I don't think we need to keep a changelog here

hs-bindgen/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 * BREAKING: occurrences of the `CFieldType`/`CBitfieldType` type families in
   class instance heads are now replaced by their definition.
+* MINOR: re-export all global definitions used by `hs-bindgen` generated code
```
jorisdral: We should settle on what we call a non-breaking change. You use `MINOR` here, but I use `NON-BREAKING` elsewhere

hs-bindgen-test-runtime/CHANGELOG.md
edsko: Yeah that is reasonable, but then it's also good to reassure people (we did promise to try and not break people's code between 0.1-alpha and 0.1). Perhaps another way of pharsing this is that explaining how users need to update their code (https://github.com/well-typed/hs-bindgen/pull/1733#discussion_r2826581337) would help, which in this case (hopefullly) means: "most users don't need to update their code at all".

hs-bindgen-test-runtime/CHANGELOG.md
edsko: Ah, yes, agreed.

hs-bindgen/CHANGELOG.md
dschrempf: Right! Actually, it turns out not the imports are affected, but the build-dependencies (ghc-prim, primitive). I added a note.

hs-bindgen-test-runtime/CHANGELOG.md
dschrempf: Yes, I accidentially added the changelog there. Meant to add it to `hs-bindgen-runtime`.

hs-bindgen/CHANGELOG.md
dschrempf: We are having a democratic vote about what would be the best partitioning scheme :-P. Thanks for bringing it up!

dschrempf:
Thanks for your reviews. I have updated the changelogs reflecting the outcome(s) of our lively discussions :-). More comments are welcome!

hs-bindgen-test-runtime/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Revision history for hs-bindgen-test-runtime
 
-## 0.1.0 -- YYYY-mm-dd
+## ?.?.? -- YYYY-mm-dd
+
+## 0.1.0-alpha -- 2026-02-06
```
jorisdral:
```suggestion
```

dschrempf: Applied.

Issue #1504, opened 2026-01-09
Incorporate all name mangling into the name mangler


dschrempf:
At the moment, not all name mangling happens in the name mangler pass.

For example, we create function wrapper names by unsafely adding `_wrapper` suffixes. Thereby we possibly introduce a conflicting name.

The name mangler should take care of these unsafe operations.

Non-exhaustive list of possible name mangler violations:

``` 
- /hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/Function.hs::mangledOrigNameWrapper = Hs.unsafeHsIdHsName $ mangledOrigId <> "_wrapper"
- /hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs::auxTypedef :: C.Typedef Final
- /hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs::getterName = Hs.unsafeHsIdHsName $ "get_" <> field.info.name.hsName
```

As stated in https://github.com/well-typed/hs-bindgen/issues/1269, checking macro expressions is out of scope for this ticket.

edsko:
Related: #1432.

edsko:
@dschrempf I'm going to mark this as a medium priority bug, and I'll assign it to the both of us. Let's coordinate who does what.

edsko:
I'll _not_ mark this for `release-0.1` because while this is a bug, it should only arise in edge cases, and even then people will have a workaround (prescriptive binding spec). The same work around as for when we'd fix this issue, in fact, just with less clear feedback (instead of a nice warning they'd end up with Haskell code that does not compile).

edsko:
I think it would be good to fix this for release-0.1 after all, though indeed not critical. But panics are not nice :) Bumping.

dschrempf:
Related: https://github.com/well-typed/hs-bindgen/issues/1734

dschrempf:
I think the name-mangler does not exhibit many panics. In fact, the only panic I found is

```haskell
instance Semigroup TypedefAnalysis where
  a <> b = TypedefAnalysis{
        map = Map.unionWithKey unexpectedOverlap a.map b.map
      }
    where
      unexpectedOverlap :: DeclId -> Conclusion -> Conclusion -> Conclusion
      unexpectedOverlap declId conclusion1 conclusion2 =
          panicPure $ concat [
              "Unexpected overlap for "
            , show declId
            , ": "
            , show (conclusion1, conclusion2)
            ]
```

[...]

PR #1736, opened 2026-02-19
Name mangler: Union getter and setter fields

dschrempf:
Houston we have a (small) problem.

Mangling the getters and setters in the name mangler actually revealed a fixture change.

Before this change, unions starting with an upper case letter were lower-cased by `FixCandidate` (which was run before adding the prefixes "get_" or "set_"). Now that we add the prefixes before fixing the candidate, the upper case letter is left in place (and I think that is correct). For example

```c
union MyUnion { int x; };
```

had a getter

```haskell
get_myUnion_x :: MyUnion -> CInt
```

which is now correctly renamed to

```haskell
get_MyUnion_x :: MyUnion -> CInt
```

(The PR does not yet contain these fixture changes).

Opinions?

#### Panics, error messages, and trace messages

This PR also reduces panics, improves error messages and a few trace messages:

Improve error messages
- Start with capital letter (error message is shown at beginning of line)
- Remove function name as it is shown in the call stack

#### Origin ticket number

The origin ticket was changed from https://github.com/well-typed/hs-bindgen/issues/1504 to https://github.com/well-typed/hs-bindgen/issues/1740.


edsko:
Feel free to merge once the backwards incompatible change is rolled back.

Issue #1714, opened 2026-02-17
Backend: Improve how we handle tuples


dschrempf:
After #1710, we will have various tuple constructors in our AST types. This is, because we have boxed and unboxed, as well as open (non-applied) and closed tuple types.

We should
- remove the closed tuples from the AST types; and instead always use open tuples together with expression or type applications;
- render tuples in _closed form_ (when possible); this requires a bit of fiddling with brackets and arguments (tuple applications need to be unfolded before we can decide how to render them).

The changes should affect both, expressions and types.

Moreover, at the moment we use `Unit`, `Solo`, and `MkSolo` to create open, boxed 0- and 1-tuples. We _don't use `Solo`/`MkSolo`_ for closed tuples!

We should get rid of 1-tuples (`Solo`/`MkSolo`), and instead just unbox the value. Maybe we can just remove 1-tuples from the AST before or while rendering?

There is also a discrepancy between how we render tuples in preprocessor and TH modes.

#1710 will add various TODO items with this ticket numner to the code; please check those.

edsko:
@dschrempf Let's do this for 0.1, just out of concern that we might accidentally introduce some backwards incompatible changes. I think it should be relatively quick?

edsko:
I've marked this as `refactor` as I _think_ it won't affect any of the generated code. I'll be important to know if that happens  not to to the case.

edsko:
Avoiding rendering this using prefix notation `(,,) ..` we could definitely do for 0.2 (like #1715), that's not a breaking change; but the `Solo`/`Unit` stuff _might_ be.

dschrempf:
Right! Removing `Unit/Solo` is definitely a breaking change. (Note, we do not have a single use case of Solo in our tests).

I am hesitating, because we might need `Unit` (i.e., `()`) for other use cases. I guess we should differentiate between N-tuples (with `N>=2`), and `Unit = ()`?

dschrempf:
I have been digging a bit about removing Solo/MkSOlo, and we have to be careful here.

In TH, we translate, for example,
```haskell
      EBoxedClosedTup xs -> TH.tupE $ mkExpr env <$> xs
```
but in preprocessor mode, we do
```haskell
    EBoxedClosedTup xs ->
      let ds = prettyExpr env 0 <$> xs
          l  = PP.hlist "(" ")" ds
      in  PP.ifFits l l $ PP.vlist "(" ")" ds
```

[...]

PR #1754, opened 2026-02-23
Remove 1-tuples from AST

dschrempf:
This fixes one part of #1714 (no `Unit/Solo/MkSOlo`).

It does so by only supporting Unit and N+2-tuples in the AST, removing support for 1-tuples.

(It does not fix the "remove closed tuples from AST, and render open tuples as closed tuples, if possible" request).


dschrempf:
Superseded by #1756

PR #1755, opened 2026-02-24
Update Nix Flake lock file

dschrempf:


dschrempf:
Minor PR, which affects Nix tooling. Putting on the queue.

PR #1763, opened 2026-02-24
Add `internal frontend` command to output passes results

### LINKED ISSUE #1731: Add CLI `internal` commands to dump after any pass
We should make the `Artefact`s around the various passes more consistent. I'm imagining something like this:

```hs/
data Pass (result :: Type) where
  ... every pass listed here with its result type ...

data Artefact (result :: Type) where
  RunPass :: Pass result -> Artefact result
```

Then `HsBindgen.Cli.InternalCmd` should be changed to something like

```hs
data Cmd where
  CmdDumpPass :: (...constraints on result so that we can print it ...) => Frontend.Opts -> Pass result -> Cmd
```

bolt12:
Closes #1731

### DIFF
--- a/hs-bindgen/app/HsBindgen/Cli/Internal/Frontend.hs
+++ b/hs-bindgen/app/HsBindgen/Cli/Internal/Frontend.hs
@@ -45,15 +45,16 @@ parseFrontendPassName s = case lookup s knownPasses of
   where
     knownPasses :: [(String, SomeFrontendPass)]
     knownPasses = [
-          mk DumpParse
-        , mk DumpSimplifyAST
-        , mk DumpAssignAnonIds
-        , mk DumpConstructTranslationUnit
-        , mk DumpHandleMacros
-        , mk DumpResolveBindingSpecs
-        , mk DumpMangleNames
-        , mk DumpAdjustTypes
-        , mk DumpSelect
+          mk ParsePass
+        , mk SimplifyASTPass
+        , mk AssignAnonIdsPass
+        , mk ConstructTranslationUnitPass
+        , mk HandleMacrosPass
+        , mk ResolveBindingSpecsPass
+        , mk MangleNamesPass
+        , mk AdjustTypesPass
+        , mk SelectPass
+        , mk FinalPass
         ]
 
     mk :: Show result => FrontendPass result -> (String, SomeFrontendPass)
@@ -62,16 +63,16 @@ parseFrontendPassName s = case lookup s knownPasses of
 -- Ensure that we handle all 'FrontendPass' constructors.
 frontendPassName :: FrontendPass result -> String
 frontendPassName = \case
-  DumpParse                    -> "parse"
-  DumpSimplifyAST              -> "simplify-ast"
-  DumpAssignAnonIds            -> "assign-anon-ids"
-  DumpConstructTranslationUnit -> "construct-translation-unit"
-  DumpHandleMacros             -> "handle-macros"
-  DumpResolveBindingSpecs      -> "resolve-binding-specs"
-  DumpMangleNames              -> "mangle-names"
-  DumpAdjustTypes              -> "adjust-types"
-  DumpSelect                   -> "select"
-  DumpFinal                    -> "final"
+  ParsePass                    -> "parse"
+  SimplifyASTPass              -> "simplify-ast"
+  AssignAnonIdsPass            -> "assign-anon-ids"
+  ConstructTranslationUnitPass -> "construct-translation-unit"
+  HandleMacrosPass             -> "handle-macros"
+  ResolveBindingSpecsPass      -> "resolve-binding-specs"
+  MangleNamesPass              -> "mangle-names"
+  AdjustTypesPass              -> "adjust-types"
+  SelectPass                   -> "select"
+  FinalPass                    -> "final"
 
 {-------------------------------------------------------------------------------
   CLI help
@@ -106,7 +107,7 @@ parseOpts =
 parseDump :: Parser SomeFrontendPass
 parseDump = option (eitherReader parseFrontendPassName) $ mconcat [
       long "pass"
-    , value (SomeFrontendPass DumpAdjustTypes)
+    , value (SomeFrontendPass AdjustTypesPass)
     , showDefaultWith (\(SomeFrontendPass d) -> frontendPassName d)
     , help "Frontend pass to dump"
     , metavar "PASS"
--- a/hs-bindgen/src-internal/HsBindgen/Artefact.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Artefact.hs
@@ -46,26 +46,25 @@ import HsBindgen.Util.Tracer
 -- Each constructor corresponds to a frontend pass carrying the result type of
 -- that pass. See "HsBindgen.Frontend" for the pass ordering and descriptions.
 data FrontendPass (result :: Star) where
--- TODO-D: Rename to ParsePass etc.
-  DumpParse
+  ParsePass
     :: FrontendPass [ParseResult Parse]
-  DumpSimplifyAST
+  SimplifyASTPass
     :: FrontendPass [ParseResult SimplifyAST]
-  DumpAssignAnonIds
+  AssignAnonIdsPass
     :: FrontendPass [ParseResult AssignAnonIds]
-  DumpConstructTranslationUnit
+  ConstructTranslationUnitPass
     :: FrontendPass (C.TranslationUnit ConstructTranslationUnit)
-  DumpHandleMacros
+  HandleMacrosPass
     :: FrontendPass (C.TranslationUnit HandleMacros)
-  DumpResolveBindingSpecs
+  ResolveBindingSpecsPass
     :: FrontendPass (C.TranslationUnit ResolveBindingSpecs)
-  DumpMangleNames
+  MangleNamesPass
     :: FrontendPass (C.TranslationUnit MangleNames)
-  DumpAdjustTypes
+  AdjustTypesPass
     :: FrontendPass (C.TranslationUnit AdjustTypes)
-  DumpSelect
+  SelectPass
     :: FrontendPass (C.TranslationUnit Select)
-  DumpFinal
+  FinalPass
     :: FrontendPass (C.TranslationUnit Final)
 
 {-------------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


TravisCardwell:
The CLI implementation is architected so that modules strictly mirror the organization of commands, but this does not follow that design.  Perhaps *the pass should be specified as an option* instead.  In addition to cleaning up the design, this would also make it easy to reuse the `--pass` option in other debugging commands, while commands are not as composable.  Also note that `internal frontend` usage is unchanged with the final pass as the default, while use of commands breaks current usage.

hs-bindgen/src-internal/HsBindgen/FrontendDump.hs
```diff
@@ -0,0 +1,101 @@
...
+
+{-------------------------------------------------------------------------------
+  Definition
+-------------------------------------------------------------------------------}
+
+-- | Frontend dumps
+--
+-- Each constructor names a frontend pass and carries the result type of running
+-- that pass. See "HsBindgen.Frontend" for the pass ordering and descriptions.
+data FrontendDump (result :: Star) where
```
bolt12:
```suggestion
data FrontendPass (result :: Star) where
```

bolt12: Move this into the Artefact.hs module and `SomeFrontendDump` and the parser for it should live in the cli. 

And then after this change we can see which other constructors/artefacts are no longer necessary because we this more general way of running any pass

edsko (commented):
@dschrempf It would be good to have your eye on this, but not just yet; @bolt12 will make some changes first and will then ping you.

bolt12:
@dschrempf All comments addressed, could you take a look please? I tried to see if we could simplify the logic with this general FrontendPass but didn't find anything worth to modify

edsko:
> I tried to see if we could simplify the logic with this general FrontendPass but didn't find anything worth to modify

This is surprising to me. A lot of the artefacts are simply "get me the result of pass such and such". Shouldn't we be able to remove all of those now?

bolt12:
Yes, but I found it a bit more ugly and less straightforward, e.g. I had to do a bit of a nested pattern matching, so I thought about delegating to @dschrempf so he could judge better for himself 🙇🏼

edsko:
I think this needs a few changes; `Artefact` now has quite a bit of redundancy. Some of the existing constructions can probably be replaced by functions (relying on the `Functor` instance) or deleted altogether. But we can wait for @dschrempf 's suggestions.

dschrempf:
Sorry, I have missed this completely. Will have a look now.

dev/internal-commands.md
```diff
@@ -0,0 +1,91 @@
...
+|---|----------------|-------------|
+| 1 | `parse` | Traverse libclang AST into Haskell C AST |
+| 2 | `simplify-ast` | Convert anonymous enums to pattern synonyms |
+| 3 | `assign-anon-ids` | Assign names to anonymous declarations |
+| 4 | `construct-translation-unit` | Order declarations by dependencies, build graphs |
+| 5 | `handle-macros` | Typecheck macros, reparse with macro info |
+| 6 | `resolve-binding-specs` | Apply external/prescriptive binding specs |
+| 7 | `mangle-names` | Assign Haskell names, squash typedefs |
+| 8 | `select` | Filter by predicates and program slicing |
+| 9 | `adjust-types` | Adjust types (e.g. function args to function pointers) |
```
dschrempf: I think we should remove this section. It is another place were we can diverge from the actual passes. Maybe we can simply refer to `Frontend.hs`?

dschrempf: (Or a moduel defining the data type for these passes).

dev/internal-commands.md
```diff
@@ -0,0 +1,91 @@
...
+  --parse-from-main-headers /usr/include/time.h
+
+# With an external binding spec
+cabal run hs-bindgen-cli -- internal frontend --pass resolve-binding-specs \
+  --external-binding-spec spec.yaml myheader.h
+```
+
+## `clang-ast-dump`
+
+Displays low-level details about the Clang AST, as libclang sees it (before
```
dschrempf:
```suggestion
Displays low-level details about the Clang AST, as `libclang` sees it (before
```

hs-bindgen/app/HsBindgen/Cli/Internal/Frontend.hs
```diff
@@ -14,28 +14,65 @@ module HsBindgen.Cli.Internal.Frontend (
...
+    knownPasses = [
+          mk DumpParse
+        , mk DumpSimplifyAST
+        , mk DumpAssignAnonIds
+        , mk DumpConstructTranslationUnit
+        , mk DumpHandleMacros
+        , mk DumpResolveBindingSpecs
+        , mk DumpMangleNames
+        , mk DumpSelect
+        , mk DumpAdjustTypes
```
dschrempf: The order has already changed in the meantime, sorry for that...

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -27,13 +30,64 @@ import HsBindgen.Frontend.Analysis.IncludeGraph qualified as IncludeGraph
...
+  DumpParse                    -> "parse"
+  DumpSimplifyAST              -> "simplify-ast"
+  DumpAssignAnonIds            -> "assign-anon-ids"
+  DumpConstructTranslationUnit -> "construct-translation-unit"
+  DumpHandleMacros             -> "handle-macros"
+  DumpResolveBindingSpecs      -> "resolve-binding-specs"
+  DumpMangleNames              -> "mangle-names"
+  DumpSelect                   -> "select"
+  DumpAdjustTypes              -> "adjust-types"
+
```
dschrempf, line 69: This seems like it should go into `Frontend.hs`?

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -42,6 +96,8 @@ import HsBindgen.Util.Tracer
 data Artefact (a :: Star) where
   -- * Boot
   HashIncludeArgs     :: Artefact [HashIncludeArg]
+  -- * Frontend passes
+  RunFrontendPass     :: FrontendPass result -> Artefact result
```
dschrempf: Maybe rename to "FrontendPass" or "DumpFrontendPass"? (Because the Artefact doesn't really "run" the passes).

hs-bindgen/src-internal/HsBindgen/Artefact.hs
```diff
@@ -117,6 +175,19 @@ runArtefacts tracer boot frontend backend artefact =
         (Lift   f)          -> f
         (Bind x f)          -> runArtefact x >>= runArtefact . f
 
+-- | Run a frontend pass by dispatching on the 'FrontendPass' GADT.
```
dschrempf: I think the comment can go.

hs-bindgen/src-internal/HsBindgen/Frontend.hs
```diff
@@ -327,6 +327,10 @@ runFrontend tracer config boot = do
     frontendDependencies <- cache "frontendDependencies" $ do
       IncludeGraph.toSortedList . (.includeGraph) <$> getCTranslationUnit
 
+    frontendDumpParse <- cache "frontendDumpParse" $ do
+      (parseResults, _, _, _, _, _) <- parsePass
+      pure parseResults
```
dschrempf: I think this can go; instead use the `parsePass` itself?

dschrempf: A'la

```
let r = (\(x, _, _, _, _, _) -> x) <$> parsePass
```

(Maybe we can introduce a record type replacing the overgrown 6-tuple)?

dev/internal-commands.md
bolt12: This is the dev docs, I don't think it hurts this being here?

hs-bindgen/app/HsBindgen/Cli/Internal/Frontend.hs
bolt12: no worries! thanks for letting me know I will update this!

hs-bindgen/src-internal/HsBindgen/Artefact.hs
bolt12: Good point!

hs-bindgen/src-internal/HsBindgen/Artefact.hs
bolt12: This is a type specific to the Artefacts module, I think this belongs here.

hs-bindgen/src-internal/HsBindgen/Frontend.hs
bolt12: Yeah, you are probably right.

dev/internal-commands.md
dschrempf: Of course! Don't feel obliged to follow all of my suggestions. I am just really careful with having duplicated information lying around. It will diverge at some point.

hs-bindgen/src-internal/HsBindgen/Artefact.hs
dschrempf: Hm, that's a bit of a bummer. I thought it so nice we finally have a type summarizing the passes. I guess that's the downside of having "open" type classes, if that makes sense.

bolt12: I see, well, I remember @edsko asking for this data type to be on the Artefact module but I could be wrong. since a couple of weeks have passed since we last spoke about this PR, maybe he could get the final word 😛  ?

edsko: @dschrempf My main question here is whether these new constructors obsolete some of the existing `Artefact` constructors. I think these should not be called _dump_ such and such a pass, but simply _get_ the (output of) such and such a pass, and I _think_ these can then obsolete quite a few of the `Artefact` constructors, as many of them as simply ways to do precisely that: get the output of a specific pass; perhaps modified in some minor way, but that can be done with the `Functor` instance. This is what I was referring to [above](https://github.com/well-typed/hs-bindgen/pull/1763#issuecomment-3965561356); am I simply wrong about that?

And yes, I do think this belongs in `Artefact`. I agree that having a closed world of passes would be nicer, and I experimented with that, but it means getting orphans everywhere.

edsko: (That said, some kind of combination of an open kind with a closed type of passes, along with functions in both directions, _could_ work to avoid the orphans I guess. But we can consider that at another time, independent from this PR.)

dschrempf: Sure. I thought the goal is to have `String` output of the passes, but if we want the _actual_ output, we can reuse these outputs in the other artefacts. I am just a bit worried that we leak a lot of implementation details. The current interface is tailored to users, while the interface we are creating seems to be tailored to developers?

edsko: In which way is `Artefact` tailored to users...?

edsko: `Artefact` is a way to query the `hs-bindgen` pipeline at various points. The `String` output is what _dump_ does, but it should not be what is hard-coded here.

edsko: If we release `hs-bindgen`-the-library and we want to expose parts of `Artefact` only, then probably the better answer is to make it opaque, and only expose a set of "blessed" functions that create specific Artefacts. That's probably a good idea regardless, hiding the specific way that `Artefact` is defined (what is primitive and what is  not).

dschrempf: Sure! I was just expressing my concerns. And you are right, we hide `Artefact` from users; somehow I thought it is closer to the interface than it actually is!

edsko: @bolt12 I've asked Dominik to push a commit to this PR, probably a bit more efficient than me discussing something with Domnik and then Dominik discussing that with you :)

Issue #1725, opened 2026-02-18
Syntax for macro "namespace"


edsko:
Macros live in a namespace of their own; selection predicates (and binding specs) currently have no way of distinguishing between a regular C declaration called `foo` and a macro called `foo`.

edsko:
@TravisCardwell @dschrempf This affects selection predicates as  well as (in principle at least) binding specifications, and it's a backwards incompatible change. Should we introduce this for `0.1`?

edsko:
We decided to 

* Extend `NameKind` with another option for macros
* User-facing syntax: `macro foo`.

edsko:
@TravisCardwell We should remember to update the `CHANGELOG`.

TravisCardwell:
We have a predicate [`hasUnsupportedType`](https://github.com/well-typed/hs-bindgen/blob/92d4a66078a25ac65487799161894960ef443bd0/hs-bindgen/src-internal/HsBindgen/Frontend/AST/Type.hs#L487-L509) that uses the following logic for `TypeRef`s:

```haskell
-- | Checks if a type is unsupported by Haskell's FFI
hasUnsupportedType :: forall tag p.
     (IsPass p, Normalize tag Canonical)
  => TypeF tag p -> Bool
hasUnsupportedType = aux . getCanonicalType
  where
    aux :: CanonicalType p -> Bool
    aux (TypeRef declId)      = auxRef (idNameKind (Proxy @p) declId)
    ...

    auxRef :: CNameKind -> Bool
    auxRef = \case
          CNameKindOrdinary              -> False
          CNameKindTagged CTagKindStruct -> True
          CNameKindTagged CTagKindUnion  -> True
          CNameKindTagged CTagKindEnum   -> False
```

Perhaps adding support for the new macro namespace is as simple as the following?

```haskell
auxRef :: CNameKind -> Bool
auxRef = \case
  CNameKindOrdinary              -> False
  CNameKindTagged CTagKindStruct -> True
  CNameKindTagged CTagKindUnion  -> True
  CNameKindTagged CTagKindEnum   -> False
  CNameKindMacro                 -> False
```

PR #1764, opened 2026-02-25
Add syntax for macro namespace

TravisCardwell:
This PR adds a macro namespace, so that macro names are distinct from other names.

This is done by adding a `CNameKindMacro` constructor to the `CNameKind` type.  Since this is not standard C, the name types are moved into `HsBindgen.Frontend.Naming`.  Since that module is intended for unqualified import, the type names now have a C prefix.

Binding specifications and selection predicates now refer to macros using `macro`.  For example a macro named `foo` is referred to as `macro foo`.

Rendered C source must *not* use a `macro` prefix.  Macro `macro foo` is rendered as just `foo` in C source.

We have special logic for handling collisions, when an ordinary declaration and a macro declaration have the same name.  This commit refactors this logic so that it works the same.  We may want to change the logic in a later commit.  The fixture changes are just changes from `foo` to `macro foo` in comments and binding specifications.

I updated the changelog, listing this as a breaking change.


TravisCardwell:
We should discuss https://github.com/well-typed/hs-bindgen/issues/1725#issuecomment-3947929330.

edsko:
> We should discuss [#1725 (comment)](https://github.com/well-typed/hs-bindgen/issues/1725#issuecomment-3947929330).

I think this code is outdated; [`TypeRef`](https://github.com/well-typed/hs-bindgen/blob/92d4a66078a25ac65487799161894960ef443bd0/hs-bindgen/src-internal/HsBindgen/Frontend/AST/Type.hs#L63) I _think_ can only point to unions and structs, so we should still do the case split but then panic on anything else.

edsko:
We discussed this at length with @TravisCardwell and @dschrempf . Travis is going to change this PR so that the conflict detection happens during the construction of the `DeclIndex` (as we did prior to this PR), instead of in the name mangler. See https://github.com/well-typed/hs-bindgen/issues/1777 for details.

TravisCardwell:
> We have special logic for handling collisions, when an ordinary declaration and a macro declaration have the same name. This commit refactors this logic so that it works the same. We may want to change the logic in a later commit. The fixture changes are just changes from `foo` to `macro foo` in comments and binding specifications.

I cleaned up and stashed my commit that removes cross-namespace conflict detection from `fromParseResults`.  When we implement #1777, we should be able to get the new `fromParseResults` from that stashed commit.  I added a `TODO` pointing to that issue, with a description.

In effect, this PR just implements the `macro` syntax, in accordance with the title.

To review the changes without "noise," see [this commit](https://github.com/well-typed/hs-bindgen/pull/1764/changes/35164054f1e1f9096b2662738ea669c1558e312d).

PR #1767, opened 2026-02-25
Flip AdjustTypes and Select passes; state why Select should come last

dschrempf:


dschrempf:
And he's not the only one... :thinking:

hs-bindgen/src-internal/HsBindgen/Frontend.hs
TravisCardwell: You might want to fix the syntax here while you are at it: `/pointer/`.

PR #1771, opened 2026-02-25
Make sure all TODOs link to issues

edsko:


hs-bindgen/test/common/Test/Common/HsBindgen/Trace/Predicate.hs
```diff
@@ -171,23 +171,30 @@ withTraceConfigPredicate ::
...
+              }
+          }
+
+        checkTraces :: IO ()
+        checkTraces = do
+          traces <- readIORef tracesRef
+          mapM_ (report . show . prettyForTrace) traces
+          case runExcept (predicate traces) of
+            Left  e -> throwIO e
+            Right _ -> pure ()
```
edsko, line 194: @dschrempf ^ This is the change to the test suite we discussed.

Issue #1696, opened 2026-02-12
Audit parse predicates in golden tests involving standard library external binding specifications


dschrempf:
In PR comment
https://github.com/well-typed/hs-bindgen/pull/1695#issuecomment-3888217386, @TravisCardwell notices that we have to add a custom parse predicate to our golden tests of the standard library external binding specifications, so that we use our own definition of `CBool`: `HsBindgen.Runtime.LibC.CBool`. If we do not use a custom parse predicate, we create bindings referring to `FC.CBool`.

We should have a look into these two tests:
```haskell
test_bindingSpecs_stdlib_instances :: TestCase
test_bindingSpecs_stdlib_return_values :: TestCase
```

What happens if we remove the parse-all predicate? (I.e., does it change the generated bindings and why?)
Do we expect such behavior?

edsko:
Let's do this after #1627 is done, because that makes how we generate imports _much_ more consistent.

edsko:
@dschrempf I've assigned this to you since you've been looking at this already anyway, and I've assigned this to `release-0.1`, primarily because of a worry that this might be indicative of a bug. If when you start looking at this you realize that this is expected behaviour, we can either just close this ticket, or if there is some work to be done, leave a summary here, and we drop the `release-0.1` label.

edsko:
Analysis:

Given

```c
#include "stdbool.h"
typedef bool Foo;
```

In `stdbool.h`, `bool` is defined as a macro

```c
#define bool _Bool
```

Therefore `clang` will inform us that it expanded a macro in the declaraiton of `Foo`, and therefore we will try to reparse it. The initial environment for the `language-c` based parser is given by

```hs
initReparseEnv :: ClangCStandard -> ReparseEnv
initReparseEnv standard = Map.fromList (bespokeTypes standard)

bespokeTypes :: ClangCStandard -> [(CName, C.Type HandleMacros)]
bespokeTypes = \case
    ClangCStandard C23 _gnu -> [("bool", C.TypePrim C.PrimBool)]
    _otherwise -> []
```

That is, it's either empty, or it contains a single entry for `bool`, mapping to `TypePrim PrimBool`, if using C23.

Now there is a case split:

**1. We do not parse everything.**

[...]

Issue #1251, opened 2025-11-06
Drop parse predicates


edsko:
Parse predicates were introduced for two reasons:

* Not having to reify every single definition could have important performance benefits. ("Could", because until we have real use cases, for big libraries, it's hard to be sure)
* If we _do_ reify everything, there is a larger change of `hs-bindgen` falling over some construct it has never seen before.

The second point here is increasingly irrelevant: `hs-bindgen` is getting better and better at simply skipping over declarations with features it doesn't recognize, whilst still generating bindings for the rest. (Indeed, if it doesn't, that should be considered a bug). The impact of the first point is unclear until we have sizable benchmarks. Once we do, we could reconsider whether we need parse predicates; if we don't, both the user interface to `hs-bindgen` as well as the implementation could benefit from some significant simplifications.

edsko:
One point in favour of removing parse predicates: https://github.com/well-typed/hs-bindgen/pull/1772, especially the discussion towards the end; therefore bumping to medium priority.

edsko:
Parse predicates are getting increasingly difficult to justify:

1. We need to parse `typedefs` (or at least record their existence) so that we can construct the correct type environment for reparsing (to recover macro use sites).
2. We need location information for all declarations, so that we can match them against external binding specifications.
3. (_New realization_): we need to know about the existence of macros, so that we can detect conflicts between macros and non-macros. This is especially critical for non-function macros, where we have no way of referring to the non-macro on the C side (the macro will always take precedence).

We decided to change the default to `--parse-all` for release 0.1 (https://github.com/well-typed/hs-bindgen/issues/1779), and then drop parse predicates altogether for release 0.2 (though we might decide to drop them for 0.1 already, depending on timing).

edsko:
Marked this as blocked by #1779; that's not _technically_ true, but it would be good to change the default first (and changing that default in turn has a dependency on #1009).

PR #1772, opened 2026-02-25
Booleans and the stdlib external binding specs

dschrempf:
Closes https://github.com/well-typed/hs-bindgen/issues/1696

- Test different C standards and parse predicates
- Document convoluted behavior and expectations

### DIFF
--- a/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden.hs
+++ b/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden.hs
@@ -867,6 +867,7 @@ test_bindingSpecs_stdlib_instances_c11_parse_all =
 test_bindingSpecs_stdlib_instances_c23_parse_def :: TestCase
 test_bindingSpecs_stdlib_instances_c23_parse_def =
     testVariant "binding-specs/stdlib/instances" "1.c23-parse-def"
+      & #clangVersion .~ Just (>= (18, 0, 0))
       & #onBoot       .~ ( #clangArgs % #argsBefore .~ ["-std=c23"] )
 
 -- **Case 4: C23; parse @stdbool.h@**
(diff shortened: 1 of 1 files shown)


TravisCardwell:
This documents and tests the issue, but we cannot *fix* it?

IMHO, we should always import `CBool` from the same module, no matter how the type was determined.  Currently, we import the same re-exported type from different modules depending on the situation.

dschrempf:
Yes, your are right.

I suggested removing the re-export of C types from the runtime internal prelude, and instead using LibC. Or we could also go the other route, and point the stdlib ext bind specs to the runtime internal prelude (and not LibC), and remove the standard C types from LibC.

(The first option may be cleaner).

edsko:
Discussed this with @TravisCardwell this morning. We have concluded that the status quo is still the best option.

Note: if we decide that we don't need parse predicates at all (#1251), the problem disappears.

**Option 1: Special-casing `bool`**

When we match type references against an external binding spec, we _could_ look at the underlying type first, and if we see that this is `PrimBool`, treat it as if it was a primitive type (which is never matched against an external binding spec). This would smooth over the difference from this ticket, but it would also mean that if a user defines something like

```c
#define FileExists bool
``` 

that `FileExists` type would disappear in the generated bindings. This violates the "preservation of semantics" principle.

**Option 2: Re-exporting `.LibC.*` from `.Internal.Prelude` (`RIP`)**

We could consider `.LibC` an _internal_ module, re-export everything from `.Internal.Prelude`. This would mean that `bool` gets mapped to `RIP.CBool` no matter if the macro is parsed or not. However, since users can choose not to use our standard library, keeping these two modules separate is much cleaner: `RIP` is the core bindgen stuff, independent from any kind of external binding specification, and `.LibC` is used _only if_ the default binding specificition is used.

**Option 3: Map `bool` to RIP.CBool from the external binding spec**

This is undesirable for the same reason: if we map `bool` to `RIP.CBool`, it suggests that this is "core" `hs-bindgen` and unaffected by external binding specs. Yet, if a user then chooses a _different_ standard library, `bool` would then suddenly be mapped elsewhere; and indeed, the very problem from this ticket would then re-appear (if the standard library is parsed, the user's binding spec would apply; if it is not, we would see it as a primtype).

There is secondary, somewhat more minor, downside to this option also. Doing this would require us to split the default libc external binding spec into two (a binding spec is always for _one_ Haskell module). This means that if users use the "dump default binding spec" feature as a starting point for creating their own, that difference would be visible there also.

**Option 4: don't include `bool` _at all_**

Our `.LibC` module is intended to be those definitions that are common to all `libc` implementations across all C versions (#1775). One could argue that `bool` does not meet that criterion. However, we _can_ provide a definition that is usable across all implementations and all C versions, and moreover _not_ doing so would mean every use of `bool` in every next library would result in new, all incompatible, bindings; that's probably not what we want.

PR #1782, opened 2026-02-26
Fix Haddock warnings in `hs-bindgen-runtime`

dschrempf:
Remaining Haddock issues:

In `LibC`, we re-exports some definitions without documentation:

```
Missing documentation for:
  CIntMax
  CUIntMax
  CIntPtr
  CUIntPtr
```

We do not find link destinations for
- declarations in `Data.Vector.Unboxed.Base`; for example, `Data.Vector.Unboxed.Base.V_Int32`
- instance declarations in `HsBindgen.Runtime.Internal.LibC.Auxiliary`; for example, `HsBindgen.Runtime.Internal.LibC.Auxiliary.D:R:CFieldTypeCTm"tm_hour"` (maybe because this is an `hsc` file?)


edsko:
@TravisCardwell Could you take a look here to see if you can figure out what's wrong?

TravisCardwell:
I am unable to spot anything.  The [instance documentation](https://haskell-haddock.readthedocs.io/latest/markup.html#instance-documentation) section of the Haddock documentation indicates that documentation for an instance with data types that are hidden is not produced.  Perhaps it always outputs warning in such cases?

BTW, I tried replacing `Auxiliary.hsc` with the generated `Auxiliary.hs`, removing the `LINE` pragmas, to see if it changed anything, it did not: we still get the same warnings.  It does not look like using `hsc2hs` is causing any issues.

edsko:
The warning seems to be about the type level _symbol_ `"some_field_name"`...? Very strange. The actual genrated Haddocks look good.

PR #1793, opened 2026-02-27
Use name mangler for macro names

### LINKED ISSUE #1269: `CheckedMacroExpr` should allow for name mangling
This will require adding a `p :: Pass` type parameter, and then distinguishing between `Name` vs `NamePair`.

edsko:
Closes #1269


edsko:
@dschrempf After this PR the only use of `fixCandidate` is in `MangleNames`.

PR #1795, opened 2026-03-02
Fix `#include MACRO` panic

TravisCardwell:
In Haddock documentation of declarations in the included header, we just document the filename of the header.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/IncludeGraph.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/IncludeGraph.hs
@@ -7,7 +7,9 @@
 module HsBindgen.Frontend.Analysis.IncludeGraph (
     IncludeGraph(..)
   , Include(..)
-  , includeArg
+  , MacroArg
+  , getIncludeArg
+  , getIncludeMacroArg
     -- * Construction
   , empty
   , register
@@ -26,6 +28,7 @@ import Data.DynGraph.Labelled (DynGraph)
 import Data.DynGraph.Labelled qualified as DynGraph
 import Data.List qualified as List
 import Data.Map.Strict qualified as Map
+import Data.Text qualified as Text
 
 import Clang.Paths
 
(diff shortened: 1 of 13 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/IncludeGraph.hs
```diff
@@ -47,11 +47,14 @@ data IncludeGraph = IncludeGraph{
...
 
 -- | Include directive as written in the source
+--
+-- When the @#include@ argument is a macro, we simply specify the filename.
 data Include =
     BracketInclude     HashIncludeArg  -- ^ @#include <...>@
   | QuoteInclude       HashIncludeArg  -- ^ @#include "..."@
   | BracketIncludeNext HashIncludeArg  -- ^ @#include_next <...>@
   | QuoteIncludeNext   HashIncludeArg  -- ^ @#include_next "..."@
+  | MacroInclude       HashIncludeArg  -- ^ Macro @#include@ argument
```
edsko: An example would help clarify these docs.

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/IncludeGraph.hs
```diff
@@ -130,3 +134,4 @@ dumpMermaid p includeGraph =
       QuoteInclude       i -> "#include \""      ++ i.path ++ "\""
       BracketIncludeNext i -> "#include_next <"  ++ i.path ++ ">"
       QuoteIncludeNext   i -> "#include_next \"" ++ i.path ++ "\""
+      MacroInclude       i -> "#include ("       ++ i.path ++ ")"
```
edsko: This of course isn't quite right, but given that it's only for the mermaid output, that doesn't really matter.

hs-bindgen/src-internal/HsBindgen/Frontend/ProcessIncludes.hs
```diff
@@ -226,12 +226,15 @@ getIncludeTo curr = do
...
+getInclude unit curr path = do
     tokens <- HighLevel.clang_tokenize unit . fmap multiLocExpansion
       =<< HighLevel.clang_getCursorExtent curr
-    let err = "Unable to parse #include: " ++ show tokens
-    maybe (panicIO err) return $ parseInclude tokens
+    return $ fromMaybe macroInclude (parseInclude tokens)
+  where
+    macroInclude :: Include
+    macroInclude = IncludeGraph.MacroInclude . RootHeader.HashIncludeArg $
+      Posix.takeFileName (getSourcePath path)
```
edsko: If this is truly only for documentation purposes, could we instead record the raw tokens, so that we can reproduce the include as-is?

(In fact, wouldn't it make sense to do that _always_...?)

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/IncludeGraph.hs
TravisCardwell: :+1: I added an example in the type documentation, with a reference to the C99 specification.

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/IncludeGraph.hs
TravisCardwell: Indeed, this is debug information shown in Mermaid graphs, so it needs to be concise.  I documented the syntax.

hs-bindgen/src-internal/HsBindgen/Frontend/ProcessIncludes.hs
TravisCardwell: From a user perspective, I would prefer to see filenames in the Haddock documentation.  For reference, we currently display documentation like this:

**defined at:** `types/structs/simple_structs.h 3:9`

I am happy to change it if you would prefer to see macros, however.  If so, what syntax would you prefer?  Example:

**included with:** `#include "types/structs/simple_structs.h"`
**source location:** `3:9`

While I pretty strongly prefer filenames in Haddock documentation, I do not have a strong preference in the debug output.  Note that keeping `#include`/`#include_next` in the debug output (not just outputting the argument) is IMHO critical, as that helps debug include issues.

TravisCardwell:
That force push just rebases and adds `#include_next MACRO` support.  Let's discuss!

hs-bindgen/src-internal/HsBindgen/Frontend/ProcessIncludes.hs
edsko: We decided to go with

**defined at**: `simple_structs.h (SOME_EXTRA_HEADER) 3:9`

Reasons:

* Only listing `simple_structs.h` might not be terribly informative when debugging, because that filename might not be very meaningful, especially when it's used deep down the bowels of some library (which seems to be the case for this kind of usage?)
* Separately listing "included as" is somewhat confusing to users, because these are not public headers; when interacting with this C code, users would (probably) never actually write that include.

Issue #1787, opened 2026-02-27
Single invocation of bindgen per golden test


edsko:
If hs-bindgen stumbles over a test case (panic, say) we currently get that exception `n` times for each golden test (one per category?). It would be better if we had a _single_ invocation that generates _all_ golden tests.

I think this is a step towards @jorisdral 's vision in https://github.com/well-typed/hs-bindgen/issues/1495, but I think won't get us _quite_ there yet (though Joris can correct me); in particular, it does not add anything clever for binding specs, or deal with multiple invocations of `hs-bindgen` for _different_ header files. It is merely about generating all _categories_ for a _single_ header at once.

jorisdral:
When working on #1495, I kind of already implemented a way to generate outputs with a single invocation. I could share that on a branch

bolt12:
Please do

dschrempf:
Just to clarify, the number of tests per golden test case depend on our
expecations.

If we expect a failure (bindgen or libclang), we run the test only once.

If we expect a success, we run

1. Template Haskell checks (1 run)
2. Preprocessor checks (5 runs, one per category)
3. Binding specification checks (1 run)

So, I think we run each test 7 times, collecting different outcomes.

ASAICS we can merge the preprocessor and binding specification tests, but we
have to keep the TH test separate, reducing the number of invocations of
`hs-bindgen` to a maximum of 2.

edsko:
Yes, I merely meant merge all invocations in (2) in your list.

Issue #1495, opened 2026-01-08
Support more elaborate Golden tests (for external binding specifications)


jorisdral:
Golden tests currently run a single invocation of `hs-bindgen` to produce some output, and then that output is compared against existing files. For the most part, this also only works for a single file at a time. Tests for binding specifications using these golden tests also exist, but for external binding specification tests the dev has to write the external binding specs by hand or include a generated one. The important bit is that the test itself does not generate an external binding spec itself, it has to be included in the golden test inputs. Moreover, the output of a golden test using an external binding spec can not be compiled using the `compile-fixtures.sh` script, because the modules that the external binding specs are written for do not exist in the fixtures directory.

My proposal would be to generalise the golden test infrastructure as follows:

* Allow golden tests to generate *directories* rather than single files. 
  * This means that the golden test should probably also be able to compare/diff directories.
* Allow running multiple invocations of `hs-bindgen` in a single test
* Allow state to be passed between invocations of `hs-bindgen`, like (filepaths to) external binding specifications.

This would allow running a test like:

* Run hs-bindgen on a header `A.h` to generate bindings a la `A.hs` and an external binding spec `A.yaml`
* Run `hs-bindgen` on a header `B.h` to generate bindings a la `B.hs` and an external binding spec `B.yaml`, using `A.yaml` as an input!
* Compare the generated output directory against a "correct" golden directory
* Because all files are placed in the same fixture directory, we can run `./compile-fixtures.sh` on the directory after

jorisdral:
I've already been experimenting with this proposal. I've so far been able to generate directories and diff them. I haven't gotten to multiple invocations of `hs-bindgen` yet, but that would be next on the list to try

edsko:
@jorisdral I've asked @bolt12 to look at https://github.com/well-typed/hs-bindgen/issues/1787, which I think is kind of a precursor to this ticket (it does not deal in any way with binding specs).

PR #1796, opened 2026-03-02
Experiment: golden tests for directories

jorisdral:
This is experimental code I wrote a while ago for #1495, and I had it lying around locally until it became relevant for #1787

A single golden test typically only deals with generating a single file and diffing it with a golden file. My idea was to extend this from *files* to *directories*. A single golden test can then generate directory and diff it with a directory.

With this new type of golden test, I changed the tests so that the pretty-printed Haskell modules are all generated in one go (for each header). This could easily be extended to generate any other outputs, like external binding specs

The code isn't documented well, and it could still be improved


jorisdral:
Note: the diff algorithm on file trees only reports the first mismatching file, but it could be changed to report all mismatches

edsko:
cc @bolt12

jorisdral:
I'll close this PR -- I have no plans for getting it over the finish line in the near future, and we can always revive the code in a later PR

PR #1797, opened 2026-03-02
Use cabal instead of ghc when compiling PP and TH fixtures

### LINKED ISSUE #1784: Use `cabal` rather than `ghc` to compile PP and TH fixtures
We currently call `cabal exec ghc --package ...`, but the names of those packages don't include hashes, causing `ghc` to potentially use the wrong package. 

Perhaps a solution is to use `cabal` on a Haskell file 

```
--
-- .. contents of the .cabal file as Haskell comment..
--
-- ..contents of the cabal.project file as a Haskell comment....
--
.. actual Haskell code ..
```

See https://cabal.readthedocs.io/en/stable/getting-started.html#running-a-single-file-haskell-script for some details.

bolt12:
Closes #1784

### DIFF
--- a/hs-bindgen/test/hs-bindgen/Test/HsBindgen/THFixtures.hs
+++ b/hs-bindgen/test/hs-bindgen/Test/HsBindgen/THFixtures.hs
@@ -16,7 +16,7 @@ import Test.Tasty.Runners (Outcome (..), Result (..))
 import Test.HsBindgen.Golden (allTestCases)
 import Test.HsBindgen.Golden.TestCase (TestCase (..))
 import Test.HsBindgen.Resources
-import Test.HsBindgen.THFixtures.Compile (setupBatchCompile, sanitizeLibName)
+import Test.HsBindgen.THFixtures.Compile (sanitizeLibName, setupBatchCompile)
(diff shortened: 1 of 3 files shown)


edsko:
Thanks @bolt12 ! Good to get this in. One thing I  noticed, it seems to wait a long time on

```
  th-fixtures
    declarations/declarations_required_for_scoping: 
```

and then after a few minutes spit out all output from all TH tests. Not sure if that's something we can do much about though (and not a big deal of course).

edsko:
I've added this to the merge queue but I'm nt sure if it doesn't conflict with #1798.

bolt12:
> Thanks @bolt12 ! Good to get this in. One thing I noticed, it seems to wait a long time on
> 
> ```
>   th-fixtures
>     declarations/declarations_required_for_scoping: 
> ```
> 
> and then after a few minutes spit out all output from all TH tests. Not sure if that's something we can do much about though (and not a big deal of course).

Yes that's because now everything is run in a cabal project, so the time it takes is cabal generating the build plan of a lot of fixtures (aka libraries). After that it should be good.

PR #1802, opened 2026-03-03
Generate bindings for `static` declarations

### LINKED ISSUE #1769: Generate bindings for `static` declarations
Given

```c
int x = 5; 
```

we generate an appropriate binding for `x`, but issue an (appropriate) warning that

```
    Bindings may result in duplicate symbols; consider using 'static' or 'extern'; or if that is not an option, consider attributing hidden visibility to the symbol
```

However, if we do what the warning suggests and make this static

```c
static int x = 5;
```

we get

```
    Parse failure:
      No bindings generated: unsupported storage class simpleEnum CX_SC_Static
```

That is problematic; for example, some device drivers for example declare statics in this way.

bolt12:
Closes #1769

### DIFF
--- a/hs-bindgen/hs-bindgen.cabal
+++ b/hs-bindgen/hs-bindgen.cabal
@@ -370,9 +370,7 @@ library test-common
     Test.Common.Util.Tasty.Golden
 
   -- Internal dependencies
-  build-depends:
-    , hs-bindgen:internal
-    , libclang-bindings
+  build-depends:   hs-bindgen:internal
 
   -- Inherited dependencies
   build-depends:


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -974,7 +975,7 @@ classifyVarDecl = \curr -> do
         case (fromSimpleEnum storage, isConst) of
           (Right CX_SC_Extern , _    ) -> return $ VarGlobal IsExtern
           (Right CX_SC_None   , _    ) -> return $ VarGlobal IsNotExtern
-          (Right CX_SC_Static , True ) -> return $ VarGlobal IsNotExtern
+          (Right CX_SC_Static , _    ) -> return $ VarGlobal IsNotExtern
```
bolt12: No need to pattern match on the isConst now

bolt12:
```c
static int counter = 0;

static int count() {
  return counter++;
}

void reset() {
  counter = 0;
}
```

have a test with a header like this and generate bindings twice (for two different Haskell modules) and then call them and verify that the static counter is indeed local to each module.

hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden.hs
```diff
@@ -424,8 +417,6 @@ test_attributes_visibility_attributes =
               Just $ Expected name
             MatchDelayed name ParseNonPublicVisibility{} ->
               Just $ Expected name
-            MatchDelayed name (MatchUnknownStorageClass CX_SC_Static) ->
-              Just $ Expected name
```
bolt12, line 428: If `MatchUnknownStorageClass` pattern synonym is not used anywhere we should delete it

edsko (approved):
Thanks for adding that test case, looks good :)

PR #1804, opened 2026-03-03
Manual section on the pointer manipulation API

### LINKED ISSUE #1288: Write text for the manual section on zero-copy bindings
A `ZeroCopy` module is added in #1247, but there is no accompanying text that explains how the approach to zero-copy bindings works. We should add some text there.

### LINKED ISSUE #1645: Use "Pointer manipulation API" instead of "Zero-copy API" throughout

jorisdral:
Resolves #1288
Resolves #1645

"Pointer manipulation API" is the new name for the "zero-copy API". There is now a textual manual section for the pointer manipulation API, which has been adapted from the old "zero-copy API" manual section in the manual executable. I've removed the more extensive examples from the manual section in favour of
simplicity. These examples will probably be revived later as tests in the `feature-tests` example.

Other changes:
* Removed the golden test for the old "zero-copy API" manual section.
* Replaced the term "zero-copy API" with "pointer manipulation API" throughout the repository.


manual/LowLevel/Translation/05-PointerManipulation.md
```diff
@@ -0,0 +1,109 @@
+# Pointer manipulation API
+
+Suppose we want to override one value nested in some C data structure. We could
+use the `Storable` instances to `peek` the value, then override the appropriate
+field, and finally `poke` the updated value. However, it may well be undesirable
```
dschrempf: "it is undesirable"? is there a case where it is desirable? I can't really think of one. (Maybe to keep referential transparency in non-monadic code?)

manual/LowLevel/Translation/05-PointerManipulation.md
```diff
@@ -0,0 +1,109 @@
...
+deriving ... instance Storable Point
+
+data Rectangle = Rectangle
+  { rectangle_topleft :: Point
+  , rectangle_bottomright :: Point
+  }
+deriving ... instance Storable Rectangle
+```
+
+If we want to modify a pointer to a `rectangle`, then we could choose to fully
```
dschrempf:
```suggestion
If we want to modify the target of a `rectangle` pointer, we could choose to fully
```
(?)

manual/LowLevel/Translation/05-PointerManipulation.md
```diff
@@ -0,0 +1,109 @@
...
+marshall the value to a Haskell value and back. We first read the value off the
+pointer, modify it, then write it back[^2]:
+
+```hs
+example1 :: Ptr Rectangle -> IO ()
+example1 rectPtr = do
+      rect <- peek rectPtr
+      poke rectPtr $ rect & #rectangle_topleft % #point_x .~ 2
```
dschrempf, line 73: Maybe we could separate the second phase "modify the value" into a `let` binding, to make it even clearer?

manual/LowLevel/Translation/05-PointerManipulation.md
jorisdral: Good question. I suppose I assume that a Haskeller's usual approach would be to read the full value off the pointer to modify it and then write it back, rather than moving around the pointer to the right offset and doing the modification there. The benefit of the former is that a developer can treat C data structures as pure Haskell values without having to care too much about pointers. And that's fine in many cases unless you care about the minutiae of performance and your structures are large, in which case I'd recommend using the pointer manipulation API

jorisdral: In any case, since there are so many ways to interact with C data structures and pointers, I would frame it as "may or may not be desirable"

manual/LowLevel/Translation/05-PointerManipulation.md
jorisdral: makes sense! will do

manual/LowLevel/Translation/05-PointerManipulation.md
jorisdral: I've added a new paragraph. If you are asking yourself the question about whether or not it is desirable to use the API, then future readers might too :smiley: 

> Whether or not it is desirable to use the pointer manipulation API depends on
your case. Our recommendation would be to use it if your data structures are
large and you care about performance.

jorisdral:
@edsko I'm going to put this PR on the merge queue to get the bulk of the changes onto `main`. If you have notes on the new manual section, let me know and I'll open a follow-up PR!

PR #1810, opened 2026-03-04
Add `--fast` flag to `test-hs-bindgen`

### LINKED ISSUE #1785: Add `--fast` command line option to `test-hs-bindgen`
As the test suite is growing, and especially with the addition of the TH compilation tests, it is getting slower, which can be detrimental for the dev cycle. Let's add a `--fast` option (using `tasty`'s support for options, see https://hackage-content.haskell.org/package/tasty-1.5.3/docs/Test-Tasty-Options.html), which we can use to disable certain tests:

* The TH fixture compilation tests
* The `.Unsafe` tests (if `.Safe` passes, `.Unsafe` almost certainly will also)
* possibly others?

Marking this for release 0.1 as it will help development.

bolt12:
This closes #1785 and #1811 and #1815

### DIFF
--- a/hs-bindgen/test/hs-bindgen/Test/HsBindgen/THFixtures.hs
+++ b/hs-bindgen/test/hs-bindgen/Test/HsBindgen/THFixtures.hs
@@ -31,7 +31,7 @@ import Test.HsBindgen.THFixtures.TestCases (THStatus (..), determineTHStatus)
 -- to this point (e.g. @[\"test-hs-bindgen\"]@). 'testPatternMatches' needs
 -- full paths to replicate tasty's @-p@ filtering.
 tests :: Seq String -> IO TestResources -> TestTree
-tests pathPrefix testResources = askOption $ \case
+tests pathPrefix getTestResources = askOption $ \case
     Fast -> testGroup "th-fixtures" []
     Full -> askOption $ \testPattern ->
       withResource (setup testPattern) (const $ return ()) $ \getResults ->
@@ -41,10 +41,10 @@ tests pathPrefix testResources = askOption $ \case
     -- Only compile fixtures that match the @-p@ pattern. Without this,
     -- 'withResource' compiles all fixtures regardless of @-p@.
     setup testPattern = do
-        testResources' <- testResources
-        let repoRoot = takeDirectory testResources'.packageRoot
+        testResources <- getTestResources
+        let repoRoot = takeDirectory testResources.packageRoot
             cases =
-              [ (tc.name, generateModule testResources' tc)
+              [ (tc.name, generateModule testResources tc)
               | tc <- allTestCases
               , determineTHStatus tc == THCompile
               , testPatternMatches testPattern


hs-bindgen/test/hs-bindgen/test-hs-bindgen.hs
```diff
@@ -36,3 +40,7 @@ main = defaultMain $ withTestResources $ \testResources ->
           ]
       , THFixtures.tests testResources
       ]
+  where
+    ingredients =
+        includingOptions [Option (Proxy :: Proxy RunMode)]
+      : defaultIngredients
```
dschrempf: Why don't we add the option to the other ones in Golden.hs?

dev/testing.md
```diff
@@ -8,6 +8,24 @@ This document covers tests and how to run them for the `hs-bindgen` project.
 
 ## Running Tests
 
+### Fast Mode
```
dschrempf, line 11: Now that you add a `--fast`-mode section (thank you!), maybe you can document the other two options `--accept` and `--debug`?

bolt12: Good point! But --accept is already documented in a different section, I should reference it here too, perhaps

hs-bindgen/test/hs-bindgen/test-hs-bindgen.hs
bolt12: Thanks for pointing it out I was over complicating!

dschrempf:
Just making sure that this doesn't get lost: Tasty filters should work also for TH tests (at the moment, the TH fixture Cabal project always builds the complete set of TH fixtures); @bolt12 and I have discussed this! (Actually, this doesn't belong here, should we create an issue @bolt12?)

bolt12:
Done! Thank you @dschrempf !

edsko:
@TravisCardwell are you happy with that final commit fixing #1815? 
(Armando tells me it depends on `-j1`, but I think more understandable debug output is probably worth it.)

hs-bindgen/test/hs-bindgen/Test/HsBindgen/THFixtures.hs
```diff
@@ -24,19 +27,28 @@ import Test.HsBindgen.THFixtures.TestCases (THStatus (..), determineTHStatus)
...
         testGroup "th-fixtures" $
-            map (mkFixtureTest getResults) allTestCases
+          map (mkFixtureTest getResults) allTestCases
   where
-    setup = do
+    -- Only compile fixtures that match the @-p@ pattern. Without this,
+    -- 'withResource' compiles all fixtures regardless of @-p@.
+    setup testPattern = do
         testResources' <- testResources
         let repoRoot = takeDirectory testResources'.packageRoot
```
TravisCardwell: The `IO` action should be named `getTestResources`, so that the value can be named just `testResources`.  I changed this recently; perhaps it was lost in a rebase.

Issue #1813, opened 2026-03-05
Nondeterministic `th-fixtures` failures


TravisCardwell:
I puzzled over various `th-fixtures` failures while testing my branch.  I rebased, however, and I confirmed that I still observe such failures in `main` ([`8dc549fd`](https://github.com/well-typed/hs-bindgen/tree/8dc549fd9060509337a6e30e22c9658146c5604d)).

```
types/typedefs/multi_level_function_pointer:                FAIL
  TH compilation failed:
  Error: [Cabal-6578]
  '.../ghc/9.10.3/bin/ghc-pkg-9.10.3' exited with an error:
  ghc-pkg-9.10.3: .../dist-newstyle-th-fixtures/packagedb/ghc-9.10.3/hs-bindgen-th-fixtures-0.0.0-inplace-th-fixture-types-qualifiers-type-qualifiers.conf: getModificationTime:getFileStatus: does not exist (No such file or directory)
  HasCallStack backtrace:
    collectBacktraces, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:169:13 in ghc-internal:GHC.Internal.Exception
    toExceptionWithBacktrace, called at libraries/ghc-internal/src/GHC/Internal/IO.hs:260:11 in ghc-internal:GHC.Internal.IO
    throwIO, called at libraries/ghc-internal/src/GHC/Internal/IO/Exception.hs:315:19 in ghc-internal:GHC.Internal.IO.Exception
    ioException, called at libraries/ghc-internal/src/GHC/Internal/IO/Exception.hs:319:20 in ghc-internal:GHC.Internal.IO.Exception

  Error: [Cabal-7125]
  Failed to build lib:th-fixture-types-typedefs-multi-level-function-pointer from hs-bindgen-th-fixtures-0.0.0. The failure occurred during the configure step.


  Use -p '/th-fixtures.types\/typedefs\/multi_level_function_pointer/' to rerun this test only.
```

Notes:

* `ghc-pkg` is failing, during the configure step.  The actual errors vary, but the root cause is always not finding a file in the package database.  Investigating, it looks like this error generally indicates a broken package database.
* Different tests fail with each run.  Sometimes all tests pass.
* I still observe such errors when using `-j1`.

I now get such errors very often locally, yet I have not seen them in CI.

TravisCardwell:
[...]

PR #1814, opened 2026-03-05
Compile TH fixtures with `-j1`

TravisCardwell:
This is an attempt to avoid nondeterministic failures where `ghc-pkg` is unable to find files in the package database during the configure step. I have not had any failures while using `-j1`, though I do not know if this makes failure impossible or just more unlikely.  This commit sets `-j1`, and we can see if the issue pops up again.

See #1813

### DIFF
--- a/hs-bindgen/test/hs-bindgen/Test/HsBindgen/THFixtures/Compile.hs
+++ b/hs-bindgen/test/hs-bindgen/Test/HsBindgen/THFixtures/Compile.hs
@@ -52,7 +52,9 @@ setupBatchCompile repoRoot cases =
             [ "build"
             , "hs-bindgen-th-fixtures"
             , "--keep-going"
-            , "-j"
+            -- Attempt to avoid nondeterministic failures; see
+            -- <https://github.com/well-typed/hs-bindgen/issues/1813>
+            , "-j1"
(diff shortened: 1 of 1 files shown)


hs-bindgen/test/hs-bindgen/Test/HsBindgen/THFixtures/Compile.hs
```diff
@@ -52,7 +52,7 @@ setupBatchCompile repoRoot cases =
             [ "build"
             , "hs-bindgen-th-fixtures"
             , "--keep-going"
-            , "-j"
+            , "-j1"
```
edsko, line 57: Let's add a comment with a link to the issue here. Feel free to merge after that :)

bolt12:
I already had a commit here that limited it to -j4 (https://github.com/well-typed/hs-bindgen/pull/1810/changes/47ed28c47ac7cd728c48feaf70731fbff6e51699).

CI is dual core so it won't be a problem there, but for dev work we should limit it to something reasonable

edsko:
The problem is that _anything other than 1_ can result in non-deterministic failures (if the hypothesis is right that it's due to that).

bolt12:
Yes, it is due to parallelism, AFAIU

PR #1823, opened 2026-03-09
Bump `s-r-p` for `libclang-bindings`

jorisdral:
Work in progress


dschrempf:
Just because I stumbled upon this; there is a script which does that canonically also for the Nix Flake sources!

jorisdral:
> Just because I stumbled upon this; there is a script which does that canonically also for the Nix Flake sources!

Cool! I'll keep that in mind and use it once I open this PR for review

PR #1824, opened 2026-03-09
Invoke hs-bindgen once per PP golden test

### LINKED ISSUE #1787: Single invocation of bindgen per golden test

bolt12:
This closes #1787

### DIFF
--- a/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden/Check/PP.hs
+++ b/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden/Check/PP.hs
@@ -2,10 +2,13 @@
 module Test.HsBindgen.Golden.Check.PP (check) where
 
 import Control.Monad (when)
+import Data.IORef
+import Data.List.NonEmpty qualified as NonEmpty
 import Optics.Core (view)
 import System.Directory (createDirectoryIfMissing)
 import System.FilePath ((</>))
 import Test.Tasty
+import Test.Tasty.HUnit (assertFailure)
 
 import HsBindgen (getBindingsMultiple)
 import HsBindgen hiding (getBindingsMultiple)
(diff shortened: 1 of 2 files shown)


edsko:
@dschrempf could I ask you to review this? @bolt12 feel free to merge once Dominik gives the green light.

hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden/Check/PP.hs
```diff
@@ -24,51 +28,71 @@ import Test.HsBindgen.Resources
...
-            <- runTestHsBindgenSuccess report getTestResources test artefacts
+          result <- getResult
+          case result of
+            Left e
+              -- The primary sub-test re-throws the original exception so that
+              -- tasty displays the full error. The other sub-tests just
+              -- reference it, avoiding redundant error output.
+              | bc == NonEmpty.head allCategories -> throwIO e
+              | otherwise   ->
+                assertFailure "hs-bindgen failed (see CType for details)"
```
dschrempf: What happens if we only test for a Term category (i.e., not for CType)?

hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden/Check/PP.hs
```diff
@@ -24,51 +28,71 @@ import Test.HsBindgen.Resources
...
+              qualifiedStyle = view #qualifiedStyle backendCfg
+            }
+          fns = view #fieldNamingStrategy backendCfg
+          artefacts = (,)
+                    <$> FinalModuleBaseName
+                    <*> getBindingsMultiple fns mrc
+      (baseName, output) <-
+        runTestHsBindgenSuccess collectTrace getTestResources test artefacts
+      msgs <- reverse <$> readIORef traceMessages
+      pure (baseName, msgs, output)
```
dschrempf: This seems like some unnecessary code duplication (but I may be wrong). Did you have a look at `runTestHsBindgen`, it returns an `Eiter` already, and doesn't expect a success like `runTestHsBindgenSuccess`.

I think you do this because you need the trace messages, right?

hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden/Check/PP.hs
bolt12: Good point, I should use the show instance for `bc`

hs-bindgen/test/hs-bindgen/Test/HsBindgen/Golden/Check/PP.hs
bolt12: Yes that's right, I will look into that! I might have overlooked something!

bolt12: addressed! Could you take a look now?

PR #1830, opened 2026-03-11
Add CEF binding generation example

### LINKED ISSUE #1791: Creating bindings for the Chromium Embedded Framework
See https://github.com/chromiumembedded .

This might be an interesting use case, because it seems to require some specific selection predicates; see  https://redirect.github.com/jwiegley/c2hsc/issues/22 for an attempt to to do this with `c2hsc`.

### LINKED ISSUE #1837: Functions with nested struct declarations should be skipped
@bolt12 I am not sure if there was already a ticket for this; this is to track the change you need to make to deal with the Chrome bindings, where we throw out function declarations if they contain struct declarations, such as

```c
void f(struct foo ... x);
```

(where `struct foo` is not defined previously, not even a forward declaration).

Marking this as a bug because not doing this results in name mangler problems.

bolt12:
This closes #1791 and closes #1837

### DIFF
--- a/hs-bindgen/fixtures/functions/decls_in_signature/Example/Unsafe.hs
+++ b/hs-bindgen/fixtures/functions/decls_in_signature/Example/Unsafe.hs
@@ -4,8 +4,6 @@
 
 module Example.Unsafe
     ( Example.Unsafe.normal
-    , Example.Unsafe.f1
-    , Example.Unsafe.f2
     )
   where
 
@@ -22,18 +20,6 @@ $(HsBindgen.Runtime.Internal.CAPI.addCSource (HsBindgen.Runtime.Internal.CAPI.un
   , ")"
   , "{"
   , "  (normal)(arg1, arg2, *arg3);"
-  , "}"
-  , "void hs_bindgen_dad6e1aa83dec458 ("
-  , "  struct named_struct *arg1"
-  , ")"
-  , "{"
-  , "  (f1)(*arg1);"
-  , "}"
-  , "void hs_bindgen_e6bb0f3956383df9 ("
-  , "  union named_union *arg1"
-  , ")"
-  , "{"
-  , "  (f2)(*arg1);"
   , "}"
   ]))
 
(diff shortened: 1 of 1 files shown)


hs-bindgen/examples/golden/functions/proto_scope_forward_decl.h
```diff
@@ -0,0 +1,2 @@
+// struct with no prior file-scope declaration — prototype-scoped forward ref
+void f(struct foo* arg);
```
bolt12: add:

```
void g(struct foo {int x; char c;}* arg);

void g(struct {int x; char c;}* arg);
```

These should be thrown out.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type.hs
```diff
@@ -134,6 +136,39 @@ typeRef :: MonadIO m => CXCursor -> CTagKind -> m (C.Type Parse)
...
+--    @void f(struct s {int x;} a)@) has @isDef = True@ and a complete type,
+--    so we allow it. A bare reference like @void f(struct s *p)@ where
+--    @struct s@ has no file-scope declaration has @isDef = False@.
+--
+-- Examples:
+--
+-- @
+-- struct opaque;
+-- void f(struct opaque *p);           -- OK: parent is TranslationUnit
+-- void g(struct s { int x; } arg);    -- OK: isDef = True
```
bolt12: This example should also be thrown away.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Msg.hs
```diff
@@ -93,6 +94,12 @@ data DelayedParseMsg =
       -- | The symbol has public visibility
       Bool
 
+    -- | Struct/union tag first declared inside a function prototype, e.g.
+    -- @void f(struct foo* arg);@
+    --
+    -- Might indicate a missing @#include@ in the header.
+  | ParseDeclarationOutOfScope CTagKind Text
```
bolt12:
```suggestion
  | ParseDeclarationNotVisible CTagKind Text
```

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Msg.hs
```diff
@@ -317,6 +324,11 @@ instance PrettyForTrace DelayedParseMsg where
           "Unusable anonymous declaration "
         , prettyForTrace anonId
         ]
+      ParseDeclarationOutOfScope _kind name -> PP.hcat [
+            "Unexpected out of scope struct/union declaration '"
+          , PP.text name
+          , "'"
```
bolt12: Make sure this is similar to what Clang says:

> test.h:1:15: warning: declaration of 'struct foo' will not be visible outside of this function [-Wvisibility]

examples/cef/hs-project/app/Main.hs
```diff
@@ -0,0 +1,96 @@
...
+                                  else exitWith (ExitFailure (fromIntegral exitCode))
+            else do
+                -- Zero-initialize settings, set required fields
+                let settingsSize = sizeOf (undefined :: C_Cef_settings_t)
+                allocaBytes settingsSize $ \(settingsPtr :: Ptr C_Cef_settings_t) -> do
+                    fillBytes settingsPtr 0 settingsSize
+                    HasCField.writeRaw (Proxy @"c_Cef_settings_t_size")
+                        settingsPtr (fromIntegral settingsSize)
+                    HasCField.writeRaw (Proxy @"c_Cef_settings_t_no_sandbox")
+                        settingsPtr 1
```
bolt12: We can use the record dot syntax here. Look into the documentation of Runtime.HasCField.

examples/cef/generate-and-run.sh
bolt12, line 1: use enable record dot flag (which has been renamed to --omit-field-prefixes) when generating the bindings

PR #1833, opened 2026-03-12
Support language-c 0.9.x and 0.10.2

### LINKED ISSUE #1662: Support language-c-0.9-*
We refer to `CBFloat16Type`, which was introduced in `language-c`.

bolt12:
This closes #1662.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/HandleMacros.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/HandleMacros.hs
@@ -29,6 +29,7 @@ import HsBindgen.Frontend.Pass.HandleMacros.Error
 import HsBindgen.Frontend.Pass.HandleMacros.IsPass
 import HsBindgen.Frontend.Pass.Parse.IsPass
 import HsBindgen.Imports
+import HsBindgen.Language.C (PrimType (PrimBool))
 
 {-------------------------------------------------------------------------------
   Top-level
@@ -463,8 +464,19 @@ parseMacro name tokens  = state     $ \st ->
               (Left $ HandleMacrosErrorParse errType errExpr, st)
   where
     updateReparseEnv :: C.Type HandleMacros -> LanC.ReparseEnv -> LanC.ReparseEnv
-    updateReparseEnv typ =
-        Map.insert name.text $
+    updateReparseEnv typ
+      -- stdbool.h defines @#define bool _Bool@, which is just an alias for
+      -- the primitive boolean type. We normalise this away: if @bool@ maps
+      -- to @PrimBool@, we store @TypePrim PrimBool@ directly instead of
+      -- wrapping it in @TypeMacro@. This ensures that @bool@ from stdbool.h
+      -- renders identically to @_Bool@, regardless of language-c version.
+      -- Genuine redefinitions like @#define bool int@ are not affected
+      -- because their underlying type is not @PrimBool@.
+      | name.text == "bool"
+      , C.TypePrim PrimBool <- typ
+      = Map.insert name.text typ
+      | otherwise
+      = Map.insert name.text $
           C.TypeMacro $ C.Ref {
               name = DeclId{name = name, isAnon = False}
             , underlying = typ
--- a/hs-bindgen/fixtures/binding-specs/stdlib/instances.1.c11-parse-all/Example.hs
+++ b/hs-bindgen/fixtures/binding-specs/stdlib/instances.1.c11-parse-all/Example.hs
@@ -58,7 +58,7 @@ import qualified HsBindgen.Runtime.Marshal as Marshal
     __exported by:__ @binding-specs\/stdlib\/instances.h@
 -}
 newtype Stdlib_CBool = Stdlib_CBool
-  { unwrapStdlib_CBool :: HsBindgen.Runtime.LibC.CBool
+  { unwrapStdlib_CBool :: RIP.CBool
   }
   deriving stock (Eq, RIP.Generic, Ord, Read, Show)
   deriving newtype
@@ -79,7 +79,7 @@ newtype Stdlib_CBool = Stdlib_CBool
     , Marshal.WriteRaw
     )
 
-instance ( ((~) ty) HsBindgen.Runtime.LibC.CBool
+instance ( ((~) ty) RIP.CBool
          ) => RIP.HasField "unwrapStdlib_CBool" (RIP.Ptr Stdlib_CBool) (RIP.Ptr ty) where
 
   getField =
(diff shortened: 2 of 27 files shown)


bolt12:
We pass the C standard in the places that need it! (grep for `cStandard`)

edsko:
A grep for `cStandard` does not show up any hits in our `LanguageC` infra?

bolt12:
Then I misunderstood, you mean that we need to pass something to language-c to tell it which standard we want? 

Looking at the code, I don't think we need to tell language-c nothing, i.e `LanC.execParser` doesn't take that as argument? AFAIU the bool support is now in language-c parser and will handle it natively irrespective of which C standard one is using

edsko:
But that's wrong. If we're using C < C23, then if somebody uses `bool`, and it's defined somewhere 

```c
#define bool int
```

then it should not be parsed as a keyword.

edsko:
So if we cannot tell `language-c` what we want, perhaps we need to change how we translate the `language-c` AST.

edsko:
Though _maaaaybe_ this "just" works, because _if_ `bool` is defined in this way (like a macro, or a typedef), then we will add it into the language-c type environment, and maybe that willl then take precendence? If so, there is nothing to fix, but we should verify that then and document it if so.

edsko:
@bolt12 Thinking some more about this.

**If the C standard is C23**

A test case would be something like

```c
// This test requires C23

#define A int
void f(A x, bool y);

#define bool int
void g(A x, bool y);
```

Now `f` and `g` should get different signatures (one `CBool`, the other a newtype wrapper we generate). 

**If the C standard is older than C23**

I guess the flipside of this now is that if code is _not_ intended for C23, and uses `bool` without defining it anywhere, _and_ we reparse it, something like

```c
// This test requires C89 (or higher, but _not_ C23)

#define A int
void f(A x, bool y);
```

then `language-c` would parse that `bool` as if it were a primitive, even though that's not justified. But I guess that's OK, because `clang` would already have rejected this test.

However, if we do

```c
// This test requires C89 (or higher, but _not_ C23)
#define bool int
void g(bool y);
```

then that should result in `g` using the `Bool` newtype wrapper we generate (around `CInt`), _not_ `CBool`.

bolt12:
I will look into that! Thanks @edsko ! It also seems I wasn't running with language-c 0.10.2 (I forgot to update my index-state), so I am re-running the tests.

bolt12:
Edit: I was still not running with language-c-0.10.2 due to nix. So I re-ran the tests and they are much different

---

## C23 — bool as keyword vs macro-redefined

```
$ cat /tmp/test_bool_c23_1.h
#define A int
void f(A x, bool y);

#define bool int
void g(A x, bool y);
```

```
$ cabal run hs-bindgen-cli -- internal frontend --pass handle-macros \
    --clang-option-after "-std=c2x" --select-all /tmp/test_bool_c23_1.h
```

- f(A x, bool y) — x: TypeMacro "A" -> PrimInt, y: TypePrim PrimBool (keyword)
- g(A x, bool y) — x: TypeMacro "A" -> PrimInt, y: TypePrim PrimBool (WRONG — should be PrimInt)

## C89 — undefined bool

```
$ cat /tmp/test_bool_c89_1.h
#define A int
void f(A x, bool y);
```

```
$ cabal run hs-bindgen-cli -- internal frontend --pass handle-macros \
    --clang-option-after "-std=c89" --select-all /tmp/test_bool_c89_1.h
```

Result: error: unknown type name 'bool'

## C89 — #define bool int

```
$ cat /tmp/test_bool_c89_2.h
#define bool int
void g(bool y);
```

```
$ cabal run hs-bindgen-cli -- internal frontend --pass handle-macros \
    --clang-option-after "-std=c89" --select-all /tmp/test_bool_c89_2.h
```

- g(bool y) — y: TypePrim PrimBool (WRONG — should be PrimInt)

## typedef bool int (C23)

```
$ cat /tmp/test_bool_typedef.h
#define A int
typedef bool int;
void f(A x, bool y);
```

```
$ cabal run hs-bindgen-cli -- internal frontend --pass handle-macros \
    --clang-option-after "-std=c2x" --select-all /tmp/test_bool_typedef.h
```

error: cannot combine with previous 'bool' declaration specifier

edsko:
> libclang does macro expansion before the tokens reach language-c's reparser, so language-c never sees bare bool when it's been #define'd away

No, this is inaccurate. The whole _point_ of reparsing with language-c is that we recover macro information. That's how we are able to see `bool` or `A` instead of `int`, after all. We get the _original_ tokens from the C source, _then_ pass hem to `language-c`. 

But your tests do seem to suggest that if we add `bool` into the `language-c` type environment, it takes precedence over its parser for primitive `bool`. That's good; let's add these test cases (especially the first one) into the test suite, with a comment that says that we're testing that if `bool` is given a definition, then it is _not_ parsed as `PrimBool`.

And while we're at it, let's just for good measure add a second, similar, test case where we use a `typedef` instead of `#define` for `bool`:

```c
#define A int
typedef bool int;
void f(A x, bool y);
```

With those test cases added, I think this PR is good to merge :)

bolt12:
> No, this is inaccurate. The whole point of reparsing with language-c is that we recover macro information. That's how we are able to see bool or A instead of int, after all. We get the original tokens from the C source, then pass hem to language-c.

I see, I thought that was the case, but this makes more sense!

Just to make sure, should the new test be run with C23 ?

edsko:
I guess C23 or not doesn't really matter for that final test.

edsko:
We should not merge this PR while there are fixture changes, unless we really understand where they are coming from.

bolt12:
I am looking into it!

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/HandleMacros.hs
```diff
@@ -463,8 +464,19 @@ parseMacro name tokens  = state     $ \st ->
...
+      -- stdbool.h defines @#define bool _Bool@, which is just an alias for
+      -- the primitive boolean type. We normalise this away: if @bool@ maps
+      -- to @PrimBool@, we store @TypePrim PrimBool@ directly instead of
+      -- wrapping it in @TypeMacro@. This ensures that @bool@ from stdbool.h
+      -- renders identically to @_Bool@, regardless of language-c version.
+      -- Genuine redefinitions like @#define bool int@ are not affected
+      -- because their underlying type is not @PrimBool@.
+      | name.text == "bool"
+      , C.TypePrim PrimBool <- typ
+      = Map.insert name.text typ
```
bolt12, line 477: With this example:

```
typedef int _Bool;
#define A int
void f(_Bool x, A y);
```

language-c would not see the typedef.

PR #1839, opened 2026-03-17
Detect implicit fields and generate bindings for them

### LINKED ISSUE #1649: Support implicit fields
Implicit fields arise when anonymous declarations are nested inside of other declarations, seemingly unused (that is, without a corresponding field name). One problem with supporting implicit fields is that while the clang C++ AST records them, they are not visible to `libclang` AST traversals. This means that solving this issue will require work-arounds (if possible at all).

Examples:

## struct-in-struct

```c
struct outer {
  ..
  struct {
    int fieldX;
    int fieldY;
  }; // <-- the anonymous inner struct is an implicit field of the outer struct  
  ..
};
```

In this case, `fieldX` and `fieldY` can be considered to be fields of the `outer` struct. 

Asking `libclang` for the offset of `fieldX` will give us the offset of `fieldX` within the _inner_ struct, and since the implicit field itself is invisible in `libclang`, we cannot ask for its offset within the _outer_ struct. However, @jorisdral believes that we can use a different libclang API to get the offset of `fieldX` within the _outer_ struct directly. If that is the case, we can support this use case.

## struct-in-union

```c
union outer {
  ..
  struct {
    int fieldX;
    int fieldY;
  }; // <-- the anonymous inner struct is an implicit field of the outer union  
  ..
}
```

This example is awkward for a different reason: the offset of the implicit field is not relevant, because all fields in a union have offset 0. However, now flattening does not make sense: it's the struct _as a whole_ that is a field of the union; flattening this so that `fieldX` and `fieldY` would become fields of the union directly would mean that `fieldX` and `fieldY` now "overlap"; that would be incorrect. This means that we'd have to invent a name for the struct.

## union-in-struct

```c
struct outer {
  ..
  union {
    int fieldX;
    int fieldY;
  }; // <-- the anonymous inner union is an implicit field of the outer struct  
  ..
};
```

In this example we have _both_ problems: we need to know of the offset of the implicit field inside the `outer` struct (though as above, if we can ask `libclang` for the offset of `fieldX` within `outer` that would suffice), _and_ we have to invent a name for the union; here too flattening is not correct.

jorisdral:
Buillds on top of #1840 and #1849

Resolves #1649

Implicit fields for nested anonymous structs and unions are not reported by `libclang`, so we detect them instead. The key observation is that the offset to a nested anonymous struct or union is equal to the offset to its first field. Using the offset to the first field, we generate a so-called implicit field and we include it as a struct/union field in the C AST. Haskell bindings are generated for such implicit fields like for any other explicit field.

The implicit field detection algorithm is the same for any nesting of structs or unions, in any order, even recursively.

The implicit field detection algorithm does rely on two conditions:

1. the anonymous nested struct or union should have at least one field
2. the anonymous nested struct or union should have only *named* fields

Concretely, this means that empty anonymous structs/unions and anonymous structs/unions with unnamed bit-fields are not supported. A warning-level trace message will be emitted if these conditions are not met.

Anonymous nested structs/unions have no name, but they need one for our Haskell bindings, so they are named after their first field. Informally, the former will be transformed to the latter:

```c
struct S {
  struct {
    int x;
  };
};
```
```c
struct S {
  struct {
    int x;
  } x;
};
```

### DIFF
--- a/hs-bindgen/CHANGELOG.md
+++ b/hs-bindgen/CHANGELOG.md
@@ -50,6 +50,11 @@
   * Generate an `IsArray` instance for each newtype of a type with an `IsArray`
 * Support unnamed bit-field declarations (of width zero), used to instruct the
   compiler to not pack any more fields into the current storage unit.
+* Generate bindings for nested struct and union declarations even if we failed
+  to generate bindings for the enclosing struct or union. See [PR
+  #1849][pr-1849].
+* Generate bindings for nested anonymous structs and unions. See [PR
+  #1839][pr-1839].
 
 ### Minor changes
 
@@ -100,6 +105,8 @@
 [pr-1712]: https://github.com/well-typed/hs-bindgen/pull/1712
 [pr-1724]: https://github.com/well-typed/hs-bindgen/pull/1724
 [issue-1806]: https://github.com/well-typed/hs-bindgen/issues/1806
+[pr-1839]: https://github.com/well-typed/hs-bindgen/pull/1839
+[pr-1849]: https://github.com/well-typed/hs-bindgen/pull/1849
 
 ## 0.1.0-alpha -- 2026-02-06
 


georgefst:
I know this is still a draft, but I couldn't resist trying it out. This _does_ allow me to remove my hack I mentioned in https://github.com/well-typed/hs-bindgen/issues/1649#issuecomment-3994425922, which is pretty awesome. And all I have to do to adapt is replace "body" with "anon0" for the names of some of the generated functions I'm calling.

jorisdral:
> I know this is still a draft, but I couldn't resist trying it out. This _does_ allow me to remove my hack I mentioned in [#1649 (comment)](https://github.com/well-typed/hs-bindgen/issues/1649#issuecomment-3994425922), which is pretty awesome. And all I have to do to adapt is replace "body" with "anon0" for the names of some of the generated functions I'm calling.

Very good to hear :smiley:

The PR was in draft mode only because I had to clean up code/documentation a bit before opening it up for review, but the feature should work already

edsko:
@georgefst We'll still make some minor changes here though (in particular, the nameless structs/unions will be named after their first field, rather than numbered).

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/ImplicitFields.hs
```diff
@@ -0,0 +1,442 @@
...
+     -- | The enclosing object
+  -> EnclosingObject
+     -- | An anonymous object nested in the enclosing object
+  -> C.Decl Parse
+  -> M m (field Parse)
+getImplicitField cntr encObj decl = do
+    targets' <- liftEither $ checkNonEmpty targets
+    offsets <- mapM offsetOf' targets'
+    name <- genImplicitFieldName cntr
+    let (origin, offset) = minimumBy (\x y -> compare (snd x) (snd y)) offsets
```
jorisdral: We should add a comment here that the minimum offset to an indirect field is also the offset to the struct because there can be no padding at the start of a struct

jorisdral (commented):
@edsko and I reviewed the PR and there are still some  TODOs that I will list here:

- [x] We should see what happens when we use `--omit-field-prefixes` with this new feature
  * EDIT: this works as intended: the prefixes are omitted
- [x] We should test Haskell bindings that we generate for structs/unions with implicit fields. These tests can live in the `feature-tests` example.
  * EDIT: added tests that check that the offsets for implicit fields in Haskell match the offsets for the same implicit fields in C
- [ ] Documentation that still needs writing:
  - [ ]  A description of the algorithm in the `ImplicitFields` module
  - [ ] The manual section on implicit fields and anonymous nested structs/unions
- [x] Implicit fields should be named after the name of the first field of the corresponding anonymous struct/union
  * EDIT: implicit fields copy the name of the first field of the referenced anonymous object. For example, if the first field of an anonymous object is called `x`, then the implicit field referencing that anonymous object is also called `x`. The `AssignAnonIds` pass then makes sure that the anonymous object gets a name based on the implicit field with name `x`.
- [ ] Unnamed bitfields can add padding at the start of anonymous structs and unions. This leads to an error currently in the implicit field detection algorithm

jorisdral:
@TravisCardwell, @edsko mentioned that we should think about how these generated implicit fields interact with binding specifications. Related issues: #1357, #1379.

TravisCardwell:
The significant change is that we now create data types for anonymous declarations that do not have field names, generating the type names.

Binding specifications use `ctypes` to specify Haskell type names as well as omit declarations, `hstypes` is used to specify details about the type.  If we can identify such an anonymous declaration within `ctypes`, then specifying names of the constructor and fields, instances, etc. in `hstypes` would be done like with all other types (#1357).  We would need `cname`s for such anonymous types, and indeed I see that `AssignAnonIds` is assigning an identifier.

Example:

```c
struct outer1 {
  char fieldA;
  struct {
    int fieldX;
    int fieldY;
  };
  int fieldC;
};
```

Binding specification:

```yaml
...
ctypes:
  - headers: types/anonymous/struct_in_struct.h
    cname: struct @outer1_anon0
    hsname: Outer1_anon0
```

Example prescriptive binding specification that customizes the names and specifies and `Ord` instance:

```yaml
...
ctypes:
  - headers: types/anonymous/struct_in_struct.h
    cname: struct @outer1_anon0
    hsname: Point
hstypes:
  - hsname: Point
    representation:
      record:
        constructor: Point
        fields:
          - x
          - y
    instances:
      - Ord
```

This works fine for external binding specifications.

One thing to consider with generated names is stability.  When using a counter, the name of a type may change if another anonymous type is inserted before it.  IMHO it is acceptable to be expected to modify the specifications for a type when modifying the type.  A comment above says that the anonymous types will be named after their first field rather than numbered.  (Is this still the plan, but it is just not implemented yet?)  A benefit of this strategy is that the generated names are stable.

Some developers likely prefer the counter names (`outer1_anon0` or just `anon0` when using `OverloadedRecordDot`, or perhaps even `field2`), but binding specifications allow them to specify their desired name in the generated API, which is what matters.

Auxiliary definitions (#1379) are more challenging because there is no corresponding C declaration.

jorisdral:
@TravisCardwell thank you for your input!

> This works fine for external binding specifications.

That's good to hear. :smiley: It sounds like the main challenge for users would be that they have to understand how anonymous structs/unions are named, so that they can refer to the right one in their prescriptive binding specification. We should document this well.

> A comment above says that the anonymous types will be named after their first field rather than numbered. (Is this still the plan, but it is just not implemented yet?)

Yes, that is still the plan, but it is not implemented yet. In your example, the anonymous struct would be named after `fieldX`

> Some developers likely prefer the counter names (`outer1_anon0` or just `anon0` when using `OverloadedRecordDot`, or perhaps even `field2`), but binding specifications allow them to specify their desired name in the generated API, which is what matters.

Makes sense! The names are still customisable

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/ImplicitFields.hs
jorisdral: Done

jorisdral:
@georgefst the feature is now merged. We hope this works for you!

Issue #1819, opened 2026-03-06
We should detect and skip over anonymous bitfields


edsko:
Given

```c
struct __attribute__((packed)) foo {
  int64_t x1 : 50;
  int64_t : 0;
  int64_t x2 : 50;
};
```

we currently try to generate a field for that anonymous bitfield (which is sometimes used to influence how the compiler packs the fields). With `--enable-record-dot` that then results in a name mangler failure (since it cannot mangle an empty name).

jorisdral:
Is this fully resolved? What if the width is non-zero? Are we handling that correctly? For example:

```c
struct foo {
  int64_t x1 : 50;
  int64_t : 13;
  int64_t x2 : 50;
} 
```

TravisCardwell:
I did not think that was valid C and would result in a Clang error.  Testing now, I see that it is allowed.  I guess it corresponds to the following in the spec:

    There may be unnamed padding within a structure object, but not at its beginning.

Thanks, @jorisdral!  I will put up a PR to handle this: #1857

PR #1841, opened 2026-03-18
Rewrite bit-field peek and poke code

TravisCardwell:
This PR resolves some critical issues that may cause runtime termination of programs that use code generated by `hs-bindgen`:

* We were reading/writing from the wrong locations in memory, sometimes resulting in segmentation faults.
* We were unable to read/write `struct` fields that cross machine word boundaries.  In such cases we would just `fail`.
* We were reading/writing to memory without concern for memory bounds.  When reading/writing the last field of a `struct`, some cases would read/write up to seven bytes into memory that is not owned.  This could theoretically cause a segmentation fault in some conditions.
* We were performing many unaligned reads/writes, especially since one part of the code normalized offsets that invalidated assumptions about alignment in another part of the code.  This results in fatal bus errors on some architectures (ARM).

Many improvements are made, including:

* The peek and poke code hard-coded alignment calculation, which is invalid for packed `struct`s.  The new implementation strictly uses the system to calculate alignment, so that it is not hard-coded.  No assumptions about alignment are made.
* Since the peek and poke code read/wrote to memory without concern for memory bounds, concurrent access to neighboring `struct` objects (in an array for example) was unsafe.  The new implementation uses memory bounds to ensure that only bytes within the `struct` are read/written.  While concurrent access to bit-fields within a single `struct` object is inherently unsafe (even in C), concurrent access to separate `struct` objects is now safe.
* A single peek/poke is used when possible, as determined by the alignment and memory bounds.  When it is not possible, we read/write bytes separately and calculate peek/poke on those bytes.  We chose to do this instead of calculating a minimum number of word reads/writes because it is much simpler, and very little is gained by the optimal solution.
* Function `loMask` returned an incorrect mask for `loMask @Int64 64`.  This is corrected.
* Tests accidentally tested with zero-width bit-fields, which are not supported.  This is no longer an issue.
* Documentation and comments are added.  The peek and poke functions explicitly test invariants so that we are alerted if any of them fail.
* The existing round-trip property tests are completely rewritten and are much more effective than the old tests.
    * The `extend . narrow` tests now test all `Word*` (unsigned) and `Int*` (signed) types.
    * The `peek . poke` tests now test bit-fields in a wide range of `struct` sizes.  Note that these tests allocate 64-bit aligned memory and hard-code alignment calculation for associated labels.
    * New property tests check consistency of the pure functions that implement peeks/pokes using bytes, for both little endian and big endian architectures.
* New tests implement C functions for setting and checking fields of a `struct`.  These tests are used to separately check that our peek and poke functions are consistent with C for various `struct`s.  These are implemented as a TH test, so the C is compiled by the C compiler that GHC is configured to use.  This is often GCC, not Clang, and this test may therefore help us discover discrepancies between Clang and GCC.  (We have not discovered any such discrepancies yet.)  Note that we will eventually generate such tests automatically.

Implementation note:

I now understand the motivation for "normalizing" the pointer and offset.  The goal is to separate a `BitfieldPtr` from the `struct`.  This is illustrated by the type `BitfieldPtr.peek :: Bitfield a => BitfieldPtr a -> IO a`.  For example, a `BitfieldPtr CUInt` references a bit-field of type `CUInt`, and this function peeks the value.  (The `Bitfield` constraint ensures that the type of the value can be extended/narrowed, required for peeking/poking bit-fields.)  A `BitfieldPtr` has no reference to the `struct` that contains the bit-field.

FWIW, I do not subscribe to this design goal.  I think that a reference to a range of bits in arbitrary memory is not a good idea; a bit-field is inherently part of a `struct`.  @edsko feels strongly about the separation, however, so I used that design.

While we want to avoid referencing the `struct` within a `BitfieldPtr`, we must know the memory bounds of the `struct` for three reasons:

* A poke of a bit-field should not mangle memory outside of the `struct`, because it may cause corruption of neighboring data when there is concurrent access.
* This allows us to use word peeks/pokes for all fields in well-designed `struct`s.  This is not possible without knowing the bounds of the `struct`.
* We should avoid peeking/poking memory that we do not own, as it could cause segmentation faults.

I added memory bounds information to `BitfieldPtr` using type `(Ptr (), Ptr ())` so that we do not reference the `struct` type, and the type can remain `BitfieldPtr a` where `a` is just the type of the bit-field.  I added documentation about the `a` type parameter.


TravisCardwell:
I added a commit that resolves #1819.

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
```diff
@@ -6,41 +6,71 @@
...
 type role BitfieldPtr nominal
--- The constructor is unsafe because the offset must be normalized (modulo 8
--- bits).
 data BitfieldPtr a = UnsafeBitfieldPtr {
+      -- | Pointer to the byte where the bit-field starts
+      --
+      -- We do /not/ assume that the pointer is aligned.
+      --
+      -- The @a@ is a lie, unless the offset is zero, but we use @a@ instead of
+      -- '()' because we do not want @a@ to be a phantom.
```
edsko: Let's change this to `Ptr ()`.

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
```diff
@@ -6,41 +6,71 @@
...
+      -- | Offset of the bit-field (0 to 7 bits)
     , _offset :: Int
-      -- | The width (in number of bits) of the bit-field
+      -- | Width of the bit-field (1 to 64 bits)
     , _width  :: Int
+      -- | Memory bounds of the @struct@
+      --
+      -- To peek/poke a bit-field, we may peek/poke any memory within these
+      -- bounds.  We must not peek/poke memory outside of these bounds.
+    , _bounds :: (Ptr (), Ptr ())
```
edsko: We should say if these bounds are inclusive or exclusive (in particular the upper bound).

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
```diff
@@ -6,41 +6,71 @@
 -- > import HsBindgen.Runtime.BitfieldPtr qualified as BitfieldPtr
 module HsBindgen.Runtime.BitfieldPtr (
     Bitfield(..)
-  , BitfieldPtr(BitfieldPtr)
+  , BitfieldPtr
```
edsko:
```suggestion
  , BitfieldPtr -- opaque
```

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
```diff
@@ -6,41 +6,71 @@
...
+      -- | Pointer to the byte where the bit-field starts
+      --
+      -- We do /not/ assume that the pointer is aligned.
+      --
+      -- The @a@ is a lie, unless the offset is zero, but we use @a@ instead of
+      -- '()' because we do not want @a@ to be a phantom.
       _ptr    :: Ptr a
-      -- | The offset (in number of bits) of the bit-field from the pointer
+      -- | Offset of the bit-field (0 to 7 bits)
     , _offset :: Int
```
edsko: Let's get rid of these underscores, and then export these accessors (but not the constructor).

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
```diff
@@ -6,41 +6,71 @@
...
--- The constructor is unsafe because the offset must be normalized (modulo 8
--- bits).
 data BitfieldPtr a = UnsafeBitfieldPtr {
+      -- | Pointer to the byte where the bit-field starts
+      --
+      -- We do /not/ assume that the pointer is aligned.
+      --
+      -- The @a@ is a lie, unless the offset is zero, but we use @a@ instead of
+      -- '()' because we do not want @a@ to be a phantom.
       _ptr    :: Ptr a
```
edsko:
```suggestion
      startingByte    :: Ptr a
```

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
```diff
@@ -6,41 +6,74 @@
 -- > import HsBindgen.Runtime.BitfieldPtr qualified as BitfieldPtr
 module HsBindgen.Runtime.BitfieldPtr (
     Bitfield(..)
-  , BitfieldPtr(BitfieldPtr)
+  , BitfieldPtr -- opaque
+  , mkBitfieldPtr
+  , startingByte
+  , offset
+  , width
+  , bounds
```
jorisdral, line 14: Removing the constructor of `BitfieldPtr`  should be marked as a breaking change in the changelog

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
```diff
@@ -6,41 +6,74 @@
...
-      -- | The width (in number of bits) of the bit-field
-    , _width  :: Int
+      -- | Pointer to the byte where the bit-field starts
+      --
+      -- We do /not/ assume that the pointer is aligned.
+      startingByte :: Ptr ()
+      -- | Offset of the bit-field (0 to 7 bits)
+    , offset :: Int
+      -- | Width of the bit-field (1 to 64 bits)
+    , width :: Int
```
jorisdral, line 40: The documentation says it should be 1-64 bits, but the implementation does not ensure it. Errors would be thrown when using `peek`/`poke` as far as I can see, but `width` is also exported as a getter so users might mistakenly assume that the `width` is 1-64 bits

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
jorisdral, line 14: Are `startingByte`, `offset`, `width`, and `bounds` exported for the user's benefit?

I would also mark these as minor changes in the changelog, or alternatively I would not export the fields

hs-bindgen-runtime/src/HsBindgen/Runtime/HasCBitfield.hs
```diff
@@ -152,7 +156,8 @@ toPtr _ ptr = BitfieldPtr (castPtr ptr) o w
 -- | Using a pointer to a C object, read from one of the object's bit-fields.
 peek ::
      forall a field. (
-       HasCBitfield a field
+       Marshal.StaticSize a
```
jorisdral, line 159: This new constraint should be marked as a breaking change in the changelog

hs-bindgen-runtime/src/HsBindgen/Runtime/HasCBitfield.hs
```diff
@@ -139,11 +140,14 @@ width = \_ _ -> bitfieldWidth# (proxy# @a) (proxy# @field)
 -- | Convert a pointer to a C object to a pointer to one of the object's
 -- bit-fields.
 toPtr ::
-     forall a field. HasCBitfield a field
+     forall a field. (
+       Marshal.StaticSize a
```
jorisdral, line 144: This new constraint should be marked as a breaking change in the changelog

hs-bindgen-runtime/src/HsBindgen/Runtime/HasCBitfield.hs
```diff
@@ -164,7 +169,8 @@ peek field ptr = BitfieldPtr.peek (toPtr field ptr)
 -- | Using a pointer to a C object, write to one of the object's bit-fields.
 poke ::
      forall a field. (
-       HasCBitfield a field
+       Marshal.StaticSize a
```
jorisdral, line 172: This new constraint should be marked as a breaking change in the changelog

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
TravisCardwell: Ah!  I forgot to update the changelog for the last-second changes!  Thanks!

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
TravisCardwell: The checks are in the `peekBitOffWidth` function [here](https://github.com/well-typed/hs-bindgen/blob/44bc78310d7d4922b0833c8d386962983100fcc3/hs-bindgen-runtime/src/HsBindgen/Runtime/Internal/Bitfield.hs#L207) and `pokeBitOffWidth` function [here](https://github.com/well-typed/hs-bindgen/blob/44bc78310d7d4922b0833c8d386962983100fcc3/hs-bindgen-runtime/src/HsBindgen/Runtime/Internal/Bitfield.hs#L288).

If we change these functions to take a `BitfieldPtr`, then we could perform the checks when constructing `BitfieldPtr` instead.  My understanding is that we want to keep the current API, however, so that users can use those functions without `BitfieldPtr`.  I am happy to make this change if we decide that it is worth it!

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
TravisCardwell: Yes, that is my understanding.

Here is what I added:

```diff
+* The `BitfieldPtr` constructor pattern is removed, making the type opaque.  A
+  smart constructor and accessor functions are exported instead.
```

BTW, I think we should be careful about how much detail we put into the changelogs.  A changelog should be user-focused, and including too much detail is generally a bad idea because it makes a changelog difficult to read.  IMHO, a changelog should be clear and concise, so that users can easily get an overview of changes, while details can be documented in the code and have `@since` annotations.

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
jorisdral: I don't think we have to change anything about `peekBitOffWidth`/`pokeBitOffWidth`. Couldn't we simply add the check to `mkBitfieldPtr`, so that you can't construct a `BitfieldPtr` with an incorrect width?

TravisCardwell: Yes, that is indeed another option.  The downside is that we would perform the same checks multiple times (but at least those checks are cheap).

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
jorisdral: I agree about taking care not to put too much detail in changelogs, though "too much detail" is hard to nail down. 

For breaking changes for example, I would argue that the user benefit from knowing quite precisely what changes are breaking in the API and how they should adjust their own code. Documentation in the code doesn't sound like the right place to me to put that information

hs-bindgen-runtime/src/HsBindgen/Runtime/BitfieldPtr.hs
jorisdral: Documenting that the width invariant is not checked by `mkBitfieldPtr` would also be an alternative

TravisCardwell: I think that documenting that it is not checked can be confusing, as the normal way that `BitfieldPtr`s are used performs the check.  I went ahead and added the check to `mkBitfieldPtr`, adding the commit to #1855.

PR #1845, opened 2026-03-18
CI: remove the whitespace check

jorisdral:
`stylish-haskell` already ensures proper use of whitespace at the end of Haskell files, so we don't have to check whitespace separately.


TravisCardwell:
Today, I noticed whitespace issues in `main`.  For example, file `manual/LowLevel/Translation/02-Structs/Nesting.md` does not have a newline at the end of the file.  We should probably add a new whitespace CI check to ensure that we do not merge such files.

PR #1849, opened 2026-03-19
Parse nested structs/unions even if the enclosing object fails to parse

jorisdral:
Take this example:

```c
struct foo {
  struct bar {
    int x1_1;
  } x1;
  long double x2;
};

extern struct bar X;
```

`bar` can be parsed, but `foo` does not parse because `long double` is not supported. Parsing of a top-level declaration was previously shortcutted using exceptions, so we would not see `bar` in the list of parsed declarations. However, it *should* be there: `bar` has file scope because it inherits the scope from `foo`, so we should still generate bindings for bar even if we don't generate bindings for `foo`.

Info on scoping: https://en.cppreference.com/w/c/language/scope.html#Notes

We achieve this by installing an exception handler when parsing nested declarations and fields, rather than only at the top level. This allows us to selectively fail (nested) declarations.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/Field.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/Field.hs
@@ -5,6 +5,8 @@ module HsBindgen.Frontend.Pass.Parse.Decl.Field (
   , getFieldInfo
   ) where
 
+import Control.Monad.IO.Class (MonadIO)
+
 import Clang.HighLevel qualified as HighLevel
 import Clang.HighLevel.Documentation qualified as CDoc
 import Clang.LowLevel.Core (CXCursor, clang_Cursor_getOffsetOfField,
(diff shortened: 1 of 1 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/Members.hs
```diff
@@ -94,31 +99,66 @@ parseMembersWith ctx parseField parseObject k =
...
+        | not (null fails)
+        -> k ParseMembersResult {
+                declMembers = declMembers
+              , fieldMembers = Left ParseNestedDeclsFailed
+              }
+        -- If any exceptions occurred during folding, only return the first one
+        --
+        -- TODO: it might be nice to report /all/ fold exceptions, but that
+        -- would require some thinking. Do we make the parse message datatype
+        -- recursive, for example?
```
jorisdral: TODO without an issue

jorisdral: Removed the TODO

Issue #1253, opened 2025-11-06
Support bit-fields in unions


jorisdral:
Currently we only support bit-fields in structs, but bit-fields are also allowed in unions. See https://en.cppreference.com/w/c/language/bit_field.html

edsko:
I'm going to lower this to low priority until we have a real use case.

jorisdral:
Note that union fields carry bit width information since #2124,  but this information is not yet used to support bit-fields in unions

PR #1857, opened 2026-03-23
Skip explicit bit-field padding

TravisCardwell:
A previous commit skipped unnamed bit-fields with zero width, but explicit padding can be specified using unnamed bit-fields with positive widths.  This commit changes the filter to skip all unnamed bit-fields.

### DIFF
--- a/hs-bindgen/examples/golden/types/anonymous/edge-cases/bitfield.h
+++ b/hs-bindgen/examples/golden/types/anonymous/edge-cases/bitfield.h
@@ -1,8 +1,8 @@
 #pragma once
 
 // Implicit field detection works the same if any of the indirect fields are
-// *named* bit-fields. Unnamed bit-fields are not supported (see
-// unnamed_bitfield.h).
+// bit-fields.  Note that unnamed bit-field declarations, which specify padding,
+// are not fields.
 //
 // Note: implicit field detection works the same for both unions and structs in
 // any order of nesting, so testing the struct-only case should be sufficient to
--- a/hs-bindgen/examples/golden/types/anonymous/edge-cases/empty_anon.h
+++ b/hs-bindgen/examples/golden/types/anonymous/edge-cases/empty_anon.h
@@ -8,14 +8,14 @@
 // <https://en.cppreference.com/w/c/language/struct.html>
 // <https://en.cppreference.com/w/c/language/union.html>
 //
-// We need at least one named (bit-)field in an anonymous struct or union to
-// determine the offset of that anonymous object within an enclosing object. In
-// other words, anonymous structs and unions should be "non-empty". A struct or
-// union with only unnamed bit-fields is also considered "empty". The parser
-// should not generate any bindings for empty structs or unions, and emit a
-// message instead. Moreover, if a nested struct or union fails to parse, for
-// example because it is empty, then the enclosing struct or union should also
-// fail to parse.
+// We need at least one (bit-)field in an anonymous struct or union to determine
+// the offset of that anonymous object within an enclosing object. In other
+// words, anonymous structs and unions should be "non-empty". A struct or
+// union with only unnamed bit-field declarations, used to specify padding, is
+// also considered "empty". The parser should not generate any bindings for
+// empty structs or unions, and emit a message instead. Moreover, if a nested
+// struct or union fails to parse, for example because it is empty, then the
+// enclosing struct or union should also fail to parse.
 //
 // This edge case was not properly handled by implicit field detection in the
 // past, so we include this as an edge case regression test.
@@ -34,7 +34,7 @@ struct S1 {
 struct S2 {
   struct {
     struct {
-      char : 3; // unnamed bit-field
+      char : 3; // unnamed bit-field declaration specifies padding
     };
     int x;
   };
--- a/manual/LowLevel/Translation/02-Structs/Nesting.md
+++ b/manual/LowLevel/Translation/02-Structs/Nesting.md
@@ -198,15 +198,10 @@ If these generated names are too unwieldy, they can always be customised using
 
 ### Limitations
 
-For technical reasons we can only generate bindings for anonymous structs if the
-following two conditions are met:
-
-1. the anonymous nested struct should have at least one field
-2. the anonymous nested struct should have only *named* fields
-
-Concretely, this means that empty anonymous structs and anonymous structs with
-unnamed bit-fields are not supported. A warning-level trace message will be
-emitted if these conditions are not met.
+For technical reasons we can only generate bindings for anonymous structs that
+have at least one field. Empty anonymous structs and anonymous structs with only
+padding (specified using unnamed bit-field declarations) are not supported.  A
+warning-level trace message will be emitted in this case.
 
 
 
--- a/manual/LowLevel/Translation/04-Unions/Nesting.md
+++ b/manual/LowLevel/Translation/04-Unions/Nesting.md
@@ -227,15 +227,10 @@ If these generated names are too unwieldy, they can always be customised using
 
 ### Limitations
 
-For technical reasons we can only generate bindings for anonymous unions if the
-following two conditions are met:
-
-1. the anonymous nested union should have at least one field
-2. the anonymous nested union should have only *named* fields
-
-Concretely, this means that empty anonymous unions and anonymous unions with
-unnamed bit-fields are not supported. A warning-level trace message will be
-emitted if these conditions are not met.
+For technical reasons we can only generate bindings for anonymous unions that
+have at least one field. Empty anonymous unions and anonymous unions with only
+padding (specified using unnamed bit-field declarations) are not supported. A
+warning-level trace message will be emitted in this case.
 
 
 
--- a/hs-bindgen/examples/golden/types/anonymous/edge-cases/padding.h
+++ b/hs-bindgen/examples/golden/types/anonymous/edge-cases/padding.h
@@ -1,7 +1,7 @@
 #pragma once
 
-// We can compute offsets for implicit fields even if the first field(s) of an
-// anonymous struct or union are unnamed bit-fields (that introduce padding).
+// We can compute offsets for implicit fields even if the anonymous struct or
+// union starts with padding (specified by an unnamed bit-field declaration).
 //
 // This edge case was not properly handled by implicit field detection in the
 // past, so we include this as an edge case regression test.
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/ImplicitFields.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/ImplicitFields.hs
@@ -83,8 +83,10 @@ data Outputs field =
 --
 -- PRECONDITION: the inputs must include all nested struct\/union\/field
 -- declarations that are present in the C source. If any failed to parse and are
--- not included, then it is unsafe to use 'withImplicitFields'. Except for
--- unnamed bit-fields, those can be omitted freely.
+-- not included, then it is unsafe to use 'withImplicitFields'.
+--
+-- Unnamed bit-field declarations, used to specify padding, are already filtered
+-- out.
 --
 -- === Algorithm description
 --
@@ -92,7 +94,7 @@ data Outputs field =
 -- @libclang@, so we detect them instead.
 --
 -- Given an enclosing object struct\/union $E$, and a nested anonymous
---struct\/union $A$, the goal is to compute the offset of the (hidden) implicit
+-- struct\/union $A$, the goal is to compute the offset of the (hidden) implicit
 -- field of $E$ that reference $A$. The key observation is that the offset to
 -- that implicit field is equal to the offset to any of $A$'s fields (let's say
 -- $F$), subtracted by the offset to $F$ with respect to $A$.
@@ -140,8 +142,9 @@ data Outputs field =
 -- The implicit field detection algorithm does rely on one condition: the
 -- anonymous nested struct or union should have at least one named field. In
 -- other words, the anonymous nested struct\/union should be "non-empty". A
--- struct\/union with only unnamed bit-fields is also considered empty. A
--- warning-level trace message will be emitted if these conditions are not met.
+-- struct\/union with only unnamed bit-field declarations, used to specify
+-- padding, is also considered empty. A warning-level trace message will be
+-- emitted if these conditions are not met.
 --
 -- Anonymous nested structs/unions have no name, but they need one for our Haskell
 -- bindings, so they are named after their first field. Informally, the former will


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -207,12 +207,12 @@ structDecl ctx info = \curr -> do
...
                 , kind = C.DeclStruct C.Struct{
-                             sizeof    = fromIntegral sizeof
-                           , alignment = fromIntegral alignment
-                           , fields    = filter isNotZeroWidth regularFields
-                           , flam      = mFlam
-                           , ann       = NoAnn
-                           }
+                      sizeof    = fromIntegral sizeof
+                    , alignment = fromIntegral alignment
+                    , fields    = filter (not . isUnnamedBitfield) regularFields
```
jorisdral: For implicit field detection (PR #1839) to work correctly, these unnamed bit-fields would have to be kept around in the `Parse` pass. It would be fine to filter them out in a later pass

The offset for implicit fields (fields referencing anonymous objects) is equal to the offset of the anonymous objects's first field. If that first field is an unnamed bit-field, and we filter it out, then implicit field detection will mistakenly assume that a different field is the first field

In the example below, `x` is not the first field but it would wrongly be assumed to be the first field if we filter out the unnamed bit-field

```c
struct foo {
  struct {
    signed char : 3;
    signed char x : 3;  
  };
};
```

TravisCardwell: From the C99 specs, 6.7.2.1 Structure and union specifiers (13):

    There may be unnamed padding within a structure object, but not at its beginning.

jorisdral: Hmm that's a good point, but I don't think Clang warns/errors when you write this though. I tried it out locally, and it was parsed without problem. I think maybe "unnamed padding" more generally means that the *compiler implementation* can not insert its own padding before the first field, but I'm not sure

jorisdral: I also just saw this in the reference:

> The following usages of bit-fields causes undefined behavior:
>
 > *  Calling [offsetof](https://en.cppreference.com/w/c/types/offsetof.html) on a bit-field.
 
 We use `offsetof` to compute implicit field offsets, which means implicit field detection won't work with *named* bit-fields either. Bummer :(

edsko: Presumably that is to avoid having to compute offsets halfway bytes though; if it\'s the _first_ field, I guess it's okay?

jorisdral: As far as I can see, the `clang_Type_getOffsetOf` is more powerful than then `offsetof` macro, even though the former says its semantics are based on the latter. So it seems we can support anonymous objects with *named* bit-fields after all. *unnamed* bit-fields are still not supported

TravisCardwell: I really wanted to test it in Clang yesterday but did not have time.  Doing so this morning, I confirm that Clang indeed accepts unnamed bit-fields as first fields without warning.

Yes, my understanding is that C's `offsetof` returns a value in bytes, so it cannot work for bit-fields (that do not start at an 8-bit boundary).  Clang's `clang_Type_getOffsetOf` returns a value in bits, so it should not have the same limitation.  The documentation references the C++11 standard, but the reference paragraph (18.2p4) does not confirm anything for me.  Searching the whole standard, I do not see any documentation of what the macro returns.  [cppreference.com documentation](https://en.cppreference.com/w/cpp/types/offsetof.html) indicates that the C++ `offsetof` also returns bytes and behavior is undefined when used on a bit-field.

Regarding implicit fields, my understanding is that we find the first field in order to name the anonymous type.  Perhaps an empty name is not very useful regardless, and we should instead use the first named field?  Regarding edge cases, a test confirms that Clang accepts a `struct` with only unnamed bit-fields without warning.  Not creating implicit fields in such a case sounds logical to me, as padding is not a bit-field.

Regarding offsets, is there an issue that I am missing?  We have the offset of the unnamed field in the parent `struct`, and the offset of bit-fields in the anonymous `struct` already takes padding into account, right?

TravisCardwell:
Test `types/anonymous/edge-cases/unnamed_bitfield` fails when I rebase, because the traces are different.

Header:

```c
struct S1 {
  struct {
    char : 3;
  };
  int x;
};

struct S2 {
  struct {
    struct {
      char : 3;
    };
    int x;
  };
};
```

Test:

```haskell
test_types_anonymous_edge_cases_unnamed_bitfield :: TestCase
test_types_anonymous_edge_cases_unnamed_bitfield =
    testTraceMulti "types/anonymous/edge-cases/unnamed_bitfield" declsWithMsgs $ \case
      MatchDelayedImplicitField "struct S1" UnsupportedUnnamedBitfield ->
        Just $ Expected "struct S1"
      MatchDelayed "struct S2" ParseNestedDeclsFailed ->
        Just $ Expected "struct S2"
      _otherwise ->
        Nothing
  where
    declsWithMsgs :: [CDeclName]
    declsWithMsgs = ["struct S1", "struct S2"]
```

I do not think that `UnsupportedUnnamedBitfield` accurately describes the issue.  C is a mess, so let me define some terminology:

* An unnamed bit-field declaration (technically "a bit-field declaration with no declarator") is a special case used to instruct the compiler about padding.
* Though the syntax uses bit-field declaration syntax, it does *not* define a field/bit-field in the `struct`/`union`.  There are no unnamed fields/bit-fields or anonymous fields/bit-fields.

In this test, the unnamed bit-field declaration is not "unsupported."  Clang supports it, and we follow Clang's lead.  The issue is that the anonymous `struct` does not have any fields/bit-fields.  We should handle this like the `empty_anon.h` test case.

@jorisdral Thoughts?

TravisCardwell:
I see the comment:

```haskell
-- There can be no unnamed padding at the start of a struct or union, so the
-- offset to the first field of an anonymous struct/union is also the offset
-- to the anonymous struct/union
```

We confirmed that Clang allows padding at the start of a `struct`/`union`, so this is not valid, right?

I did a quick test with the following header, using byte-sized fields to keep it simple.

```c
struct foo {
  unsigned char a;
  struct {
    unsigned char   : 8;
    unsigned char x : 8;
    unsigned int    : 16;
  };
  unsigned char b;
};
```

Running `clang-ast-dump`, I observe:

* The size of `struct foo` is 6 bytes, as expected.
* I do not see an offset for the anonymous `struct` within `struct foo`.  Things would be easy if we had that information...
* The size of the anonymous `struct` is 4 bytes, as expected
* The offsets of the fields, including unnamed field declarations, in the anonymous `struct` are 0, 8, and 16 bits respectively, as expected.  These are offset from the anonymous `struct`, so we must know the offset of the anonymous `struct` within `struct foo` to access them given a address of a `struct foo` object.
* The offsets of the fields in `struct foo` are 0 and 40 bits respectively, as expected.

I do not think that initial padding poses a problem, because we are given the offsets of the fields in the anonymous `struct`.  Determining the offset of the anonymous `struct` within the enclosing object is the challenge.  This is not affected by padding within the anonymous `struct`; it is affected by padding within the enclosing object.

Consider the following minimal example:

```c
struct foo {
  unsigned char : 8;
  struct {
    unsigned char a;
  };
};
```

Running `clang-ast-dump`, I observe:

* The size of `struct foo` is 2 bytes, as expected.
* The offset of field `a` is 0 bits from the anonymous `struct`.

If there is no way to get the offset of the anonymous `struct` within the enclosing object from Clang, do we need to (attempt to) calculate it ourselves?  The good news is that only integral types are packed, and tests thankfully show that the anonymous the `struct` always has a byte offset as expected.  To calculate this ourselves, we would need to take the following into account:

* If a field precedes the anonymous `struct`/`union`:
    * The offset of the field
    * The size of the field
    * The alignment of the anonymous `struct`/`union`, *unless* it is in a packed `struct`
* If a bit-field precedes the anonymous `struct`/`union`:
    * The offset of the bit-field
    * The width of the bit-field
    * The alignment of the anonymous `struct`/`union`, *unless* it is in a packed `struct`
* If a padding declaration precedes the anonymous `struct`/`union`:
    * The offset of the padding
    * The width of the padding, taking the zero special case into account
    * The alignment of the anonymous `struct`/`union`, *unless* it is in a packed `struct`

Note that we would indeed need to retain padding information to implement this, not just filter out the padding declarations.

Calculating this seems like quite a poor solution, but is there any other way?

Ah, If I interpret the comment correctly, `clang_Type_getOffsetOf` can give us the offset of "implicit fields," not just fields of the `struct`/`union`!  Using print debugging to test the minimal example above, it indeed calculates that field `a` has an offset of 8 bits in `struct foo`, as expected.  Great!

Testing with some padding in the anonymous `struct`, it calculates that field `a` has an offset of 16 bits in `struct foo`, as expected.  Great!

```c
struct foo {
  unsigned char : 8;
  struct {
    unsigned char : 8;
    unsigned char a;
  };
};
```

TravisCardwell:
Is [this](https://github.com/well-typed/hs-bindgen/blob/4b6febb5cc6196835bd2890a3ab27a88dab1806b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/ImplicitFields.hs#L295-L297) comment obsolete?

```haskell
-- | When the field is implicit and we want to ask for its offset using its
-- name, then we should ask for the offset to the first field of the referenced
-- anonymous object instead.
```

For translating an anonymous type to a Haskell type, the offsets of the fields are those reported by Clang as usual.

For creating "implicit fields," `clang_Type_getOffsetOf` gives us the offset relative to the containing type.

We should not have to base anything on the first field (perhaps field declaration, which may be for padding), right?

TravisCardwell:
Discussing a bit with @edsko (to determine when we should tag the next release), I understand better now.  Checking the fixtures, I now see that we are considering the whole anonymous `struct` to be an "implicit field," *not* the fields within the anonymous `struct`.  For this, we must determine the offset of the anonymous `struct` within the parent type.

Just to explain my misunderstanding, I thought that we would implement "implicit fields" following how they are used in C.  Given the following header, type `Foo` would have implicit fields `foo_x` and `foo_y`, both of types `CInt`.

```c
struct foo { struct { int x, y; }; };
```

With out implementation, type `Foo` has implicit field `foo_x` of type `Foo_x`, which has fields `foo_x_x` and `foo_x_y`.

Given that we have a way to get the offset of `x` in `struct foo`, I thought of a way to get the offset of the anonymous `struct` even when padding is involved: subtract the offset of a field within the anonymous `struct` from the offset of the same field in the outer `struct`.

jorisdral:
> Just to explain my misunderstanding, I thought that we would implement "implicit fields" following how they are used in C. 

I can understand the misunderstanding. The terminology is not super well-defined here, I could've done a better job there. If we take your example, I've been calling `x` and `y` *indirect* fields of `foo`. I've been calling the unnamed field referencing the anonymous struct an *implicit* field.

The original plan was to generate bindings for the indirect fields, but if we go that route then we have to implement special cases for the struct-in-union, union-in-union, and union-in-struct variants of this example as well, and then recursive nesting makes it more tricky. Generating bindings for the implicit field is a uniform approach that works for all nestings of structs and unions, even recursively without problem. 

Note: generating bindings for the implicit field rather than the indirect fields is also the only approach that works for union-in-struct and struct-in-union variants, see the description of #1649. As I was in the process of implementing the code for implicit fields anyway, why not use it for the union-in-union and struct-in-struct variants as well, was my thought. Then we don't need separate logic for the different variants

And yeah, `libclang`'s `visitChildren` does not report indirect fields or implicit fields, so we have calculate the offset for the implicit field ourselves

> Given that we have a way to get the offset of `x` in `struct foo`, I thought of a way to get the offset of the anonymous `struct` even when padding is involved: subtract the offset of a field within the anonymous `struct` from the offset of the same field in the outer `struct`.

Ah! That makes sense, simple yet clever :smile: I'm going to give this a try

TravisCardwell:
This needs to be merged ASAP so that we can tag `alpha2`.  In the rebase, I changed the `types/anonymous/edge-cases/unnamed_bitfield` test case to check for a `UnsupportedEmptyAnon` trace.  I reverted all of my tentative changes to the `ImplicitFields` implementation, as issue #1866 is opened to address them.  Note that `UnsupportedUnnamedBitfield` should be removed completely.

edsko:
Blocked by https://github.com/well-typed/hs-bindgen/issues/1866.

TravisCardwell:
Issue #1866 is addressed in PR #1869, which has been merged.  Without this PR, it is generating fields for padding.  [Example](https://github.com/well-typed/hs-bindgen/blob/5822144f6f4c1840b57b67e13bf69c4f5201be38/hs-bindgen/fixtures/types/anonymous/edge-cases/padding/Example.hs#L59-L65):

```haskell
data SS_y = SS_y
  { sS_y_ :: RIP.CChar
    {- ^ __C declaration:__ @@

         __defined at:__ @types\/anonymous\/edge-cases\/padding.h 16:5@

         __exported by:__ @types\/anonymous\/edge-cases\/padding.h@
    -}
  ...
```

With this PR, I confirmed that we no longer have that issue.

TravisCardwell:
TIL that a `union` can have bit-fields and explicit padding.  While it makes sense for a `union` to have a `struct` field that contains bit-fields, I have yet to think of a practical reason for a `union` to have bit-fields, as bit-fields only really makes sense when grouped together using a `struct`, and searches have not been enlightening.  A `union` with explicit padding is nonsensical.

What do we (want to) do about such cases?  Our `StructField` implementation uses `width` to support bit-fields, but `UnionField` does not.  Is this something that we need to think about?

We already have an issue: #1253

TravisCardwell:
FYI: I refactored `ImplicitFields` a bit because [this case](https://github.com/well-typed/hs-bindgen/blob/5822144f6f4c1840b57b67e13bf69c4f5201be38/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/ImplicitFields.hs#L348-L349) is no longer possible:

```haskell
getOffsetOfTarget ::
     IsField field
  => field Parse
  -> Maybe Target
getOffsetOfTarget (Field -> field)
  | Text.null field.info.name.text
  = Nothing
  ...
```

TravisCardwell:
I went ahead and audited the terminology so that we only refer to an "unnamed bit-field declaration" (C syntax used to specify padding), never "unnamed bit-field."  Similarly, we should never use "named bit-field" because it implies the existence of an "unnamed bit-field."

I do not mean to nit-pick; I think that this is important.  I talked about this before yet the incorrect terminology persists, so I feel that I need to justify it.  [Here](https://github.com/well-typed/hs-bindgen/blob/5822144f6f4c1840b57b67e13bf69c4f5201be38/manual/LowLevel/Translation/02-Structs/Nesting.md?plain=1#L199-L209) is an example from the manual:

```md
For technical reasons we can only generate bindings for anonymous structs if the
following two conditions are met:

1. the anonymous nested struct should have at least one field
2. the anonymous nested struct should have only *named* fields

Concretely, this means that empty anonymous structs and anonymous structs with
unnamed bit-fields are not supported. A warning-level trace message will be
emitted if these conditions are not met.
```

Using the invalid terminology, calling an unnamed bit-field declaration an "unnamed bit-field," the second second condition is misleading/incorrect.  An anonymous nested `struct` may contain unnamed bit-field declarations, used to specify padding, while this makes it sound like we do not support *any* such unnamed declarations.  The sentence after that is definitely incorrect; perhaps "with unnamed bit-fields" was meant to be "with *only* unnamed bit-fields."

I changed this to the following, which avoids invalid terminology and is IMHO easier to understand.

```md
For technical reasons we can only generate bindings for anonymous structs that
have at least one field. Empty anonymous structs and anonymous structs with only
padding (specified using unnamed bit-field declarations) are not supported. A
warning-level trace message will be emitted in this case.
```

(Note that I purposefully kept the language style, not revising punctuation/grammar, for consistency with the rest of the file.)

hs-bindgen/examples/golden/types/anonymous/edge-cases/bitfield.h
```diff
@@ -1,8 +1,8 @@
 #pragma once
 
 // Implicit field detection works the same if any of the indirect fields are
-// *named* bit-fields. Unnamed bit-fields are not supported (see
-// unnamed_bitfield.h).
+// bit-fields.  Note that unnamed bit-field declarations, which specify padding,
+// are not fields.
```
jorisdral, line 5: Why is it not a field? A bit-field declaration in a struct/union member list declares a member/field regardless of whether it is named or not, as I understand it. Having a name or not affects whether you can modify the bit-field by name, having a name is not a prerequisite for being a field

Arguably it's not so well-defined in the manual what it means to be a field. But this passage from the "Struct declaration" page of the C reference might substantiate my case: 

> A pointer to a struct can be cast to a pointer to its first member (or, if the member is a bit-field, to its allocation unit). Likewise, a pointer to the first member of a struct can be cast to a pointer to the enclosing struct.

I think this implies that an unnamed bit-field at the start of a struct/union should be considered the first field/member

hs-bindgen/examples/golden/types/anonymous/edge-cases/empty_anon.h
```diff
@@ -8,14 +8,14 @@
...
 //
-// We need at least one named (bit-)field in an anonymous struct or union to
-// determine the offset of that anonymous object within an enclosing object. In
-// other words, anonymous structs and unions should be "non-empty". A struct or
-// union with only unnamed bit-fields is also considered "empty". The parser
-// should not generate any bindings for empty structs or unions, and emit a
-// message instead. Moreover, if a nested struct or union fails to parse, for
-// example because it is empty, then the enclosing struct or union should also
-// fail to parse.
+// We need at least one (bit-)field in an anonymous struct or union to determine
```
jorisdral, line 11: Also relevant for the "named field" discussion

manual/LowLevel/Translation/02-Structs/Nesting.md
```diff
@@ -198,20 +198,15 @@ If these generated names are too unwieldy, they can always be customised using
...
-1. the anonymous nested struct should have at least one field
-2. the anonymous nested struct should have only *named* fields
-
-Concretely, this means that empty anonymous structs and anonymous structs with
-unnamed bit-fields are not supported. A warning-level trace message will be
-emitted if these conditions are not met.
+For technical reasons we can only generate bindings for anonymous structs that
+have at least one field. Empty anonymous structs and anonymous structs with only
+padding (specified using unnamed bit-field declarations) are not supported.  A
+warning-level trace message will be emitted in this case.
```
jorisdral, line 204: Also relevant for the "named field" discussion

manual/LowLevel/Translation/04-Unions/Nesting.md
```diff
@@ -227,20 +227,15 @@ If these generated names are too unwieldy, they can always be customised using
...
-following two conditions are met:
-
-1. the anonymous nested union should have at least one field
-2. the anonymous nested union should have only *named* fields
-
-Concretely, this means that empty anonymous unions and anonymous unions with
-unnamed bit-fields are not supported. A warning-level trace message will be
-emitted if these conditions are not met.
+For technical reasons we can only generate bindings for anonymous unions that
+have at least one field. Empty anonymous unions and anonymous unions with only
```
jorisdral, line 231: Also relevant for the "named field" discussion

hs-bindgen/examples/golden/types/anonymous/edge-cases/bitfield.h
TravisCardwell: You access a (bit-)field in C using `obj.field` or `obj->field` syntax.  Unnamed bit-field declarations provide a way to explicitly specify padding, using bit-field syntax so that new language keywords are not needed, but the padding data is not accessed like a field.  Indeed, without a field name, field syntax would not work even if we could think of padding as a field.

> I think this implies that an unnamed bit-field at the start of a struct/union should be considered the first field/member

I think that padding at the start of a `struct`/`union` is not the design.  It is something that the compilers decided to implement regardless.

TravisCardwell: I indeed found "unnamed bit-field" in the C spec, *and* Clang uses the terminology as well, [here](https://github.com/llvm/llvm-project/blob/7b52dd6ad9ac16987d1646b651e5b1fd13b7cb41/clang/include/clang/AST/Decl.h#L3280-L3281):

```cpp
/// Determines whether this is an unnamed bitfield.
bool isUnnamedBitField() const { return isBitField() && !getDeclName(); }
```

I have addressed this in #1880.

PR #1858, opened 2026-03-24
nix: keep haskell.lib.* when extending haskell.lib.compose

jian-lin:
Previously, after the `default` or `hsBindgen` overlay was applied, other lib functions under haskell.lib.* were dropped:

```
nix-repl> builtins.attrNames haskell.lib == [ "compose" ]
true
```

This patch fixes that.

cc @dschrempf for review

---

- [ ] Did you update the changelog? Please be sure to give a "migration hint" if this is a breaking change.


dschrempf:
Ah great, thank you! I was not aware of that.

I will review as soon as I have time!

PR #1859, opened 2026-03-24
nix: use composeExtensions to compose overlays instead of manually

jian-lin:
`lib.composeExtensions' is semantically correct.  The manual way is not totally correct in all cases (but is correct in these use cases).

In addition, `lib.composeExtensions` improves readability a bit.

cc @dschrempf for review

---

- [ ] Did you update the changelog? Please be sure to give a "migration hint" if this is a breaking change.

### DIFF
--- a/nix/overlay/hs-bindgen.nix
+++ b/nix/overlay/hs-bindgen.nix
@@ -19,13 +19,12 @@ let
 in
 {
   haskell = prev.haskell // {
-    packageOverrides =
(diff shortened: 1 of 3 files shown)


dschrempf:
Thank you! I will also put this one on the queue, although it may require a rebase since it is touching lines next to #1858.

dschrempf:
Closing this, since I rebased and opened #1863, thank you!

Issue #1846, opened 2026-03-19
Handle Doxygen `JAVADOC_BANNER` comments


TravisCardwell:
As documented [here](https://www.doxygen.nl/manual/docblocks.html), Doxygen supports Javadoc banner-style comments (when [`JAVADOC_BANNER`](https://www.doxygen.nl/manual/config.html#cfg_javadoc_banner) is configured).

```c
/*******************************************************************************
 * This is a Javadoc banner-style comment.
 ******************************************************************************/
```

We currently parse these comments, and we get lots of `*` characters in the Haddock documentation.

```haskell
{-| ***************************************************************************** This is a Javadoc banner-style comment. ****************************************************************************

...
-}
```

Perhaps we should detect this and handle it more gracefully?

edsko:
@bolt12 As your resident Doxygen expert, could you take a look at this?

bolt12:
This is addressed by #1861, which replaces the libclang CXComment parser with doxygen's own XML output. The old approach parsed comment text naively via libclang, which included the decorative `*` characters verbatim. The new approach delegates parsing to the doxygen binary itself, which understands Javadoc banner syntax and strips the decoration before producing its XML. We also enable `JAVADOC_BANNER = YES` in the generated doxygen configuration so that banner-style comments (with three or more leading asterisks) are correctly recognized as documentation rather than silently ignored.

Issue #945, opened 2025-07-24
Release version 0.1


edsko:
This will be the first public release of `hs-bindgen`. We initially thought that this would cover only storable instances, but this is not going to be the case, and will cover both storable instances and the low-level API.

Important features for this first release:

- [ ] [All of C supported](https://github.com/well-typed/hs-bindgen/issues?q=is%3Aissue%20state%3Aopen%20label%3Amissing-c-feature%20-label%3A%22priority%3A%20low%22), except issues explicitly marked as low priority (or issues that require the high level API)
- [ ] Clear documentation https://github.com/well-typed/hs-bindgen/issues/500
- [ ] [List of open bugs](https://github.com/well-typed/hs-bindgen/issues?q=is%3Aissue%20state%3Aopen%20label%3Abug) should be carefully considered, though it will not be necessary to fix all of these before this release.
- [ ] **Syntax** for binding specifications should ideally be stable, so that we can _extend_ it in later releases, but don't have to _change_ it needlessly. 
- [ ] Similarly, the **syntax** for TH integration should be stable. We might want to implement more features, but the main calls (`withHsBindgen`, `hashInclude`) should ideally not change in the very next release; see https://github.com/well-typed/hs-bindgen/issues/728 (and possibly others)

Features I'm not completely sure about:

* Build process improvements (tickets like https://github.com/well-typed/hs-bindgen/issues/419); we can probably release before all that, but we should think a bit about it.
* Test suite generation (e.g. https://github.com/well-typed/hs-bindgen/issues/22). We can probably release without, though we should mention it, especially n relation to `gcc` vs `clang`.

Features we can release separately, relatively soon after the first release:

* The zero-copy API. This is in principle separate; it's not a ton of work, but we can release it separately with a separate announcement.
* Support for duplicate fields, and record syntax in general. This too could be a separate release.
* Use of `hs-bindgen` as a library: the first release should release only the executable, not yet the library. This means that we don't commit to a particular API yet

Features that require more thought:

* Patterns for manually written high-level bindings
* Features related to tracking versions of generated bindings
* Automatic generation of a high-level API

edsko:
We should do due diligence on all cabal fils also (version bounds etc.)

PR #1861, opened 2026-03-24
Replace libclang CXComment with Doxygen binary XML parsing

bolt12:


hs-bindgen/src-internal/HsBindgen/Doxygen.hs
```diff
@@ -0,0 +1,794 @@
...
+
+-- | Parsed doxygen state for a single C header
+--
+-- Map keys are C names used for lookup only (never mangled).
+-- The @DoxyComment Text@ values contain @Text@ at cross-reference
+-- positions (@\<ref\>@ elements in the XML). These @Text@ values are
+-- C names that will be wrapped in @CommentRef ref Nothing@ during
+-- the Parse pass, and then resolved to Haskell names by MangleNames.
+--
+data DoxygenState = DoxygenState {
```
bolt12: Rename this type to just `Doxygen`

bolt12: Have a Key type so that we can merge some of the fields (maps)

```hs
data DoxygenKey = 
    KeyDecl Text
  | KeyField Text 
  | KeyEnum Text
  ..
```

and then `index :: Map DoxygenKey (DoxygenComment Text)`

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -973,3 +999,56 @@ visibilityCanCauseErrors ::
...
+    PrelimDeclId.Named cname -> Just cname.text
+    PrelimDeclId.Anon _      -> Nothing
+
+-- | Enrich anonymous nested declarations with doxygen comments
+--
+-- For anonymous inner structs\/unions, doxygen stores the struct-level comment
+-- as a field comment of the parent (e.g., @data_union_t::as_parts@). After
+-- parsing members, we match nested declarations to their corresponding
+-- fields using type dependencies, look up the field comment, and attach it.
+enrichAnonDeclComments ::
```
bolt12: Make this into a separate pass so that we decouple the parser from the doxygen comment enriching/creation.

bolt12: Add to DeclInfo the parent name, so that we can retrieve it inside the pass

bolt12:
@jorisdral maybe you have some thoughts about how the new doxygen to haddock comments should fit in the whole parser/passes workflow in hs-bindgen. 

We are using `doxygen` executable, so we now have a `Doxygen` state that has all the mappings C name -> Comment. The way I did it in this PR (which will have to change) is that, at the parser level, when we are generating the struct decls, e.g., I get the `Doxygen` state from the parser state, query the state to look for the structs comment and then I do everything that's required to create the comment and add it to the struct decl. See: [hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs](https://github.com/well-typed/hs-bindgen/pull/1861/changes/6c241fd98e47c1968a1e92b81d8a533b7a4d91e9#diff-99799ef54c69cbceb68c63d2608ba344c79030d6078e73d89488a9f80f368004).

@edsko suggested I added a new pass to enrich the Declarations with the comment so that I decouple the parse decls from the comment enrichment. Maybe you have a better/different opinion?

edsko:
@joris Specifically, we would need to know nesting structure of the C code. I was thinking perhaps adding `declParent :: Maybe (Id p)` to `DeclInfo`, but I don't know if that makes sense or not; you've thought a lot more about nesting lately, which is why I figured maybe you have some thoughts on this topic.

doxygen-parser/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Revision history for doxygen-parser
+
+## 0.1.0 -- 2026-03-26
```
jorisdral:
```suggestion
## 0.1.0 -- Unreleased
```

doxygen-parser/doxygen-parser.cabal
```diff
@@ -0,0 +1,83 @@
...
+    DerivingStrategies
+    DuplicateRecordFields
+    NoFieldSelectors
+    OverloadedRecordDot
+    OverloadedStrings
+
+library
+  import:          lang
+  hs-source-dirs:  src
+  ghc-options:     -Wmissing-export-lists
```
jorisdral: The warning can also go into `lang`

doxygen-parser/doxygen-parser.cabal
```diff
@@ -0,0 +1,83 @@
...
+    , doxygen-parser
+    , filepath        >=1.4     && <1.6
+    , tasty           >=1.5     && <1.6
+    , tasty-golden    >=2.3     && <2.5
+    , tasty-hunit     >=0.10.2  && <0.11
+    , text            >=1.2     && <2.2
+    , xml-conduit     >=1.9     && <1.11
+
+-- Internal dependencies
+-- External dependencies
```
jorisdral: These comments look like they were moved around by `cabal-fmt`. If you don't want that, put comments before fields, not inside them. i.e., use something like:

```
  -- Internal dependencies
  build-depends: ...
  -- External dependencies
  build-depends: ...
```

instead of

```
  build-depends: 
    -- Internal dependencies
    ...
    -- External dependencies
    ...
```

The latter will be reformatted, the former won't

hs-bindgen/CHANGELOG.md
```diff
@@ -27,6 +27,13 @@
 
 ### New features
 
+* Extract Haddock documentation from C headers using doxygen. When the `doxygen`
+  binary is available on `PATH`, `hs-bindgen` invokes it to parse structured
+  documentation comments (Javadoc/Doxygen-style `/** ... */`) and translates
+  them into Haddock comments on the generated Haskell bindings. If `doxygen`
+  is not installed, `hs-bindgen` silently falls back to generating
+  metadata-only comments (source location and C name) with no documentation
+  content.
```
jorisdral: Should be placed in a new release section (alpha2 was recently tagged)

hs-bindgen/CHANGELOG.md
```diff
@@ -58,6 +65,9 @@
 
 ### Minor changes
 
+* New build dependency on `xml-conduit` for parsing doxygen XML output.
+  The `doxygen` binary is an optional runtime dependency — if absent, comment
+  extraction is silently skipped.
```
jorisdral: Should be placed in a new release section (alpha2 was just recently tagged)

doxygen-parser/doxygen-parser.cabal
jorisdral, line 1: I think it would make sense to move `doxygen-parser` to its own repo, in part so that we can test it with many `doxygen` versions

hs-bindgen/CHANGELOG.md
jorisdral: I see only a new `doxygen-parser` dependency

hs-bindgen/CHANGELOG.md
bolt12: Yeah, I need to rebase, probably.

hs-bindgen/CHANGELOG.md
bolt12: Thanks!

doxygen-parser/doxygen-parser.cabal
bolt12: Yeah, this is just so we can review everything together, once it is ready I think that's the most sensible thing to do too!

hs-bindgen/CHANGELOG.md
bolt12: I need to update things!

doxygen-parser/src/Doxygen/Parser/Internal.hs
```diff
@@ -0,0 +1,980 @@
...
+--
+data DoxygenKey
+  = KeyDecl Text
+    -- ^ Function, typedef, variable, or enum (keyed by C name)
+  | KeyStruct Text
+    -- ^ Struct\/union compound (keyed by C type name)
+  | KeyField Text Text
+    -- ^ Struct\/union field: @(parent struct name, field name)@
+  | KeyEnumValue Text Text
+    -- ^ Enum value: @(parent enum name, enumerator name)@
```
bolt12: Make the constructors records and use NoFieldSelectors.

doxygen-parser/src/Doxygen/Parser/Internal.hs
```diff
@@ -0,0 +1,980 @@
...
+lookupStruct :: Text -> Doxygen -> Maybe (DoxyComment Text)
+lookupStruct name = lookupComment (KeyStruct name)
+
+-- | Look up a struct\/union field comment
+lookupField :: Text -> Text -> Doxygen -> Maybe (DoxyComment Text)
+lookupField structName fieldName = lookupComment (KeyField structName fieldName)
+
+-- | Look up an enum value comment
+lookupEnumValue :: Text -> Text -> Doxygen -> Maybe (DoxyComment Text)
+lookupEnumValue enumName valueName = lookupComment (KeyEnumValue enumName valueName)
```
bolt12: No need for these, delete

doxygen-parser/src/Doxygen/Parser/Internal.hs
```diff
@@ -0,0 +1,980 @@
...
+{-------------------------------------------------------------------------------
+  Result
+-------------------------------------------------------------------------------}
+
+-- | Result of parsing doxygen XML output
+data Result = Result {
+    doxygen        :: Doxygen
+    -- ^ The parsed comments and group information
+  , warnings       :: [Warning]
+    -- ^ Warnings about unsupported or degraded content
```
bolt12: Reword that this are our warnings, reference DoxygenException for doxygen errors.

doxygen-parser/src/Doxygen/Parser/Internal.hs
```diff
@@ -0,0 +1,980 @@
...
+{-------------------------------------------------------------------------------
+  Errors
+-------------------------------------------------------------------------------}
+
+-- | Exception thrown when doxygen invocation or XML parsing fails
+data DoxygenException
+  = DoxygenNotFound
+    -- ^ The @doxygen@ binary was not found on @PATH@
+  | DoxygenFailed Int String
+    -- ^ @doxygen@ exited with a non-zero exit code.
```
bolt12, line 231: Mention in the comment that this String is the stderr from the doxygen executable call

doxygen-parser/src/Doxygen/Parser/Internal.hs
```diff
@@ -0,0 +1,980 @@
...
+  | DoxygenXMLParseError FilePath String
+    -- ^ Failed to parse a doxygen XML output file.
```
bolt12, line 235: This String is the error as reported by xml-conduit

doxygen-parser/src/Doxygen/Parser/Internal.hs
```diff
@@ -0,0 +1,980 @@
...
+
+-- | Parse C\/C++ header files using doxygen
+--
+-- Invokes the @doxygen@ binary on the given header files, parses the XML
+-- output, and returns a 'Result' containing all extracted comments, group
+-- information, and any warnings about unsupported content.
+--
+-- Throws 'DoxygenException' on failure.
+--
+parse :: Config -> [FilePath] -> IO Result
```
bolt12: Make this list non empty and change doxygenVersion field to not be a Maybe anymore

doxygen-parser/src/Doxygen/Parser/Internal.hs
```diff
@@ -0,0 +1,980 @@
...
+-------------------------------------------------------------------------------}
+
+-- | Parse @index.xml@ to extract the declaration→group mapping and version.
+--
+-- For each member in a file compound, the @refid@ attribute tells us which
+-- group (if any) the member belongs to.  The doxygen version is extracted
+-- from the root element's @version@ attribute (e.g., @\"1.15.0\"@).
+--
+parseIndexXML :: FilePath -> IO (Map Text Text, Maybe Text)
+parseIndexXML path = do
```
bolt12: Use a similar structure to the inlineCommand parser, where we pattern match on each different element, e.g. compound, member, and decide what to do from there, while capturing the "unsupported" ones as a warning.

doxygen-parser/src/Doxygen/Parser.hs
```diff
@@ -0,0 +1,46 @@
...
+--
+module Doxygen.Parser (
+    -- * Configuration
+    Config(..)
+  , defaultConfig
+    -- * Parsing
+  , parse
+  , Result(..)
+    -- * State type
+  , Doxygen
```
bolt12: Expose the group sections

bolt12:
```suggestion
  , Doxygen -- Opaque
```

doxygen-parser/test/Main.hs
bolt12, line 1: Add a test that feeds an XML string that doesn't have the structure that the parser is expecting, e.g. <noncompound> tag and check that the parser does what is supposed to

doxygen-parser/src/Doxygen/Parser/Internal.hs
```diff
@@ -0,0 +1,980 @@
...
+    blockElements =
+      [ "parameterlist", "simplesect", "programlisting"
+      , "itemizedlist", "orderedlist", "xrefsect"
+      , "table"
+      ]
+
+-- | Parse inline content from an XML element
+parseInline :: XML.Element -> Cursor -> ([Warning], [DoxyInline Text])
+parseInline el cursor = case XML.nameLocalName (XML.elementName el) of
+  "bold"           -> let (w, is) = parseInlineChildren cursor in (w, [DoxyBold is])
```
bolt12: check if when using \\b or \<b> if doxygen outputs \<bold> xml elements, the same for other HTML inline tags.

doxygen-parser/src/Doxygen/Parser/Types.hs
```diff
@@ -0,0 +1,158 @@
...
+  | DoxyItemizedList [[DoxyBlock ref]]
+    -- ^ Bullet list: @\<itemizedlist\>@. Each element is one list item.
+  | DoxyOrderedList [[DoxyBlock ref]]
+    -- ^ Numbered list: @\<orderedlist\>@. Each element is one list item.
+  | DoxyXRefSect Text [DoxyBlock ref]
+    -- ^ Cross-reference section: @\<xrefsect\>@
+    --
+    -- Used for @\@deprecated@ and similar. First field is the title
+    -- (e.g., \"Deprecated\"), second is the description.
+  | DoxyUnsupported Text
```
bolt12: Rename this to DoxyTag, this constructor should be used following these guidelines:

- If there is a doxygen tag/(inline) command it will be represented as an explicit constructor
- otherwise, e.g. HTML tags that do not overlap/or have a common doxygen tag/command will be represented by DoxyTag.

Make DoxyTag recursive so that we offer a better constructor for unsupported XML/HTML tags that will still recognise other block level contents.

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -673,15 +673,15 @@ typedefFunPtrDecs supInsts fns haddockConfig sizeofs origInfo n (args, res) orig
...
-            , Clang.InlineRefCommand $
-                C.CommentRef
-                  origInfo.id.cName.name.text
-                  (Just origInfo.id)
+    auxComment = C.Comment $ Doxy.DoxyComment {
+          brief = [
+              Doxy.DoxyText "Auxiliary type used by"
+            , Doxy.DoxyRef
+                (C.CommentRef origInfo.id.cName.name.text (Just origInfo.id))
+                origInfo.id.cName.name.text
```
bolt12: This makes me realise that we no longer need to attach C.Comment's to C.Decl's. Previously we needed to do that because we were getting the comments by libclang via CXComment and we would have to mangle the comments and translate them, but now we get all the comments from doxygen-parser and store then in `Doxygen` state. EDIT: Nevermind, we need to mangle the references in C comments.

See #1885

hs-bindgen/src-internal/HsBindgen/Config/Internal.hs
```diff
@@ -23,6 +23,8 @@ import HsBindgen.Frontend.Predicate (Boolean, SelectPredicate)
 import HsBindgen.Imports
 import HsBindgen.Util.Tracer
 
+import Doxygen.Parser (Config, defaultConfig)
```
bolt12: import this qualified.

doxygen-parser/src/Doxygen/Parser/Types.hs
bolt12, line 1: We should rename the types and their constructors and avoid the Doxy prefix. These should be imported qualified anyway.

Add a comment on the module header saying this is intended for qualified import with an example.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Monad/Decl.hs
```diff
@@ -47,6 +47,7 @@ import HsBindgen.Frontend.RootHeader (HashIncludeArg, RootHeader)
 import HsBindgen.Imports
 import HsBindgen.Util.Tracer
 
+
```
bolt12: no needed

doxygen-parser/src/Doxygen/Parser/Types.hs
```diff
@@ -0,0 +1,158 @@
...
+{-------------------------------------------------------------------------------
+  Comment types
+-------------------------------------------------------------------------------}
+
+-- | A Doxygen comment with brief and detailed sections
+--
+-- Corresponds to the @\<briefdescription\>@ and @\<detaileddescription\>@
+-- elements in Doxygen XML output.
+--
+data DoxyComment ref = DoxyComment {
```
bolt12: Explain why we have a `ref` type variable here, explain what it is.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -38,6 +36,7 @@ import HsBindgen.Frontend.RootHeader (HashIncludeArg)
 import HsBindgen.Imports
 import HsBindgen.Language.C qualified as C
 
+
```
bolt12: No need

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -219,7 +218,7 @@ structDecl ctx info = \curr -> do
 
         -- Recursively parse all members of the struct. These members include
         -- field declarations and nested struct/union declarations.
-        parseStructMembersWith ty ctx parseDeclNested $ \membersResult ->
+        parseStructMembersWith ty ctx parseDeclNested $ \membersResult -> do
```
bolt12: no need

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -294,7 +293,7 @@ unionDecl ctx info = \curr -> do
 
         -- Recursively parse all members of the union. These members include
         -- field declarations and nested struct/union declarations.
-        parseUnionMembersWith ty ctx parseDeclNested $ \membersResult ->
+        parseUnionMembersWith ty ctx parseDeclNested $ \membersResult -> do
```
bolt12: no need

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -978,3 +976,4 @@ visibilityCanCauseErrors ::
   -> Bool
 visibilityCanCauseErrors NonPublicVisibility ExternalLinkage False = True
 visibilityCanCauseErrors _ _ _ = False
+
```
bolt12: no need

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/EnrichComments.hs
bolt12, line 1: Revisit this code and check that it works for arbitrary depths (e.g. a::b::c::d), the AST might really need to be extended, we probably don't have the all the information in the AST.

bolt12: What happens on doxygen-parser with:

```c
struct s1 {
  struct {
    int x;
    int y;
  }
}
```

what gets into the Doxygen state map? How does the XML look like? What about other variations and nesting.

doxygen-parser/src/Doxygen/Parser/Internal.hs
bolt12: That's right, I have confirmed that doxygen normalises this and a bunch of other commands

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/EnrichComments.hs
```diff
@@ -0,0 +1,276 @@
...
+--         struct {
+--             int a;
+--             struct inner_inner_inner {
+--                 int c;
+--             };
+--         } inner_inner;
+--     } field;
+--     int z;
+-- };
+-- @
```
bolt12, line 24: Adding the doxygen dump here would be much easier to understand the example!

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/EnrichComments.hs
```diff
@@ -0,0 +1,276 @@
...
+
+-- | Index for following 'declParent' chains
+type DeclIndex = Map PrelimDeclId (C.Decl Parse)
+
+buildDeclIndex :: [ParseResult Parse] -> DeclIndex
+buildDeclIndex results = Map.fromList
+    [ (success.decl.info.id, success.decl)
+    | pr <- results
+    , ParseResultSuccess success <- [pr.classification]
+    ]
```
bolt12: Review this, because we already build our own DeclIndex, we should re-use it. This means declParent should probably be Id p and not PrelimDeclId

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -66,26 +64,26 @@ topLevelDecl = foldWithHandler handleParseExceptions parseDeclTopLevel
...
 parseDeclTopLevel :: Parser
-parseDeclTopLevel = parseDecl' Nothing
+parseDeclTopLevel = parseDecl' Nothing Nothing
 
--- | Auxiliary function; use 'parseDecl' or 'parseDeclTopLevel'
-parseDecl' :: HasCallStack => Maybe ParseCtx -> Parser
-parseDecl' mCtx = withCursorKindNoCtx $ \case
+-- | Auxiliary function; use 'parseDeclNested' or 'parseDeclTopLevel'
+parseDecl' :: HasCallStack => Maybe PrelimDeclId -> Maybe ParseCtx -> Parser
+parseDecl' mParent mCtx = withCursorKindNoCtx $ \case
```
bolt12: @jorisdral could you take a look at what I did here and check if its sound/makes sense to pass the parent through the parser in order to set declParent ?

hs-bindgen/hs-bindgen.cabal
```diff
@@ -236,6 +237,7 @@ library internal
   build-depends:
     , c-expr-dsl
     , c-expr-runtime
+    , doxygen-parser
```
jorisdral, line 241: Not really related to the changes in this PR, but shouldn't all these "internal dependencies" get version bounds eventually? Does that make them external dependencies? @edsko

.github/workflows/examples.yml
```diff
@@ -69,6 +69,17 @@ jobs:
...
+    - name: 🛠️ Install doxygen
+      env:
+        DOXYGEN_VERSION: "1.15.0"
+        DOXYGEN_SHA256_LINUX: "0ec2e5b2c3cd82b7106d19cb42d8466450730b8cb7a9e85af712be38bf4523a1"
+      run: |
+        curl -fsSL -o /tmp/doxygen.tar.gz "https://github.com/doxygen/doxygen/releases/download/Release_${DOXYGEN_VERSION//./_}/doxygen-${DOXYGEN_VERSION}.linux.bin.tar.gz"
+        echo "${DOXYGEN_SHA256_LINUX}  /tmp/doxygen.tar.gz" | sha256sum --check
+        tar xz -C /tmp -f /tmp/doxygen.tar.gz
+        sudo cp /tmp/doxygen-${DOXYGEN_VERSION}/bin/doxygen /usr/local/bin/
+        doxygen --version
```
jorisdral: This install step might be nice to put in its own action, so that the install logic is not copied across multiple workflow files

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/EnrichComments.hs
```diff
@@ -0,0 +1,301 @@
...
+
+import Doxygen.Parser (Doxygen, DoxygenKey (..), lookupComment)
+import Doxygen.Parser.Types qualified as Doxy
+
+{-------------------------------------------------------------------------------
+  Top-level
+-------------------------------------------------------------------------------}
+
+-- | Enrich parsed declarations with doxygen comments
+enrichComments :: Doxygen -> [ParseResult AssignAnonIds] -> [ParseResult AssignAnonIds]
```
jorisdral: Is it a conscious decision to not define a separate `EnrichComments` tag and use it here? There is no `IsPass` instance for this pass and that makes it inconsistent with the other passes

```suggestion
-- | Enrich parsed declarations with doxygen comments
enrichComments :: Doxygen -> [ParseResult AssignAnonIds] -> [ParseResult EnrichComments]
```

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Decl.hs
```diff
@@ -109,6 +110,14 @@ data DeclInfo p = DeclInfo{
     , headerInfo   :: HeaderInfo
     , availability :: Availability
     , comment      :: Maybe (Comment p)
```
jorisdral: If this is always `Nothing` until after the `EnrichComments` pass, might it make sense to add an associated type to `IsPass`? For example, something like:

```hs
class IsPass p where 
   type CommentType p :: Type
```
with instances like so:
```hs
instance IsPass Parse where
  type CommentType Parse = ()

instance IsPass AssignAnonIds where
  type CommentType AssignAnonIds = ()
  
instance IsPass EnrichComments where
  type CommentType EnrichComments = Comment EnrichComments
```

This ensures statically that comments can't exist until after the `EnrichComments` pass

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Decl.hs
```diff
@@ -109,6 +110,14 @@ data DeclInfo p = DeclInfo{
...
+    , declParent   :: Maybe (Id p)
+      -- ^ Immediate parent declaration, if this declaration is nested.
+      --
+      -- Set during parsing for declarations nested inside another
+      -- declaration (e.g., anonymous or named structs\/unions inside a
+      -- parent struct\/union).  'Nothing' for top-level declarations.
+      -- Used by 'EnrichComments' to build doxygen-qualified names and
+      -- look up parent field comments in the doxygen state.
```
jorisdral: So this parent is set for *all* nested declarations, right? Not just anonymous structs/unions? Makes sense. Do we only use this information for nested structs/unions, or also in other cases?

jorisdral: The C reference doesn't use the term *parent* declaration. It instead uses the term *enclosing* declaration. Maybe we should use that here too? And if we want to emphasise that it is the *direct*  / *immediate* parent, we use *directly enclosing* or *immediately enclosing* declaration? See also this new terminology section of the manual: https://github.com/well-typed/hs-bindgen/blob/main/manual/LowLevel/Terminology.md#enclosing-structunion

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
jorisdral: Looks sensible to me!

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/EnrichComments.hs
bolt12: Well that's because this is not really a pass like the others, that's why, it keeps things simple, but with this being said, if it does make more sense to make it a full pass then I will. cc @edsko

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Decl.hs
bolt12: I think that's a good argument to make this a full pass!

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Decl.hs
bolt12: only in nested declarations. other cases do not need to construct a qualified doxygen name

bolt12: Well noted!

bolt12:
This also closes #1846

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/AssignAnonIds/IsPass.hs
```diff
@@ -27,10 +27,11 @@ type family AnnAssignAnonIds ix where
...
 instance IsPass AssignAnonIds where
-  type MacroBody  AssignAnonIds = UnparsedMacro
-  type ExtBinding AssignAnonIds = Void
-  type Ann ix     AssignAnonIds = AnnAssignAnonIds ix
-  type Msg        AssignAnonIds = WithCallStack ImmediateAssignAnonIdsMsg
+  type MacroBody   AssignAnonIds = UnparsedMacro
+  type ExtBinding  AssignAnonIds = Void
+  type Ann ix      AssignAnonIds = AnnAssignAnonIds ix
+  type Msg         AssignAnonIds = WithCallStack ImmediateAssignAnonIdsMsg
+  type CommentType AssignAnonIds = ()
```
bolt12: Rename CommentType to just Comment

hs-bindgen/hs-bindgen.cabal
bolt12: There's an issue that is tracking this #945

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/EnrichComments.hs
```diff
@@ -0,0 +1,304 @@
...
+  | not info.id.isAnon = do
+      qualName <- resolveQualifiedName index info
+      lookupComment (KeyDecl qualName) doxy
+        <|> lookupComment (KeyStruct qualName) doxy
+  | otherwise = do
+      enclosingId   <- info.declEnclosing
+      enclosingDecl <- DeclIndex.lookup enclosingId index
+      fieldName  <- findFieldName enclosingDecl info.id
+      enclosingName <- resolveQualifiedName index enclosingDecl.info
+      lookupComment (KeyField enclosingName fieldName) doxy
```
bolt12: This can go, because doxygen only associates a comment with the field in the case of an anonymous declaration. Make sure we don't use `depsOfField` anywhere in this module.

bolt12:
The XML dump should dump our parsed AST instead of the raw XML.

PR #1865, opened 2026-03-24
Add contributor

TravisCardwell:
This PR updates the README to add a new contributor.

Should we list contributors' names in natural name order, or should we mangle names to follow Western conventions (given name(s) before family name(s))?  In this case, I went with Western conventions only because the GitHub ID uses that order, but I am of course happy to change to natural name order if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ bug reports, feature requests, blog posts, etc., including:
 * [Jonathan Lorimer](https://github.com/JonathanLorimer)
 * [Joris Dral](https://github.com/jorisdral)
 * [`Krantz98`](https://www.reddit.com/user/Krantz98)
+* [Lin Jian](https://github.com/jian-lin)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,6 +89,7 @@ bug reports, feature requests, blog posts, etc., including:
 * [Edsko de Vries](https://github.com/edsko)
 * [Finley McIlwaine](https://github.com/FinleyMcIlwaine)
 * [George Thomas](https://github.com/georgefst)
+* [Jian Lin](https://github.com/jian-lin)
```
jian-lin: I kind of prefer the order in my country which puts the family name before the given name, i.e., "LIN, Jian" or "Lin Jian".

> I went with Western conventions only because the GitHub ID uses that order

That's because my preferred GitHub ID is not available when I register my GitHub account 😅

TravisCardwell: Thank you for the feedback!

PR #1870, opened 2026-03-26
Make bindings compile even when `NoFieldSelectors` is enabled globally

### LINKED ISSUE #1657: Generated code fails to build if `NoFieldSelectors` is enabled globally
We should probably include the inverse (`NoNoFieldSelectors`? `FieldSelectors?`) as required language extensions (unless using `--enable-record-dot`).

jorisdral:
Resolves #1657

Rather than using field selectors, we use `getField` calls unconditionally. Generated bindings should then compile regardless of whether `NoFieldSelectors` is enabled.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/SHs/Translation.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/SHs/Translation.hs
@@ -23,7 +23,6 @@ import HsBindgen.Backend.Level
 import HsBindgen.Backend.SHs.AST
 import HsBindgen.Backend.SHs.Macro
 import HsBindgen.Backend.SHs.Translation.Common
-import HsBindgen.Config.Prelims (FieldNamingStrategy (..))
 import HsBindgen.Errors
 import HsBindgen.Frontend.Naming
 import HsBindgen.Imports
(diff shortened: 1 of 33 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/SHs/Translation.hs
```diff
@@ -647,17 +646,12 @@ translateCEnumInstance struct fTyp vMap isSequential fieldNamingStrategy mbComme
...
+
     fromCEnumE :: ClosedExpr
-    fromCEnumE =
-      case fieldNamingStrategy of
-        OmitFieldPrefixes ->
-          eBindgenGlobal HasField_getField `ETypeApp` translateType (Hs.HsStrLit "unwrap")
-        AddFieldPrefixes ->
-          EFree fname
+    fromCEnumE = ELam "x" $
+      EApp (eBindgenGlobal HasField_getField `ETypeApp` translateType (Hs.HsStrLit fnameStr)) (EBound IZ)
```
jorisdral:
```suggestion
    fromCEnumE = eBindgenGlobal HasField_getField `ETypeApp` translateType (Hs.HsStrLit fnameStr)

PR #1881, opened 2026-03-31
Summary of terminology related to anonymous structs/unions

### LINKED ISSUE #1871: Terminology: "anonymous struct/union" vs "implict field"
What we've been calling an "implicit field" inside a struct or a union is referred to as an "anonymous struct" in C. 
I am unsure if there is a name in the spec for what we've been calling an "anonymous" struct/union. If not, perhaps "unnamed" would be a better choice, to go along "unnamed field"; or perhaps "untagged", since structs have a _tag_, not a name?

@TravisCardwell @jorisdral Thoughts here?

jorisdral:
Resolves #1871

### DIFF
--- a/manual/LowLevel/Terminology.md
+++ b/manual/LowLevel/Terminology.md
@@ -1,0 +1,144 @@
+# Terminology
+
+## Public
+
+### Anonymous struct/union
+[t:anon]: #anonymous-structunion
+
+A [field][t:field] that declares a nested [untagged][t:untagged] struct or union
+can optionally omit the [field name][t:named_field]. In this case, the [nested
+struct/union][t:nested] is called an *anonymous struct/union*.
+
+<details>
+<summary>WARNING</summary>
+
+> In our code and documentation, the term *anonymous* is sometimes used
+> erroneously for "things that lack names" in general, such as untagged structs
+> and unions. It is on our TODO list to audit the code and documentation to
+> consistently apply the *anonymous* terminology correctly. See [issue
+> #1893][issue-1893].
+
+[issue-1893]: https://github.com/well-typed/hs-bindgen/pull/1893
+
+</details>
+
+<details>
+<summary>Example</summary>
+
+```c
+struct S {
+  union {
+    int x;
+    char y;
+  };
+};
+```
+
+</details>
+
+<details>
+<summary>Binding generation</summary>
+
+* [Binding generation for anonymous structs](./Translation/02-Structs/Nesting.md#example-e)
+* [Binding generation for anonymous unions](./Translation/04-Unions/Nesting.md#example-e)
+
+</details>
+
+### Bit-field
+[t:bit_field]: #bit-field
+
+A *bit-field* is a special kind of [field][t:field] that has a bit-width, used
+for packing and [padding][t:padding].
+
+### Enclosing struct/union
+[t:enclosing]: #enclosing-structunion
+
+A [field][t:field] is declared inside a struct/union definition. This parent
+definition is called the *enclosing struct/union*.
+
+### Field
+[t:field]: #field
+
+A *field* is described in the C reference as a "member" of a struct/union.
+Fields are declared using variable declarations or [bit-field][t:bit_field]
+declarations.
+
+### File scope
+[t:file_scope]: #file-scope
+
+https://en.cppreference.com/w/c/language/scope.html#File_scope
+
+### Indirect field
+
+[Fields][t:field] of an [anonymous struct/union][t:anon] can be accessed as if
+they were fields of the [enclosing struct/union][t:enclosing]. Such a field is
+called an *indirect field* with respect to the enclosing struct/union. Such
+fields are still fields of the anonymous struct/union as well.
+
+<details>
+<summary>Binding generation</summary>
+
+Haskell bindings are generated for the field declaring the [anonymous
+struct/union][t:anon] rather than for the indirect fields.
+
+</details>
+
+### Named field
+[t:named_field]: #named-field
+
+A [field][t:field] with a name is called a *named field*. All fields should be
+named, with some exceptions: see [unnamed field][t:unnamed_field].
+
+### Nested struct/union
+[t:nested]: #nested
+
+A [field][t:field] can declare a new struct or union type, in which case it is
+called a *nested struct/union*. The nested struct/union can be
+[untagged][t:untagged] as usual. The nested struct/union typically has [file
+scope][t:file_scope] as long as its [enclosing struct/union][t:enclosing] has
+file scope, which is usually if not always the case.
+
+### Padding
+[t:padding]: #padding
+
+Unnamed padding may be inserted by the compiler in between [fields][t:field].
+Unnamed padding can be inserted by the using an [unnamed][t:unnamed_field]
+[bit-field][t:bit_field].
+
+<details>
+<summary>Binding generation</summary>
+
+Unnamed bit-fields are not translated to fields in the corresponding Haskell
+record in the generated Haskell bindings.
+</details>
+
+### Tag
+[t:tag]: #tag
+
+A struct/union/enum (optionally) declares a name in the tag namespace. We refer
+to this name as a *tag*.
+
+### Tagged struct/union/enum
+
+A struct or union or enum with a [tag][t:tag] is called *tagged*.
+
+### Unnamed field
+[t:unnamed_field]: #unnamed-field
+
+A [field][t:field] without a name is called an *unnamed field*.
+[bit-fields][t:bit_field] are allowed to be unnamed. Fields that declare an
+[anonymous struct/union][t:anon] are unnamed by definition.
+
+### Untagged struct/union/enum
+[t:untagged]: #untagged-structunionenum
+
+A struct or union or enum without a [tag][t:tag] is called *untagged*.
+
+## Internal
+
+### Implicit field
+
+Any field that is not directly accessible through the `libclang` API is
+represented as an *implicit field* in `libclang`'s C AST. Fields that declare an
+anonymous struct/union are such implicit fields. Indirect fields are also
+represented in `libclang` internally as implicit fields.
--- a/manual/README.md
+++ b/manual/README.md
@@ -6,6 +6,7 @@
 ## Generating low-level bindings
 
 * [Introduction](LowLevel/Introduction.md)
+* [Terminology](LowLevel/Terminology.md)
 
 ### Usage
 


manual/LowLevel/Translation/09-Terminology.md
```diff
@@ -0,0 +1,45 @@
+# Terminology
+
+**tag name**: struct/union/enum (optionally) declare a name in the tag
+namespace. We can refer to this name as "tag name", or just "tag", or "name" if
+it is clear from the context that it is a tag name.
```
edsko: Let's just call these "tags". "Tag names" sounds odd to me.

manual/LowLevel/Translation/09-Terminology.md
```diff
@@ -0,0 +1,45 @@
...
+
+**tagged struct/union/enum**: a struct or union or enum with a tag name is
+called "tagged" (or "named")
```
edsko: And then let's be consistent here also: these are "tagged", not "named" (and here and elsewhere, "tag" rather than "tag name").

manual/LowLevel/Translation/09-Terminology.md
```diff
@@ -0,0 +1,45 @@
...
+
+**untagged struct/union/enum**: a struct or union or enum without a tag name is
+called "untagged" (or "unnamed")
```
edsko: untagged (not unnamed).

manual/LowLevel/Translation/09-Terminology.md
```diff
@@ -0,0 +1,45 @@
...
+**unnamed field**: a field without a name is called "unnamed". Only bit-fields
+are allowed to be unnamed.
+
+**enclosing struct/union**: a field is declared inside a struct/union
+definition. This parent definition is called the "enclosing" struct/union.
+
+**nested struct/union**: a field can declare a new struct or union type, in
+which case it is called a "nested" struct or union. The nested struct/union can
+be untagged as usual. The nested struct/union typically has file scope as long
+as its enclosing struct/union has file scope (usually if not always the case)
```
edsko: Should we introduce "file scope" somewhere?

manual/LowLevel/Translation/09-Terminology.md
```diff
@@ -0,0 +1,45 @@
...
+
+**anonymous struct/union**: a field that declares a nested untagged struct or
+union can optionally omit the field name. In this case, the nested struct/union
+is called "anonymous".
```
edsko: (Here and elsewhere) It might be helpful to have some examples for some of these.

manual/LowLevel/Translation/09-Terminology.md
```diff
@@ -0,0 +1,45 @@
...
+
+**indirect field**: Fields of an anonymous struct/union can be accessed as if
+they were fields of the enclosing struct/union. Such fields are called
+"indirect" fields with respect to the enclosing struct/union. They are still
+fields to the anonymous struct/union as well.
```
edsko:
```suggestion
fields of the anonymous struct/union as well.
```

manual/LowLevel/Translation/09-Terminology.md
```diff
@@ -0,0 +1,45 @@
...
+
+**implicit field**: any field that is not directly accessible through the
+`libclang` API is represented as an "implicit field" in `libclang`'s C AST.
+Fields that declare an anonymous struct/union are such implicit fields. Indirect
+fields are also represented in `libclang` internally as implicit fields.
```
edsko: Hmm this definition doesn't feel very satisfying, very implementation specific. Could we define this without such a direct reference to the libclang implementation?

manual/LowLevel/Translation/09-Terminology.md
TravisCardwell: What is "file scope?"

I think of "file scope" in the quoted documentation as meaning that the nested `struct`/`union` is only in scope within the file that it is declared.  With this interpretation, it would not be in scope in files that include the file where it is declared, nor in files that are later included.  That is not the case, though, so I must be misunderstanding.

Confirming my understanding of C scope:

<details><summary>Previous include</summary>

```c
// foo.h

#include "bar.h"

struct rect_tl {
  struct pt tl;
  double width, height;
};
```

```c
// bar.h

struct rect {
  struct pt {
    double x, y;
  } tl, br;
};
```

```
    --unique-id ex \
    --hs-output-dir ex-src \
    --create-output-dirs \
    --overwrite-files \
    --select-from-main-header-dirs \
    -I . \
    foo.h
```

This works fine, as I expect.

</details>

<details><summary>Later include</summary>

```c
// foo.h

struct rect {
  struct pt {
    double x, y;
  } tl, br;
};

#include "bar.h"
```

```c
// bar.h

struct rect_tl {
  struct pt tl;
  double width, height;
};
```

```
    --unique-id ex \
    --hs-output-dir ex-src \
    --create-output-dirs \
    --overwrite-files \
    --select-from-main-header-dirs \
    -I . \
    foo.h
```

This works fine, as I expect.

</details>

My understanding is that files are essentially irrelevant with CPP.

manual/LowLevel/Translation/09-Terminology.md
```diff
@@ -0,0 +1,45 @@
...
+fields are declared using variable declarations or bit-field declarations
+
+**bit-field**: a special kind of field that has a bit-width, used for packing
+and padding
+
+**named field**: a field with a name is called "named". All fields except for
+bit-fields should be named.
+
+**unnamed field**: a field without a name is called "unnamed". Only bit-fields
+are allowed to be unnamed.
```
TravisCardwell: Perhaps we should mention that unnamed bit-fields are used to specify padding.  (They are not translated to fields in the Haskell record.)

TravisCardwell (commented):
I am a huge fan of such glossary documentation! :+1:

I am currently working on the binding specifications documentation in the manual, and one of the things I am working on is documentation about C names for various things.  I kept it concise initially, with instructions to generate binding specifications to confirm C names for things that one is unsure about.  I then moved that to a separate section, where more examples can be provided.  I started to wonder if it might be better to put it in a separate file, so that we can document C names used for selection and binding specifications in the same place, though most examples do not apply to selection.  Now I wonder if it might be good to combine with this documentation ("C Terminology and Names" or similar).

(This has been in my list of things to discuss with Edsko for over a week, but other topics have taken priority and we have not discussed it yet.)

manual/LowLevel/Translation/09-Terminology.md
jorisdral: Done

manual/LowLevel/Translation/09-Terminology.md
jorisdral: Done

manual/LowLevel/Translation/09-Terminology.md
jorisdral: Done

manual/LowLevel/Translation/09-Terminology.md
jorisdral: If a declaration has file scope then it means that it is in scope of its translation unit. The C preprocessor includes the contents of `#include`d headers (verbatim), so it does not really matter where the declaration is placed. If the declaration is in an included header or not, the declaration would be part of the same translation unit.

manual/LowLevel/Translation/09-Terminology.md
jorisdral: Done

manual/LowLevel/Translation/09-Terminology.md
jorisdral: I think because it is a `libclang` term with no real meaning in the C reference, it's hard to define it with referring to the implementation. I've put this entry in an "Internal" section for now, maybe that helps? The term implicit field is only used (or should only be used) internally in `hs-bindgen`, so it's important that developers can understand what it means, but it's not so important for users

manual/LowLevel/Translation/09-Terminology.md
jorisdral: Done

manual/LowLevel/Translation/09-Terminology.md
jorisdral: I included one for anonymous structs/unions. Are there others you think need an example?

jorisdral:
@TravisCardwell @edsko thank you for your comments, I pushed a new version, could you take a look? One major change is that I use headers now for each term, so that we can easily refer to them with hyperlinks.

> Now I wonder if it might be good to combine with this documentation ("C Terminology and Names" or similar).

That might make sense :smiley:

manual/LowLevel/Translation/09-Terminology.md
```diff
@@ -0,0 +1,104 @@
...
+[untagged][t:untagged] as usual. The nested struct/union typically has [file
+scope][t:file_scope] as long as its [enclosing struct/union][t:enclosing] has
+file scope, which is usually if not always the case.
+
+### Anonymous struct/union
+[t:anon]: #anonymous-structunion
+
+A [field][t:field] that declares a nested [untagged][t:untagged] struct or union
+can optionally omit the [field name][t:named_field]. In this case, the nested
+struct/union is called an *anonymous struct/union*.
```
TravisCardwell: To confirm: We should no longer refer to an untagged `struct`/`union` as "anonymous" in general?  (Example: an untagged `struct` declared as the type for a named field)

If so, this is a significant terminology change, and we should audit the code.  One of the reasons why we chose the `@` syntax is that the circled "a" is short for "anonymous."

manual/LowLevel/Translation/09-Terminology.md
```diff
@@ -0,0 +1,104 @@
...
+  }
+}
+```
+
+### Indirect field
+
+[Fields][t:field] of an [anonymous struct/union][t:anon] can be accessed as if
+they were fields of the [enclosing struct/union][t:enclosing]. Such a field is
+called an *indirect field* with respect to the enclosing struct/union. Such
+fields are still fields of the anonymous struct/union as well.
```
TravisCardwell: Perhaps we should write that we do *not* translate these to the Haskell API, where they must be accessed via the implicit field.

manual/LowLevel/Translation/09-Terminology.md
```diff
@@ -0,0 +1,104 @@
...
+Unnamed padding may be inserted by the compiler in between [fields][t:field].
+Unnamed padding can be inserted by the user using an [unnamed][t:unnamed_field]
+[bit-field][t:bit_field], and they are not translated to fields in the Haskell
+record.
+
+### Named field
+[t:named_field]: #named-field
+
+A [field][t:field] with a name is called a *named field*. All fields except for
+[bit-fields][t:bit_field] should be named.
```
TravisCardwell: Implicit fields are not named.

TravisCardwell (commented):
Great!

[`manual/README.md`](https://github.com/well-typed/hs-bindgen/blob/jdral/anon-terminology/manual/README.md) needs to be updated to list the terminology page.  Regarding organization, how about moving the terminology page out, as follows?  It will be referenced from usage as well as translation pages.

```md
...
## Generating low-level bindings

* Introduction
* Terminology

### Usage

* Invocation
...
```

TravisCardwell:
Looking at the [rendered page](https://github.com/well-typed/hs-bindgen/blob/jdral/anon-terminology/manual/LowLevel/Translation/09-Terminology.md) now, it is quite concise.  I like it!  One example sicks out as an exception, though.  If we would like to keep it concise, perhaps we could use `<details>` for such examples?

BTW, the example is missing some semicolons.

----

**Anonymous struct/union**

A [field][t:field] that declares a nested [untagged][t:untagged] struct or union can optionally omit the [field name][t:named_field]. In this case, the nested struct/union is called an *anonymous struct/union*.

[t:field]: https://github.com/well-typed/hs-bindgen/blob/jdral/anon-terminology/manual/LowLevel/Translation/09-Terminology.md#field
[t:untagged]: https://github.com/well-typed/hs-bindgen/blob/jdral/anon-terminology/manual/LowLevel/Translation/09-Terminology.md#untagged-structunionenum
[t:named_field]: https://github.com/well-typed/hs-bindgen/blob/jdral/anon-terminology/manual/LowLevel/Translation/09-Terminology.md#named-field

<details><summary>Example</summary>

```c
struct S {
  union {
    int x;
    char y;
  };
};
```

</details>

TravisCardwell:
>> Now I wonder if it might be good to combine with this documentation ("C Terminology and Names" or similar).
>
> That might make sense 😃

If we would like to keep the terminology page concise, then perhaps keeping the C names documentation separate is a good idea.  I like being able to look over terminology without being distracted by too many details.

One idea is to put both on the same page, with a "Glossary" section at top that is concise and links to details in sections below.  Anchors would not be as clean, though, while separate pages would keep anchors clean.

manual/LowLevel/Translation/09-Terminology.md
jorisdral: Right, if we follow the C reference's definition exactly, then untagged structs/unions should not be called anonymous anymore. For now I'll include a warning on the terminology page that the *anonymous struct/union* terminology is not applied consistently, and then I'll open an issue to audit the code after

manual/LowLevel/Translation/09-Terminology.md
jorisdral: Good point, amended

manual/LowLevel/Translation/09-Terminology.md
jorisdral: I added `<details>` sections for notes on binding generation

manual/LowLevel/Translation/09-Terminology.md
jorisdral: I opened #1893

jorisdral:
I pushed a new version with all your suggestions. I also alphabetised the sections for easy searching

> Left some thoughts. I also wonder if we should have some nomenclature that distinguishes between
> 
> ```c
> struct {
>   int x;
>   int y;
> };
> ```
> 
> vs
> 
> ```c
> typedef struct {
>   int x;
>   int y;
> } foo;
> ```
> 
> I think (thought?) that clang referred to the former as "anonymous" but not to the latter; though I could be wrong about that.

@edsko As far as I can tell from the [C reference](https://en.cppreference.com/w/c.html) and [`libclang`'s documentation](https://clang.llvm.org/doxygen/group__CINDEX.html), *anonymous* is really only used for unnamed struct/union fields that declare untagged structs/unions. So if we're using the terminology as outlined in `Terminology.md`, your first example is *an untagged struct*, and the second example is *a typedef of an untagged struct*. 

> Great!
> 
> [`manual/README.md`](https://github.com/well-typed/hs-bindgen/blob/jdral/anon-terminology/manual/README.md) needs to be updated to list the terminology page. Regarding organization, how about moving the terminology page out, as follows? It will be referenced from usage as well as translation pages.
> 
> ...

@TravisCardwell  I applied your suggestion!

> Looking at the [rendered page](https://github.com/well-typed/hs-bindgen/blob/jdral/anon-terminology/manual/LowLevel/Translation/09-Terminology.md) now, it is quite concise. I like it! One example sicks out as an exception, though. If we would like to keep it concise, perhaps we could use `<details>` for such examples?

@TravisCardwell  done!

> BTW, the example is missing some semicolons.

@TravisCardwell done!

jorisdral:
AFAICT the comments were resolved, so I'll put this PR on the merge queue once CI passes. We can of course still make changes to the text in a follow-up PR if needed

PR #1888, opened 2026-04-02
Ask `libclang` for anonymous structs and unions directly

jorisdral:


hs-bindgen/src-internal/HsBindgen/Frontend/AST/Decl.hs
```diff
@@ -159,6 +159,9 @@ data DeclKind p =
 data Struct p = Struct {
       sizeof    :: Int
     , alignment :: Int
+      -- | A struct is anonymous if it is untagged and if it is referenced by a
+      -- single unnamed field.
+    , isAnon    :: Bool
```
jorisdral: Maybe this should be moved into the `ann` field, because they (technically?) won't be anonymous anymore after `AssignAnonIds`

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Decl.hs
```diff
@@ -177,6 +180,9 @@ data StructField p = StructField {
 data Union p = Union {
       sizeof    :: Int
     , alignment :: Int
+      -- | A union is anonymous if it is untagged and if it is referenced by a
+      -- single unnamed field.
+    , isAnon    :: Bool
```
jorisdral: Maybe this should be moved into the `ann` field, because they (technically?) won't be anonymous anymore after `AssignAnonIds`

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/ImplicitFields.hs
```diff
@@ -229,47 +227,52 @@ classifyInputs inputs = Classification {
...
+isReferenced decl fields decls =
+       decl.info.id `elem` map snd (concatMap depsOfField fields)
+    || decl.info.id `elem` map snd (concatMap depsOfStructOrUnion decls)
   where
-    deps = concatMap (\d -> depsOfDecl d.kind) decls ++
-           concatMap depsOfField fields
+    depsOfStructOrUnion d = case d.kind of
+        C.DeclStruct struct -> depsOfStruct struct
+        C.DeclUnion  union  -> depsOfUnion  union
+        _ -> []
```
jorisdral: This version still uses `depsOf`-style functions, but without relying on anything macro related (I think). Would this help you out @dschrempf?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/ImplicitFields.hs
```diff
@@ -250,29 +244,23 @@ isAnonymous decl = case decl.kind of
...
+    not (decl.info.id `elem` concatMap childrenOfStructOrUnion decls)
   where
-    depsOfStructOrUnion d = case d.kind of
-        C.DeclStruct struct -> depsOfStruct struct
-        C.DeclUnion  union  -> depsOfUnion  union
-        _ -> []
+    childrenOfStructOrUnion d = case d.kind of
+          C.DeclStruct struct -> struct.ann
+          C.DeclUnion  union  -> union.ann
+          _                   -> []
```
jorisdral: This version does not use any `depsOf` functions anymore @dschrempf

dschrempf: Of course, this would be the best case. But even the approach before will work I think (depsOfField only uses the type, which directly references the macro IDs).

Even so, we must not forget that `MacroId` is only populated in `TypecheckMacros` (a new pass replacing part of `HandleMacros`). So we wouldn't see any macro dependencies of field types (etc.) here anyways. Also, we do not see any macro dependencies in declarations with macro expansions at this state in the pipeline.

Does this also affect your second version which doesn't use `depsOfField` (directly or indirectly)?

I hope this is OK for your use case.

jorisdral: I think it's fine: we don't need any information about macro dependencies when determining whether we want to generate an implicit field. So it's fine if `MacroId`s are populated only later

hs-bindgen/src-internal/HsBindgen/Frontend/AST/Deps.hs
```diff
@@ -40,6 +38,15 @@ depsOfDecl (C.DeclFunction function) =
 depsOfDecl (C.DeclGlobal ty) =
     depsOfType ty
 
+depsOfStruct :: IsPass p => C.Struct p -> [(ValOrRef, Id p)]
+depsOfStruct struct = concat [
+      concatMap depsOfField struct.fields
```
dschrempf, line 43: Just for the record, `depsOfField` uses `depsOfType`, and `depsOfType` does not know macro dependencies before reparse at the moment.

In theory, we could know about these now (because we know which macros were expanded in a declaration with macro expansions)... @edsko thoughts?

dschrempf (approved):
Thank you! This makes implementation of `depsOfDecl` a lot easier, because we don't have to know about macros _while_ parsing the unit.

PR #1889, opened 2026-04-06
Capture callstacks at trace message creation

bolt12:
Closes #1883


hs-bindgen/src-internal/HsBindgen/Util/Tracer.hs
```diff
@@ -94,6 +95,21 @@ traceWith tracer =
     . ContraTracer.traceWith (unwrap tracer)
     . MsgWithCallStack callStack
 
+-- | Like 'traceWith', but for messages that already have a 'CallStack'
+--
+-- Useful when trace messages are collected in pure code and emitted later.
+traceDelayedWith :: MonadIO m => Tracer e -> MsgWithCallStack e -> m ()
+traceDelayedWith tracer =
```
bolt12:
```suggestion
traceWithExplicitCallStack :: MonadIO m => Tracer e -> MsgWithCallStack e -> m ()
traceWithExplicitCallStack tracer =
```

bolt12: and change `traceWith` to `traceWithImplicitCallStack`

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/AdjustTypes.hs
```diff
@@ -38,7 +39,7 @@ import HsBindgen.Imports (Identity (Identity))
 --
 adjustTypes ::
      C.TranslationUnit MangleNames
-  -> (C.TranslationUnit AdjustTypes, [Msg AdjustTypes])
+  -> (C.TranslationUnit AdjustTypes, [MsgWithCallStack (Msg AdjustTypes)])
```
bolt12: I should change Msg X to be Msg X = MsgWithCallStack X

edsko (approved):

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/IsPass.hs
```diff
@@ -39,7 +40,7 @@ instance IsPass Parse where
   type MacroBody  Parse = UnparsedMacro
   type ExtBinding Parse = Void
   type Ann ix     Parse = AnnParse ix
-  type Msg        Parse = WithLocationInfo ImmediateParseMsg
+  type Msg        Parse = MsgWithCallStack (WithLocationInfo ImmediateParseMsg)
```
dschrempf: For consistency, we could frop the `Msg` prefix from `MsgWithCallStack`. It would also make this a bit shorter.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ResolveBindingSpecs.hs
```diff
@@ -63,7 +65,7 @@ resolveBindingSpecs hsModuleName extSpecs pSpec unit =
             UseDeclGraph.deleteRevDeps (Map.keys state.extTypes)
           . UseDeclGraph.deleteDeps (Set.toList state.opqTypes)
           $ unit.ann.useDeclGraph
-        notUsedErrs = ResolveBindingSpecsTypeNotUsed <$> Map.keys state.noPTypes
+        notUsedErrs = map withCallStack $ ResolveBindingSpecsTypeNotUsed <$> Map.keys state.noPTypes
```
dschrempf: Maybe avoidthe doube map here, so it is more readable?

dschrempf (approved):
Thank you!

In general, I feel the tracing infrastructure is already quite heavy, and it gets a bit heavier by this PR; but we need this so call stacks are actually usable...

`traceWith` is completley gone, right? We now have to decide at the call site whether we want to emit a trace message with an explicit call stack or create an implicit call stack during trace emission.

We could also _remove the function_ `traceWithImplicitCallStack`, and always force the caller to use `traceWith (withCallStack msg)`, which is not much longer. (This would remove the `withExplicitCallstack` suffixes when tracing with explicit call stacks).

(Also, I am unsure whether it is call stack -> `CallStack`, or callstack -> `Callstack`...)

Great work, Armando, thank you!

edsko:
I like that suggestion. Return to `traceWith`, but then always take a `WithCallStack` argument. Makes sense.

(Capitalization should be `CallStack`, not `Callstack`, to be consistent with `base`, even though personally I'd have preferred `Callstack` :slightly_smiling_face: )

bolt12:
I agree! Thanks for the great comments!

PR #1890, opened 2026-04-06
Derive --module from file path in literate mode

### LINKED ISSUE #1659: `--module` argument should not be necessary in `literate` mode
We should be able to figure it out from the input filename. 

_Perhaps_ we do something for the `--unique-id` also.

bolt12:
Closes #1659

### DIFF
--- a/examples/literate-example/README.md
+++ b/examples/literate-example/README.md
@@ -53,12 +53,14 @@ literate-example/
 2. The `SimpleBindings.lhs` file contains a Haskell list of command-line arguments:
    ```haskell
    [ "-I", "./c-lib"
-   , "--module=SimpleBindings"
    , "--unique-id", "org.hs-bindgen.literate-example"
    , "--enable-program-slicing"
    , "simple.h"
    ]
    ```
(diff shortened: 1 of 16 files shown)


bolt12:
How does GHC does it? if we have all capital letter directory names, how does GHC know which module name to expect?

bolt12:
I looked into how GHC handles this. GHC never derives module names from file paths, it works the other direction: given a module name (from the cabal file), it constructs the expected path using `-i` source directories and checks if the file exists.

The `-pgmL` preprocessor interface only receives the file path, not the module name or source directories. I confirmed this by tracing the actual arguments and environment variables GHC passes to the preprocessor. GHC has both the module name and `-i` flags at invocation time but doesn't forward them.

This means any path-to-module derivation is a heuristic, and it will break for uppercase source directories. Since we can't solve this correctly without parsing the `.cabal` file, having a flag is the way to go.

I am closing this.

edsko:
Ok, sounds good. Thanks for checking @bolt12 !

PR #1892, opened 2026-04-08
Use the `clang` executable to preprocesses tokens before reparsing

### LINKED ISSUE #1225: Macro type semantics lost when using `__attribute__`
I ran into this with a more complex header (the [C API](https://github.com/taglib/taglib/blob/master/bindings/c/tag_c.h) for [TagLib](https://taglib.org/)), but the following is a minimal example:

```c
#define EXPORT __attribute__ ((visibility("default")))

#define BOOL int

void foo(BOOL);

EXPORT void bar(BOOL);
```

The header uses the [`visibility`](https://gcc.gnu.org/onlinedocs/gcc-15.2.0/gcc/Common-Function-Attributes.html#index-visibility-function-attribute) attribute to mark functions as exported.  It also uses macros to provide cross-platform support for a `BOOL` type, and it is simply `#define BOOL int` on my system.

With this minimal example, the parameter for function `foo` has type `BOOL`:

```haskell
foreign import ccall unsafe "hs_bindgen_ex_88f2e771bb2c9def" foo ::
     BOOL
  -> IO ()
```

The parameter for function `bar` has type `CInt`:

```haskell
foreign import ccall unsafe "hs_bindgen_ex_dca13970f6a30840" bar ::
     FC.CInt
  -> IO ()
```

Expected/desired behavior: the parameter for function `bar` should also have type `BOOL`.

jorisdral:
Resolves #1225


jorisdral:
I'm probably going to park this PR for a bit until #1862 is done

hs-bindgen/src-internal/HsBindgen/Clang/Macros/Floating.hs
```diff
@@ -0,0 +1,77 @@
...
+--
+-- Macro invocations can be floated inwards as long as the reference to the
+-- macro name is not captured by a new macro definition. By extension, macro
+-- invocations can be fully floated inwards as long as the macro name is not
+-- captured by any new macro definition.
+--
+-- TODO <?>: for simplicity we have picked a temporary approximation of this
+-- property. All macro invocations are considered not inwards-floatable as long
+-- as there is a macro name that has multiple definitions. The goal is to write
+-- a better approximation later.
```
jorisdral, line 27: For simplicity we have picked a temporary approximation of this property. All macro invocations are considered not inwards-floatable as long as there is a macro name that has multiple definitions. The goal is to write a better approximation later.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/PrepareReparse/IsPass.hs
```diff
@@ -0,0 +1,238 @@
...
+    --
+    -- The generated Haskell bindings might not fully support macro types as a
+    -- result.
+  | PrepareReparsePreprocessorFailed
+
+    -- * Info
+
+    -- | There are macro invocations that are not inwards floatable, so we can
+    -- not expand them.
+  | PrepareReparseInvocationsNotFloatable
```
jorisdral:
```c
#define A int
#undef A
#define A car
#define B int y

/* global */ A x; /* global */
/* global */ B; /* global */
```

edsko (approved):

edsko:
Will be great to get this in, a significant upgrade to how we deal with macros!

hs-bindgen/src-internal/HsBindgen/Clang/Macros/Floating.hs
jorisdral: opened an issue for this, see #2010

PR #1895, opened 2026-04-09
Fix reversed Haddock order of TH-generated declarations

### LINKED ISSUE #1832: Unintuitive order of Haddocks in TH mode
I have the following C file:

```c
struct A {
  int dup;
};

struct B {
  int dup;
};
```

When generating bindings in TH mode, and running `cabal haddock`, the order of Haddocks is reversed (`B` before `A`). This is unintuitive.

I checked, and the actual order of declarations provided by `withHsBindgen` is correct (`A` before `B`). It seems that the order of Haddocks does not correspond to the order of the declarations.

bolt12:
Closes #1832


bolt12:
This fix is fragile, because it's compensating for undocumented GHC behavior. If GHC ever changes, our reversal would break Haddock ordering. Also one thing to note is that the TH generated code is reversed now, but I think that's okay. It won't affect tests because we run our own Qu monad.

edsko:
Is this happening because we attach documentation by registering omething to run "at the end", and by doing that repeatedly, we effectively add documentation in reverse order?

If that's the reason, we should probably add a comment to the "add finalizer" thing, to remind ourselves that if we ever remove the use of a finalzier, that the order might change.

bolt12:
> Is this happening because we attach documentation by registering omething to run "at the end", and by doing that repeatedly, we effectively add documentation in reverse order?
> 
> If that's the reason, we should probably add a comment to the "add finalizer" thing, to remind ourselves that if we ever remove the use of a finalzier, that the order might change.

Yes, I think that's essentially why GHC behaves like this.

edsko (approved):

PR #1899, opened 2026-04-10
Speed up PP/TH fixture compilation test suites

### LINKED ISSUE #1612: Turn the `compile-fixtures.sh` script into a Haskell script
placeholder

bolt12:
Fixture compilation tests (TH and PP) previously generated a cabal package with ~340 internal libraries and ran cabal build -j1. The -j1 was needed because concurrent ghc-pkg registrations corrupt the shared package database (#1813). This made the full test suite take a lot of time (~9 minutes in my machine). The new approach reads the .ghc.environment file that cabal build already generates (via `write-ghc-environment-files: always` in cabal.project) and invokes `ghc --make` directly for each fixture in parallel. This eliminates the `ghc-pkg` race entirely and brings the full suite down to ~26 seconds (in my machine). A ~20x speedup! Haddock mode uses haddock --ghc --html via the same path.

This closes #1612


hs-bindgen/test/hs-bindgen/Test/HsBindgen/Fixtures/Utils.hs
```diff
@@ -0,0 +1,207 @@
...
+
+    -- Compile each fixture in parallel, bounded by available cores
+    caps <- max 1 <$> getNumProcessors
+    sem <- newQSem caps
+    results <- forConcurrently jobs $
+      \(libName, srcDir, moduleNames, extraArgs) ->
+        bracket_ (waitQSem sem) (signalQSem sem) $ do
+          let outDir = workDir </> "out" </> libName
+          createDirectoryIfMissing True outDir
+          let allGhcOpts = ["-XGHC2021"]
```
bolt12: This should be Haskell 2010

edsko (approved):
Approving modulo checking with some people if environment files have downsides that are important to us.

PR #1907, opened 2026-04-14
Haskell CI: take into account the manual's build plan for the cabal store cache

jorisdral:
This PR makes sure to include the manual's build plan hash in the cache keys

See the commit messages for a detailed description of the changes


.github/workflows/windows.yml
```diff
@@ -231,16 +235,28 @@ jobs:
       - name: "[setup] Generate Cabal plan"
         run: cabal build all --dry-run
 
+      - name: "[manual] Configure manual"
```
TravisCardwell: The bracketed strings at the start of each step name are section names.  There are 6 sections:

1. `init`
2. `checkout`
3. `setup`
4. `build`
5. `test`
6. `manual`

When dispatching a workflow, you can (optionally) run a `tmate` debug session at the end of any of these sections.

With this change, steps for section `manual` are interspersed in other sections.  The step names should instead be renamed to match the sections that they are in.  Suggestions:

* `[manual] Configure manual` should be `[setup] Configure manual`
* `[manual] Generate Cabal plan` should be `[setup] Generate manual Cabal plan`
* `[manual] Build dependencies` should be `[build] Build manual dependencies`

manual/configure.sh
```diff
@@ -0,0 +1,179 @@
...
+
+
+### PATH ###
+
+# Add manual library directory to PATH so that the DLLs are found.
+set_path_windows() {
+  local C_DIR_WINDOWS
+  C_DIR_WINDOWS=$(cygpath -aw "$C_DIR")
+  export PATH="$C_DIR_WINDOWS:${PATH:-}"
+}
```
TravisCardwell, line 65: I have had all kinds of trouble with the Windows `PATH` using GitHub actions.  I am *not* confident that this is a good way to do it.

The previous `PATH` configuration configured the `PATH` for all subsequent steps in the job, while this just configures the `PATH` for the current process (and any sub-processes).  I was curious to how that allows builds to pass and see that it only works because the `generate-and-run.sh` script is *sourcing* the configure script.  We are running `configure.sh` twice: first to configure so that we can hash, and again before building and running.  It is hacky, but I think that is fine because Windows is hacky.

TravisCardwell (changes requested):
Looks good!  I just think we should fix the Windows step names.

.github/workflows/windows.yml
jorisdral: Done! And what about the "save cache" step? I changed it to `[build]` now

dschrempf (approved):
Thank you.

The manual builds and runs fine on NixOS.

(I didn't check the other CI script changes in detail).

manual/configure.sh
jorisdral: Right, we do set env variables twice and that is a bit hacky, though they are only required to be set just before building and running the manual. The same thing happens for `BINDGEN_EXTRA_CLANG_ARGS`,  `LD_LIBRARY_PATH`, and `DYLD_LIBRARY_PATH`. If we really want to persist the changes to environment variables like `PATH` in the workflow file, then we could do something like this:

```diff
diff --git a/.github/workflows/windows.yml b/.github/workflows/windows.yml
index 2d1b08706..4ab1a8ddf 100644
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -239,7 +239,8 @@ jobs:
         working-directory: ${{ env.manual-dir }}
         run: |
           cat hs/cabal.project.ci | tee hs/cabal.project
-          ./configure.sh
+          source ./configure.sh
+          echo "$PATH_ADDITIONS" | tee -a "${GITHUB_PATH}"
 
       - name: "[setup] Generate manual Cabal plan"
         working-directory: ${{ env.manual-hs-dir }}
diff --git a/manual/configure.sh b/manual/configure.sh
index 604c3883e..a6849bc9c 100755
--- a/manual/configure.sh
+++ b/manual/configure.sh
@@ -57,11 +57,13 @@ is_windows() {
 
 ### PATH ###
 
+PATH_ADDITIONS=""
+
 # Add manual library directory to PATH so that the DLLs are found.
 set_path_windows() {
   local C_DIR_WINDOWS
   C_DIR_WINDOWS=$(cygpath -aw "$C_DIR")
-  export PATH="$C_DIR_WINDOWS:${PATH:-}"
+  PATH_ADDITIONS="$C_DIR_WINDOWS:${PATH_ADDITIONS:-}"
 }
 
 if is_windows; then
@@ -69,7 +71,8 @@ if is_windows; then
   set_path_windows
 fi
 
-
+export PATH_ADDITIONS
+export PATH="$PATH_ADDITIONS:${PATH:-}"
 
 ### BINDGEN_EXTRA_CLANG_ARGS ###
```

but for now it's not really necessary

TravisCardwell (approved):
Looks good!  Thanks!

PR #1908, opened 2026-04-14
Replace `Dyngraph` with `Digraph`

TravisCardwell:
The original `DynGraph` module was originally written to provide a simple API for graphs that are constructed incrementally.  Our usage of the data structure changed over time, introducing various issues in addition to loss of consistency.  A new `Digraph` module replaces `DynGraph`.  Some details are in the commit message.

After experimenting with various designs and discussing with @edsko, I went with a design that is not too different from `DynGraph`.  Integral indices are still used internally, so that the data structure provides decent performance, and that makes vertex insertion order the natural ordering, as before.  A benefit of keeping this ordering is that this PR has minimal changes to fixtures, and we can separately improve ordering by making use of the new `topSortBy` function.  Also note that `Digraph` properly supports vertex deletion.

Topological sorting is rewritten so that it does not use post-order traversal, making it easier to reason about.  One significant difference is that the new version detects cycles.  The signature of `topSortBy` is as follows, where the returned `Set v` is a set of vertices that are in one or more cycles.  Returning a pair allows the caller to decide how to deal with the situation.

```haskell
topSortBy :: forall e v.
     Ord v
  => (v -> v -> Ordering)
  -> Digraph e v
  -> (Set v, [v])
```

I found that our topological sort of the include graph was completely messed up.  It worked before, so I suspect that some attempt to tweak the order refactored incorrectly.  All of the changes to fixtures are due to correcting this issue.

Note that vertices that form cycles in include graphs are appended to the sorted list, in `Ord` order.  This somewhat arbitrary handling of cycles will go away if/once we switch to sequence order.

We previously constructed the `UseDeclGraph` first, and then constructed the `DeclUseGraph` by "reversing" it.  With this change, we now construct the `DeclUseGraph` first, and then construct the `UseDeclGraph` by "reversing" it.  The `DeclUseGraph` is now used to construct the ordered list of declarations.  There are no changes to any fixtures in this commit!

Note that cycles in the `DeclUseGraph` currently cause a `panicPure`.

`DynGraph` is removed.


dschrempf (commented):
Hi Travis, thank you!

Just to confirm: In the graph representation, you have only changed how edges are represented, right? I think having a layered `IntMap` makes much more sense, than having the inner `Set (Int, l)` of targets.

Thinking out of the box: Did you consider using a `HashMap`, to avoid the indexing? I understand that the integer indices are used so we can use `IntMap`s instead of normal `Map`s. In my experience, `HashMap`s are considerably faster than normal `Map`s, albeit slower than `IntMap`s, so there is a tradeoff. Asking the other way around, is there a reason that we _cannot use_ `HashMap`s?

As stated in an internal communication, this will lead to non-trivial conflicts with #1862 , where I have changed how we construct the use-decls graph, and fixed topological sort; as well as made some other minor changes to the `DynGraph` modules (such as changing how insertion of edges works). That's not a big deal, but it's good to keep it in mind.

I didn't have a detailed look at the new implementation of `topSort`!

TravisCardwell:
> As stated in an internal communication, this will lead to non-trivial conflicts with #1862 , where I have changed how we construct the use-decls graph, and fixed topological sort; as well as made some other minor changes to the `DynGraph` modules (such as changing how insertion of edges works). That's not a big deal, but it's good to keep it in mind.

I hope that conflicts in the usage of the graph module will be easy to fix.

I took a look at the changes to the graph module:

* `insertEdge` is changed to panic if edges do not exist.  I added a new `unsafeInsertEdge` function to `Digraph` to implement this.  (This imports from `HsBindgen` to use `panicPure`, which is unfortunate, as `Data` modules should not import from `HsBindgen`.  I am leaving it for now.)
* `hasVertex` is added. I added it to `Digraph`.
* `topSort` changes the location of the reverse call.  The new implementation of `topSort` avoids the post-order traversal and reverse complications altogether.  New function `topSortBy` implements Edsko's requested `v -> v -> Ordering` feature, allowing users to specify ordering.

I tried to minimize conflict.  Please let me know if there are any issues.

> Just to confirm: In the graph representation, you have only changed how
edges are represented, right? I think having a layered `IntMap` makes much
more sense, than having the inner `Set (Int, l)` of targets.

Yes.  `Set (Idx, e)` is isomorphic to `IntMap (Set e)`, but the `IntMap` works better in practice.  BTW, one of the improvements is to tighten up the terminology.  A graph has vertices and edges, while visualizations of graphs may use different terminology, which even differs in different implementations.  Mermaid diagrams represent vertices using "nodes" and edges using "links," and both of these can have "labels."  I am happy to add Graphviz visualization if/when requested.

Another change is the addition of `nextIdx`, which tracks the next internal vertex index to allocate.  This is required for the graph to support vertex deletion.

> Thinking out of the box: Did you consider using a `HashMap`, to avoid the indexing? I understand that the integer indices are used so we can use `IntMap`s instead of normal `Map`s. In my experience, `HashMap`s are considerably faster than normal `Map`s, albeit slower than `IntMap`s, so there is a tradeoff. Asking the other way around, is there a reason that we cannot use `HashMap`s?

There is no reason why a graph design could not be implemented using `HashMap`s.  The representation would look something like the following:

```haskell
data Digraph e v = Digraph {
      vertexSet :: !(HashSet v)
    , edgeMap   :: !(HashMap v (HashMap v (Set e)))
    }
```

Comparing performance of `HashMap` and `Map` varies based on the types, usage, and size of the set (of keys).  Because computation of hashes is relatively expensive, `Map` generally outperforms `HashMap` for small sets.  It is just one data point, but the last time I benchmarked this, `HashMap` provided better performance for 13 or more entries (of the types that I was using, on my laptop).  The above sketch uses `Set e` instead of `HashSet e` because there are usually few edges between the same vertices.

Using `HashMap` like in the above sketch would likely provide better performance than the current `IntMap` implementation (when dealing with large sets) for some functions.  One of the benefits of using an internal integral domain, however, is that we can use `Control.Monad.ST` with an array of bits to maintain flags for each entry.  For such functions, the `IntMap` implementation greatly outperforms a `HashMap` implementation.  In the `Digraph` implementation, we still do this in anything that does depth-first traversal.  (We no longer do this for `topSort`/`topSortBy` because of the new way that we handle ordering requires `Set.minView`.)

A (IMHO major) drawback to using `HashSet`/`HashMap` is ordering.  Ordering is relevant to much of the API.  For example, ordering is relevant to any depth-first traversal; the same depth-first search may return different results given different orderings.  With the above sketch, ordering is dependent on the internal implementation of `Hashable`, *which may change* across versions.  In contrast, use of `IntMap` gives us complete control over ordering, and it is also an ordering that makes sense to users.

BTW, I also experimented with a version that uses `Ord` instances for ordering.  I think that it is also a useful way to implement it, though it has the worse performance.

TravisCardwell:
(Rebased)

TravisCardwell:
The `DynGraph` implementation of `topSort` is implemented using a post-order traversal of the depth-first forest and is unable to detect cycles.  A graph with cycles has no topological sort; the returned graph simply does not maintain topological invariants for the vertices that form a cycle.

The new `Digraph` implementation detects cycles.  In my first implementation, it even returned an `Either (Set v) [v]`, where the error case is the (minimal) set of vertices that contain cycles.  It is possible to have working circular includes, however, because macro conditions can break the cycle.  I therefore changed topological sorting to return the tuple `(Set v, [v])`, which allows the caller to decide what to do about the situation.  In the include graph case, the problematic vertices are simply appended (in `Ord` order)  to the end of the sorted vertices (minus the root header, which may never be part of a cycle):

```haskell
toSortedList :: IncludeGraph -> [SourcePath]
toSortedList includeGraph =
    let (cycles, paths) = Digraph.topSort includeGraph.graph
    in  List.delete RootHeader.name paths ++ Set.elems cycles
```

For sorting declarations, now done using the `DeclUseGraph`, cycles are possible in macros.  We do not support such macros, but we should give meaningful error traces.  In the `program-analysis/circular_macro` test, we fail to type check the circular macros, tracing that transitive dependencies are missing.  Sorting declarations uses code like the above so that we continue to get meaningful error traces.

TravisCardwell:
This PR has changes to (only) two fixtures.

`program-analysis/circular_includes`: Haddock "exported by" documentation listed the same header more than once.  The header is now only listed once.

`macros/parse/elaborate`: This example has an "outer" header that includes an "inner" header in between declarations.  We still sort declarations so that declarations in included headers come first (modulo dependencies), so declarations in the inner header that do not depend on declarations in the outer header should come first.  The existing fixture did not follow this, but now the result is as expected.

dschrempf:
Good morning Travis! Thank you for your rigor!

~~I sorted the declarations in `inner` specifically to the location where they are imported in `outer`.~~ (I think this sorting was incomplete; Travis's solutiuon below is the best way to go).
 
Rephrasing the order you are suggesting:

1. Independent declarations in `inner`
2. Declarations in `outer` before import of `inner`
3. Declarations in `inner` depending on declarations in `outer` before the
   import of `inner`
4. Declarations in `outer` after import of `inner`

Is this correct?

TravisCardwell:
Currently, `elaborate_inner.h` is ordered before `elaborate.h`:

```c
// elaborate_inner.h

#define INNER_A OUTER_A
#define INNER_B 2
typedef outer_int inner_int;
```

```c
// elaborate.h

#define OUTER_A 1
typedef int outer_int;

#include "elaborate_inner.h"

#define OUTER_B INNER_A
#define OUTER_C INNER_B
```

We get the following order:

1. `iNNER_B`
2. `oUTER_A`
3. `iNNER_A`
4. `Outer_int`
5. `Inner_int`
6. `oUTER_B`
7. `oUTER_C`

With the "Travis index," the headers would be ordered as follows:

```c
// elaborate.h (up to the #include)

#define OUTER_A 1
typedef int outer_int;

#include "elaborate_inner.h"
```

```c
// elaborate_inner.h

#define INNER_A OUTER_A
#define INNER_B 2
typedef outer_int inner_int;
```

```c
// elaborate.h (after the #include)

#define OUTER_B INNER_A
#define OUTER_C INNER_B
```

We would get the following order:

1. `oUTER_A`
2. `Outer_int`
3. `iNNER_A`
4. `iNNER_B`
5. `Inner_int`
6. `oUTER_B`
7. `oUTER_C`

IMHO, that is the optimal order, and it is unfortunate that we do not get enough information from `libclang` to be able to implement the "Travis index" in all cases.  Perhaps `clang_isBeforeInTranslationUnit` provides the same order, and it is unfortunate that it only works in newer releases of LLVM/Clang.

TravisCardwell:
(Rebased)

edsko (approved):
Feel free to merge once you figured out what is happening with the example `A - {B,C}-loop - D`.

TravisCardwell:
Preempted by linking issues, this PR has not gotten attention for a while, but I am popping my stack and got to it.

Realizing that we should properly handle cycles, not just treat them as edge cases, I implemented a new `Digraph.sort` function.  It organizes vertices of the graph into groups of strongly connected components and does a topological sort of those groups.  Vertices with cycles are no longer put at the end.  It is not named `topSort` because such a function does not satisfy the invariants of topological sorting; there is no topological sort for a graph with cycles.

I had previously implemented a `Digraph.topSortBy` function using an `ExtOrdIdxSet` data structure, an unsatisfactory hack that allows you to maintain a set (balanced binary tree) using a specified ordering function.  I rewrote as `ExtOrdSet`, simplifying it, but then I realized that there is a better way...

I implemented a `Digraph.reIndex` function that transforms the (internal) indexes so that the vertices of the graph are ordered according to a specified ordering function.  Using the resulting graph gives consistent ordering across all of the API functions.  Function `Digraph.sortBy` is now just a simple implementation using this function, for cases where the new order is only needed for that function.  The `Ord` constraint is needed for the internal `Map`.  The `ExtOrdSet` hack is no longer needed!

```haskell
sortBy :: Ord v => (v -> v -> Ordering) -> Digraph e v -> [v]
sortBy cmp = sort . reIndex cmp
```

Graphs with cycles are now handled without issue.  Our test fixtures have the exact same changes as before (described above); there are no additional changes to any fixtures.

PR #1909, opened 2026-04-15
Move bit-field C tests into `feature-tests`

### LINKED ISSUE #1856: Run `feature-tests` with both Clang and GCC
It currently only runs with the C compiler that is shipped with `GHC`

TravisCardwell:
This PR also includes a commit that changes the `feature-tests` `Makefile` to use `clang` to compile C (by default).


examples/feature-tests/c/Makefile
```diff
@@ -1,5 +1,7 @@
 default: libfeature-tests.so
 
+CC = clang
```
jorisdral, line 3: Nice :) does this mean we can consider #1856 resolved when this PR is merged?

jorisdral (approved):
All looks good to me!

examples/feature-tests/c/Makefile
TravisCardwell: I added a commit so that we can indeed close #1856.  In the CI run, I confirmed that it is building with both Clang ([here](https://github.com/well-typed/hs-bindgen/actions/runs/24477789987/job/71534434154?pr=1909#step:25:21)) and GCC ([here](https://github.com/well-typed/hs-bindgen/actions/runs/24477789987/job/71534434154?pr=1909#step:25:321)).

Issue #1910, opened 2026-04-15
Unexpected cursor kind CXCursor_DLLImport


jorisdral:
I noticed that this "unexpected cursor kind CXCursor_DLLImport" message is traced a lot on runs of Windows CI. For example [here](https://github.com/well-typed/hs-bindgen/actions/runs/24399899153/job/71267381648#step:27:114). I'm not sure why we didn't catch this earlier. Is it because these Bug-level messages don't lead to a non-zero exit status? Also, why does the manual not fail to run when these messages are traced? Is it because the declarations where these bug messages are traced are not selected by the `Select` pass?

jorisdral:
@dschrempf would you have ideas?

dschrempf:
Hm, `Bug` traces are to inform users about problems we must fix. They should not happen.

Why we didn't catch this earlier? `Bug` traces are like elevated warnings. They are shown in RED (I don't know why this is not the case for CI); but they do not lead to a non-zero exit code. That's why CI is green.

In my opinion, we should not change how `Bug` messages behave, but we should change our CI scripts. They should fail when they trigger a `Bug` or an `Error`. (Maybe we even want them to fail when they trigger `Warning`s)?

The easiest way to do so, I believe, is to "elevate" `Warning`s and `Bug`s to be `Error`s.

There is already a command line flag for `Warning`s, but not for `Bug`s... I could implement it fast though. (Let me do that right now).

We would need to use `--log-as-error-warnings` (we may think about changing the name of the command line option).

dschrempf:
https://github.com/well-typed/hs-bindgen/pull/1911

(This PR only introduces the `--log-as-error-bugs` flag, it does not use it in CI scripts).

PR #1911, opened 2026-04-15
CLI: log bugs as errors flag

dschrempf:
Add a CLI option `--log-as-error-bugs`.
We need this option for our CI scripts. CI should fail when it encounters a `BUG`. 

See discussion in #1910.


dschrempf:
@edsko I think we can merge this minor change, but wanted to check with you (see PR description).

edsko (approved):

PR #1913, opened 2026-04-15
Fix haddock out-of-scope and ambiguous identifier warnings

### LINKED ISSUE #769: Check Haddocks
Prior to release, we should check that the generated Haddocks are good for

- [x] `hs-bindgen`
- [x]  `clang`
- [x]  `c-expr`  

And possibly others also (`hs-bindgen-runtime`). Ideally `haddock` should run without warnings.

bolt12:
This PR fixes all haddock out-of-scope and ambiguous identifier warnings across the codebase.
Here's a version with structure:

---

While going through the doc-comment rewrites I also ran into a handful of stale identifier references. Some I was able to fix; others I left flagged for someone with more context on the original intent.

**Left as `@X@` monospace, possibly stale:**

- `@Slice@` in `HsBindgen/Frontend/AST/Decl.hs:62` — no such pass exists; program slicing is actually done by the `Select` pass driven by `ProgramSlicing`, so the sentence probably just needs rewriting. (Should be SelectPass).
- `@PiTy@` in `c-expr-dsl/src/C/Expr/Typecheck/Expr.hs:293` — appears alongside `TyConAppTy` in an example of incompatible types, but `Type` only has `TyVarTy | FunTy | TyConAppTy | NomEqPred`, so this was probably meant to be `FunTy`. (Remove the example)
- `@hasFieldInstanceVia@` in `HsBindgen/Backend/Hs/AST.hs:730` (deriveVia is the new field name - if haddocks cant reference it write: @deriveVia@ see field deriveVia of 't'HasFieldInstance') and `@structFieldDecls@` in `HsBindgen/Backend/Hs/Translation/Structure.hs:290` (getFieldDecls that's the new name HsBindgen.Backend.Hs.Translation.Structure, getFieldDecls)

## What's still missing

Running `cabal haddock all --haddock-for-hackage` still produces 124 warnings on stock cabal 3.16.1. They fall into three classes, and none of them are broken doc comments:

**`hs-bindgen:internal` (31 warnings).** `HsBindgen.TH` re-exports names from the `hs-bindgen:internal` sublibrary, and haddock can't resolve them across the sibling-sublibrary every export produces a `HsBindgen.TH: Couldn't find .haddock for export <Name>` warning (`withHsBindgen`, `hashInclude`, `Config`, and ~28 others). This is the same root cause the `c-expr-core` sublibrary had before it was inlined into `c-expr-runtime` in this PR. Inlining fixed `c-expr-core` cleanly (3 modules), but `hs-bindgen:internal` is ~150 modules and inlining it would cost the linker-enforced separation between `src/` and `src-internal/`.

**`Rep_*` auto-derivation (~805 unresolved items across 80 warnings).** Modules that use `deriving Generic` on an exposed type cause haddock to emit `C.Char: could not find link destinations for: - C.Char.Rep_CharValue`, e.g. . `Rep_TypeName` is a synthetic name haddock generates when rendering the `Rep` associated type family instance. Initially I assumed this was another sublibrary symptom, but inlining `c-expr-core` did not change the count. No haddock flag suppresses it. This is a haddock-level limitation, not a project-level structural problem.

**Cross-package references.** References to types from sibling packages (`HsBindgen.BindingSpec.*`, `C.Expr.*`) or from external dependencies (`Data.Vector.*`, `Data.Attoparsec.*`, `GHC.Internal.*`) appear as unresolved because `--haddock-for-hackage` documents each package in isolation. These should resolve on Hackage once dependency docs are installed.

### What was investigated

- **cabal HEAD (3.17.0.0 from `master`).** The `--haddock-for-hackage` counts didn't move.
- **Alternative commands and flags.** `cabal haddock all --haddock-all` on cabal HEAD and `cabal haddock-project` both eliminate the 31 `HsBindgen.TH` warnings, but neither produces a Hackage tarball. Combining `--haddock-all` with `--haddock-for-hackage` reintroduces every warning.

This closes #769


bolt12:
Try cabal header (the latest cabal version) in order to try and see if the internal libraries problem goes away. This means that the comprehensive CI suite should also pull this cabal version.

edsko (approved):

PR #1914, opened 2026-04-16
CI filesystem workflow: EOF newlines and symbolic links

TravisCardwell:
This PR adds a new CI workflow that checks for EOF newlines (required for text files) and symbolic links (forbidden).

Existing test files without EOF newlines are fixed.  Note that this is usually caused by use of Microsoft software (OS or editor).  VS Code users should configure their editors, as the default configuration is not POSIX-compatible even on POSIX OSes.

As discussed, I moved the manual headers into the manual directory, removing those symbolic links.  We no longer test them in the golden tests.  The `omit_field_prefixes.h` example was stored in both the golden test directory and the manual C directory, and the files had diverged.  Investigating with `git blame`, I am pretty sure that the one in the golden test directory was updated and the one in the manual C directory was forgotten.  I kept the newer version, discarding the older version.


TravisCardwell:
(Rebased)

The new check is already proving useful! :smile:

TravisCardwell:
(Rebased)

edsko:
@jorisdral I'm going to go ahead and merge this; if there is anything here that you feel needs tweaking, let me know!

Issue #1898, opened 2026-04-10
Consider loading the `libclang` dynamic library using `dlopen`


TravisCardwell:
In chat, @edsko wrote:

> ... might there by any advantage to using `dl_open` to link to `libclang` at runtime, so that *we* are in control rather than the linker? ...

TravisCardwell:
That is indeed an interesting idea that I had not considered before either.

Currently, `hs-bindgen` is linked to a `libclang` dynamic library that is loaded by the system when executed.  (If the system is unable to load a library, it fails before the executable is executed.)  Our bindings to `libclang` are determined at build-time.  At runtime, we check that the version of the loaded dynamic library matches the version of `libclang` that the bindings were built for.

The situation is completely different if we instead implement our own loading of dynamic libraries.  `hs-bindgen` would not be linked to `libclang` at all.  `hs-bindgen` could load a `libclang` dynamic library and provide an API that depends on the LLVM/Clang version of the library.  A build of `libclang-bindings` (cached in the Cabal store) *could* work with any version of LLVM/Clang that we support, not a specific one that was used to build the library.

When loading a `libclang` dynamic library, we could search for candidates, perhaps prioritize them based on filename, and iterate through them to find one that we can use.  For example, on Linux, the (default) candidates may be `/usr/lib/libclang.*`.  We could similarly search common locations on macOS, and we could traverse the `PATH` on Windows.  We would need to provide a way for users to specify a library path, and Nix users would probably always use that option.

Note that this is platform-specific.  On POSIX systems, [`withDL`](https://hackage-content.haskell.org/package/unix-2.8.8.0/docs/System-Posix-DynamicLinker.html#v:withDL) and [`dlsym`](https://hackage-content.haskell.org/package/unix-2.8.8.0/docs/System-Posix-DynamicLinker.html#v:dlsym) could be used to load the library and get function pointers to specific symbols.  On Windows, [`loadLibrary`](https://hackage-content.haskell.org/package/Win32-2.14.2.1/docs/System-Win32-DLL.html#v:loadLibrary), [`getProcAddress`](https://hackage-content.haskell.org/package/Win32-2.14.2.1/docs/System-Win32-DLL.html#v:getProcAddress), and [`freeLibrary`](https://hackage-content.haskell.org/package/Win32-2.14.2.1/docs/System-Win32-DLL.html#v:freeLibrary) could be used.

[...]

PR #1918, opened 2026-04-21
Add "Cabal store maintenance" section to the manual

### LINKED ISSUE #1244: Cached `libclang-bindings` may not match current configuration

TravisCardwell:
[Rendered](https://github.com/well-typed/hs-bindgen/blob/tcard/manual-cabal-store/manual/Installation.md#cabal-store-maintenance)

Note that this PR updates the `libclang-bindings` revision so that the `CLANG_VERSION` macro is checked.  See https://github.com/well-typed/libclang/pull/57

While the whole situation is still unsatisfying, this PR closes #1244, as we now document two ways for users to work around the issue (clear the cache or use the `CLANG_VERSION` macro).  As noted in https://github.com/well-typed/hs-bindgen/issues/1244#issuecomment-4264401283, we are keeping the `dlopen` solution (#1898) on the table and may implement it in the future if the issue proves particularly problematic for users.

### DIFF
--- a/manual/Installation.md
+++ b/manual/Installation.md
@@ -111,6 +111,84 @@ should be set in the `.cabal` file.
   This is only needed when you want to use a version of LLVM/Clang that is
   not in your current `PATH`.
 
+### Cabal store maintenance
+
+The `libclang` bindings used by `hs-bindgen` are implemented in package
+`libclang-bindings`.  When built, it is linked to the `libclang` and `libLLVM`
+shared objects for the version of LLVM/Clang that you are using, and the built
+`libclang-bindings` library is cached in the Cabal store.  Cabal does not
+consider such system dependencies to determine when a package needs to be
+rebuilt, however, so you might continue to use a library built with an older
+version of LLVM/Clang even after a system upgrade installs a newer version.
+
+Links to LLVM/Clang shared objects generally specify the minor version.  If an
+upgrade replaces an older version of LLVM/Clang with a newer version that has
+the same minor version (and a newer patch version), then the link in the
+cached `libclang-bindings` build is not broken.  The newer version *should*
+still work, though we recommend rebuilding `hs-bindgen`/`libclang-bindings`
+using the newer version.  In this case, `hs-bindgen` displays a warning like
+the following, where the "compile time version" is the version of LLVM/Clang
+used to to build the cached `libclang-bindings` library, and the "runtime
+version" is the version for the shared object that is dynamically linked at
+runtime.
+
+```
+[Warning] [HsBindgen] [boot] clang version mismatch:
+  clang compile time version: 22.1.2
+  clang runtime version:      22.1.3
+```
+
+If an upgrade installs a new major version and removes the older version, then
+the link in the cached `libclang-bindings` build is broken.  In this case, the
+dynamic library loader fails with an error like the following, preventing
+`hs-bindgen` from running at all.
+
+```
+libclang.so.20.1: cannot open shared object file: No such file or directory
+```
+
+Cabal does not provide an easy way to resolve this issue.  Note that this is an
+issue for any Haskell package that links to system libraries that do not have
+`pkg-config` support (in which case Cabal can track the dependency versions).
+It is particularly problematic with LLVM/Clang, however, because LLVM/Clang has
+frequent releases.  Currently, we suggest two ways to work around the issue.
+
+The easiest way to force Cabal to rebuild out-of-date/broken packages is to
+clear the cache.  You can determine your Cabal store directory by running
+`cabal path --store-dir`.  The Cabal store has separate caches for each version
+of GHC.  Recursively remove a GHC directory to clear the cache for that version.
+Note that deleting just a package directory is not advised because it can break
+the package database.  The downside to clearing your cache is that doing so may
+result in time-consuming recompilation of many other packages.
+
+Since we understand that clearing the cache might not be desirable, `hs-bindgen`
+offers a bespoke workaround, in the form of a compile-time setting specifying
+the LLVM/Clang version.  This forces Cabal to distinguish `libclang-bindings`
+builds that link to different versions, forcing a new build if one for the
+specified version is not already in the Cabal store.  Do this by adding the
+following to a `cabal.project.local` file for your project, with the desired
+version.
+
+```
+package libclang-bindings
+  ghc-options: -optc=-DCLANG_VERSION=22.1.3
+```
+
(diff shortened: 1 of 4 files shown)


TravisCardwell:
Note that I did not add a new entry to the `hs-bindgen` changelog, as there are no changes to `hs-bindgen`.  The same `CLANG_VERSION` macro hack works for the whole history of `hs-bindgen`.  The only change is that we now *check* the version when building `libclang-bindings`, and that change is of course in the `libclang` changelog.

This is the kind of information that I would generally put in release notes.

If you would like me to put this in the `hs-bindgen` changelog, please let me know.  ("New features?")

manual/Installation.md
```diff
@@ -111,6 +111,82 @@ should be set in the `.cabal` file.
...
   not in your current `PATH`.
 
+### Cabal store maintenance
+
+The `libclang` bindings used by `hs-bindgen` are implemented in package
+`libclang-bindings`.  When built, it is linked to the `libclang` and `libLLVM`
+shared objects for the version of LLVM/Clang that you are using, and the built
+`libclang-bindings` library is cached in the Cabal store.  Cabal does not
+consider such system dependencies to determine when a package needs to be
+rebuilt, however, so you may continue to use a library built with one version
```
edsko:
```suggestion
rebuilt, however, so you might continue to use a library built an older version
```

manual/Installation.md
```diff
@@ -111,6 +111,82 @@ should be set in the `.cabal` file.
...
+dynamic library loader fails with an error like the following, preventing
+`hs-bindgen` from running at all.
+
+```
+libclang.so.20.1: cannot open shared object file: No such file or directory
+```
+
+Cabal does not provide an easy way to resolve this issue.  Note that this is an
+issue for any Haskell package that links to system libraries (without using
+`pkg-config`), but it is particularly problematic with LLVM/Clang because it
```
edsko: "without using `pkg-config`" -> "unless they provide `pkg-config` support, in which case cabal can track their dependencies just as it does for Haskell packages"

manual/Installation.md
```diff
@@ -111,6 +111,82 @@ should be set in the `.cabal` file.
...
+
+The easiest way to force Cabal to rebuild out-of-date/broken packages is to
+clear the cache.  You can determine your Cabal store directory by running
+`cabal path --store-dir`.  The Cabal store has separate caches for each version
+of GHC.  Recursively remove a GHC directory to clear the cache for that version.
+Note that deleting just a package directory is not advised because it can break
+the package database.  The downside to clearing your cache is that doing so may
+result in time-consuming recompilation of many other packages.
+
+An alternative is to specify the LLVM/Clang version in `ghc-options`.  This
```
edsko:
```suggestion
Since we understand that clearing the cache might not be desirable, `hs-bindgen` offers a bespoke workaround, in the form of a compile-time setting specifying the LLVM/Clang version.  This
```

manual/Installation.md
```diff
@@ -111,6 +111,82 @@ should be set in the `.cabal` file.
...
+forces Cabal to distinguish `libclang-bindings` builds that link to different
+versions, forcing a new build if one for the specified version is not already
+in the Cabal store.  Do this by setting the following in a `cabal.project.local`
```
edsko:
```suggestion
in the Cabal store.  Do this by adding the following to a `cabal.project.local`
```

edsko (approved):

manual/Installation.md
```diff
@@ -111,6 +111,84 @@ should be set in the `.cabal` file.
   This is only needed when you want to use a version of LLVM/Clang that is
   not in your current `PATH`.
 
+### Cabal store maintenance
```
jorisdral, line 114: Is this text also relevant for MacOS and Windows (I think so, right)? Because it is only a sub-section of the Linux section now. Maybe it should be moved out into its own top-level section

TravisCardwell: Yes, it is indeed relevant for macOS and Windows.  With the recent change to only check the major/minor version for compatibility, ignoring the patch version, the issue is on Windows differs from Linux and macOS.  I have moved the documentation to the (new) `libclang-bindings` manual, referenced from the `hs-bindgen` manual.  The documentation is rewritten to make it clear how the issue manifests on the different OSes.

Issue #797, opened 2025-06-26
Manual: Installation


edsko:

TravisCardwell:
[How to use GHC's LLVM backend](https://minoki.github.io/posts/2025-01-06-ghc-llvm-backend.html)

Note that each GHC version supports a specific range of LLVM versions.

PR #1931, opened 2026-04-28
Update and Cleanup Manual

### LINKED ISSUE #821: Manual: functions
https://github.com/well-typed/hs-bindgen/blob/main/manual/LowLevel/Functions.md

bolt12:
Closes #821 and #797


manual/LowLevel/Translation/06-Functions.md
```diff
@@ -500,3 +500,67 @@ and we will generate
 foreign import ccall safe "<userland CAPI wrapper>"
   mod_10 :: CInt -> IO CInt
 ```
+
+## Function attributes
+
+C functions can carry compiler-specific attributes such as `__attribute__((const))`
+or `__attribute__((pure))`. `hs-bindgen` recognises common GCC and Clang function
+attributes and generates bindings normally regardless of which attributes are
+present. For example, given
```
bolt12: We do interpret attributes, e.g. pure, see hash function lives in IO. But double check that there are attributes for which we do not generate bindings.

manual/LowLevel/Translation/06-Functions.md
```diff
@@ -500,3 +500,67 @@ and we will generate
...
+We do not generate bindings for variadic functions, i.e., functions using `...`
+or `va_list` parameters ([#53][issue-53]):
+
+```c
+void f(const char* fmt, ...);
+void g(const char* fmt, va_list args);
+```
+
+These functions are silently skipped during binding generation. Other
+non-variadic functions in the same header are not affected. For example, given
```
bolt12:
```suggestion
These functions are skipped (with a warning) during binding generation. Other
non-variadic functions in the same header are not affected. For example, given
```

manual/LowLevel/Translation/06-Functions.md
```diff
@@ -500,3 +500,67 @@ and we will generate
...
+void f(const char* fmt, ...);
+void h();
+```
+
+`hs-bindgen` will generate a binding for `h` but not for `f`.
+
+The Haskell FFI does not have direct support for variadic calling conventions,
+so supporting variadic functions would require a fundamentally different
+approach such as generating fixed-arity wrappers for each call site.
+
```
bolt12, line 568: Add a note that if one really wants, it is possible to add a manual binding to a fixed arity function, if needed.

edsko (approved):

PR #1933, opened 2026-04-28
Reflect Doxygen kindref attribute to narrow NameMap search

### LINKED ISSUE #1924: Avoid searches of the name map for comment references in `MangleNames`
In the `MangleNames` pass, we also have to mangle references to other declarations in the documentation (`C.Comment`, `C.CommentRef`). At the moment, we do not have any information about the `DeclKind` of the reference targets, and so, need to search the complete `NameMap`.

We can improve on that by reflecting the `DeclKind` so it is available when searching the `NameMap`. Then, we only have to search for a specific name in a specific map (type constructors, data constructors, or variables).

This should only be done after #1922 is merged.

@bolt12

bolt12:
Closes #1924

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames.hs
@@ -39,6 +39,8 @@ import HsBindgen.Imports
 import HsBindgen.Language.Haskell qualified as Hs
 import HsBindgen.Util.Tracer (WithCallStack, withCallStack)
 
+import Doxygen.Parser.Types qualified as Doxy
+
 {-------------------------------------------------------------------------------
   Top-level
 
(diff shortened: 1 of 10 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames.hs
```diff
@@ -821,58 +823,74 @@ instance Mangle C.Comment where
...
+            DeclIdPair declId . Hs.demoteNs <$> lookupType declId nameMap
+
+        -- | Try all three sub-maps: for members we don't know the Haskell
+        -- namespace (typedef → type constr, function → var, enum const →
+        -- data constr).
+        lookupAnyNs :: DeclId -> NameMap -> Maybe DeclIdPair
+        lookupAnyNs declId nameMap =
+            DeclIdPair declId <$> asum
+              [ Hs.demoteNs <$> lookupType declId nameMap
+              , Hs.demoteNs <$> lookupData declId nameMap
```
bolt12: We can skip this one, because these are Haskell data types.

edsko (approved):
If this works out as we discussed, feel free to merge; if not, let's discuss once more.

PR #1934, opened 2026-04-29
Deduplicate trace callstacks

jorisdral:
What I found was that we were creating trace messages with multiple callstacks. For example, a message from the `Parse` pass lifted into the `FrontendMsg` type would look like this when traced:

```hs
WithCallStack {
    callStack = ...
  , traceMsg = FrontendParse WithCallStack {
        callStack = ...
      , traceMsg = ... -- a parse message
      }
  }
```

This is in part because `Msg p` is now set to `WithCallStack ...` for most passes, but our tracer type also separately requires `WithCallStack` on top of trace messages.

```hs
newtype Tracer e = Wrap (ContraTracer.Tracer IO (WithCallStack e))
```

I think it might be better if `Msg p` would be a plain, non-callstacked message type. And then we enforce that plain messages are always annotated with a callstack using a different type synonym, like:

```hs
-- | Trace messages with annotations (like a call stack)
type AMsg p :: Star
type AMsg p = WithCallStack (Msg p)
```

Moreover, I think we can replace `extendCallStackMsg` with `contramap`, but I may be wrong

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass.hs
@@ -1,6 +1,7 @@
 module HsBindgen.Frontend.Pass (
     Pass
   , IsPass(..)
+  , AMsg
   , NoAnn(..)
   , NoMsg
   ) where
@@ -178,6 +179,10 @@ class (
   macroIdId :: Proxy p -> MacroId p -> Id p
   default macroIdId :: MacroId p ~ Void => Proxy p -> MacroId p -> Id p
   macroIdId _ = absurd
+
+-- | Trace messages with call-stack annotations
+type AMsg :: Pass -> Star
+type AMsg p = WithCallStack (Msg p)
 
 {-------------------------------------------------------------------------------
   Defaults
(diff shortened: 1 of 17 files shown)


edsko:
@dschrempf could you review this please? @jorisdral feel free to merge once Dominik gives the green light.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass.hs
```diff
@@ -144,6 +144,10 @@ class (
   -- | Trace messages possibly emitted by the pass
   type Msg p :: Star
 
+  -- | Trace messages with annotations (like a call stack)
+  type AMsg p :: Star
```
dschrempf: Do we this as part of the `IsPass` class? I think it is always the same definition, correct?

dschrempf (approved):
Nice!

Did you check that the callstacks are correctly created at the "trace message creation" site, not the "trace message emission" site?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass.hs
jorisdral: It's always the same definition. I've moved it out of the class

jorisdral:
> Nice!
> 
> Did you check that the callstacks are correctly created at the "trace message creation" site, not the "trace message emission" site?

I did, and it all looks as expected

PR #1941, opened 2026-05-04
Use cabal-configured GHC for fixture compilation

### LINKED ISSUE #1940: Fixture tests fail when the GHC version is configured locally
I use `ghcup` to manage my GHC versions. On my current system I have many versions installed, but `ghc-9.8` is set as the default GHC version. In my `cabal.project.local` file in `hs-bindgen`, I can configure my local GHC version, like so:

```
with-compiler: ghc-9.4
```

If this local version is  different from the system-wide GHC version, then the fixtures tests fail to run with output like this:

```
    types/typedefs/typedefs:                                    FAIL
      PP compilation failed:
      <command line>: cannot satisfy -package-id c-expr-dsl-0.1.0.0-inplace:
          c-expr-dsl-0.1.0.0-inplace is unusable due to missing dependencies:
            base-4.17.2.1 c-expr-runtime-0.1.0.0-inplace containers-0.6.7 fin-0.3.2-c7afe6c3455ee998e41700f36fcf9bfe0f18a534cf25605a6d0336d133a3ea7b libclang-bindings-0.1.0-e9751b34b70a52f9f1671f979e5372e96024b1a8c8024e5d3b46a6d7b6272ba8 mtl-2.2.2 parsec-3.1.16.1 scientific-0.3.8.1-38a7e87430f44abe2b645e99cf0f7719c436742dded015be86b5a54f255af1e0 some-1.0.6-fad533da6bc97df40c5abf3d679723503a1d7a8aa3099a9fd962d59044a86f68 text-2.0.2 vec-0.5.1-387a14866d964e052c8f28132121102870d580119fbde70a327c50789f3775c8
          (use -v for more information)

      Use -p '/pp/&&/pp-fixtures.types\/typedefs\/typedefs/' to rerun this test only.
```

If I locally configure `ghc-9.8` again, the tests succeed again.

bolt12:
Resolve the compiler by parsing the version suffix off the `.ghc.environment.<arch>-<os>-<version>` file cabal already writes, then search PATH for `ghc-X.Y.Z`, then `ghc-X.Y`, then plain `ghc`. Same thing for `haddock-ghc-X.Y.Z` -> `haddock`. Falls back to bare `ghc`/`haddock` when no version-specific binary exists, so CI environments with a single GHC keep working unchanged.

Closes #1940

### DIFF
--- a/hs-bindgen/CHANGELOG.md
+++ b/hs-bindgen/CHANGELOG.md
@@ -58,6 +58,10 @@
 
 ### Bug fixes
 
+* Fixture tests now invoke the same GHC that cabal used to build the test
+  suite, so they no longer fail when `with-compiler` is set to a non-default
+  version and the system's `ghc` on `PATH` is something else. See [issue
+  #1940](https://github.com/well-typed/hs-bindgen/issues/1940).
 * Fix reversed order of Haddock documentation in TH mode
 * Fix `DeclIndex` construction when macros and ordinary declarations share a
   name.
--- a/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Fixtures/Utils.hs
+++ b/hs-bindgen/test/hs-bindgen/Test/HsBindgen/Fixtures/Utils.hs
@@ -7,6 +7,11 @@
 -- package databases and visible packages from the main project build.
 -- Individual fixtures are then compiled in parallel.
 --
+-- The compiler binary is selected from the env file's version suffix (e.g.
+-- @ghc-9.10.3@ from @.ghc.environment.x86_64-linux-9.10.3@), so fixture
+-- compilation uses the same GHC that cabal built the test suite with even
+-- when @with-compiler@ differs from the system default on @PATH@.
+--
 module Test.HsBindgen.Fixtures.Utils (
     -- * Name sanitization
     sanitizeLibName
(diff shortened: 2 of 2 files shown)


bolt12:
What happens if we have different GHC env files for different GHC versions? Does cabal delete the current Env file and creates a new one?

> ghc-X.Y.Z, then ghc-X.Y, then plain ghc

Don't do this fallback, just use the ghc-X.Y.Z

edsko (approved):
Feel free to merge once that changes we discussed are done.

bolt12:
Cabal does not delete stale env files when the active GHC changes, so multiple .ghc.environment.* can coexist. Given this, we can read `compiler-id` from `dist-newstyle/cache/plan.json` and pin fixture compilation to whatever GHC cabal used to build the test suite. I think this is more robust: plan.json gives us the right version, and the env file the package databases and visible packages. What do you think @edsko  ?

edsko:
That works for me, though perhaps we are overengineering this. @jorisdral , any thoughts here?  
(WHat if we actually just deleted the env file at the end of the test run..? :thinking: )

bolt12:
> WHat if we actually just deleted the env file at the end of the test run..?

That might indeed just be a lot simpler! But I think it is not as elegant, the problem was that I was calling `ghc` and it could pick up a different GHC version than what cabal was configured with. So we should make an effort to use the same version that cabal compiled the project with. This means if we have two env files we should use the one that cabal used the last time it compiled the code, which will be in the plan.json.

PR #1944, opened 2026-05-05
Ignore storage and function specifiers in macro reparse

### LINKED ISSUE #1891: Reparse of inline functions fails
Assume

```c
#define ID(X) (X)

static inline int bar(int x) {
  int y = 10;
  return ID(y);
}
```

We fail to reparse this declaration with

```
   'bar' at "/home/dominik/work/hs-bindgen/hs-bindgen/examples/golden/macros/reparse_functions.h 5:19":
                During parse:
                  Failed to reparse:  Encountered unexpected node in the language-c AST:  CStatic NodeOmitted
```

Instead, we should ignore this node, and go on, only reparsing the function header declaration, not the function body.

bolt12:
Closes #1891

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/LanguageC/PartialAST/FromLanC.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/LanguageC/PartialAST/FromLanC.hs
@@ -61,17 +61,12 @@ instance Apply (LanC.CDeclarationSpecifier a) PartialDecl where
       LanC.CTypeSpec x -> overM #typ $ apply x
       LanC.CTypeQual x -> overM #typ $ apply x
 
-      LanC.CStorageSpec x ->
-        case x of
-         -- We ignore the @typedef@ specifier: when reparsing typedefs it adds
-         -- no information, and elsewhere we don't expect it at all (we /could/
-         -- in principle check in such cases that it's not there, to catch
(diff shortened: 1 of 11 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/LanguageC/PartialAST/FromLanC.hs
```diff
@@ -61,17 +61,12 @@ instance Apply (LanC.CDeclarationSpecifier a) PartialDecl where
...
-          LanC.CTypedef _a -> return
-
-          -- TODO <https://github.com/well-typed/hs-bindgen/issues/1768>
-          -- We could deal other storage specifiers (auto, static, extern, ..).
-          other -> \_ -> unexpectedF other
+      -- We only care about the name and type of the declaration. The
+      -- remaining specifiers say how the symbol is linked or compiled,
+      -- not what its type is, so we leave the partial declaration
+      -- unchanged.
+      LanC.CStorageSpec _ -> return  -- @static@, @extern@, @typedef@, ...
```
bolt12: We should perhaps pattern match only on the things we know and should ignore. I.e. add only the cases we need to add (static, extern, typedef, only).

Make sure we have test cases.

edsko (approved):

PR #1945, opened 2026-05-05
`AnonId`: add spelling field for `Eq`/`Ord` disambiguation

### LINKED ISSUE #1860: Macro-expanded code breaks uniqueness of `AnonId`s
For C code like this which is defined with macros:

```c
#define NestedUnnamed(t1, n1, t2)\
  typedef t1 {\
    t2 {\
      int fieldX;\
    } fieldY;\
  } n1;\

NestedUnnamed(struct, UU, struct)
```

We would generate these Haskell bindings (I included only relevant bindings):

```hs
{-| __C declaration:__ @struct \@UU_fieldY@

    __defined at:__ @macros\/wrong_source_location\/uu.h 8:1@

    __exported by:__ @macros\/wrong_source_location\/uu.h@
-}
data UU_fieldY = UU_fieldY
  { uU_fieldY_fieldX :: RIP.CInt
    {- ^ __C declaration:__ @fieldX@

         __defined at:__ @macros\/wrong_source_location\/uu.h 8:1@

         __exported by:__ @macros\/wrong_source_location\/uu.h@
    -}
  }
  deriving stock (Eq, RIP.Generic, Show)

{-| __C declaration:__ @struct UU@

    __defined at:__ @macros\/wrong_source_location\/uu.h 8:1@

    __exported by:__ @macros\/wrong_source_location\/uu.h@
-}
data UU = UU
  { uU_fieldY :: UU_fieldY
    {- ^ __C declaration:__ @fieldY@

         __defined at:__ @macros\/wrong_source_location\/uu.h 8:1@

         __exported by:__ @macros\/wrong_source_location\/uu.h@
    -}
  }
  deriving stock (Eq, RIP.Generic, Show)
```

All the source code locations are reported as `8:1`, which is the start of the line with `NestedUnnamed(struct, UU, struct)` . This probably means that `libclang` can not reliably report source locations for macro-expanded code, so it defaults source locations to where the macro expansion occurs. 

This can be problematic with `AnonId`s. Anon IDs assign preliminary names to unnamed declarations based on source locations. Anon IDs are also assumed to uniquely identify unnamed declarations. `AssignAnonIds` pass will trace a message of the form "Conflicting use sites for AnonId" if they aren't. This was observed in the wild on commit https://github.com/well-typed/hs-bindgen/pull/1839/commits/ba3c004510eea071b3665a81932edc63d837437a, but only with LLVM15 for some peculiar reason ([CI output](https://github.com/well-typed/hs-bindgen/actions/runs/23445673777/job/68208682652#step:17:3118))

bolt12:
Closes #1860


edsko (changes requested):
Let's understand the problem a bit better first, I'm a bit confused about the situation :)

bolt12:
When a macro expands to multiple anonymous tag declarations, libclang reports the same expansion location for all of them.

Example:

```c
#define TwoAnons               \
    struct { int a; } x;       \
    struct { int b; } y;

TwoAnons // <- both structs "live" here, per expansion location
```

Both `struct`s share an expansion location (the line where `TwoAnons` was invoked). The old `AnonId` keyed only on that, so the two anonymous structs were treated as the same declaration.

libclang exposes a second location per cursor: the spelling location, which points back to where the token was originally written.

| decl in expansion     | expansion loc | spelling loc                |
| --------------------- | ------------- | --------------------------- |
| `struct { int a; } x` | call site     | macro body, line of `int a` |
| `struct { int b; } y` | call site     | macro body, line of `int b` |

Adding `spelling :: SingleLoc` to `AnonId` and including it in the derived `Eq`/`Ord` gives us distinct IDs in this case. `loc` (expansion) is unchanged and still drives traces/Haddock, so user-facing output is unaffected.

A thing to note is that `clang_getSpellingLocation` only returns a distinct value on **`llvm >= 19.1.0`**; older libclang reports spelling == expansion for macro-expanded cursors and the collision returns. The new `wrong_source_location` golden is gated on that version (see `Test/HsBindgen/Golden/Macros.hs`).

edsko (approved):
Thanks for the second example that is simple enough even I can understand it 😬

PR #1954, opened 2026-05-08
`boot`: discover the path to the `clang` executable

jorisdral:
Finding the `clang` executable is a prerequisite for #1892. Since the boot phase already performs the search as part of the  `BuiltinIncDir` module, we can reuse that bit of code and pass on the discovered path to the frontend.

### DIFF
--- a/hs-bindgen/hs-bindgen.cabal
+++ b/hs-bindgen/hs-bindgen.cabal
@@ -140,9 +140,9 @@ library internal
     HsBindgen.Boot
     HsBindgen.Cache
     HsBindgen.Clang
-    HsBindgen.Clang.BuiltinIncDir
     HsBindgen.Clang.CompareVersions
     HsBindgen.Clang.CStandard
+    HsBindgen.Clang.Discover
     HsBindgen.Clang.ExtraClangArgs
     HsBindgen.Clang.Macos
     HsBindgen.Clang.Sizeof
--- a/hs-bindgen/src-internal/HsBindgen/Boot.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Boot.hs
@@ -61,6 +61,9 @@ runBoot tracer config uncheckedHashIncludeArgs = do
     getCStandard <- cache "cStandard" $ withTrace BootStatusCStandard $
       (.cStandard) <$> getClangArtefacts'
 
+    getClangExe <- cache "clangExe" $ withTrace BootStatusClangExe $
+      (.clangExe) <$> getClangArtefacts'
+
     getClangArgs <- cache "clangArgs" $ withTrace BootStatusClangArgs $
       (.clangArgs) <$> getClangArtefacts'
 
@@ -87,6 +90,7 @@ runBoot tracer config uncheckedHashIncludeArgs = do
     pure BootArtefact {
           baseModule              = config.boot.baseModule
         , cStandard               = getCStandard
+        , clangExe                = getClangExe
         , clangArgs               = getClangArgs
         , hashIncludeArgs         = getHashIncludeArgs
         , externalBindingSpecs    = getExternalBindingSpecs
@@ -111,6 +115,7 @@ runBoot tracer config uncheckedHashIncludeArgs = do
 
 data ClangArtefacts = ClangArtefacts {
     cStandard :: ClangCStandard
+  , clangExe  :: Maybe ClangExe
   , clangArgs :: ClangArgs
   }
 
@@ -122,15 +127,16 @@ getClangArtefacts ::
 getClangArtefacts tracer config0 = do
     compareClangVersions (contramap BootCompareClangVersions tracer)
     extraClangArgs <- getExtraClangArgs tracerExtraClangArgs
-    mBuiltinIncDir <- getBuiltinIncDir tracerDiscover config0.builtinIncDir
+    paths <- getPaths tracerDiscover config0.builtinIncDir
     let config =
             applyExtraClangArgs extraClangArgs
-          . applyBuiltinIncDir  mBuiltinIncDir
+          . applyPaths paths
           $ config0
         clangArgs' = ClangArgs.clangArgsConfigToClangArgs config
     cStandard' <- getClangCStandard' tracer clangArgs'
     return ClangArtefacts{
         cStandard = cStandard'
+      , clangExe  = paths.clangExe
       , clangArgs = clangArgs'
       }
   where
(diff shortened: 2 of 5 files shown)


jorisdral:
@TravisCardwell could you maybe take a look if the general approach here would cause any problems? I've not polished the changes at all, it is just a proof of concept. If we want to flesh out this concept then maybe we should generalise the `BuiltinIncDir` module into a `Paths` module (or some other name), and rename types like `BuiltinIncDirMsg` accordingly?

TravisCardwell:
I do not know of any problems with the approach.  Yeah, and we should probably rename the module as well.

The builtin include directory resolution using Clang is one of the dirty corners of the codebase.  It is unfortunate that `libclang` is insufficient and we need to use Clang directly, and the way that we find Clang is unsatisfactory.  My current branches update the manual, adding a `libclang-bindings` manual.  I have updated it to make it clear that `hs-bindgen` runs the Clang executable, and I will revise to emphasize that even more now that we will use Clang more.

Last week, @edsko and I discussed the configure script, as I recently rewrote it (and will merge soon).  The `LLVM_PATH` and `LLVM_CONFIG` environment variables have been used since before I joined the project.  I think that `LLVM_PATH` was introduced (just) because `llvm-config.exe` is not distributed with some Windows installations of LLVM/Clang, *but* those target MSVC and we now know that we should not use them anyway.  Note that action [KyleMayes/install-llvm-action](https://github.com/KyleMayes/install-llvm-action) sets `LLVM_PATH`.  All installations that we currently test with have `llvm-config`.  Perhaps we should remove `LLVM_PATH` support and always use `llvm-config`...  As for `LLVM_CONFIG`, there are no clues about why it was implemented.  Perhaps it is just a convenience for people who do not want to update the `PATH`.  Now that we use the Clang executable, however, I think it should be removed...

Resolving the Clang executable path uses `LLVM_PATH` and `LLVM_CONFIG` in order to be compatible with our configure script options.  The (updated) manual makes it clear that these environment variables should remain set when using `hs-bindgen`, not just used for configuration.

Another note is that we have previously considered using static linking to create a standalone `hs-bindgen-cli` executable.  Static linking was brought up as a possible solution to our linking issues, but it is not really a solution because it would not help with TH mode.  Use of the Clang executable makes static linking even less helpful.  To ensure consistency, developers would need to install the exact version of LLVM/Clang that `hs-bindgen-cli` was statically linked with anyway!

Always using `llvm-config` and only using `PATH` to find it would simplify things, making it less messy.  *If* we decide to simplify things, it is something that I would like to do, as I have all of this cached in.

jorisdral:
> I do not know of any problems with the approach. Yeah, and we should probably rename the module as well.

Alright! Then I'll update the PR a bit more and open it up for review.

> ...
> 
> Always using `llvm-config` and only using `PATH` to find it would simplify things, making it less messy. _If_ we decide to simplify things, it is something that I would like to do, as I have all of this cached in.

Thank you for the in-depth info! If we make changes to how we discover the executable, then it looks like we'd only have to change the `findClangExe`, right?

jorisdral:
@TravisCardwell I updated the PR, let me know what you think

TravisCardwell:
> As for `LLVM_CONFIG`, there are no clues about why it was implemented. Perhaps it is just a convenience for people who do not want to update the `PATH`. Now that we use the Clang executable, however, I think it should be removed...

Coincidentally, @bolt12 used `LLVM_CONFIG` in getting cross-compilation to work in TH mode (#1958).  The implementation cross-compiles `libclang-bindings`, running the configure script on the target, and `LLVM_CONFIG` is used to specify the target LLVM/Clang installation (instead of the host).  Perhaps this would work by changing `PATH` instead of `LLVM_CONFIG`.

In that branch, the Clang executable is just used to determine the builtin include directories, and that functionality is explicitly disabled using `BINDGEN_BUILTIN_INCLUDE_DIR=disable`.  Using the Clang executable for other things (#1892) means that the target Clang executable will need to be used.  The logic for finding the Clang executable takes `LLVM_CONFIG` into account, so maybe it would work as-is.

TravisCardwell:
> Thank you for the in-depth info! If we make changes to how we discover the executable, then it looks like we'd only have to change the `findClangExe`, right?

Yes, that is all we have to change in the `hs-bindgen` code.

The task that I would prefer to do myself is changing the `libclang-bindings` configure script and documentation.

TravisCardwell (commented):
Looks good.  The only thing that I am wondering about regards if the Clang executable is required.  In `Discover`, a `Maybe` type is used and `DiscoverClangNotFound` is still `Debug`.  In the `BootArtefact`, it is still a `Maybe` type.  Will `hs-bindgen` still work even if the Clang executable is not found?  If not, where should we emit the error?

jorisdral:
> ...
> Coincidentally, @bolt12 used `LLVM_CONFIG` in getting cross-compilation to work in TH mode (#1958). 
> ...

Ah, interesting, maybe it will work out of the box indeed because `findClangExe` takes `LLVM_CONFIG` into account

> The task that I would prefer to do myself is changing the `libclang-bindings` configure script and documentation.

That's okay! I won't touch any of that

> Looks good. The only thing that I am wondering about regards if the Clang executable is required. In `Discover`, a `Maybe` type is used and `DiscoverClangNotFound` is still `Debug`. In the `BootArtefact`, it is still a `Maybe` type. Will `hs-bindgen` still work even if the Clang executable is not found? If not, where should we emit the error?

In the new pass I'm writing in #1892, I was planning to emit a warning when the `clang` executable can not be found. If the executable is found, we use it to expand macro invocations that we can't parse as a type (like `#define EXPORT __attribute__((visibility("hidden")))`. If we can't find the executable, then I was planning to have fallback behaviour that simply does not expand macro invocations at all. Is it okay if I update trace message levels as part of #1892?

TravisCardwell:
> In the new pass I'm writing in #1892, I was planning to emit a warning when the `clang` executable can not be found. If the executable is found, we use it to expand macro invocations that we can't parse as a type (like `#define EXPORT __attribute__((visibility("hidden")))`. If we can't find the executable, then I was planning to have fallback behaviour that simply does not expand macro invocations at all.

I see!  Sounds good.

> Is it okay if I update trace message levels as part of #1892?

Yes, of course!  No problem!

bolt12:
As I said in the other PR, I will experiment with setting the PATH instead and document it, env variables are tricky with cross compilation via iserv-on-QEMU because we have to make sure they are inherited/propagated to the target architecture.

edsko (approved):

TravisCardwell:
@jorisdral Please go ahead and merge this without removal of `LLVM_CONFIG` support.

I implemented removal of `LLVM_CONFIG` in `libclang-bindings` in https://github.com/well-typed/libclang-bindings/pull/62, but we are not ready to update the commit in `hs-bindgen` yet.  Once ready, I will update in `hs-bindgen`, update the discovery logic, and update the documentation.

PR #1955, opened 2026-05-08
`ReparseMacroExpansions`: skip over attribute specifier sequences

jorisdral:
Previously, declarations carrying macro expansions and attribute specifier sequences failed to reparse and silently fell back to the un-reparsed type, dropping macro typedef names from generated bindings.


jorisdral:
Note: this is a prelude to #1892

PR #1958, opened 2026-05-11
TH Mode cross compilation

### LINKED ISSUE #1607: Experiment with and document the cross-compilation workflow in TH mode
placeholder

bolt12:
Closes #1607.

Here's a breakdown of my understanding of how to get TH mode to cross compile:

To cross compile TH mode we need everything that Preprocess mode needs and a bit more.

In TH mode, the splice runs on the target architecture (via iserv and QEMU), which means the full `hs-bindgen` package needs to be cross-built. That transitively pulls in a bunch of dependencies (e.g. `libclang-bindings`, `doxygen-parser`, `aeson`, `yaml`, and (down the chain) `zlib`). Preprocess mode only needs `hs-bindgen-runtime`, which is rather small (4 deps).

This complicates things because some of those dependencies require special attention when cross-compiling, namely `zlib` and `libclang-bindings`.

`zlib`'s `.hsc` files do `#include <zlib.h>`. The package's default cabal config is `pkgconfig-depends: zlib`, and in our cross setup `pkg-config` isn't target-aware, which means when the cross-CC tries to preprocess `Stream.hsc` it fails with `zlib.h: No such file or directory`. The fix I went with is to add the target zlib's include dir under `package *` in `cabal.project.local`. An equally good alternative is the `+bundled-c-zlib` flag, which swaps the system-library dep for the `zlib-clib` package (zlib's C source compiled inline).

`libclang-bindings` has a configure step that calls `llvm-config` to figure out the right paths for LLVM and clang, but when cross-compiling it picks up the host's `llvm-config` and gets host paths instead. The configure script declares `AC_ARG_VAR(LLVM_CONFIG, ...)` precisely for this case (autoconf's mechanism for user-overridable env vars), so we override `LLVM_CONFIG` to point at a stub script that returns target paths.

Regarding clang at runtime, when the splice runs inside iserv-on-QEMU, it loads the cross-built `hs-bindgen` object code which dlopens `libclang.so` to actually parse the C header. We make target `libclang.so` reachable by injecting `LD_LIBRARY_PATH` into the emulated process via QEMU's `-E` flag. The same target `libclang.so` is also needed at the final executable's link step (because `libclang-bindings` declares `extra-libraries: clang`), so one needs to wire it via `extra-lib-dirs:` entries in `cabal.project.local`.

The splice's own include paths can't come from cabal. libclang has its own include-path mechanism (the -I flags passed to `clang_parseTranslationUnit`), which cabal never wires up. So we have to tell libclang directly via the `Config` value passed to `withHsBindgen`: `extraIncludeDirs = [Pkg "../c-src"]`.

We also need to pass the right `default-extensions` cabal option in the Haskell project to match the splice's generated code.

Given all these quirks, I have decided to include a TH mode cross compilation example that gets built in CI, so that we can track this not only in docs but also in a working example!


TravisCardwell:
> `libclang-bindings` has a configure step that calls `llvm-config` to figure out the right paths for LLVM and clang, but when cross-compiling it picks up the host's `llvm-config` and gets host paths instead. The configure script declares `AC_ARG_VAR(LLVM_CONFIG, ...)` precisely for this case (autoconf's mechanism for user-overridable env vars), so we override `LLVM_CONFIG` to point at a stub script that returns target paths.

Would configuring `PATH` (so that the target LLVM/Clang installation executables are found) instead of configuring `LLVM_CONFIG` work?

Context:

Unfortunately, `hs-bindgen` needs to run the Clang executable for things that cannot be done via `libclang`.  Currently, it is just used to determine builtin include directories, and your example explicitly sets `BINDGEN_BUILTIN_INCLUDE_DIR=disable` so that this is not done.  We will soon use it for other things (#1892) as well, though.  The logic for finding the Clang executable takes `LLVM_CONFIG` into account, so maybe it would work as-is.

`LLVM_CONFIG` was not a bad design when we just interfaced with LLVM/Clang via `libclang`.  It is questionable now that we use the Clang executable, though.  The current implementation where `LLVM_PATH` and `LLVM_CONFIG` are used to search for the Clang executable works, but it is convoluted and is not normal behavior that users expect.  We can keep them if they are necessary, but we are considering removal of one or both of these to simplify things, making behavior less unexpected.

bolt12:
@TravisCardwell I think it could work, the reason I went with this was because llvm-config was used and I didn't delve too much on alternatives. I don't think this should block any of the efforts to improve the clang executable discovery because these are just guidelines to get TH to cross compile, it is expected that thingsmay  change a bit depending on the setup. For instance, much of what I did assumes a nix managed project.

TravisCardwell:
@edsko and I discussed this a bit, and we think it would be a good idea to include something like the following in the README for the example.  (First draft:)

> This example demonstrates how cross-compilation of a package that uses the `hs-bindgen` Template Haskell API can be done, but note that there are drawbacks to doing this.
>
> A library/executable that uses the `hs-bindgen` Template Haskell API must have `hs-bindgen` as a dependency (in addition to `hs-bindgen-runtime`).  The built library/executable is therefore linked to `libclang`, and the same major/minor version of LLVM/Clang must be installed on any system the library/executable is used.  One generally does not want compile-time dependencies to be dependencies of the built library/executable, but this is a cost of using the Template Haskell API.  The situation might be improved/resolved in the future with the introduction of explicit level imports, *if* system dependencies of compile time dependencies are not linked in the built libraries/executables.
>
> This cost is likely much more significant in situations when cross-compilation is required.  When cross-compiling, you likely want to use the CLI, so that `hs-bindgen` is not a dependency of your library/executable and `libclang` is not linked.

BTW, I suspect that this kind of setup is fairly fragile, and we probably do not want to spend much time on maintaining it.

examples/cross-compilation/generate-and-run.sh
```diff
@@ -171,75 +219,197 @@ HS_EOF
...
   ${AARCH64_GMP_LIB:+extra-lib-dirs: $AARCH64_GMP_LIB}
+  ${LIBCLANG_DIR_AARCH64:+extra-lib-dirs: $LIBCLANG_DIR_AARCH64}
+  ${ZLIB_DIR_AARCH64:+extra-lib-dirs: $ZLIB_DIR_AARCH64}
+  -- zlib's hsc files \`#include <zlib.h>\`. With the default
+  -- \`+pkg-config\` flag, cabal asks pkg-config for the include path,
+  -- but pkg-config is not target-aware in our cross setup, so the
+  -- preprocessing fails with \`zlib.h: No such file or directory\`.
+  -- Adding the target zlib's include dir here lets the cross-CC find
+  -- it without overriding cabal flags.
+  ${AARCH64_ZLIB_INCLUDE:+extra-include-dirs: $AARCH64_ZLIB_INCLUDE}
```
bolt12: Have different cabal.project.local files for Preprocess mode cross compilation and TH Mode cross compilation so that we have independent files for each one and we don't miss some unnecessary dependencies on the non-TH mode examples.

bolt12:
Why is there no path necessary for zlib.so?

examples/cross-compilation/hs-project/cross-compilation-example.cabal
bolt12, line 1: Here it would be better to have a separate .cabal file, maybe even a different project to keep the Preprocess mode and TH mode cross compilation examples really separated.

edsko:
@TravisCardwell Once @bolt12 has rebased and addressed my review comments , could you review just the manual changes part of this PR please? 

And perhaps you and I can briefly discuss your comments regarding paths above, as I don't fully understand it.

@bolt12 Could you ping @TravisCardwell once you feel this is ready?

edsko (approved):

bolt12:
> Why is there no path necessary for zlib.so?

from what I understood the compiler needs to be told where the headers live but the linker does a better job of finding the .so files, but we do add the zlib library paths to the cabal.project.local too!

examples/cross-compilation/generate-and-run.sh
```diff
@@ -171,75 +201,215 @@ HS_EOF
...
+LLVM_CONFIG_STUB="$HS_PROJECT_TH_DIR/llvm-config-aarch64-stub.sh"
 
-exe_path=$(find "$BUILDDIR" -type f -name "$EXE_NAME" -executable 2>/dev/null | head -1)
+write_if_changed "$LLVM_CONFIG_STUB" << STUB_EOF
+#!/usr/bin/env bash
+case "\$1" in
+    --version)    echo "19.1.7" ;;  # value unused beyond non-empty check
+    --libdir)     echo "$LIBCLANG_DIR_AARCH64" ;;
+    --includedir) echo "$LIBCLANG_INC_AARCH64" ;;
+    --prefix)     echo "${LIBCLANG_DIR_AARCH64%/lib}" ;;
```
edsko, line 376: @bolt12 I discussed this with @TravisCardwell . I understand his comment better now. The suggestion is to use `PATH` on the host machine instead of `LLVM_CONFIG`, if possible, so that when we go looking for `llvm-config`, we'll find the shim instead of the normal executable; the reason is that Travis feels we should get rid of the `LLVM_CONFIG` environment variable. Could you give that a go? Mind: if it turns out that this is difficult for some reason (or indeed impossible), then that would be interesting in its own right, as it would tell us that we should not get rid of `LLVM_CONFIG` after all.

TravisCardwell:
Regarding why somebody would want to cross-compile Haskell instead of just build under QEMU, I think that the primary reason is speed.  Search for `Slower is an understatement` in [this /r/haskell post](https://old.reddit.com/r/haskell/comments/k0qtrz/a_tale_of_template_haskell_and_cross_compilation/) for an anecdote about going from ~3-4 days to ~2 hours.

(Using `hs-bindgen-cli` to generate sources *before* cross-compiling is likely what people want to do.)

examples/cross-compilation/generate-and-run.sh
bolt12: It worked!

TravisCardwell: Thank you for checking!  I created issue #1988 for removing `LLVM_CONFIG` support.

PR #1960, opened 2026-05-12
Recover from `CXType_Unexposed` via `clang_getCanonicalType`

### LINKED ISSUE #1901: `c-expr-runtime` binary operation `-` tests fail with Clang 22
Environment:

```
$ git rev-parse --short HEAD
9e3e45be6
$ cabal run hs-bindgen-cli -- --version
hs-bindgen 0.1.0
binding specification 1.0
clang version 22.1.2
```

Test:

```
$ cabal test c-expr-runtime
...
Binary operators
...
-
   FAILED:
   (ty_1 *,ty_1 *)
     - computed type: ptrdiff_t
     -  Clang's type: <n/a>
```

The tests consistently pass with Clang 21.

bolt12:
This closes #1901

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type.hs
@@ -74,7 +74,7 @@ cxtype ty = do
       CXType_Record          -> fromDecl
       CXType_Typedef         -> fromDecl
       CXType_Void            -> const (pure C.TypeVoid)
-      CXType_Unexposed       -> unexposed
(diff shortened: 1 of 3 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Type.hs
bolt12, line 1: Make sure we really don't need this change.

edsko (approved):
Feel free to merge if this only touches `c-expr`. If it turns out that is not enough, then perhaps we don't understand the clang-22 "unexposed" problem fully yet after all.

Issue #1732, opened 2026-02-19
Static executable example


edsko:
It would probably be worthwhile to create a (Linux) static executable example.  After using `hs-bindgen` to generate bindings for a C library, we build the example so that the library is statically linked.

Note that we can check that the executable is statically linked as follows:

```
$ ldd path/to/executable
        not a dynamic executable
```

edsko:
Marking this as low priority until we have a specific use case.

bolt12:
Create a CI job (for the comprehensive CI) that runs on an Alpine container and build a static executable of one of the hs-bindgen examples and then make sure it has no dynamic linked libraries.

PR #1961, opened 2026-05-12
Add Alpine CI workflow that builds a static executable

### LINKED ISSUE #1817: Test with Alpine Linux
Alpine containers are often used to build static executables for Haskell projects, and it is centered around `musl`. It might therefore be a good test case, to see if we're making any assumptions in `hs-bindgen` that we shouldn't be, and to have a case study where we use the generated bindings to make a static executable.

bolt12:
Closes #1817 and #1732


examples/blocktest/c/Makefile
```diff
@@ -1,5 +1,9 @@
+# `-lBlocksRuntime` must come AFTER iterator.o; some Linux linkers
+# (Alpine/musl with --as-needed default) drop libraries that come before any
+# object referencing their symbols, leading to undefined
+# `_Block_copy`/`_NSConcreteStackBlock` etc. at link time.
 libsimple_iterators.so: iterator.o
-	clang -Wall -shared -lBlocksRuntime -o libsimple_iterators.so iterator.o
+	clang -Wall -shared -o libsimple_iterators.so iterator.o -lBlocksRuntime
```
bolt12, line 6: @TravisCardwell could you double check this here, I found that when building blocktest on Alpine I run into this issue, but don't fully understand it.

.github/workflows/comprehensive-ci.yml
bolt12, line 1: Can we trigger or even configure this CI job with a fixed set of runners available?

edsko (approved):

.github/workflows/comprehensive-ci.yml
bolt12: We can add a job-level `concurrency: { group: comp-ci-${{ matrix.runner }}, cancel-in-progress: false }` to every job in `comprehensive-ci.yml` (substituting a literal OS for single-OS jobs, e.g. `comp-ci-windows-latest`), so each runner-OS gets its own per-repo queue, meaning at most one ubuntu job, one macos job, and one windows job from this workflow runs at a time while the rest wait, leaving the rest of the GitHub-hosted runner pool free for regular PR CI.

bolt12: cc @jorisdral @edsko

examples/blocktest/c/Makefile
TravisCardwell: The answer to [this SO question](https://stackoverflow.com/questions/28786873/ordering-of-object-files-and-libraries-in-static-linking) provides a concise explanation.

PR #1979, opened 2026-05-14
`ReparseMacroExpansions`: add underlying types to macro references

### LINKED ISSUE #1953: Avoid conversion to `C.Type` in `TypecheckMacros`
This is a prerequisite for #942.

At the moment, we typecheck macrosm, and immediately convert type-like macro expressions to their C type. Then, we pull the C type through the pipeline, finally translating to an Hs type.

Instead, we should only pull along the checked type-like macro expression, directly converting it to an Hs type in the backend. This conversion must be done by the external library providing the macro interface (`c-expr-dsl` by default).

See the TODO comments in the code.

jorisdral:
The goal of #1953 is to avoid converting parsed macros to C types. As a result, macro-defined types (or just macro types) would only have to be translated to Haskell. The C type language is not very powerful, but the Haskell type language is. Without the intermediate conversion to C types we could parse macros as polymorphic types for example:

```c
// A can not be written as a C typedef!
#define A(x) x * 
```
```hs
-- but A can be translated to Haskell!
newtype A x = A (Ptr x) 
```

Removing the conversion to C types is somewhat tricky because we have an invariant on C types that we can always look through "sugar" like references to macro-defined types (or typedefs) to inspect the *underlying* C type of that reference. If we don't translate macros to C types directly, then we also lose the ability to easily get the underlying C type of a reference to a macro type.

For monomorphic macro types, the underlying type of a reference to a macro type is always going to be the same. However, with polymorphic macro types (like `A`), the underlying type depends on the use site. For example, the (underlying) type of global variable `x` is an `int`, whereas the underlying type of global variable `y` is `char`, even though both global variables reference the same macro. So at the point where we parse a macro, we may not know the underlying type anyway. Only at use sites the underlying type becomes clear.

```c
A(int) x;
A(char) y;
```

Underlying types are important for `hs-bindgen`. Underlying types allow us to desugar C types, which we inspect to determine requirements for the Haskell bindings. For example, we look at underlying types to decide whether a function argument/result has to be passed by a pointer or not. Such is the case if the underlying type is a struct. So it is important to still be able desugar types after we resolve #1953, or we might produce incorrect Haskell bindings.

The `ReparseMacroExpansions` pass takes care of reparsing raw `libclang` tokens. The parse results (primarily types) are used to replace parts of the C AST. These new types may now contain references to macro types for which we do not know the underlying type. One important observation is this: we know what the C AST was before and after replacing parts of the C AST. If we replace a type with a reference  to a macro-defined type, then the underlying type *must* be the the type we replaced!

This PR makes use of this observation. We add a `fixup` function that compares the C AST before and after reparsing, and enhances references to macro types in the after-AST with underlying types taken from the before-AST. 

Note: this `fixup` function is generally useful as a sanity check, even if we don't go with this approach of adding underlying types to macro references. For example, it checks that we don't do anything silly like replacing an `int` with a reference to a struct. The check even found a bug where a macro expanion is parsed as an argument name:

```c
#define A const
void foo (int A);
```

The reparser thinks `A` is an argument name and discards the `const` qualifier. The `fixup` function catches this case and prevents us from adopting the wrong type

EDIT: note that `fixup` is now renamed to `align`

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions.hs
@@ -26,6 +26,7 @@ import HsBindgen.Frontend.Naming
 import HsBindgen.Frontend.Pass
 import HsBindgen.Frontend.Pass.Parse.IsPass
 import HsBindgen.Frontend.Pass.Parse.Msg
+import HsBindgen.Frontend.Pass.ReparseMacroExpansions.Align
 import HsBindgen.Frontend.Pass.ReparseMacroExpansions.IsPass
 import HsBindgen.Frontend.Pass.TypecheckMacros.IsPass
 import HsBindgen.Imports
@@ -330,34 +331,31 @@ processFunction ::
      C.DeclInfo ReparseMacroExpansions
   -> C.Function TypecheckMacros
   -> M (C.Decl ReparseMacroExpansions)
-processFunction info function =
-    reparseWith info.id LanC.reparseFunDecl function.ann withoutReparse withReparse
+processFunction info function = do
+    function' <- reparseWith info.id LanC.reparseFunDecl function.ann withoutReparse withReparse
+    pure C.Decl {
+        info = info
+      , ann = NoAnn
+      , kind = C.DeclFunction function'
+      }
   where
-    withoutReparse :: C.Decl ReparseMacroExpansions
-    withoutReparse = C.Decl{
-          info = info
-        , ann  = NoAnn
-        , kind = C.DeclFunction C.Function{
-              args  = map coercePass function.args
-            , res   = coercePass function.res
-            , attrs = function.attrs
-            , ann   = NoAnn
-            }
+    withoutReparse :: C.Function ReparseMacroExpansions
+    withoutReparse = C.Function{
+          args  = map coercePass function.args
+        , res   = coercePass function.res
+        , attrs = function.attrs
+        , ann   = NoAnn
         }
 
     withReparse ::
          (([(Maybe Text, CType)] , CType) , Text)
-      -> M (C.Decl ReparseMacroExpansions)
-    withReparse ((tys, ty), _name) = pure C.Decl{
-           info = info
-         , ann  = NoAnn
-         , kind = C.DeclFunction C.Function{
-               args  = map (uncurry mkFunctionArg) tys
-             , res   = ty
-             , attrs = function.attrs
-             , ann   = NoAnn
-             }
-         }
+      -> M (C.Function ReparseMacroExpansions)
+    withReparse ((tys, ty), _name) = pure C.Function{
+          args  = map (uncurry mkFunctionArg) tys
+        , res   = ty
+        , attrs = function.attrs
+        , ann   = NoAnn
+        }
 
     mkFunctionArg :: Maybe Text -> CType -> C.FunctionArg ReparseMacroExpansions
     mkFunctionArg mname typ = C.FunctionArg{
@@ -370,27 +368,24 @@ processGlobal ::
      C.DeclInfo ReparseMacroExpansions
   -> C.Global TypecheckMacros
   -> M (C.Decl ReparseMacroExpansions)
-processGlobal info global =
-    reparseWith info.id LanC.reparseGlobal global.ann withoutReparse withReparse
+processGlobal info global = do
+    global' <- reparseWith info.id LanC.reparseGlobal global.ann withoutReparse withReparse
+    pure C.Decl {
+        info = info
+      , ann = NoAnn
+      , kind = C.DeclGlobal global'
+      }
   where
-    withoutReparse :: C.Decl ReparseMacroExpansions
-    withoutReparse = C.Decl{
-          info = info
-        , ann  = NoAnn
-        , kind = C.DeclGlobal C.Global{
-              typ = coercePass global.typ
-            , ann = NoAnn
-            }
+    withoutReparse :: C.Global ReparseMacroExpansions
+    withoutReparse = C.Global{
+          typ = coercePass global.typ
+        , ann = NoAnn
         }
 
-    withReparse :: CType -> M (C.Decl ReparseMacroExpansions)
-    withReparse ty = pure C.Decl{
-          info = info
-        , ann  = NoAnn
-        , kind = C.DeclGlobal C.Global{
-              typ = ty
-            , ann = NoAnn
-            }
+    withReparse :: CType -> M (C.Global ReparseMacroExpansions)
+    withReparse ty = pure C.Global{
+          typ = ty
+        , ann = NoAnn
         }
 
 {-------------------------------------------------------------------------------
@@ -455,10 +450,14 @@ runM cStd knownTypes knownMacroTypes (WrapM ma) = runReader (runStateT ma s) e
 
 -- | Run reparser if needed; use fallback on failure or if not needed.
 --
--- On failure, records @(declId, ParseMacroErrorReparse e)@ in
+-- On reparsing failure, records @(declId, ParseMacroErrorReparse e)@ in
+-- 'ReparseState.reparseWarnings' and uses the fallback value.
+--
+-- On alignment failure, records @(declId, ParseMacroErrorReparseAlign e)@ in
 -- 'ReparseState.reparseWarnings' and uses the fallback value.
 reparseWith ::
-     DeclId
+     Align r
+  => DeclId
   -> LanC.Parser a
   -> ReparseInfo
   -> r
@@ -481,8 +480,15 @@ reparseWith declId parser reparseInfo fallback onSuccess = case reparseInfo of
           reparseEnv = LanC.ReparseEnv env.knownTypes usedKnownMacros
       case parser reparseEnv tokens of
         Right a -> do
-          modify $ #reparseSuccesses %~ Set.insert declId
-          onSuccess a
+          r <- onSuccess a
+          case alignEither fallback r of
+            Left msgs -> do
+              forM_ msgs $ \msg ->
+                modify $ #reparseWarnings %~ ((declId, msg) :)
+              pure fallback
+            Right r' -> do
+              modify $ #reparseSuccesses %~ Set.insert declId
+              pure r'
         Left  e -> do
           modify $ #reparseWarnings %~ ((declId, ParseMacroErrorReparse e) :)
           pure fallback
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Align.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Align.hs
@@ -1,0 +1,308 @@
+-- | Align the C ASTs before and after reparsing
+--
+-- Intended to be imported qualified:
+--
+-- > import HsBindgen.Frontend.Pass.ReparseMacroExpansions.Align.Error qualified as Align
+module HsBindgen.Frontend.Pass.ReparseMacroExpansions.Align (
+    AlignResult (..)
+  , Align (..)
+  , alignEither
+  ) where
+
+import Control.Monad (zipWithM)
+
+import HsBindgen.Frontend.AST.Decl qualified as C
+import HsBindgen.Frontend.AST.Type qualified as C
+import HsBindgen.Frontend.Pass (IsPass (Ann))
+import HsBindgen.Frontend.Pass.Parse.Msg (DelayedParseMsg (ParseMacroErrorReparseAlign))
+import HsBindgen.Frontend.Pass.ReparseMacroExpansions.Align.Error qualified as Align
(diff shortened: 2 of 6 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Fixup.hs
```diff
@@ -0,0 +1,264 @@
...
+      (C.TypeQual q1 t1', C.TypeQual q2 t2') ->
+        C.TypeQual <$> fixup q1 q2 <*> fixup t1' t2'
+
+      -- edge case: some builtins are reparsed as typedefs
+      (t1', C.TypeTypedef m2) -> C.TypeTypedef <$> (C.Ref <$> pure m2.name <*> fixup t1' m2.underlying)
+
+      -- fail
+      (_, _) -> failure
+    where
+      _coveredAllCases :: C.Type Post -> ()
```
dschrempf: This is just here to let the compiler convince developers to add new branches to the case splits, correct?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Fixup.hs
```diff
@@ -0,0 +1,264 @@
...
+
+instance Fixup (C.Type Pre) (C.Type Post) where
+  fixup t1 t2 = case (t1, t2) of
+      -- impossible case: macro references don't exist in our C AST before
+      -- reparsing
+      (C.TypeMacro{}, C.TypeMacro{}) -> failure
+
+      -- interesting case: an arbitrary type is replaced by a macro reference
+      (_, C.TypeMacro m2) -> do
+        underlying' <- fixup t1 m2.underlying
```
dschrempf: At the moment, this still assumes that we know the underlying type for macros. We could, however, not set this type, and only set it here in the fixup stage. This would avoid the additional conversion to C type for macros entirely.

(We would lose the conversion to C type test you mention in the PR description, but I think this is not a big deal; we would just test a function we don't really need).

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Fixup.hs
```diff
@@ -0,0 +1,264 @@
...
+      _coveredAllCases :: C.PrimType -> ()
+      _coveredAllCases = \case
+        C.PrimChar{}     -> ()
+        C.PrimIntegral{} -> ()
+        C.PrimFloating{} -> ()
+        C.PrimBool{}     -> ()
+
+instance Fixup C.PrimSignChar C.PrimSignChar where
+  fixup csign1 csign2 = case (csign1, csign2) of
+      -- interesting case: the reparser does not know the sign of a character
```
dschrempf: I see that the `Fixup` also fixes other problems?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions.hs
```diff
@@ -480,8 +476,14 @@ reparseWith declId parser reparseInfo fallback onSuccess = case reparseInfo of
           reparseEnv = LanC.ReparseEnv env.knownTypes usedKnownMacros
       case parser reparseEnv tokens of
         Right a -> do
-          modify $ #reparseSuccesses %~ Set.insert declId
-          onSuccess a
+          r <- onSuccess a
+          case fixupEither fallback r of
```
dschrempf: Maybe we should rename this from "fallback" to something like "originalDeclMacrosExpanded" (bad name, but I guess it conveys my point, maybe we can find a better one?).

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Fixup.hs
```diff
@@ -0,0 +1,264 @@
...
+import HsBindgen.Frontend.Pass.Parse.Msg (DelayedParseMsg (ParseMacroErrorReparseFixup))
+import HsBindgen.Frontend.Pass.ReparseMacroExpansions.IsPass
+import HsBindgen.Language.C qualified as C
+
+{-------------------------------------------------------------------------------
+  Shorthands
+-------------------------------------------------------------------------------}
+
+type Pre = ReparseMacroExpansions
+type Post = ReparseMacroExpansions
```
dschrempf: I am also always defining these shortcuts. I was thinking we could do this more systematically at some point to allow shuffling around passes in an easier way.

I called it `PreviousPass` which is too verbose, and probably too specific. `Pre`, `Post` is nice.

dschrempf (approved):
If I am not mistaken, this allows us to _avoid conversion_ of macros to C types in `TypecheckMacros`, but only in `ReparseMacros`, that would be _great_.

I also really enjoyed your PR description, maybe we can move some of this into the developer manual, or the code? (Or the actual test cases)?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Fixup.hs
jorisdral: Indeed, it's just there so that the compiler lets us know to update the cases when we add or modify constructors of `C.Type` in the future

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Fixup.hs
jorisdral: Yeah if we decide to adopt this "fixup" function, then the idea would be to only add the underlying type here, and not do that before

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Fixup.hs
jorisdral: Right, there a few small things the the "fixup" function does in addition to adding underlying types to macro references. These cases caused some golden tests to fail, so I had to update the "fixup" function

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Align.hs
```diff
@@ -0,0 +1,312 @@
...
+-- allowed to differ from \( t_{pre} \), then this algorithm might not work
+-- anymore.
+--
+-- NOTE: the description of PR #1979 describes historically why we implemented
+-- the 'align' function, which mainly had to do with supporting underlying types
+-- for macro references.
+--
+-- <https://github.com/well-typed/hs-bindgen/pull/1979>
+--
+class Align a where
```
jorisdral: @dschrempf @edsko could you give this a read and let me know if you think it describes the algorithm well enough? 

Rendered version:
image

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Align/Error.hs
```diff
@@ -0,0 +1,107 @@
...
+        , "Sub-tree after reparsing:" PP.<+> PP.text rhs
+        ]
+    SubTreesNotEqual lhs rhs -> PP.vsep [
+          PP.hsep [
+              "Expected two sub-trees in the C ASTs before and after reparsing to be"
+            , "equal, but they were not:"
+            ]
+        , "Sub-tree before reparsing:" PP.<+> PP.text lhs
+        , "Sub-tree after reparsing:" PP.<+> PP.text rhs
+        ]
```
jorisdral: @dschrempf @edsko I'm unsure about how to best report errors in these cases. Is it useful to report the subtrees, or should I leave them out?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions.hs
jorisdral: for now I'm sticking with `fallback`, but we can change it in the future of course

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Align/Error.hs
jorisdral: Ah, maybe I should emit debug-level messages for the subtrees?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Align.hs
```diff
@@ -0,0 +1,310 @@
...
+checkEq :: (Show a, Eq a) => a -> a -> AlignResult a
+checkEq x y
+  | x == y = success y
+  | otherwise = failure $ Align.errNotEqual (show x) (show y)
+
+{-------------------------------------------------------------------------------
+  Class
+-------------------------------------------------------------------------------}
+
+-- | Align the C AST from before reparsing with the C AST from after reparsing
```
dschrempf:
```suggestion
-- | Align the C AST before reparse with the C AST after reparse
```

(Suggestion)

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Align.hs
```diff
@@ -0,0 +1,310 @@
...
+--
+-- The reparser parses raw @libclang@ tokens and replaces parts of the C AST
+-- with the parse results. This replacement can not be performed arbitrarily,
+-- because that can lead to incorrect Haskell bindings. For example, if we have
```
dschrempf:
```suggestion
-- with the parse results. To ensure correct Haskell bindings, this replacement must adhere to certain rules. For example, if we have
```

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Align.hs
```diff
@@ -0,0 +1,310 @@
...
+-- a global variable of type @int@ before reparsing, then it is fine if the
+-- reparser replaces that @int@ with some syntactical sugar (e.g., a typedef)
+-- representing an @int@. However, it is not okay to replace that @int@ by a
+-- reference to a struct, for example.
```
dschrempf: Maybe also put `typedef` and `struct` in verbatim?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Align.hs
```diff
@@ -0,0 +1,310 @@
...
+--
+-- The 'align' function operates recursively on sub-trees of C ASTs. It takes a
+-- tree from before reparsing, \( t_{pre} \), and a tree from after reparsing,
+-- \( t_{post} \), and produces an aligned tree, \( t_{aligned} \). Firstly, the
+-- 'align' function checks that \( t_{pre} \) has undergone a small set of
+-- allowed modifications with respect to \( t_{post} \). Moreover, 'align'
+-- combines information from both trees to produce a new \( t_{aligned} \). The
+-- reason for this is that the reparser is slightly "dumber" than the @libclang@
```
dschrempf: Maybe switch the comparison?

"@libclang@ has more information available than @language-c@", or 
"@libclang@ can leverege the information content of the complete translation unit, while @language-c@ is constrained to reparsing a single declaration"?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Align.hs
```diff
@@ -0,0 +1,310 @@
...
+-- parser (see @Parse@ pass), and so \( t_{post} \) has to be modified a bit
+-- using \( t_{pre} \) to insert missing information. The result is a new
+-- sub-tree \( t_{aligned} \).
+--
+-- For the most part, the process of 'align' is simple:
+--
+-- * It checks that that the root nodes of \( t_{pre} \) and \( t_{post} \) are
```
dschrempf:
```suggestion
-- * It checks that the root nodes of \( t_{pre} \) and \( t_{post} \) are
```

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ReparseMacroExpansions/Align/Error.hs
dschrempf: I can think of three scenarios:
- Sub-tree mismatches are expected, and frequent; they do not change the generated bindings: Make them `Debug` traces.
- Sub-tree mismatches may happen; they are less frequent; they do affect bindings: Make them nicely formatted `Warning`s
- Sub-tree mismatches are unexpected; they indicate a bug in `hs-bindgen`: Make them `Bug` traces

dschrempf (approved):
Thanks @jorisdral! I think the explanation of `Align` can be shortened a bit, but the generous comment also adds a lot of clarity, thank you. I have left some suggestions.

PR #1980, opened 2026-05-14
Add `enableExtraClangArgs` to opt out of `BINDGEN_EXTRA_CLANG_ARGS`

### LINKED ISSUE #875: Golden tests C include path sensitivity to `BINDGEN_EXTRA_CLANG_ARGS`
The golden tests must run using the `musl-include/x86_64` standard headers.  If the environment has `BINDGEN_EXTRA_CLANG_ARGS` set, however, those settings can make different headers be used and many tests fail.

bolt12:
Closes #875

### DIFF
--- a/hs-bindgen/CHANGELOG.md
+++ b/hs-bindgen/CHANGELOG.md
@@ -6,6 +6,12 @@
 
 ### New features
 
+* New `ClangArgsConfig.enableExtraClangArgs` field (default `True`) and
+  matching `--no-extra-clang-args` CLI flag, allowing library users and the
+  CLI to opt out of reading the `BINDGEN_EXTRA_CLANG_ARGS` environment
+  variable. The test suite now opts out so that `BINDGEN_EXTRA_CLANG_ARGS`
+  in the developer's shell no longer interferes with golden tests
+  ([#875](https://github.com/well-typed/hs-bindgen/issues/875)).
 * When using the Template Haskell backend, external types referenced via
   external binding specifications that are not in scope now produce a helpful
   compile error suggesting the missing import.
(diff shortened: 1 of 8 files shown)


dschrempf:
I am still a bit unsure if this is the correct way of solving this issue. A better option to me seems

env -u BINDGEN_EXTRA_CLANG_ARGS hs-bindgen-cli ...

If we need to unset this variable for parts of the test suite, we can still do so in the test suite, programmatically.

Implementing this switch feels like implementing a workaround for a workaround.

The implementation is well done, though. I think the changelog entry is too verbose.

TravisCardwell:
@dschrempf makes a good point.  `BINDGEN_EXTRA_CLANG_ARGS` is mostly a way to configure Clang defaults when in an environment where defaults are insufficient (such as Nix).  If, in such an environment, `BINDGEN_EXTRA_CLANG_ARGS` options are required for Clang to work at all, then disabling it is likely problematic.  Do our tests pass on Nix when opting out of `BINDGEN_EXTRA_CLANG_ARGS`?

Issue #875 was created before we changed the behavior so that `BINDGEN_EXTRA_CLANG_ARGS` options are lower in priority.  From the manual, option priority is now:

1. `--clang-option-before` options
2. Clang options managed by `hs-bindgen` (include directories, etc.)
3. `--clang-option` options
4. `BINDGEN_EXTRA_CLANG_ARGS` options (see below)
5. `--clang-option-after` options

Configuration of Musl include directories for tests (should) now take priority over anything configured in `BINDGEN_EXTRA_CLANG_ARGS`.  I just did a test on `main`, and tests pass even with `BINDGEN_EXTRA_CLANG_ARGS` configuring system directories (at least on my system), as expected.

I agree with @dschrempf -- we probably do not need this option/configuration anymore.

bolt12:
Closing then! Thanks for reviewing! 😄

edsko:
Sorry about that @bolt12 !

PR #1981, opened 2026-05-15
Update comment in hs-bindgen-dev.nix with the new nixos wiki link

znmz:
I'm working on replacing all references to `nixos.wiki` with `wiki.nixos.org` across all nix-related repos. This increases visibility of the official wiki so that someday there will no longer be any confusion.


dschrempf:
Would you like to be added to the list of "contributors" (see the readme). If so, how should we refer to you?

ghost:
> Would you like to be added to the list of "contributors" (see the readme). If so, how should we refer to you?

Nah, no need to do that. Have a good day/evening :)

PR #1983, opened 2026-05-18
Remove `sameMacro`

jorisdral:
`sameMacro` is used in `DeclIndex` to decide whether two macro definitions are the same, so that we can select one definition and throw out the other without generatic incorrect code.

`sameMacro` checks that two macros are *synatically* equal, and if so, it considers the macros to be the same. However, syntactically equal macro definitions are not necessarily *semantically* equal.

Let's consider an example:

```c
#define A int
#define T A
void foo (T x);
#undef A
#define A char
#define T A
void bar (T x);
```

The definitions for `A` are considered unusable because they are not syntactically equal. The definitions for `T` are considered usable because the definitions are syntactically equal. However, the defininitions for `T` are not *semantically* equal. In the example above, the function argument of function `foo` is of type `int`, while the function argument of function `bar` is of type `char`. Luckily, this is not causing any bugs in `hs-bindgen` because the `Select` pass will not select declarations for `A` because they are unusable, and the declarations for `T` are thrown out because their dependency `A` is not selected. Still, we should probably not use `sameMacro` anymore.


dschrempf:
This reduces the interface we require from our macro language framework.

I think it would be great if can merge this (or a reviewed version) so we reduce the interface.

edsko (approved):

PR #1987, opened 2026-05-18
Add contributors, fix links, configure Lychee

TravisCardwell:
* Added three contributors
* Fixed a few links
* Updated Lychee configuration


TravisCardwell:
@jorisdral I fixed a link issue that was reported by the nightly Lychee job.  If it is running so frequently, we should prune false-positives or we run the risk of ignoring the job that always fails.  I am updating the `.lycheeignore`, adding a link to the reference (which is not great, unfortunately) as well as organizing using comments.  There are two URLs already in the file.  Perhaps we should ignore the entire `dl.acm.org` and `www.reddit.com` domains instead of just these specific URLs?  Perhaps just `https://registry.khronos.org/OpenGL/` remains, which I am not configuring since I do not know if it the current failure is intermittent or due to blocking automated access.

TravisCardwell:
Note that the author of #1981 is *not* added as a contributor; see the comments in that PR.

PR #1990, opened 2026-05-19
 `c-expr-dsl`: parse macros as function-like only if there is no whitespace after the name

### LINKED ISSUE #1903: Some object-like macros are treated as function-like macros
> In the example below, F is a function-like macro definition while G is an object-like macro definition. The difference is in the whitespace between the identifier and the opening parenthesis. hs-bindgen treats F and G both as function-like macro definitions, which is wrong.
> 
> ```c
> #define F(x, y) x + y
> #define G (x, y) x + y
> ```

Posted originally in #1902

jorisdral:
Resolves #1903

Function-like macros are only function-like if there is *no* whitespace between the macro name and the opening parenthesis of the parameter list. We were sometimes parsing object-like macros as function-like macros because we were not considering this. This is now fixed.

### DIFF
--- a/c-expr-dsl/src/C/Expr/Parse/Expr.hs
+++ b/c-expr-dsl/src/C/Expr/Parse/Expr.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE OverloadedRecordDot #-}
+
 module C.Expr.Parse.Expr (parseMacro, parseMacroType) where
 
 import Control.Monad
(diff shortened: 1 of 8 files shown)


dschrempf (approved):
Thank you!

edsko:
I feel like half of my depression about C comes from you.

edsko (approved):

jorisdral:
> I feel like half of my depression about C comes from you.

I'm honored ;)

PR #1991, opened 2026-05-19
Un-index Hs.Struct

### LINKED ISSUE #1576: Revisit defintion of `DeclData` (Is it necessary to track the number of fields on the type level at this stage?)
In `HsBindgen.Backend.Hs.AST` we have

```haskell
data Decl where
    ...
    DeclData                 :: SNatI n => Struct n -> Decl
    ...
```

That is, we track the number of struct fields on the type level. In comparison, we do not track the number of function arguments on the type level at this stage.

To simplify some code in `HsBindgen.Backend.Hs.Translation` and sub-modules, we could "postpone" generating the type-level information until we create the DeBruin expressions.

bolt12:
Closes #1576

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST.hs
@@ -132,10 +132,10 @@ data Field = Field{
 -- TODO <https://github.com/well-typed/hs-bindgen/issues/1757>
 -- For enums we generate /both/ a newtype /and/ a struct, and then define
 -- instances only for the struct. We should get rid of this nasty hack.
-data Struct (n :: Nat) = Struct{
+data Struct = Struct{
       name      :: Hs.Name Hs.NsTypeConstr
     , constr    :: Hs.Name Hs.NsConstr
-    , fields    :: Vec n Field
+    , fields    :: [Field]
     , origin    :: Maybe (Origin.Decl Origin.Struct)
     , instances :: Set Inst.TypeClass
     , comment   :: Maybe HsDoc.Comment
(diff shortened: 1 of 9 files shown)


hs-bindgen/src-internal/HsBindgen/Backend/SHs/Translation/Common.hs
```diff
@@ -31,7 +34,18 @@ translateElimStruct f (Hs.ElimStruct x struct add k) = ECase
...
+-- | Convert a list to a 'Vec' driven by the shape of an 'Add' proof.
+--
+-- The list must have at least as many elements as 'n'. Surplus elements are
+-- discarded. Invariant established by all current 'Hs.ElimStruct' construction
+-- sites: the field list determines the @n@ in 'Add' (via 'Vec.reifyList' in
+-- 'Hs.makeElimStruct', or by direct construction with matching arity).
+listToVecBy :: Add n ctx ctx' -> [a] -> Vec n a
+listToVecBy AZ     _      = VNil
+listToVecBy (AS a) (x:xs) = x ::: listToVecBy a xs
+listToVecBy (AS _) []     = panicPure "listToVecBy: list shorter than Add"
```
bolt12: we can use `fromList` (https://hackage-content.haskell.org/package/vec-0.5.1.1/docs/Data-Vec-Lazy.html#v:fromList)

hs-bindgen/src-internal/HsBindgen/Backend/Hs/AST.hs
```diff
@@ -793,7 +796,7 @@ type ElimStruct :: (Ctx -> Star) -> (Ctx -> Star)
 data ElimStruct t ctx where
     ElimStruct ::
          Idx ctx
-      -> Struct n
+      -> Struct
```
bolt12: This probably just needs the fields (as a type indexed vector)

edsko (approved):

PR #1992, opened 2026-05-19
`c-expr-dsl`: use `tasty-golden` instead of a custom test runner

jorisdral:
Branches off from #1990

PR #1993, opened 2026-05-19
`c-expr-dsl`: support multi-line macro definitions

jorisdral:
Branches off #1992

### DIFF
--- a/c-expr-dsl/test/fixtures/macros.h
+++ b/c-expr-dsl/test/fixtures/macros.h
@@ -151,6 +151,24 @@
 // Ternary operator is not yet in the expression grammar (see below).
 // #define FUNC_MAX(a, b)       ((a) > (b) ? (a) : (b))
 
+// FUNC_MULTILINE is like FUNC_SINGLELINE, but the former's definition is spread
+// over multiple lines. Both macros should lead to the same parse result.
+#define FUNC_SINGLELINE(xx, yy) xx * yy
+#define FUN\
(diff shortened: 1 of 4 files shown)


c-expr-dsl/test/fixtures/macros.h
dschrempf, line 1: Umph. Do we really want to parse this :-P?

dschrempf (approved):
Thank you.

The code looks good. I am a bit ambivalent to the change though, because it slows down macro parsing (we traverse the string to remove multilines).

jorisdral:
> The code looks good. I am a bit ambivalent to the change though, because it slows down macro parsing (we traverse the string to remove multilines).

~~I don't think it's going to slow down parsing noticeably. Laziness makes sure that we don't actually traverse the whole string. We only traverse the string until the first non-multiline character. In practice that's going to be a negligible cost I think~~

Oops, forgot about the text conversion. I pushed a new version that should have the better lazy behaviour

c-expr-dsl/test/fixtures/macros.h
jorisdral: The current status quo is to parse it as an object-like macro, which would be wrong. The alternative is to detect multilines and then return a parse failure, but at that point where we can detect multilines, it's also easy to correctly parse it

jorisdral: So I'm not sure there is a better alternative to supporting multi-line macro definitions

PR #1996, opened 2026-05-19
Move hlint --path to .hlint.yaml

### LINKED ISSUE #1995: HLint configuration for local runs
Could we make it as easy to run HLint locally as `hlint -j .`?

https://github.com/well-typed/hs-bindgen/blob/7d70866aba6e53965e382c5257b38e825fa546a2/.github/workflows/check-hlint.yml#L28-L32

The GitHub workflow checks three directories.

philderbeast:
Fixes #1995.

An alternative approach would be to exclude everything we don't want to run HLint over with `--ignore-glob=...` entries under `arguments`.

### DIFF
--- a/.github/workflows/check-hlint.yml
+++ b/.github/workflows/check-hlint.yml
@@ -28,5 +28,5 @@ jobs:
     - name: 'Run HLint'
       uses: haskell-actions/hlint-run@v2
       with:
-        path: '[ "hs-bindgen/src/", "hs-bindgen/app/", "hs-bindgen/src-internal/"  ]'
+        path: '.'
         fail-on: suggestion
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -100,3 +100,8 @@
 
 # To generate a suitable file for HLint do:
 # $ hlint --default > .hlint.yaml
+
+- arguments:
+  - --path=hs-bindgen/src
+  - --path=hs-bindgen/app
+  - --path=hs-bindgen/src-internal
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -102,6 +102,7 @@
 # $ hlint --default > .hlint.yaml
 
 - arguments:
-  - --path=hs-bindgen/src
-  - --path=hs-bindgen/app
-  - --path=hs-bindgen/src-internal
+  - --ignore-glob=alternatives/**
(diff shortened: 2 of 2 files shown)


dschrempf (approved):
Thank you for your contribution!

dschrempf:
Actually, coming back to this; I noticed that only the first --path is used, consecutive --path statements are simply ignored...

So we need to use --ignore-glob patterns.

I suggest using
```
- arguments:
  - --ignore-glob=alternatives/**
  - --ignore-glob=c-expr-runtime/**/HostPlatform.hs
  - --ignore-glob=examples/**
  - --ignore-glob=hs-bindgen/test-artefacts/**
```

dschrempf (changes requested):
See above, I am happy to implement the changes, if you like @philderbeast !

dschrempf:
Hi @philderbeast; I have picked your last commit and changed some other `hlint` options (more ignore patterns) #2002.

philderbeast:
Thanks for checking on the behaviour of `--path` @dschrempf.

I've added your suggested `--ignore-glob` list, regenerated the "Warnings currently triggered by your code" but excluded "Use panicPure" from that list. If you run `hlint -j .` these will fire.

> [!NOTE]
> To regenerate the ignore list, those under the heading "Warnings currently triggered by your code", I deleted them all and then ran `hlint --default . > .hlint-cmp.yaml` and manually pasted the resulting list back to `.hlint.yaml`.

dschrempf:
Yes, the `panic*` functions are only available within the sub-project `hs-bindgen`, so we should completely ignore them elsewhere. See #2002, which supersedes this PR, I hope this is OK!

philderbeast:
That's as much as I should do on this for now as I'm only starting to get familiar with this codebase. Happy to help with the HLint side of things further if need be.

Watched the [Haskell'25 talk](https://www.youtube.com/watch?v=SeLI3lMnhnA) about hs-bindgen so am very interested in this project.

dschrempf:
Superseded by #2002, thank you!

If you would like to be added as a contributor (see the Readme), let us know. If yes, how should we refer to you?

philderbeast:
Thanks @dschrempf. I see I'm already in the contributor list with c005c5253dc0aabaeb9df4bd8cbc7d88f2eb22c5. Things are moving fast with this project!

PR #1997, opened 2026-05-19
Add c-yaml example project with bindings to LibYAML

### LINKED ISSUE #1165: Create bindings for `libyaml`

mheinzel:
Resolves #1165.

Work in progress. The example program is very basic, maybe I'll come up with something more interesting that's still not too complicated.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -98,6 +98,7 @@ jobs:
           echo '{"runner": "ubuntu-latest" , "ghc-version": "9.4", "cabal-version": "3.16", "llvm-version": "15", "example": "c-qrcode"},'
           echo '{"runner": "ubuntu-latest" , "ghc-version": "9.4", "cabal-version": "3.16", "llvm-version": "15", "example": "c-rogueutil"},'
           echo '{"runner": "ubuntu-latest" , "ghc-version": "9.4", "cabal-version": "3.16", "llvm-version": "15", "example": "c-rpm"},'
+          echo '{"runner": "ubuntu-latest" , "ghc-version": "9.4", "cabal-version": "3.16", "llvm-version": "15", "example": "c-yaml"},'
           echo '{"runner": "ubuntu-latest" , "ghc-version": "9.4", "cabal-version": "3.16", "llvm-version": "15", "example": "cef"},'
           echo '{"runner": "ubuntu-latest" , "ghc-version": "9.4", "cabal-version": "3.16", "llvm-version": "15", "example": "libpcap"},'
           echo EOF
@@ -331,6 +332,10 @@ jobs:
     - name: 🧪 Build and run c-rpm example
       if: ${{ matrix.example == 'c-rpm' }}
       uses: ./.github/actions/examples/c-rpm
+
+    - name: 🧪 Build and run c-yaml example
+      if: ${{ matrix.example == 'c-yaml' }}
+      uses: ./.github/actions/examples/c-yaml
 
     - name: 🧪 Build and run cef example
       if: ${{ matrix.example == 'cef' }}
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -61,7 +61,7 @@ jobs:
       if: ${{ env.comprehensive == 'false' }}
       run: |
         { echo 'MATRIX_COMBINATIONS<<EOF'
-          # default
+          # default (ordered alphabetically)
           echo '{"runner": "ubuntu-latest" , "ghc-version": "9.4", "cabal-version": "3.16", "llvm-version": "15", "example": "blocktest"},'
           echo '{"runner": "ubuntu-latest" , "ghc-version": "9.4", "cabal-version": "3.16", "llvm-version": "15", "example": "bundled-c"},'
           echo '{"runner": "ubuntu-latest" , "ghc-version": "9.4", "cabal-version": "3.16", "llvm-version": "15", "example": "constructor-import-issue"},'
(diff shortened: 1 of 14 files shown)


.github/workflows/examples.yml
```diff
@@ -290,6 +293,10 @@ jobs:
       if: ${{ matrix.example == 'libpcap' }}
       uses: ./.github/actions/examples/libpcap
 
+    - name: 🧪 Build and run c-yaml example
+      if: ${{ matrix.example == 'c-yaml' }}
+      uses: ./.github/actions/examples/c-yaml
+
```
mheinzel: Is there a reason behind the order the examples are listed in? In the matrix above the entries are almost alphabetically ordered, but not quite. Here, I don't see a pattern.

I can just put the new example somewhere, but I can also apply some ordering.

dschrempf: I think we should order alphabetically.

manual/low-level/translation/macros.md
dschrempf, line 1: Thanks for finding these leftovers. I missed those when we changed the representation from `Word8` to `CChar`.

manual/low-level/translation/generated-names.md
```diff
@@ -56,10 +56,24 @@ data Triple = Triple {
...
 (The capitalization is a result of candidate name _fixing_, see below.)
 
-> [!NOTE]
-> Alternatively we could take advantage of `DuplicateRecordFields` or
-> `OverloadedRecordDot`/`OverloadedRecordUpdate`.
-> <https://github.com/well-typed/hs-bindgen/issues/69>
+#### Duplicate record fields
+
+Instead of using a prefix to make field labels globally unique, we can use
+take advantage of
```
dschrempf: I think this is a leftover; choose either "use" or "take advantage".

examples/c-yaml/generate.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+# Resolve the repository root if not already exported by a parent script
+# (generate-and-run.sh exports PROJECT_ROOT before sourcing/calling us).
+if [ -z "${PROJECT_ROOT:-}" ]; then
+    SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+    PROJECT_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
+    export PROJECT_ROOT
+fi
```
dschrempf, line 9: Do we want to get the project root again? I am not sure how we do this in the other examples, but it looks like a library file exposing PROJECT_ROOT (or get_project_root()) is in order.

I don't think you should do this right now, just mentioning.

examples/c-yaml/generate.sh
```diff
@@ -0,0 +1,22 @@
...
+
+LIBYAML_INCLUDE_DIR="${LIBYAML_INCLUDE_DIR:-./yaml/include}"
```
dschrempf, line 11: Is a relative path enough here? Does this mean, we cannot execute this script from somewhere else (do we even need to?)?

examples/c-yaml/generate-and-run.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
+
+# Exit on first error
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
```
dschrempf, line 7: I think we should use `git` to get the project root. (Sorry to be so picky, I think we do should do this across all examples, and it is most likely not your call to fix it

```
  PROJECT_ROOT=$(git rev-parse --show-toplevel)
```

examples/c-yaml/generate.sh
dschrempf: I am also just seeing that ./generate-and-run.sh already sets LIBYAML_INCLUDE_DIR, so we set it twice?

dschrempf (approved):
Hi Matthias! Thanks for taking care of this.

I left a few comments below, I am not sure if you should actually fix them in the course of this PR; but we are a bit sloppy with some Bash code involved...

The Haskell code reads well, thank you! I don't know enough about libYaml to review it extensively, but if it works for you and CI, that is great!

examples/c-yaml/generate.sh
mheinzel: The idea for splitting out `generate.sh` is that it can be run independently. In that case `LIBYAML_INCLUDE_DIR` is not set already. I think originally Armando introduced that split in `libpcap` to build it with static linking and system deps on Alpine Linux, but I also found it useful for my own development workflow.

I generally followed the `libpcap` and `c-minisat` setup quite closely. I do think some parts of the example build scripts could be improved/unified/simplified, though (e.g. at times making an effort to work when being executed from a different directory, but sometimes not), and have it on my list to open another PR for that.

examples/c-yaml/generate-and-run.sh
mheinzel: I agree we should follow a more principled approach. :+1: 
And I can take care of it across all the examples, but probably in a separate PR.

examples/c-yaml/generate.sh
dschrempf: Sure. In general, I think we should have a lib script that setups paths; and this script can be parsed from either ./generate-and-run.sh or ./generate.sh.

(The lib script could also just defined functions that can be brought into scope and run in other scripts). Again, not saying you should do this in the course of this PR; what you could do is open a ticket.

mheinzel:
I fixed the small issue, rebased and now removed the commit that runs this example as part of the default jobs. It should still work.

I also saw that there are `HasField` instances for unions as well, so I changed my code to use record dot access instead of getters.

examples/c-yaml/generate.sh
mheinzel: I created an issue: https://github.com/well-typed/hs-bindgen/issues/2088

I wasn't quite sure which labels to add, as it just relates to the examples, so feel free to update them.

Issue #942, opened 2025-07-24
Make macro infrastructure standalone


edsko:
This involves a number of steps:

- [ ] Rename `c-expr` to `c-expr-runtime`
- [ ] Introduce `c-expr-dsl`, which defines the macro AST, the type checker, etc. Initially, `hs-bindgen:internal` would depend on `c-expr-dsl`
- [ ] Remove dependency on `c-expr-dsl` from `hs-internal:internal`. Instead, introduce an abstraction for "macro language", so that people can plug in different DSLs if they so desire. This abstraction would need to capture the parser, the type checker, and the translation to `hs-bindgen` internal Haskell AST.
- [ ] Then reintroduce the dependency on `c-expr-dsl` but from the main `hs-bindgen` library, as the _default_ DSL to be used for macros.

edsko:
This is somewhat blocked on @jorisdral 's work on separating out the constructors of `HsType` that are only used for macros.

edsko:
Moving this to `release-0.1` as we're not planning a Hackage release for the alpha.

jorisdral:
See #1608

dschrempf:
Working on this; leaving some notes on the API.

Summary of what we likely will "require" from type-like macros (TExpr) and value-like macros (VExpr):

*TExpr*

```haskell
-- We need 'Foldable' when getting the dependencies
depsOfTExpr :: Foldable f => Proxy p -> f (Id p) -> [Id p]
depsOfTExpr _ = toList

-- When mangling names, we need to traverse 'TExpr' (and likely also in other places)
mangleTExpr :: Traversable tExpr => tExpr a -> ...

-- Translation
translateType ::
     KnownTypes
  -> TExpr a (?)
  -> Either Error HsType
     -- ^ This translation may fail (and it needs to lookup 'KnownTypes',
     -- something we still need to specify).

-- Complications
-- - ResolveBindingSpecs (try to use Traversable instance)
-- - Instance resolution still a question
```

*VExpr*

- We will change how `VExpr` treats local arguments (they are not free variables).

```haskell
-- We need 'Foldable' when getting the dependencies
depsOfVExpr :: Foldable f => Proxy p -> f (Id p) -> [Id p]
depsOfVExpr _ = toList

-- When mangling names, we need to traverse 'VExpr' (and likely also in other places)
mangleVExpr :: Traversable vExpr => vExpr a -> ...

-- Translation
translateType :: VExpr DeclIdPair -> ClosedType
[...]

PR #2000, opened 2026-05-20
Remove macros from ParseState

### LINKED ISSUE #1999: Remove macros from parse state
Prerequisite for #942 

When we introduce the macro framework, all passes/data types carrying macros will need a macro language type parameter. Not having macros in the parse state greatly reduces the burden of sprinkling type parameters.

dschrempf:
Prerequisite for #942 

Closes #1999

For the curious ones, we will celebrate PR 2024, and probably PR 2003.

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse.hs
@@ -29,19 +29,19 @@ parseDecls parseEnv = do
     root <- clang_getTranslationUnitCursor parseEnv.unit
     ParseDecl.run parseEnv $ do
       resultsWithLocs <- HighLevel.clang_visitChildren root topLevelDecl
-      allMacros <- reverse <$> ParseDecl.getMacroDefinitions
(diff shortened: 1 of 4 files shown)


jorisdral:
Just FYI, for #1892 I need to store the tokens for each macro so that I can use them in a new pass, so even though you are removing the `macroDefinitions` field now, I'm probably going to reintroduce it later. I think that's fine but I just wanted to let you know

edsko:
_Tokens_ are fine, _parsed_ macros are not.

dschrempf:
If we store tokens, then we might want to parse them later, as we discussed yesterday @edsko ?

dschrempf:
Conclusion after meeting with @jorisdral:

- We do not create a "parse macros pass" for now.
- Joris uses the ParseState to store tokens and directly uses the output in the addiitonal pass he is writing between TypecheckMacros and ReparseMacros.
- If need be, we add the "parse macros pass"; then, we have to be careful to also register tokens of macros that do not parse into the DeclIndex, so that the pass of Joris has access to them. (He needs access to tokens of all macros, not just the ones that parse).

edsko (approved):

Issue #812, opened 2025-06-27
Manual: macros


edsko:
We should describe macros for types and macros for expressions.

We should also outline what we do and do not want to support in terms of macro handling; where minor additional features are possible and what is definitely out of reach. For example, supporting macros that define structs is definitely out of scope since we do want to rely on `clang` to tell us things like memory layout, but adding minor extensions to the expression language is possible.

edsko:
@dschrempf Mind taking this one on? It's one of the few remaining issues for release 0.1.

edsko:
See also #2011.

PR #2011, opened 2026-05-26
`Manual`: macro terminology

jorisdral:
A first foray into recording terminology related to macros (and reparsing).

### DIFF
--- a/manual/terminology.md
+++ b/manual/terminology.md
@@ -66,6 +66,13 @@ declarations.
 
 <https://en.cppreference.com/w/c/language/scope.html#File_scope>
 
+### Function-like macro
+[t:function_like]: #function-like-macro
+
+A *function-like* macro replaces each occurrence of that macro's name with its
+definition, additionally taking a list of named arguments, which replace
+corresponding occurrences of the argument name in the definition.
+
 ### Indirect field
 
 [Fields][t:field] of an [anonymous struct/union][t:anon] can be accessed as if
@@ -80,6 +87,45 @@ Haskell bindings are generated for the field declaring the [anonymous
 struct/union][t:anon] rather than for the indirect fields.
 
 </details>
+
+### Macro
+[t:macro]: #macro
+
+A (text) macro is a name associated with a replacement text.
+
+### Macro definition
+[t:macro_def]: #macro-definition
+
+A macro *definition* declares a name in the macro namespace with a replacement
+text. A macro can be [object-like][t:object_like] or
+[function-like][t:function_like]. Macro definitions might be
+[parsable][t:parsable_macro].
+
+### Macro expansion
+[t:macro_expansion]: #macro-expansion
+
+A C preprocessor *expands* macro [invocations][t:macro_inv], replacing
+invocations by their definition, which may be processed additionally depending
+on the type of the macro [definition][t:macro_def].
+
+### Macro invocation
+[t:macro_inv]: #macro-invocation
+
+A macro *invocation* references a macro definition by name so that it can be
+replaced by its definition. Macro invocations are [expanded][t:macro_expansion]
+by a C preprocessor. An [object-like][t:object_like] macro is invoked by
+referencing the macro's name. A [function-like][t:function_like] macro is
+invoked as if it is a function call.
+
+### Macro type
+[t:macro_type]: #macro-type
+
+A [macro][t:macro] that is parsed by `hs-bindgen` as a Haskell type.
+
+### Macro value
+[t:macro_value]: #macro-value
+
+A [macro][t:macro] that is parsed by `hs-bindgen` as a Haskell value.
 
 ### Named field
 [t:named_field]: #named-field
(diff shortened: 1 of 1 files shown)


manual/low-level/terminology.md
```diff
@@ -81,6 +92,45 @@ struct/union][t:anon] rather than for the indirect fields.
...
+A macro *definition* declares a name in the macro namespace with a replacement
+text. A macro can be [object-like][t:object_like] or
+[function-like][t:function_like]. Macro definitions might be
+[parsable][t:parsable_macro].
+
+### Macro expansion
+[t:macro_expansion]: #macro-expansion
+
+A C preprocessor *expands* macro [invocations][t:macro_inv], replacing
+invocations by their definition, which can optionally be additionally processed
```
dschrempf:
```suggestion
invocations by their definition, which may be processed additionally,
```

manual/low-level/terminology.md
```diff
@@ -81,6 +92,45 @@ struct/union][t:anon] rather than for the indirect fields.
...
+depending on the type of macro [definition][t:macro_def].
```
dschrempf:
```suggestion
depending on the type of the macro [definition][t:macro_def].
```

manual/low-level/terminology.md
```diff
@@ -81,6 +92,45 @@ struct/union][t:anon] rather than for the indirect fields.
...
+
+A [macro][t:macro] that is parsed by `hs-bindgen` as a Haskell expression.
+
+### Macro invocation
+[t:macro_inv]: #macro-invocation
+
+A macro *invocation* references a macro definition by name so that it can be
+replaced by its definition. Macro invocations are [expanded][t:macro_expansion]
+by a C preprocessor. An [object-like][t:object_like] macro is invoked by
+referencing the macro's name. A [function-like][t:function_like] macros is
```
dschrempf:
```suggestion
referencing the macro's name. A [function-like][t:function_like] macro is
```

manual/low-level/terminology.md
```diff
@@ -110,6 +166,18 @@ Unnamed bit-fields are not translated to fields in the corresponding Haskell
 record in the generated Haskell bindings.
 </details>
 
+### Parsable macro
+[t:parsable_macro]: #parsable-macro
+
+A *parsable* macro is a macro definition that can be parsed as a [macro
+type][t:macro_type] or as a [macro expression][t:macro_expr].
```
dschrempf: We are explicitly careful in `c-expr-dsl`, and say "type-like macro expression" or "value-like macro expression", because technically speaking, both of these are parsed as "macro expressions" (we just don't know yet whether they are types or values), if that makes sense.

Maybe we can also convey this detail?

manual/low-level/terminology.md
```diff
@@ -110,6 +166,18 @@ Unnamed bit-fields are not translated to fields in the corresponding Haskell
...
+
+### Reparsing
+[t:reparsing]: #reparsing
+
+`hs-bindgen` parses headers twice. The second *reparse* inspects the source C
```
dschrempf: AFAIK, `hs-bindgen` only parses a subset of declarations a second time (the ones containing macro expansions).

manual/low-level/terminology.md
jorisdral: Hmm, I wrote this from the perspective of the user: they see a binding for a Haskell type or Haskell expression/function, hence the names were picked that way. But I also understand what you mean -- between macro parsing and type checking the parse results are all technically (type-like/value-like) macro expressions. Maybe we should consider that an implementation detail of the macro language though? Once macro infrastructure is standalone, I assume `hs-bindgen` won't know about this intermediate stage anymore

jorisdral: Or maybe we should rename "macro expression" to "macro function"?

dschrempf: `hs-bindgen` does know about "macro types" and "macro values" (not mentioning expression deliberatly). Maybe we can use those? If not, it's not a big deal, I just wanted to convery that we struggled somewhat with choosing consistent names (especially since "expression" is somewhat overloaded).

jorisdral: Renamed it to macro value!

TravisCardwell:
(At @edsko's request,) I moved the terminology page to the top, not specific to low-level bindings.  I did not make any changes to the content of the file, so it should be trivial to resolve the merge conflict. :bow:

jorisdral:
@dschrempf is this PR still in sync with the macro terminology you've been using? If not, do you have suggestions for updates?

dschrempf:
I think, in general, yes.

I didn't use the words "Object-like macro" "Parsable macro", and "Macro invocation". Out of the top of my head, I think sometimes I used "Macro declaration" instead of "Macro definition". The difference between declaration and definition is blurred, in my opinion; and we use declaration much more often than definition, I feel.

dschrempf:
@jorisdral do you mind if I rebase this, and maybe make small amendments? (Coming from https://github.com/well-typed/hs-bindgen/issues/812).

jorisdral:
@dschrempf this can now be closed, right?

dschrempf:
Superseded by #2252

PR #2017, opened 2026-05-27
Support untagged types indirectly referenced by globals

### LINKED ISSUE #1429: Support top-level anonymous structs

jorisdral:
Follow-up to #1614

We were supporting cases like:

```c
struct { int x; } global1;
```

But not:

```c
struct { int x; } * global2;
```


bolt12 (approved):
LGTM

edsko (approved):

Issue #8, opened 2024-08-01
Enumerate the high-level patterns supported by `c2hs`


edsko:
The main thing that `c2hs` offers beyond what `hsc2hs` offers is support for high-level patterns. We should enumerate them, and support them in our high-level API wherever appropriate, or list the ones we cannot/don't want to support. We should create tickets for these, and associate them with [Milestone 3](https://github.com/well-typed/hs-bindgen/milestone/4). 

References:

* https://github.com/haskell/c2hs/wiki/User-Guide
* [c2hs.pdf](https://github.com/user-attachments/files/16452139/c2hs.pdf)  

Have attached the original paper describing the tool to the ticket as it is a bit hard to find online.

edsko:
Going to repurpose this ticket for enumerating theses patterns and constructing a set of type classes that captures these patterns, before we look at _automatic_ generation; so tool-assisted high-level binding first.

bolt12:
# High-level binding DSL for `hs-bindgen`

## Introduction and context

Currently, `hs-bindgen` reads a C header and generates low-level Haskell bindings. In these bindings, a C `int` becomes `CInt`, a `const char *` becomes `PtrConst CChar`, a flexible array becomes `IncompleteArray a`, etc.

What most users want, though, is a Haskell-native API where strings are `String` or `ByteString`, bools are `Bool`, nullable pointers are `Maybe`, arrays are `IncompleteArray` (mappable to `Vector`). The QR-code example in `examples/c-qrcode/` shows the gap in practice. By hand, calling `qrcodegen_encodeText` looks like this:

```haskell
basicDemo :: IO ()
basicDemo =
  F.withCAString "Hello, world!" $ \text ->
    F.allocaArray (fromIntegral QR.qrcodegen_BUFFER_LEN_MAX) $ \tmp ->
      F.allocaArray (fromIntegral QR.qrcodegen_BUFFER_LEN_MAX) $ \qr -> do
        ok <- QR.qrcodegen_encodeText
                (PtrConst.unsafeFromPtr text) tmp qr
                QR.Qrcodegen_Ecc_LOW QR.qrcodegen_VERSION_MIN
                QR.qrcodegen_VERSION_MAX QR.Qrcodegen_Mask_AUTO
                (F.fromBool True)
        ...
```

[...]

PR #2018, opened 2026-05-27
High Level Binding Combinators Design

bolt12:
Proposal in https://github.com/well-typed/hs-bindgen/issues/8#issuecomment-4535882902


bolt12:
Closed in favour of moving the hs-bindgen-highlevel to [well-typed/binding-combinators](https://github.com/well-typed/binding-combinators) and renaming it.

Issue #2012, opened 2026-05-26
Improve `invocationCanFloatInwards`


jorisdral:
For simplicity we have picked a temporary approximation of this property. All macro invocations are considered not inwards-floatable as long as there is a macro name that has multiple definitions. The goal is to write a better approximation later.

_Originally posted by @jorisdral in https://github.com/well-typed/hs-bindgen/pull/1892#discussion_r3302673640_

jorisdral:
Maybe we can rename the function to `isExpansionUnique`? That is more descriptive for users of `hs-bindgen`, I think

jorisdral:
```c
#define A int
#define B bool
#undef B
#define B char

A foo (void);
void bar (A, B);
```
once we relax the constraint, warn about `bar`

```c
#define A int
#define B bool
#undef B
#define arg A x

void baaz (arg, B);
```

but we can't warn about `baaz`

PR #2032, opened 2026-06-01
`Parse`: add `SourceRangeMap` abstraction for recording macro expansions

jorisdral:
Builds on top of #2031, so that PR should be merged first.

This PR is made in preparation for #2012.

In the spirit of encapsulating functionality, we add a an abstraction for (performantly) recording the source location ranges for macro expansions. We are likely to modify this abstraction in the future in the context of other changes to macros.


dschrempf:
I need a bit more context before I can meaningfully review the PR. What are we trying to achieve here?

jorisdral:
@dschrempf  I'm going to make some modifications to how we record macro expansions in the context of #2012. I thought it would be nice to factor out this "source range map" into its own module in preparation for that. Nothing about the functionality changes at the moment. I figured you might want to be notified since you worked on this code recently

edsko (approved):

PR #2034, opened 2026-06-02
`PrepareReparse`: expand macro invocations if their expansion is unique

### LINKED ISSUE #2012: Improve `invocationCanFloatInwards`

jorisdral:
Builds on top of #2032, so that PR should be merged first

Resolves #2012

NOTE: `invocationCanFloatInwards` is now called `isExpansionUnique`. 

The `PrepareReparse` pass expands macro invocations before reparsing given that these macro invocations and the corresponding macro definitions satisfy some conditions that ensure we can expand these macro invocations unambiguously. This allows us to run the `clang` preprocessor on an artificially constructed header to obtain the (partially) expanded source code of declarations that we want to reparse. It can be partially expanded because we do not expand macro invocations for macro definitions that we parsed as types. These are handled by the reparser as if they were references to typedefs.

The temporary implementation of the expansion condition is simple: we expand macros if and only if there are no macros that are defined more than once in the same translation unit. The expansion of an invocation of a multiply defined macro (or any macro that transitively uses that multiply defined macro) is not unique: where the macro is invoked in the file influences what the invocation expands to. This incurs ambiguity. 

The downside of the temporary implementation is that it prevents macro invocations from being expanded even if that macro invocation does not (transitively) refer to a multiply defined macro. We want a more fine-grained approach, where we can decide *per invocation* whether it has a unique expansion. We can then decide per declaration whether we can expand any macro invocations used therein. This is implemented in this PR using `isExpansionUnique`.


edsko (approved):
Approved modulo the graph implementation.

Issue #2022, opened 2026-05-28
Some `DelayedParseMsg`s are not parse messages


dschrempf:
In particular,

- 'ParseMacroErrorReparse' should be a delayed `ReparseMacroExpansions` message.
- 'ParseMacroErrorReparseAlign' should be a delayed `Align` message (after #2015).
- 'ParseMacroReparseUnknownType' should be a delayed `ReparseMacroExpansions` message.

See TODO items in the code. (On PR #2015).

jorisdral:
There is also a `PrepareReparse`-related message now, and I'm planning on adding more delayed messages

edsko:
Let's fix this after the first release.

jorisdral:
In #2109, I moved the three constructors mentioned in the original issue into their own `DelayedReparseMacroExpansionsMsg` datatype. 

I'm not sure if there are any messages left in `DelayedParseMsgs` that have to move but it would be good to keep this issue open regardless (or we create a new one?). We could review how we currently register delayed messages, and then maybe we can come up with a more principled approach. For example, we currently have separate fields in the `Success` type for each pass:

```hs
data Success l p = Success {
    decl                              :: C.Decl l p
  , delayedParseMsgs                  :: [DelayedParseMsg]
  , delayedPrepareReparseMsgs         :: [DelayedPrepareReparseMsg]
  , delayedReparseMacroExpansionsMsgs :: [DelayedReparseMacroExpansionsMsg]
  }
```

Maybe we could define a sum type of delayed trace messages like how we have `TraceFrontend` for immediate trace messages? Maybe we use `sop-core` combinators? 😛

jorisdral:
I opened #2110 with my suggestion from above, and I'll close this issue as part of #2109

Issue #1580, opened 2026-01-21
Consider making the `DeclIndex` a central "database"


edsko:
The `DeclIndex` has grown in importance in `hs-bindgen`. Initially it was merely a reflection of the parse result, and not updated after that, but over time we've been adding more and more information to it, updating it through the various frontend passes. However, since that was not its original purpose, some aspects of the `hs-bindgen` internal design could and/or should be reconsidered:

* There is an awkward duplication between the _list_ of declarations that we hand from pass to pass, and the `DeclIndex`. We could instead regard the list as a "view" on the `DeclIndex` (that is, a `[DeclId]`). Most passes don't actually need to process declarations in order at all. 

* We sometimes lose information when we update the `DeclIndex`. For example, when we replace a declaration by an external binding spec, we currently lose all location information (#1577). Perhaps we should not remove entries from the "database", but instead "add" more information; "here is the C declaration, and by the way, here is the external declaration that will replace it".

* There is additional information that we'd want to store in this database; for example, it would make sense to use it for a generalized concept of delayed trace messages (#1554).

This would need to be considered carefully. A "database oriented" way of working comes with draw-backs of its own:

* Types cannot guarantee that lookups will succeed
* If we want to store different types of information for different kinds of declarations, it might be much more convenient to store those in the declarations themselves rather than as separate "tables", though this would fit well with our trees-that-grow approach
* We might introduce new consistency problems. For example, if we consider the "list" of declarations as a view on the database, as suggested above, then if we compute this "order" list in one pass, and then in a later pass change declarations in any way, it might be even easier to forget to also update the ordered list. (Perhaps we should recompute it every time we need it, though that will have performance consequences.)

TravisCardwell:
This topic came up in discussion today, in regards to #1564.  We attempted to query the `DeclIndex` for information that was not there until a later pass.  The function in question just returned empty lists, and it was not obvious that there was an issue.  This kind of issue does not happen with TTG-style annotations, where we would get a type error if we attempt to use something that is not available yet.  See https://github.com/well-typed/hs-bindgen/issues/1564#issuecomment-3781560468 for details.

edsko:
This came up once more, #1699. @TravisCardwell , @dschrempf and I discussed this for a while. Some conclusions.

## Ordering

There are a few different kinds of orderings of declarations, which may be relevant for different kinds of processing:

[...]

Issue #1554, opened 2026-01-15
Delay more messages (such as NameMangler messages)


dschrempf:
This idea generalizes the concept of delayed parse messages which we have already implemented.

We could manage a separate map `Map DeclId [DelayedTrace]` of delayed trace messages. Only when we select a declaration to be translated, we emit all messages accumulated for that declaration.

We do this already for parse messages (albeit using a different internal representation/data type), but not for messages from other passes (such as `MangleNames`).

edsko:
Marking this as a medium priority enhancement. I initially thought low priority would suffice, since this _mostly_ applies to `Info` messages. However, now that we've made the squashing message level `Notice` it becomes more important: it's a bit strange, and confusing, to have this message shown if such a type is not selected.

edsko:
Marking this for `release-0.1`, but _only_ for the new `Notice`-level "squash" message. We can downgrade to low priority after that, I think. (More generally, it's important for messages _above_ Info level, but low priority for Info-level messages).

dschrempf:
Related: #1734

edsko:
> Marking this for `release-0.1`, but _only_ for the new `Notice`-level "squash" message. We can downgrade to low priority after that, I think. (More generally, it's important for messages _above_ Info level, but low priority for Info-level messages).

Since @dschrempf has now done this, moving this to 0.2.

dschrempf:
Some thoughts about immediate and delayed parse messages I am having while
implementing the `ParseMacros` pass:

**Discrepancies between passes**

Before we construct the translation unit, we have three passes `SimplifyAST`,
`AssignAnonIds`, and now also `ParseMacros`.

`SimplifyAST` and `AssignAnonIds` still exclusively use "immediate" traces. An
immediate trace is even necessary in the `AssignAnonIds` pass because it may
_remove_ parse results from the list.

On the other hand, the new `ParseMacros` pass will use a delayed parse message.

This is a tiny discrepancy that may be useful to know when handling this ticket.

**Delayed parse messages may be lost**

Another aspect is that when a `ParseSuccess` is replaced by a `ParseFailure`, we
loose the delayed parse messages attached to the `ParseSuccess`.

jorisdral:
I added delayed `PrepareReparse` messages to the select pass. See also https://github.com/well-typed/hs-bindgen/pull/2036#discussion_r3346813647

dschrempf:
Related: #1851

PR #2036, opened 2026-06-02
Delayed trace messages for the `PrepareReparse` pass

jorisdral:
Previously they were included in the `DelayedParseMsg` datatype. Now they are in their own `DelayedPrepareReparseMsg` datatype.

Note that there is also a general open issue (#2022) to make sure delayed messages for different passes are in their own datatypes rather than the `DelayedParseMsg` datatype


hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclIndex.hs
```diff
@@ -204,6 +209,15 @@ unusableToLoc = \case
     UnusableTypecheckMacrosError loc _ -> [loc]
     UnusableOmitted loc                -> [loc]
 
+data Success p = Success {
+    decl                      :: C.Decl p
+  , delayedParseMsgs          :: [DelayedParseMsg]
+  , delayedPrepareReparseMsgs :: [DelayedPrepareReparseMsg]
+  }
+  deriving stock (Generic)
```
jorisdral: `ParseResult` only holds `DelayedParseMsg`. I tried to put `DelayedPrepareReparseMsg` in there too, but that causes import cycles. I added this `Success` type to avoid that

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclIndex.hs
```diff
@@ -461,6 +483,28 @@ registerMacroTypecheckFailure
...
+  -> DeclIndex
+registerDelayedPrepareReparseMsg (declId, msg) (DeclIndex i) = DeclIndex $
+    Map.adjust addMsg declId i
+  where
+    addMsg :: Entry -> Entry
+    addMsg (UsableE (UsableSuccess ps)) =
+      UsableE $ UsableSuccess ps{
+          delayedPrepareReparseMsgs = ps.delayedPrepareReparseMsgs ++ [msg]
+        }
+    addMsg entry = entry
```
jorisdral, line 513: Function for registering `DelayedPrepareReparseMsg`

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
```diff
@@ -390,19 +393,31 @@ getSelectMsgsDeclId
...
+  where
+    mkAMsg :: HasCallStack => SelectMsg -> AMsg Select
+    mkAMsg msg = withCallStack WithLocationInfo{
+          loc = declIdLocationInfo declId [success.decl.info.loc]
+        , msg = msg
+        }
+
+    _coveredAllCases :: Success EnrichComments -> ()
+    _coveredAllCases = \case
+        DeclIndex.Success _ _ _ -> ()
```
jorisdral: Helper function so that we don't forget to trace messages if we ever add delayed trace messages for other passes

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclIndex.hs
dschrempf: Yes, I always thought it is not nice that we associate successful declaration IDs with parse successes. I think the success type captures this nicely. In the end, it could contain a lot of different types of delayed messages that are emitted _when this declaration is selected_.

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclIndex.hs
dschrempf: Great, this works for `PrepareReparseMsg`; in the future, we may want something more general: "Register a delayed trace message with this declaration ID".

I think we don't need to "register delayed parse messages" because they are directly added when we assembel the declaration index.

This is all related to #1580. Maybe we can leave a comment?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select/IsPass.hs
```diff
@@ -210,30 +215,32 @@ instance PrettyForTrace SelectMsg where
...
+    SelectDeprecated{}               -> "select"
+    SelectDelayedParseMsg x          -> "select-" <> getTraceId x
+    SelectParseNotAttempted{}        -> "select-parse"
+    SelectParseFailure x             -> "select-" <> getTraceId x
+    SelectConflict{}                 -> "select"
+    SelectMangleNamesFailure{}       -> "select-mangle-names-failure"
+    SelectMangleNamesSquashed{}      -> "select-mangle-names-squashed"
+    SelectMacroTypecheckFailure x    -> "select-" <> getTraceId x
+    SelectDelayedPrepareReparseMsg x -> "select-" <> getTraceId x
+    SelectNoDeclarationsMatched      -> "select"
```
dschrempf, line 245: I think adding new constructors for Select messages is the correct way to go now, but in the end, we may want to have a single "DelayedTraceMsg" constructor?

Some of these delayed messages are special though. Not sure how we can simplify, while not losing these special connotations.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
dschrempf: What if you directly pattern match above? Then we don't need this special case.

dschrempf (approved):

edsko (approved):

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
jorisdral: Makes sense, changed that!

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select/IsPass.hs
jorisdral: Right, I also think it would be good to have a more general constructor for delayed messages. I'm not much familiar with the `Select` pass, so I also don't know how to do that if there are delayed messages that have to be treated specially

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclIndex.hs
jorisdral: I tried some approaches with a more general design but I ended up hitting cyclic imports a lot of the times. #1554 is probably where we could attempt to do this for real

Issue #2020, opened 2026-05-28
Enum tag clashes with enumerator name


jorisdral:
For an `enum` definition like so:

```c
enum A { a };
```

We generate:

```hs
newtype A = A ...
pattern A :: A
pattern A = A 0
```

Which fails to compile because `A` is both defined as a constructor and pattern synonym

dschrempf:
The name manger should usually take care of that. I think the reason is that the pattern names are created in Pass 2 of the name mangler.

See #1432.

I think we should fix this soon. It is not too hard to fix, I believe.

TravisCardwell:
This one can be resolved using a prescriptive binding specification.

Example:

```yaml
version:
  hs_bindgen: 0.1.0
  binding_specification: '1.0'

ctypes:
  - headers: ex.h
    cname: enum A
    hsname: EnumA
```

Command:

```
$ hs-bindgen-cli preprocess \
    --unique-id ex \
    --create-output-dirs \
    --hs-output-dir ex-src \
    --prescriptive-binding-spec ex.yaml \
    -I "$(pwd)" \
    ex.h
```

Result:

```haskell
newtype EnumA = EnumA ...
pattern A :: EnumA
pattern A = EnumA 0
```

There are other cases where we generate conflicting names like this as well.  I think that we decided to *not* check for such collisions, as users should use a prescriptive binding specification to resolve them.

#1432 exists because we currently have no way to write specifications for auxiliary types.  We need to fix that so that users can resolve auxiliary type name clashes using a prescriptive binding specification.

dschrempf:
#2089 will fix this, I am also adding a golden test.

Issue #1184, opened 2025-10-04
Misleading error message when declaration is missing


dschrempf:
Coming from #1068, where we have trouble with circular dependencies. I tried a different ordering:

```c
typedef void (*fun_ptr)(struct forward_declaration *s);

struct forward_declaration;

struct forward_declaration {
  fun_ptr f;
};
```

and we get a `mangle-names` error, which is misleading:

```
[Warning] [Libclang ] [clang] ./test.h:2:32: warning: declaration of 'struct forward_declaration' will not be visible outside of this function [-Wvisibility]
[Warning] [HsBindgen] [decl-index-error] 'forward_declaration' (tag) declared at "./test.h:2:32" was redeclared at "./test.h:6:8". No binding generated.
[Error  ] [HsBindgen] [mangle-names] Missing declaration: 'struct forward_declaration'; did you select the declaration?
```

Should we "fix this" (i.e., improve the error message)? If not, we can close the issue.

edsko:
Going to mark this as a low priority enhancement, this this is essentially invalid C code. However, since this is related to @dschrempf 's current work on selection, let's see what the situation is once that is done, and then decide whether to close this as `wontfix` or not.

dschrempf:
I was repeating this example, we now get different output:

```
[Warning] [Libclang ] [clang] ./test.h:1:32: warning: declaration of 'struct forward_declaration' will not be visible outside of this function [-Wvisibility]
[Warning] [HsBindgen] [select] 'fun_ptr' at "./test.h 1:16"
  Could not select declaration (direct select predicate match):
    Transitive dependency unusable:
      'struct forward_declaration' at ("./test.h 1:32", "./test.h 5:8")
        Conflicting declarations
[Warning] [HsBindgen] [select] 'struct forward_declaration' at ("./test.h 1:32", "./test.h 5:8")
  Could not select declaration:
    Conflicting declarations
[Notice ] [HsBindgen] [artefact-no-bindings-multiple] No output generated (all binding categories with base module name "Generated" are empty)
```

However, I think this is also not quite right. Why are there "conflicting" declarations? Let me know if I should investigate further.

edsko:
This is actually related to something @bolt12 is working on: when we have a declaration of a struct inside a function declaration, we should throw out the function declaration, as this results in name mangler panic. @bolt12 is there a ticket for that issue? I suspect that once we close that, we can close this one also (we should verify).

dschrempf:
[...]

PR #2043, opened 2026-06-05
Consistently check for duplicate names

### LINKED ISSUE #1432: Think about name clashes arising from the introduction of `_Aux` types
We introduce `_Aux` (formerly `_Deref`; #1427) types in at least two places. For 

```c
typedef struct foo { .. } * foo;
```

we generate

```hs
data Foo_Aux = ...
newtype Foo = Foo Foo_Aux
```

and for 

```c
typedef void (*f)(int);
```

we generate

```hs
newtype F_Aux = F_Aux (Int -> IO ())
newtype F = F (FunPtr F_Aux)
```

In principle these names could clash, if the C code would have `foo_Aux` or `f_Aux` as a type declaration.

Related: #1373.

### LINKED ISSUE #1445: Avoid name clashes between tagged type and surrounding typedef
Given

```c
typedef struct foo { int x; } foo[50];
```

we currently generate _two_ types, both called `Foo`; one of these should be renamed to `Foo_Aux`. We do this correctly for _pointers_, but not for other kinds of direction.

### LINKED ISSUE #1552: Suggested example: Struct with field for which name mangler fails
For example,

```c
struct bad {
   int something_for_which_the_name_mangler_fails;
};
```

The failure should propagate and the `struct bad` be registered as failed (i.e., removed from list of declarations, and the decl-index should know about name mangle failure).

### LINKED ISSUE #1734: Name mangler does not detect all name collisions yet
### Example of name collision

The C code
```c
struct Collision {
  int x;
};

// Collides.
void collision_x();
```

may lead to collisions that we do not detect. In particular, we will generate (in multi-file mode)

CTypes.hs

```haskell
data Collision = Collision { collision_x :: CInt }
  }
```

Safe.hs

```haskell
collision_x :: IO ()
```

In single-file mode, and without suffix (_which is the default for TH mode_), these definitions collide. As expected, the resulting error is, e.g.

```
Multiple declarations of ‘collision_x’
Declared at: /home/dominik/work/hs-bindgen/hs-bindgen/test/th/Test/TH/Test02.hs:29:1
             /home/dominik/work/hs-bindgen/hs-bindgen/test/th/Test/TH/Test02.hs:29:1
```

### Reason

In the name mangler, we do check for name collisions in the first pass, when generating top-level names.

In the second pass, we generate field names and such, not checking for collisions anymore.

dschrempf:
The `MangleNames` pass now consistently checks for name duplicates in both of its passes. In Pass 2, we distinguish between top-level names that can collide, and other names that do not lead to collisions. An example of the second kind are record field names when `OmitFieldPrefixes` and `DuplicateRecordFields` are enabled.

Closes #1734 #1432 #1445

Related issue: #1520 

```hs
-- TODO <link to issue>
-- Brief description
```

(If this does not apply, feel free to delete this template text.)

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/Typedefs.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/Typedefs.hs
@@ -151,45 +151,82 @@ analyseTypedef ::
   -> C.Typedef ResolveBindingSpecs
   -> TypedefAnalysis
 analyseTypedef declUseGraph typedefInfo typedef =
-    case taggedPayload typedef.typ of
-      Nothing      -> mempty
-      Just payload ->
-        typedefOfTagged typedefInfo payload $
-          DeclUseGraph.getUseSitesNoSelfReferences declUseGraph payload.id
-
--- | Typedef around tagged payload
-typedefOfTagged ::
-     C.DeclInfo ResolveBindingSpecs    -- ^ Typedef info
-  -> TaggedPayload                     -- ^ Payload
-  -> [(DeclId, usage)]                 -- ^ Use sites of the payload
-  -> TypedefAnalysis
-typedefOfTagged typedefInfo payload useSites
-  | shouldSquash
-  = mconcat [
-        conclude typedefInfo.id $ Squash $ SquashTypedef {
-            typedefLoc = typedefInfo.loc
-          , targetId   = payload.id
-          }
-      , conclude payload.id $ UseNameOf typedefInfo.id
-      ]
-
-  | shouldAddSuffix
-  = conclude payload.id $ AddSuffix "_Aux"
-
-  | otherwise
-  = mempty
-  where
-    shouldSquash, shouldAddSuffix :: Bool
-    shouldSquash = and [
-          payload.isDirect
-        , or [ typedefInfo.id.name.text == payload.id.name.text
-             , length useSites == 1
+    case directPayload of
+      -- The typedef is a direct alias of a tagged type: a candidate for
+      -- squashing.
+      Just payload
+        | shouldSquash payload
+        -> mconcat [
+               conclude typedefInfo.id $ Squash $ SquashTypedef {
+                   typedefLoc = typedefInfo.loc
+                 , targetId   = payload.id
+                 }
+             , conclude payload.id $ UseNameOf typedefInfo.id
              ]
+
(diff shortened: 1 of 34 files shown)


jorisdral:
I'm going to work on https://github.com/well-typed/hs-bindgen/issues/1520 soon. Do you think that overlaps a bit with your work here?

dschrempf:
@jorisdral Ah yes, I think so. I have done quite some work on #1445 now, and I think it includes many of those cases. I didn't know of #1520.

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/Typedefs.hs
```diff
@@ -204,27 +241,41 @@ data TaggedPayload = TaggedPayload{
...
-        C.TypeEnum ref       -> typeRef direct ref.name
-        C.TypePointers _n ty -> go False ty
+        C.TypeRef declId         -> typeRef direct declId
+        C.TypeEnum ref           -> typeRef direct ref.name
+        C.TypePointers _n ty     -> go False ty
+        C.TypeConstArray _n ty   -> go False ty
+        C.TypeIncompleteArray ty -> go False ty
+        C.TypeBlock ty           -> go False ty
+        C.TypeFun args res       ->
+          concatMap (go False . (.typ)) args ++ go False res
```
jorisdral, line 284: Should this also recurse into type qualifiers? Something like:

```hs
    TypeQual _qual ty -> go False ty
```

dschrempf: I guess, yes, that should be the case... I will have a look, and also think about a test case.

dschrempf: Yes, this was definitely a bug; fixing atm.

dschrempf: I have fixed this discrepancy; it also affected two other fixtures in a positive way (we now squash in cases, we did not squash erronously).

edsko:
Let's make sure that this fixes https://github.com/well-typed/hs-bindgen/issues/2020 also.

edsko:
@dschrempf See https://github.com/well-typed/hs-bindgen/issues/1184#issuecomment-4751300963 .

dschrempf:
Superseded by #2089 .

PR #2065, opened 2026-06-17
`HasField`: why we can't get rid of `(~)`

jorisdral:
The `Elem` type family can otherwise appear in class instance heads, which is not valid Haskell. We also can't expand the type family because instances can be defined by users as part of an external binding specification, because of which we should consider the type to be "opaque" in a way


jorisdral:
This PR just intended to show that we can't get rid of `(~)`, so closing it now

edsko:
Might be nice to add a comment somewhere in the manual?

jorisdral:
> Might be nice to add a comment somewhere in the manual?

Done in #2082

PR #2073, opened 2026-06-18
Generate StaticSize instances for non-opaque emptydata types

### LINKED ISSUE #2014: Generate `StaticSize` instances for `emptydata` types that are not opaque in C
Some C `struct`s have fields that should not be exposed in the public API, and those fields may have types that pull internal dependencies that should not be exposed.  If the `struct` is mean to be allocated on the stack in C, then there may not be functions to allocate and free the `struct` in the API.

When generating Haskell bindings, using representation `emptydata` makes the Haskell representation opaque, so that it can only be used via a pointer, as well as makes it possible to omit/exclude the unwanted dependencies.  Initialization is currently problematic, however.  We cannot allocate within Haskell because we do not know the size and alignment of the type.  To allocate in C, one would need to write C helper functions and generate bindings for those.

The `StaticSize` class methods use proxies, so it is possible for an `emptydata` type to have an instance.  Knowing the size and alignment would enable users to manage the memory in Haskell, without having to write C helper functions.

Proposal:

* When a C type is opaque, we do not know the size and alignment, so we cannot generate a `StaticSize` instance for the corresponding Haskell type.
* We should generate `StaticSize` instances for all other `emptydata` types, by default.
* Users may omit the instance via a prescriptive binding specification in cases where it is not wanted.  The binding specification design/syntax will support this without change.

    ```yaml
    ctypes:
      - headers: foo.h
        cname: struct lexer
        hsname: Lexer

    hstypes:
      - hsname: Lexer
        representation: emptydata
        instances:
          - omit: StaticSize
    ```

bolt12:
Closes #2014


hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation.hs
```diff
@@ -262,6 +264,32 @@ opaqueDecs haddockConfig info spec = do
...
+                }
+            ]
+          )
+      | otherwise = (Set.empty, [])
+
+    staticSizeOmitted :: Bool
+    staticSizeOmitted =
+      case spec.hsSpec >>= \hs -> Map.lookup Inst.StaticSize hs.instances of
+        Just BindingSpec.Omit -> True
+        _otherwise            -> False
```
bolt12: Not necessary!

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/ResolveBindingSpecs.hs
```diff
@@ -295,16 +295,20 @@ applyPrescriptive decl cTypeSpec = \case
...
-            C.DeclOpaque{}    -> True
+      -- A complete C type (struct, union, enum) keeps its size and alignment, so
+      -- that a 'StaticSize' instance can be generated for the empty data type.
+      -- Types that are opaque in C, and kinds whose size is not directly
+      -- available (typedefs, macro types), keep 'Nothing'.
+      let (isValid, mSize) = case decl.kind of
+            C.DeclStruct s    -> (True,  Just (C.OpaqueSize s.sizeof s.alignment))
+            C.DeclUnion  u    -> (True,  Just (C.OpaqueSize u.sizeof u.alignment))
+            C.DeclEnum   e    -> (True,  Just (C.OpaqueSize e.sizeof e.alignment))
+            C.DeclTypedef{}   -> (True,  Nothing)
```
bolt12: Get the underlying type here and for macros!

hs-bindgen/test-artefacts/fixtures/binding-specs/rep/emptydata/omit_staticsize/bindingspec.yaml
bolt12, line 1: lets actually check that the type is generated for typedefs and macros (essentially every place in the code where we have a `Just (OpaqueSize mSize)` value.

edsko (approved):

PR #2074, opened 2026-06-19
IR refactoring: Pass, C

TravisCardwell:
This PR refactors the pass and C AST, putting them under `HsBindgen.IR`.  These submodules should *never* (transitively) import from `Frontend` or `Backend`.

As we translate Haskell types in the frontend, `Pass` is more and more not specific to C.  `IsPass` is split up into separate modules, so that we can avoid import cycles by importing only what is needed, when necessary.  Module `HsBindgen.IR.Pass` re-exports, so modules outside of `IR` can just import it.

This is almost exactly as I would have implemented `Types` refactoring, just renaming `Types` to `IR`.  The only difference is `HashIncludeArg`, which I probably would have put outside of the `C` hierarchy, but it definitely makes sense in the `C` hierarchy since it is a C concept.

All `HsBindgen.IR.C` submodules are designed to import `qualified as C`, for consistency.  I anticipated that others may want a `HsBindgen.IR.C` that re-exports everything and experimented with it, but IMHO it is not worth it.  I think it is better to explicitly import only the needed modules and decrease likelihood of running into import cycles.

Making this work involved moving some things:

* The `CoercePass` class is now defined alongside `Pass`, and instances are defined in relevant modules instead of a single `Coerce` pass.  This helps avoid import issues, and it will work better as we move Haskell translation into the frontend.

* The `ResolvedExtBinding` type is moved into `HsBindgen.BindingSpec`, where it fits in nicely and already has the needed imports.

* The `referencesUntagged` function is moved into `HsBindgen.Bindgen.Hs.Translation`, where it is used, so that we can avoid undesired imports.

* The `TranslationUnit` is moved to `HsBindgen.Frontend.TranslationUnit`.  It is not really part of the C IR; it is more of a `FrontendState`.  It depends on many things that I assume we want to keep in the frontend.

* The `Deps` modules is moved to `HsBindgen.Frontend.Deps`.  It stays in the frontend because it is implemented in terms of specific passes.


TravisCardwell:
Changes:

* Moved `Deps` module under `Analysis`
* Refactored to *always* import `HsBindgen.IR.C qualified as C` (not from submodules) when outside of `HsBindgen.IR.C`
* Added/updated usage instructions to the `IR` modules
* Added (IR) import rules to Cabal file
* Added `includeMacroArg` example
* Renamed `AMsg` to `AnnMsg`

Issue #1977, opened 2026-05-14
Nightly CI reports: Examples


jorisdral:
The [Examples](https://github.com/well-typed/hs-bindgen/actions/workflows/examples.yml) workflow is run nightly at 18:30 UTC. GitHub Actions does not support enabling notifications for scheduled workflows that fail, but users can subscribe to notifications from issues. So if any jobs in this workflow fail, the results will be reported here, and anyone who is subscribed to this issue will get a notification.

NOTE: this issue should not be closed

jorisdral:
> Nightly CI for the Examples workflow failed. See the results at https://github.com/well-typed/hs-bindgen/actions/runs/35787068409.

Nothing stands out about the build failure. There is too little information in the log to see what went wrong. Might just be a transient failure. So I'll let this fly for now and if it fails again tonight I'll investigate further

PR #2076, opened 2026-06-19
`CI`: fix `Examples` failures

jorisdral:
Let's merge this after the weekend so that we can first see if the notifications for failing nightly CI work

### DIFF
--- a/examples/botan/hs-project/cabal.project
+++ b/examples/botan/hs-project/cabal.project
@@ -1,4 +1,4 @@
 import: ../../../cabal.project.base
 
 packages: .
-          ../../../hs-bindgen-runtime,
+          ../../../hs-bindgen-runtime


examples/libpcap/hs-project/libpcap.cabal
```diff
@@ -35,6 +35,7 @@ library
   build-depends:
     , c-expr-runtime
     , hs-bindgen-runtime
+    , bytestring
```
jorisdral: @dschrempf this looks to be a dependency incurred by `c-expr`. Do you think it's okay to add the dependency here? Or should `ByteString` be re-exported by a `hs-bindgen-runtime` prelude?

jorisdral: Opened #2078 to prevent needing to directly depend on `bytestring`

jorisdral:
> Let's merge this after the weekend so that we can first see if the notifications for failing nightly CI work

Confirmed that the workflow failed and notifications were posted at https://github.com/well-typed/hs-bindgen/issues/1977#issuecomment-4763002282. Time to merge this PR!

PR #2075, opened 2026-06-19
Provide `Compat.HasField` instances for structs

jorisdral:
Resolves #2059

PR #2087, opened 2026-06-23
Provide `Compat.HasField` instances for `typedef`s, `enum`s, and macro types

### LINKED ISSUE #2083: Provide `Compat.HasField` instances for `typedef`s, `enum`s, and macro types

jorisdral:
Resolves #2083


hs-bindgen/src-internal/HsBindgen/Instances.hs
jorisdral, line 1: @TravisCardwell same as https://github.com/well-typed/hs-bindgen/pull/2075#discussion_r3452987979, did I do this right?

TravisCardwell: I did not see https://github.com/well-typed/hs-bindgen/pull/2075#discussion_r3452987979!  Sorry about that!

This looks fine to me.  The `SupportedInstances` configuration is correct, and the implementation follows the current status quo.

(When we move instance resolution into the frontend, prescriptive binding specifications will be taken into account correctly.  Current `knownInsts` does not do this, so it is not yet possible to `omit: CompatHasField`.)

Issue #1925, opened 2026-04-23
Conditional generation of auxiliary names in the name mangler may diverge from the requirements in the backend


dschrempf:
After #1922, we generate all names in the name mangler.

However, some (auxiliary) names we do not always require. For example, we may not require the name of the auxiliary typedef of a structure with a flexible array member, because the structure doesn't contain one.

The status quo is that check the condition independently in the name mangler and the backend.

- Is this a FLAM? If yes, create the auxiliary name.
- Is this a FLAM? If yes, generate code for the auxiliary typedef _assuming_ the name for it was generated in the name mangler.

Instead, we should tie these two checks together.

For example, for FLAMS we maybe store a `StructField`

```haskell
data Struct p = Struct {
    ...
    , flam      :: Maybe (StructField p) -- ^ FLAM element type, if any
    ...
    }
```

We should enrich this field by the names it requires.

This is easy for FLAMs, but probably more involved for typedefs to function pointers.

dschrempf:
PR #2089 will solve the FLAM-part of this issue. For the typedef part, we need to assure that the AST reflects if we need an auxiliary name for a typedef, or not. Then, we can use an annotation, like we do for FLAMs after #2089.

See TODO comment in the code.

PR #2089, opened 2026-06-25
Name mangler: split into three traversals

### LINKED ISSUE #1432: Think about name clashes arising from the introduction of `_Aux` types

### LINKED ISSUE #1445: Avoid name clashes between tagged type and surrounding typedef

### LINKED ISSUE #1552: Suggested example: Struct with field for which name mangler fails

### LINKED ISSUE #1734: Name mangler does not detect all name collisions yet

### LINKED ISSUE #2020: Enum tag clashes with enumerator name

dschrempf:
Supersedes #2043, using three traversals. Details will follow soon.

Closes #1734 #1432 #1445 #2020 #1552

Fixes the FLAM part of #1925

Related issue: #1520

---

```hs
-- TODO <link to issue>
-- Brief description
```

### DIFF
--- a/hs-bindgen/CHANGELOG.md
+++ b/hs-bindgen/CHANGELOG.md
@@ -40,6 +40,27 @@
   and supply the new `resolve` field. Error types are consolidated in
   `HsBindgen.Macro.Error`; `MacroTypecheckResult` is renamed `TypecheckResult`
   with shortened constructors.
+* Tagged types that would clash with the name of an enclosing typedef are now
+  disambiguated with a `_struct`/`_union`/`_enum` suffix (mirroring the C
+  keyword) instead of the previous `_Aux`. Clash detection now also covers
+  arrays, blocks, and function types (argument and result types), not just
+  pointers, so e.g. `typedef struct {…} foo[10];` no longer produces clashing
+  names. Qualifiers (e.g. `const`) are treated as transparent, so
+  `typedef const struct {…} foo;` is now squashed like its unqualified
+  counterpart. This renames affected generated identifiers. See
+  [#1445](https://github.com/well-typed/hs-bindgen/issues/1445).
+* The `MangleNames` pass now checks names minted in its second pass (data
+  constructors, record-field selectors, union getters/setters, enum constants,
+  and `_Aux` types for FLAMs and function pointers) for collisions. A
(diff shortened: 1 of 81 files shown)


dschrempf:
Finally pulling the plug; fingers crossed!

PR #2107, opened 2026-07-05
Add example with multi-line `COMPLETE` pragma

TravisCardwell:
This is a trivial PR that just adds a `closed` `enum` example that results in a mutli-line `COMPLETE` pragma.  I figure that it is good to have this in our fixtures so that any changes in how this is rendered will be explicitly detected.


TravisCardwell:
Merges are failing due to an Ubuntu server 404.  I will try again later.

Issue #2080, opened 2026-06-19
Expose the topological header order


bolt12:
`IncludeGraph.toSortedList` / `getDependencies` already compute the topological order and are used internally (Select pass, TH path), but the CLI only emits an alphabetically-ordered Mermaid graph. Adding a --toposort flag would be nice!

edsko:
@bolt12 If you have a specific use case for this, feel free to self-assign.

PR #2123, opened 2026-07-09
Add libgit2 high-level bindings example

bolt12:
This PR adds three worked examples under `examples/`: libgit2, libsodium, and libclang-ffi. Each is a high-level binding hand-written with the `ToHighLevel` combinators over a C library of a deliberately different shape (a handle-based API, a stateless buffer API, and a raw by-value FFI), plus small programs written twice, once against the low-level layer and once against the high-level one, so the two styles sit side by side. A good chunk of the gaps below have already been fixed in `hs-bindgen-runtime` as a result; the checklists mark what has landed (`[x]`) and what is still open (`[ ]`).

---

## examples/libgit2

Low-level bindings to libgit2 generated by hs-bindgen, plus a hand-written high-level layer, and two programs written twice. The bindings were generated in topological header order; the `include-graph` hs-bindgen option only gives a mermaid graph, which I had to `tsort` (this can be improved, see #2080).

libgit2 is handle-based. Almost every function takes an opaque handle (`git_repository *`, `git_commit *`) or produces one through an out-parameter, and each handle is acquired and released by a pair: `git_repository_open(git_repository **out, ...)` hands you a handle and `git_repository_free` releases it. The high-level layer maps that onto Haskell:

- each opaque handle becomes a newtype over `ForeignPtr` with the C `_free` installed as the finalizer, so handles free themselves at GC and the user never calls free by hand;
- errors become a `GitError` exception (the wrapper reads libgit2's thread-local `git_error_last` on failure);
- there is no state-threading monad. [`withLibgit2`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/examples/libgit2/hs-project/src/LibGit2/Git.hs#L28-L34) runs an action on a bound thread (libgit2's error state is thread-local) and brackets the `git_libgit2_init` / `git_libgit2_shutdown` pair, and `withRepository` opens a repo and hands the handle to a callback, so callers write plain `IO`.

The high-level `git log` is 19 lines against 81 low-level. What the example surfaced:

- [x] **Recognise managed handles.** A pointer-to-a-pointer out-parameter with a matching `_free` is always the same shape. The runtime now offers `outForeignPtr`, which allocates the `git_X **out` slot, peeks the `git_X *`, and wraps it in a `ForeignPtr` with a caller-supplied `_free` finalizer; the example's `outHandle` collapses to `fromFP <$> outForeignPtr fin`, and the `Handle` class stays only to map the raw `ForeignPtr` into each handle newtype. (Its `CRep` is now injective, so `newHandle` can bake in the strip below and still infer the handle type from the finalizer alone.)
- [x] **Ptr / PtrConst conversion.** libgit2 often takes a `const` pointer where a sibling takes a non-const one, which meant two marshallers per handle. The [`asConstArg` / `asMutableArg`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/hs-bindgen-runtime/src/HsBindgen/Runtime/HighLevel/Marshaller.hs#L125-L139) pair now lives in the runtime; the example imports it so one definition serves both (e.g. `handleInC = asConstArg handleIn`).
- [x] **Enum default instances.** The runtime gained a `DefaultIn` [identity catch-all](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/hs-bindgen-runtime/src/HsBindgen/Runtime/HighLevel/Defaults.hs#L149-L150), so an enum that is a newtype over a C integer now passes through `auto` with no instance. (See https://github.com/well-typed/hs-bindgen/issues/2129)
- [x] **The trailing `()`.** The runtime's `dropTrailingUnit` strips a void closer's trailing `()` as a forward-only pass at the head of the spec. It cannot be folded into the injective `Cons` family (a drop-the-unit clause breaks the output-slot inference the `()` disambiguates; a reverse fundep is either rejected or unsound), so it stays a separate pass. `newHandle` bakes it in, so every constructor drops its `fst`.

---

## examples/libsodium

[libsodium](https://libsodium.org/) is stateless and buffer-oriented, a good contrast with libgit2's handles: generated low-level bindings, a hand-written high-level layer, and two programs (secret-key encryption and Ed25519 signatures) written once against each layer with byte-identical output.

Almost every function writes into a caller-allocated output buffer and reads fixed-size keys and nonces from `const unsigned char *`. The high-level layer maps that onto Haskell:

- keys, nonces, and signatures become newtypes over `ByteString` with length-checked smart constructors (`Key`, `Nonce`, `PublicKey`, `SecretKey`, `Signature`, `Seed`);
- errors are hybrid: a `SodiumError` exception for setup and key generation, but `Maybe`/`Bool` for authentication and verification, since a rejected ciphertext or signature is expected control flow, not an exception;
- there is no monad and no `-threaded`. The library is stateless with no thread-local error state, so a bare [`sodiumInit :: IO ()`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/examples/libsodium/hs-project/src/LibSodium/Init.hs#L22-L25) called once at startup is the whole lifecycle, a nice contrast with libgit2's init/shutdown bracket and bound-thread requirement.

The high-level `secretbox` is 43 lines against 85 low-level; `sign` is 65 against 127. 

- [x] **A `fixed` combinator** for a constant/persistent C argument (`fixed c = scratch (\k -> k c)`), needed for `randombytes_buf`'s size and the multipart signing state. Prototyped here, now in the [runtime](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/hs-bindgen-runtime/src/HsBindgen/Runtime/HighLevel.hs#L199-L201); it's also the combinator libgit2 wanted for its NULL arguments.
- [x] **A sized byte-buffer output combinator.** `auto` fills every fixed-size key/nonce input but can't fill the output buffer every encrypt/sign/keygen writes, because `DefaultOut` only covers scalars. [`byteStringOut`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/hs-bindgen-runtime/src/HsBindgen/Runtime/HighLevel/Marshaller/Utils.hs#L109-L111) (polymorphic in the pointer element) is now in the runtime and the example imports it.
- [x] **A `ByteString` input marshaller that fits.** libsodium messages are `(const unsigned char *, unsigned long long)`, which the stock `useAsByteStringLenIn` (`PtrConst CChar`, `CSize`) doesn't match. `constByteStringLenIn` now lives in the runtime and the example uses it via `input2`; a `DefaultIn ByteString` so `auto` reaches it would be the remaining nicety.
- [x] **A closer that maps a status to a value.** The runtime now has [`throwOn` / `throwOnOut`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/hs-bindgen-runtime/src/HsBindgen/Runtime/HighLevel.hs#L283-L314), an `Either`-based classifier with no `Num` requirement, so [`open`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/examples/libsodium/hs-project/src/LibSodium/SecretBox.hs#L123-L136) and [`verifyDetached`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/examples/libsodium/hs-project/src/LibSodium/Sign.hs#L145-L154) classify status to `Maybe`/`Bool` in one step.
- [x] **A purity combinator.** libsodium's crypto primitives are deterministic functions of their inputs, so [`assertPure`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/hs-bindgen-runtime/src/HsBindgen/Runtime/HighLevel/Result.hs#L12-L37) drops the `IO` from a wrapper over such a call, and [`LibSodium.Pure`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/examples/libsodium/hs-project/src/LibSodium/Pure.hs) mirrors every operation as a pure function (`encrypt :: Key -> Nonce -> ByteString -> ByteString`, and so on).
- [x] **The trailing `()`, again.** Fixed by the same `dropTrailingUnit` (see libgit2); `newKey`, `encrypt`, and `randomBytes` drop their `fst`. (Value-closer wrappers like `open`, which keep a real status, are untouched.)

---

## examples/libclang-ffi

Instead of generating fresh bindings, this one writes the high-level layer over libclang's *raw* FFI and compares against a high-level layer that already exists. `libclang-bindings` ships two low-level layers: `Clang.LowLevel.FFI`, the raw `foreign import`s (structs passed by value, `Ptr` out-parameters, raw `FunPtr` callbacks, status enums), and `Clang.LowLevel.Core`, a high-level API someone hand-wrote on top of it. The AST-walk tutorial (parse a header, walk the cursor tree printing each cursor's kind, spelling, and `line:col`) is written twice, against the raw FFI (`app/WalkLow.hs`) and the wrapper layer (`app/WalkHigh.hs`), with byte-identical output.

libclang passes its structs (cursors, types, locations, strings) *by value* and hides everything else behind opaque handles:

- Where every argument is lifted, the combinators reproduce `Core` almost line for line. ([`parseTU`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/examples/libclang-ffi/hs-project/src/ClangFFI/Wrappers.hs#L80-L93), [`spellingLineCol`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/examples/libclang-ffi/hs-project/src/ClangFFI/Wrappers.hs#L153-L167)).
- Where an argument is by-value, the combinators now reach it through the representation-polymorphic `bracketUnlifted` / `outputUnlifted` (an `R` / `W` struct pointer is *unlifted*, and the vocabulary is now polymorphic over that). The cursor/string/location functions are full specs; only `clang_visitChildren` still needs the C trampoline.
- The visitor callback needs a small C trampoline to turn the by-value cursor into a pointer ([`hs_visit.c`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/examples/libclang-ffi/hs-project/cbits/hs_visit.c)), but on the Haskell side it reuses the runtime's generated `ToFunPtr` and so avoids the hand-written `foreign import ccall "wrapper"` that `Core` carries ([`visitChildren`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/examples/libclang-ffi/hs-project/src/ClangFFI/Wrappers.hs#L193-L209)). This is the one place the combinator layer beats the hand-written one.

The high-level walk is 39 lines against 118.

- [x] **A status closer not tied to `Num`.** `throwOnNonZero` needs `Num c`, but the parse status is a `SimpleEnum (Maybe CXErrorCode)`. The runtime's new `throwOn` covers it (an `Either`-based classifier with no `Num` requirement); the local [`statusClose`](https://github.com/well-typed/hs-bindgen/blob/5212e938293c3c572d28c104596c30f5a23f331a/examples/libclang-ffi/hs-project/src/ClangFFI/Wrappers.hs#L106-L110) predates it and could be swapped for it. This is the same closer libsodium wanted.
- [x] **A string-array input marshaller.** `withCStringArrayIn` (a `[String] -> const char *const *` marshaller) and a matching `DefaultIn [String]` now live in the runtime; `parseTU` uses `input withCStringArrayIn` and the hand-rolled `argvIn` is gone.
- [x] **Callbacks with domain types.** `withFunPtrAs` (and the marshaller form `funPtrInAs`) now live in the runtime: they `coerce` a domain-typed callback onto a `ToFunPtr`-covered signature. The local `ClangFFI.Callback` prototype is deleted and `visitChildren` imports the runtime one. (A generated binding has a per-callback `ToFunPtr` and uses `withFunPtr` directly; this bridge is for the not-generated case.)
- [x] **An unlifted-aware marshaller vocabulary.** Done: `Unmarshaller` and `ThreadOut` are now representation-polymorphic in the C-side type (a free kind generalization, since they never bind the C value), and the runtime's `HighLevel.Unlifted` adds `outputUnlifted` / `bracketUnlifted` for the two spots that must bind it. `cxStringOut :: Unmarshaller (W CXString_) Text` now exists, and all six cursor/location wrappers are full `toHighLevel` specs instead of hand-rolled `onHaskellHeap` / `preallocate_`. The one limit: an unlifted output must sit in tail position (a by-value out-parameter always does).
- [x] **The trailing `()`.** The same `dropTrailingUnit`: `parseTU` and `spellingLineCol` use it, and the local `dropUnit` is gone.


examples/libgit2/hs-project/src/LibGit2/Error.hs
```diff
@@ -0,0 +1,69 @@
...
+  | n < 0     = throwIO =<< gitError n
+  | otherwise = pure ()
+
+-- | A 'ToHighLevel' closer that consumes the C @int@ status: throws on failure,
+-- yields @()@ on success. Closes most fallible libgit2 wrappers.
+--
+-- It is 'resultIO', not the runtime's 'HsBindgen.Runtime.HighLevel.throwOnNonZero':
+-- the error /detail/ is thread-local state reached through an 'IO' call
+-- (@git_error_last@), which a pure @c -> e@ classifier cannot read.
+checkStatusResult :: ToHighLevel (IO CInt) (IO ())
```
bolt12: Experiment again with having a newtype `Trivial` that can be erased when flattening the tuples in the return type. This will clear up the `fst` calls

examples/libgit2/hs-project/app/CommitHigh.hs
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+main :: IO ()
+main = do
+  args <- getArgs
+  let path = case args of { (p : _) -> p; [] -> "/tmp/libgit2-commit-high" }
+  removePathForcibly path
+  withNewRepository path $ do
+    blobOid <- inRepo (\r -> blobCreateFromBuffer r (BS8.pack "hello, libgit2\n"))
+    treeOid <- inRepo $ \r -> do
```
bolt12: move the r to the top level

```suggestion
  withNewRepository path $ \r do
    blobOid <- blobCreateFromBuffer r (BS8.pack "hello, libgit2\n")
    treeOid <- do
```

examples/libgit2/hs-project/app/CommitHigh.hs
```diff
@@ -0,0 +1,47 @@
...
+    cOid <- inRepo (\r -> commitCreate r "HEAD" demoSig demoSig "Initial commit\n" tree)
+    -- read the commit back
+    commit <- inRepo (\r -> commitLookup r cOid)
+    msg    <- liftIO (commitMessage commit)
+    author <- liftIO (commitAuthor commit)
+    liftIO $ do
+      T.putStrLn ("commit  " <> oidToHex cOid)
+      T.putStrLn ("tree    " <> oidToHex treeOid)
+      T.putStrLn ("author  " <> sigName author <> " <" <> sigEmail author <> ">")
+      T.putStr   ("message " <> msg)
```
bolt12: remove the liftIO

examples/libgit2/hs-project/src/LibGit2/Repository.hs
```diff
@@ -0,0 +1,45 @@
...
+import Generated.Repository.Safe qualified as RS
+import LibGit2.Error (checkStatusResult)
+import LibGit2.Marshal (borrowedText, handleIn, outHandle, textIn)
+import LibGit2.Types (Reference, Repository)
+
+-- | @git_repository_open@: open the repository at @path@.
+repositoryOpen :: Text -> IO Repository
+repositoryOpen path =
+    fst <$> toHighLevel
+      ( output (outHandle RF.git_repository_free)  -- git_repository **out
```
bolt12: come up with a combinator specific to libgit2 that abstracts over the outHandle pattern and allows to use auto more effectively

examples/libsodium/hs-project/app/SecretBoxHigh.hs
bolt12, line 1: This libsodium example might be a good place to introduce a `toHighLevelPure` so we can offer pure versions of the crypto primitives.

examples/libclang-ffi/hs-project/src/ClangFFI/Callback.hs
bolt12, line 1: remove this since `withFunPtrAs` is implemented already on hs-bindgen

examples/libclang-ffi/hs-project/src/ClangFFI/Lifted.hs
```diff
@@ -0,0 +1,70 @@
...
+-- 'ByteString' is at the call site (done immediately, so the borrow is valid).
+getFileContents :: CXTranslationUnit -> CXFile -> IO ByteString
+getFileContents tu file = do
+    (n, p) <-
+      toHighLevel
+        (input (scalar id) $ input (scalar id) $ output defaultOut $ resultPure id)
+        nowrapper_getFileContents
+        tu
+        file
+    BS.packCStringLen (unConstPtr p, fromIntegral (n :: CSize))
```
bolt12: This can be cleaned up and written in a better, simpler manner.

examples/libclang-ffi/hs-project/src/ClangFFI/Wrappers.hs
```diff
@@ -0,0 +1,218 @@
...
+-- | @clang_disposeTranslationUnit@.
+disposeTranslationUnit :: CXTranslationUnit -> IO ()
+disposeTranslationUnit = toHighLevel auto nowrapper_disposeTranslationUnit
+
+-- | @clang_parseTranslationUnit2@: 'output' for the out-parameter translation
+-- unit, 'withCStringIn' for the source path, 'argvIn' for the argument array, and
+-- 'fixed' for the arguments we pin. 'statusClose' classifies the result.
+parseTU :: CXIndex -> FilePath -> [String] -> IO CXTranslationUnit
+parseTU idx path args =
+    fst <$> toHighLevel
```
bolt12: We should use dropTrailingUnit here instead.

examples/libclang-ffi/hs-project/src/ClangFFI/Wrappers.hs
```diff
@@ -0,0 +1,218 @@
...
+              nowrapper_parseTranslationUnit2 idx path args
+  where
+    -- @[String]@ to a @const char *const *@. There is no ready-made marshaller for
+    -- a string array, so build one from the 'Marshal' constructor; the length is a
+    -- separate 'fixed' at the call site.
+    argvIn :: Marshal [String] (ConstPtr (ConstPtr CChar) -> lo') lo'
+    argvIn = Marshal $ \as lo k ->
+        withMany withCString as $ \cstrs ->
+          withArray (map ConstPtr cstrs) $ \arr ->
+            k (lo (ConstPtr arr))
```
bolt12: hs-bindgen already offers this marshaller.

examples/libclang-ffi/hs-project/src/ClangFFI/Wrappers.hs
```diff
@@ -0,0 +1,218 @@
...
+              ( scratch alloca    -- CXFile *file   (ignored)
+              $ output defaultOut -- unsigned *line
+              $ output defaultOut -- unsigned *column
+              $ scratch alloca    -- unsigned *offset (ignored)
+              $ discardResult
+              )
+              (wrap_getSpellingLocation r)
+  where
+    dropUnit :: (a, b, ()) -> (a, b)
+    dropUnit (a, b, ()) = (a, b)
```
bolt12: We can use the unlifted version of the marshallers to abstract over this pattern.

bolt12: use `dropTrailingUnit`

examples/libgit2/hs-project/src/LibGit2/Commit.hs
```diff
@@ -0,0 +1,60 @@
...
+import Generated.Commit.Safe qualified as CS
+import LibGit2.Marshal (borrowedOid, borrowedScalar, borrowedText, handleIn,
+                        handleInC, newHandle, oidInC, peekTextConst)
+import LibGit2.Signature (peekSignatureConst)
+import LibGit2.Types (Commit, Oid, Repository, Signature)
+
+-- | @git_commit_lookup@: the commit object for @oid@.
+commitLookup :: Repository -> Oid -> IO Commit
+commitLookup repo oid =
+    fst <$> newHandle CF.git_commit_free (input handleIn . input oidInC)
```
bolt12: Use dropTrailingUnit here.

examples/libgit2/hs-project/src/LibGit2/Error.hs
```diff
@@ -0,0 +1,69 @@
...
+gitError :: CInt -> IO GitError
+gitError code = do
+  ep <- git_error_last
+  let p = PtrConst.unsafeToPtr ep
+  if p == nullPtr
+    then pure (GitError (fromIntegral code) 0 (T.pack "no libgit2 error detail"))
+    else do
+      Git_error msgPtr klass <- peek p
+      msg <- if msgPtr == nullPtr then pure T.empty else T.pack <$> peekCString msgPtr
+      pure (GitError (fromIntegral code) (fromIntegral klass) msg)
```
bolt12, line 58: Can't we specify this using high-level combinators, maybe leverage the Maybe marshal type for pointers.

examples/libgit2/hs-project/src/LibGit2/Marshal.hs
```diff
@@ -0,0 +1,197 @@
...
+  p <- peek pp
+  fromFP <$> newForeignPtr fin p
+
+-- | Build a managed-handle constructor. @newHandle fin inputs cfn@ fills the
+-- @git_X **out@ slot (freed by @fin@), applies the caller's @inputs@ chain, and
+-- checks the libgit2 status. The result carries a trailing @()@ from the status
+-- check, dropped at the call site with @fst@:
+--
+-- > repositoryOpen path =
+-- >   fst <$> newHandle git_repository_free (input textIn) git_repository_open path
```
bolt12: this should mention dropTrailingUnit

bolt12: should it use it by default? Does it affect inference?

examples/libgit2/hs-project/src/LibGit2/Marshal.hs
```diff
@@ -0,0 +1,197 @@
...
+-------------------------------------------------------------------------------}
+
+-- | Marshal 'Text' as a NUL-terminated @const char *@.
+textIn :: Marshal Text (PtrConst CChar -> lo) lo
+textIn = at T.unpack withCStringIn
+
+-- | Marshal 'Text' as a NUL-terminated non-@const@ @char *@ (e.g. a
+-- @git_signature@ name/email field).
+textInPtr :: Marshal Text (Ptr CChar -> lo) lo
+textInPtr = bracket (\t k -> withCString (T.unpack t) k)
```
bolt12: Why aren't these default marshallers defined in hs-bindgen?

examples/libgit2/hs-project/src/LibGit2/Marshal.hs
```diff
@@ -0,0 +1,197 @@
...
+-- left polymorphic so it unifies with @void@ at the call site). The zero-copy
+-- counterpart of the runtime's @constByteStringLenIn@: 'BSU.unsafeUseAsCStringLen'
+-- passes the buffer without copying, so it is not NUL-terminated and is valid
+-- only during the call and only alongside the length (an empty input may alias a
+-- shared buffer). Safe here because the callee honours the length and does not
+-- retain the pointer.
+bufferIn :: Marshal ByteString (PtrConst void -> CSize -> lo) lo
+bufferIn = Marshal $ \bs lo k ->
+  BSU.unsafeUseAsCStringLen bs $ \(p, n) ->
+    k (lo (PtrConst.unsafeFromPtr (castPtr p)) (fromIntegral n))
```
bolt12: Same here, why isn't this a default marshaller already defined in hs-bindgen?

examples/libgit2/hs-project/src/LibGit2/Marshal.hs
```diff
@@ -0,0 +1,197 @@
...
+borrowedText :: Handle h => (PtrConst (CRep h) -> IO (PtrConst CChar)) -> h -> IO Text
+borrowedText = toHighLevel (input handleInC $ resultIO peekTextConst)
+
+-- | A @const T *@ accessor returning a borrowed @const git_oid *@, copied to 'Oid'.
+borrowedOid :: Handle h => (PtrConst (CRep h) -> IO (PtrConst Git_oid)) -> h -> IO Oid
+borrowedOid = toHighLevel (input handleInC $ resultIO peekOidConst)
+
+-- | A @const T *@ accessor returning a C scalar, converted with 'fromIntegral'.
+borrowedScalar :: (Handle h, Integral c, Num n) => (PtrConst (CRep h) -> IO c) -> h -> IO n
+borrowedScalar = toHighLevel (input handleInC $ resultPure fromIntegral)
```
bolt12: Style this code such that:

```hs
foo :: (a -> b)
    -> ...
```

examples/libgit2/hs-project/src/LibGit2/Marshal.hs
```diff
@@ -0,0 +1,197 @@
...
+-- 'asArgument' fills a non-@const@ @Ptr@, then 'asConstArg' retags it @const@.
+asArgumentC
+  :: (StaticSize s, WriteRaw s)
+  => MarshalStruct hi s
+  -> Marshal hi (PtrConst s -> lo) lo
+asArgumentC = asConstArg . asArgument
+
+-- | A NULL @const@ pointer, for optional/absent C arguments.
+nullConst :: PtrConst a
+nullConst = PtrConst.unsafeFromPtr nullPtr
```
bolt12: why aren't these already defined in hs-bindgen? look like they should be part of the API

examples/libgit2/hs-project/src/LibGit2/Object.hs
```diff
@@ -0,0 +1,27 @@
...
+import Generated.Revparse.Safe qualified as RPS
+import LibGit2.Marshal (borrowedOid, handleIn, newHandle, textIn)
+import LibGit2.Types (Object, Oid, Repository)
+
+-- | @git_revparse_single@: resolve a revision spec (e.g. @HEAD@, @main~3@, a
+-- sha) to an object.
+revparseSingle :: Repository -> Text -> IO Object
+revparseSingle repo spec =
+    fst <$> newHandle OF.git_object_free (input handleIn . input textIn)
+              RPS.git_revparse_single repo spec
```
bolt12: use dropTrailingUnit, unless newHandle can already do that implicitly

examples/libgit2/hs-project/src/LibGit2/Repository.hs
```diff
@@ -0,0 +1,35 @@
...
+
+-- | @git_repository_open@: open the repository at @path@.
+repositoryOpen :: Text -> IO Repository
+repositoryOpen path =
+    fst <$> newHandle RF.git_repository_free (input textIn) RS.git_repository_open path
+
+-- | @git_repository_head@: the resolved @HEAD@ reference.
+repositoryHead :: Repository -> IO Reference
+repositoryHead repo =
+    fst <$> newHandle RefF.git_reference_free (input handleIn) RS.git_repository_head repo
```
bolt12: use dropTrailingUnit unless newHandle can already handle that implicitly.

examples/libgit2/hs-project/src/LibGit2/Revwalk.hs
```diff
@@ -0,0 +1,66 @@
...
+import Generated.Revwalk.FunPtr qualified as WF
+import Generated.Revwalk.Safe qualified as WS
+import LibGit2.Error (checkStatusResult, gitError)
+import LibGit2.Marshal (handleIn, newHandle, oidOut)
+import LibGit2.Types (Oid, Repository, Revwalk)
+
+-- | @git_revwalk_new@: a fresh walker over @repo@.
+revwalkNew :: Repository -> IO Revwalk
+revwalkNew repo =
+    fst <$> newHandle WF.git_revwalk_free (input handleIn) WS.git_revwalk_new repo
```
bolt12: same here, use dropTrailingUnit unless newHandle can do that implicitly

examples/libgit2/hs-project/src/LibGit2/Revwalk.hs
```diff
@@ -0,0 +1,66 @@
...
+revwalkNext walk = do
+    (oid, status) <- toHighLevel
+      ( output oidOut       -- git_oid *out (peeked unconditionally)
+      $ input  handleIn     -- git_revwalk *walk
+      $ resultPure id       -- keep the raw status
+      ) WS.git_revwalk_next walk
+    case fromIntegral status :: Int of
+      0   -> pure (Just oid)
+      -31 -> pure Nothing                          -- GIT_ITEROVER
+      n   -> throwIO =<< gitError (fromIntegral n)
```
bolt12: Can we deal with the error inside the toHighLevel call?

examples/libgit2/hs-project/src/LibGit2/Signature.hs
```diff
@@ -0,0 +1,74 @@
...
+      GitTime
+  <$> unmarshalFieldPure git_time_time   fromIntegral
+  <*> unmarshalFieldPure git_time_offset fromIntegral
+
+sigUnmarshal :: UnmarshalStruct Git_signature Signature
+sigUnmarshal =
+      Signature
+  <$> unmarshalField  git_signature_name  peekText
+  <*> unmarshalField  git_signature_email peekText
+  <*> unmarshalNested git_signature_when  timeUnmarshal
```
bolt12, line 47: Is there a way we can make it easier to write unmarshallers using/inferring defaults, e.g. using auto?

examples/libgit2/hs-project/src/LibGit2/Signature.hs
```diff
@@ -0,0 +1,74 @@
...
+timeMarshal :: MarshalStruct GitTime Git_time
+timeMarshal =
+    struct Git_time
+      (     at gitTimeEpoch     (scalar fromIntegral)  -- time
+        >>> at gitTimeOffsetMin (scalar fromIntegral)  -- offset
+        >>> scalar signChar                            -- sign
+      )
+  where
+    signChar gt =
+      if gitTimeOffsetMin gt < 0 then castCharToCChar '-' else castCharToCChar '+'
```
bolt12, line 66: Is there a way we can make it easier to write marshallers using/inferring defaults, e.g. using auto?

examples/libgit2/hs-project/src/LibGit2/Types.hs
```diff
@@ -0,0 +1,132 @@
...
+newtype TreeBuilder = TreeBuilder (ForeignPtr G.Git_treebuilder)
+newtype Config      = Config      (ForeignPtr G.Git_config)
+
+-- | A high-level handle backed by a 'ForeignPtr' to an opaque libgit2 object.
+-- One method pair replaces a @withForeignPtr@ / @newForeignPtr@ wrapper for
+-- every one of the ten handle types above.
+class Handle h where
+  type CRep h
+  toFP   :: h -> ForeignPtr (CRep h)
+  fromFP :: ForeignPtr (CRep h) -> h
```
bolt12, line 72: I think this class is already defined in hs-bindgen. If not it should.

examples/libgit2/hs-project/src/LibGit2/Write.hs
```diff
@@ -0,0 +1,113 @@
...
+regularFileMode :: Git_filemode_t
+regularFileMode = GIT_FILEMODE_BLOB
+
+-- | @git_repository_init@ at @path@ (non-bare). Creates the directory.
+repositoryInit :: Text -> IO Repository
+repositoryInit path =
+    fst <$> newHandle RF.git_repository_free
+              ( input textIn         -- const char *path
+              . fixed (0 :: CUInt) )  -- unsigned is_bare = 0
+              RS.git_repository_init path
```
bolt12: use dropTrailingUnit

examples/libgit2/hs-project/src/LibGit2/Write.hs
```diff
@@ -0,0 +1,113 @@
...
+
+-- | @git_blob_create_from_buffer@: write @content@ as a blob; returns its oid.
+blobCreateFromBuffer :: Repository -> ByteString -> IO Oid
+blobCreateFromBuffer repo content =
+    fst <$> toHighLevel
+      ( output oidOut      -- git_oid *id (out)
+      $ input  handleIn    -- git_repository *repo
+      $ input2 bufferIn    -- const void *buffer, size_t len
+      $ checkStatusResult
+      ) B.git_blob_create_from_buffer repo content
```
bolt12: same here, use dropTrailingUnit.

Can we use auto?

examples/libgit2/hs-project/src/LibGit2/Write.hs
```diff
@@ -0,0 +1,113 @@
...
+
+-- | @git_treebuilder_new@ with no source tree.
+treebuilderNew :: Repository -> IO TreeBuilder
+treebuilderNew repo =
+    fst <$> newHandle TF.git_treebuilder_free
+              ( input handleIn
+              . fixed nullConst )   -- const git_tree *source = NULL
+              TS.git_treebuilder_new repo
```
bolt12: Same here, use dropTrailingUnit

examples/libgit2/hs-project/src/LibGit2/Write.hs
```diff
@@ -0,0 +1,113 @@
...
+treebuilderInsert :: TreeBuilder -> Text -> Oid -> Git_filemode_t -> IO ()
+treebuilderInsert =
+    toHighLevel
+      ( fixed  nullPtr      -- const git_tree_entry **out = NULL
+      $ input  handleIn     -- git_treebuilder *bld
+      $ input  textIn       -- const char *filename
+      $ input  oidInC       -- const git_oid *id
+      $ input  defaultIn    -- git_filemode_t filemode (DefaultIn, LibGit2.Enums)
+      $ checkStatusResult
+      ) TS.git_treebuilder_insert
```
bolt12: can we use auto here?

examples/libgit2/hs-project/src/LibGit2/Write.hs
```diff
@@ -0,0 +1,113 @@
...
+
+-- | @git_treebuilder_write@: persist the tree; returns its oid.
+treebuilderWrite :: TreeBuilder -> IO Oid
+treebuilderWrite bld =
+    fst <$> toHighLevel
+      ( output oidOut $ input handleIn $ checkStatusResult ) TS.git_treebuilder_write bld
```
bolt12: use dropTrailingUnit

examples/libgit2/hs-project/src/LibGit2/Write.hs
```diff
@@ -0,0 +1,113 @@
...
+
+-- | @git_tree_lookup@.
+treeLookup :: Repository -> Oid -> IO Tree
+treeLookup repo oid =
+    fst <$> newHandle TF.git_tree_free (input handleIn . input oidInC)
+              TS.git_tree_lookup repo oid
```
bolt12: use dropTrailingUnit, unless newHandle can do it implicitly

examples/libgit2/hs-project/src/LibGit2/Write.hs
```diff
@@ -0,0 +1,113 @@
...
+      $ input  textIn                    -- const char *update_ref
+      $ input  (asArgumentC sigMarshal)  -- const git_signature *author
+      $ input  (asArgumentC sigMarshal)  -- const git_signature *committer
+      $ fixed  nullConst                 -- const char *message_encoding = NULL
+      $ input  textIn                    -- const char *message
+      $ input  handleInC                 -- const git_tree *tree
+      $ fixed  (0 :: CSize)              -- size_t parent_count = 0
+      $ fixed  nullPtr                   -- parents = NULL
+      $ checkStatusResult
+      ) C.git_commit_create repo updateRef author committer message tree
```
bolt12: use dropTrailingUnit

examples/libgit2/README.md
```diff
@@ -0,0 +1,27 @@
+# Generate `libgit2` Haskell bindings with `hs-bindgen`
+
+This example demonstrates generating Haskell bindings for
+[libgit2](https://libgit2.org/), a portable C implementation of Git. It also
+hand-writes a high-level API over the generated bindings; see
+[FINDINGS.md](FINDINGS.md) for the write-up.
```
bolt12: get rid of the FINDINGS mentions

examples/libsodium/hs-project/src/LibSodium/Init.hs
```diff
@@ -0,0 +1,36 @@
...
+import LibSodium.Error (SodiumError (SodiumError))
+
+-- | Initialise libsodium. Idempotent and thread-safe (libsodium 1.0.11+); call
+-- once before any other operation. @sodium_init@ returns @-1@ on failure, @0@ on
+-- first success, and @1@ if already initialised, so success is @>= 0@ (a naive
+-- @== 0@ test is a latent bug).
+sodiumInit :: IO ()
+sodiumInit = do
+  r <- sodium_init
+  when (r < 0) $ throwIO (SodiumError "sodium_init" (fromIntegral r))
```
bolt12, line 25: Can we design this in such a way that sodiumInit returns a singleton value that will allow all other calls to type check. This means we can be sure the user called sodium_init before calling any other function. How cumbersome/complicated would be to refactor the code?

examples/libsodium/hs-project/src/LibSodium/Marshal.hs
```diff
@@ -0,0 +1,26 @@
...
+
+-- | A 'ByteString' as a @const unsigned char *@ (pointer only). libsodium's
+-- fixed-size arguments (key, nonce, signature, public\/secret key) carry no
+-- length: C reads a fixed number of bytes, so the caller must ensure the
+-- 'ByteString' is at least that long (a short one is an out-of-bounds read, not a
+-- logic error). Zero-copy: the pointer is live only for the call.
+bytesConstIn :: Marshal ByteString (PtrConst.PtrConst CUChar -> lo') lo'
+bytesConstIn = bracket $ \bs use ->
+  BSU.unsafeUseAsCStringLen bs $ \(p, _len) ->
+    use (PtrConst.unsafeFromPtr (castPtr p))
```
bolt12: Why isn't this marshaller already part of the hs-bindgen ToHighLevel combinators library API, as a default marshaller perhaps?

examples/libsodium/hs-project/src/LibSodium/Pure.hs
bolt12, line 1: Regarding my idea of the sodiumInit initialization function proof token, it might not be very ergonomic to use with pure functions..

examples/libsodium/hs-project/src/LibSodium/Random.hs
```diff
@@ -0,0 +1,22 @@
...
+
+-- | @n@ cryptographically random bytes from libsodium's CSPRNG
+-- (@randombytes_buf@).
+randomBytes :: Int -> IO ByteString
+randomBytes n =
+  fst <$> toHighLevel
+    ( output (byteStringOut n)          -- void *buf   (out)
+    $ fixed  (fromIntegral n :: CSize)  -- size_t size
+    $ discardResult
+    ) randombytes_buf
```
bolt12: use dropTrailingUnit

examples/libsodium/hs-project/src/LibSodium/SecretBox.hs
```diff
@@ -0,0 +1,136 @@
...
+  | BS.length bs == nonceBytes = Just (Nonce bs)
+  | otherwise                  = Nothing
+
+-- | A fresh random secret key (@crypto_secretbox_keygen@).
+newKey :: IO Key
+newKey =
+  Key . fst <$> toHighLevel
+    ( output (byteStringOut keyBytes)  -- unsigned char k[32] (out)
+    $ discardResult
+    ) crypto_secretbox_keygen
```
bolt12: use dropTrailingUnit

examples/libsodium/hs-project/src/LibSodium/SecretBox.hs
```diff
@@ -0,0 +1,136 @@
...
+randomNonce :: IO Nonce
+randomNonce = Nonce <$> randomBytes nonceBytes
+
+-- | Encrypt and authenticate @message@ under @key@ and @nonce@
+-- (@crypto_secretbox_easy@). The ciphertext is @'macBytes' + length message@
+-- bytes. Throws 'SodiumError' only on a setup failure (which @easy@ does not
+-- signal in practice).
+encrypt :: Key -> Nonce -> ByteString -> IO ByteString
+encrypt key nonce message =
+  fst <$> toHighLevel
```
bolt12: use dropTrailingUnit

examples/libsodium/hs-project/src/LibSodium/SecretBox.hs
```diff
@@ -0,0 +1,136 @@
...
+      classify <$> toHighLevel
+        ( output (byteStringOut (BS.length ciphertext - macBytes))  -- m (out)
+        $ input2 constByteStringLenIn                               -- c, clen
+        $ input  defaultIn                                          -- n
+        $ input  defaultIn                                          -- k
+        $ resultPure id                                             -- raw status
+        ) crypto_secretbox_open_easy ciphertext nonce key
+  where
+    classify (plaintext, status) =
+      if status == 0 then Just plaintext else Nothing
```
bolt12: can we move the classification to inside the toHighLevel call?

examples/libsodium/hs-project/src/LibSodium/Sign.hs
```diff
@@ -0,0 +1,204 @@
...
+-- 'signatureBytes' for Ed25519).
+signDetached :: SecretKey -> ByteString -> IO Signature
+signDetached secretKey message =
+  mkSignature <$> toHighLevel
+    ( output (byteStringOut signatureBytes)   -- sig
+    $ output (unmarshalOutPure fromIntegral)  -- siglen_p -> Int
+    $ input2 constByteStringLenIn                  -- m, mlen
+    $ input  defaultIn                        -- sk
+    $ throwOnNonZero (sodiumError "crypto_sign_detached")
+    ) crypto_sign_detached message secretKey
```
bolt12: If we'd use dropTrailingUnit we wouldn't need the ad-hoc mkSignature/mkKeyPair functions.

examples/libsodium/hs-project/src/LibSodium/Sign.hs
```diff
@@ -0,0 +1,204 @@
...
+verifyDetached :: PublicKey -> Signature -> ByteString -> IO Bool
+verifyDetached publicKey signature message =
+  classify <$> toHighLevel
+    ( input  defaultIn        -- sig
+    $ input2 constByteStringLenIn  -- m, mlen
+    $ input  defaultIn        -- pk
+    $ resultPure id           -- raw status
+    ) crypto_sign_verify_detached signature message publicKey
+  where
+    classify status = status == 0
```
bolt12: Can we use auto and move the classification to inside the toHighLevel call?

examples/libsodium/hs-project/src/LibSodium/Sign.hs
```diff
@@ -0,0 +1,204 @@
...
+signMultipart secretKey chunks =
+  alloca $ \st -> do
+    initState st chunks
+    mkSignature <$> toHighLevel
+      ( fixed  st                               -- crypto_sign_state *state
+      $ output (byteStringOut signatureBytes)   -- sig
+      $ output (unmarshalOutPure fromIntegral)  -- siglen_p -> Int
+      $ input  defaultIn                        -- sk
+      $ throwOnNonZero (sodiumError "crypto_sign_final_create")
+      ) crypto_sign_final_create secretKey
```
bolt12: Why is there an explicit alloca call here? maybe we should pass `scratch`? Also use dropTrailingUnit

examples/libsodium/hs-project/src/LibSodium/Sign.hs
```diff
@@ -0,0 +1,204 @@
...
+  alloca $ \st -> do
+    initState st chunks
+    classify <$> toHighLevel
+      ( fixed st          -- crypto_sign_state *state
+      $ input defaultIn   -- sig
+      $ input defaultIn   -- pk
+      $ resultPure id     -- raw status
+      ) crypto_sign_final_verify signature publicKey
+  where
+    classify status = status == 0
```
bolt12: Same question here. Can we refactor this module such that everything is kept high-level and designs a nice UI for the user? Why do we need initState here?

Can we move the classification to inside the toHighLevel call?

examples/libsodium/hs-project/src/LibSodium/Sign.hs
```diff
@@ -0,0 +1,204 @@
...
+initState :: Ptr Crypto_sign_state -> [ByteString] -> IO ()
+initState st chunks = do
+  toHighLevel (fixed st $ throwOnNonZero (sodiumError "crypto_sign_init"))
+    crypto_sign_init
+  forM_ chunks $ \c ->
+    toHighLevel
+      ( fixed  st                -- state
+      $ input2 constByteStringLenIn   -- m, mlen
+      $ throwOnNonZero (sodiumError "crypto_sign_update")
+      ) crypto_sign_update c
```
bolt12: I am not familiar with criptography, so I don't know why this is needed, why do we receive a `Ptr Crypto_sign_state` that the other function allocate? Can't we call scratch here and simplify things?

examples/libsodium/README.md
```diff
@@ -0,0 +1,27 @@
+# Generate `libsodium` Haskell bindings with `hs-bindgen`
+
+This example demonstrates generating Haskell bindings for
+[libsodium](https://libsodium.org/), a modern, easy-to-use cryptography library.
+It also hand-writes a small high-level API over the generated bindings; see
+[FINDINGS.md](FINDINGS.md) for the write-up.
```
bolt12: get rid of the FINDINGS.md references

bolt12:
Closed in favour of moving the hs-bindgen-highlevel to https://github.com/well-typed/binding-combinators and renaming it.

PR #2125, opened 2026-07-09
Skip CXCursor_StaticAssert in the parse pass

jtnuttall:
I vendored `hs-bindgen` for some SDL3 bindings and parse panicked on `CXCursor_StaticAssert`. Since C11 `_Static_assert` declarations don't declare anything bindable, I skipped them in the parse pass. I thought I'd try upstreaming the change.

This was actually the only alteration I had to make to get a rendering triangle from raw bindings, which is really neat.

(In SDL3 this affected `SDL_COMPILE_TIME_ASSERT` in `SDL_stdinc.h` )

---

- [X] Did you update the changelog? Please be sure to give a "migration hint" if this is a breaking change.


edsko:
Thanks @jtnuttall ! And great to hear that it's working well for you :) Let us know if you have any further problems.

PR #2134, opened 2026-07-14
`CI`: run `cabal check`

jorisdral:


edsko:
this is run by the comprhensive CI check. do we really need it enabled all the time?

jorisdral:
> this is run by the comprhensive CI check. do we really need it enabled all the time?

Ah I didn't know that -- I have this job in most of my other repos just because it is cheap to run

Issue #560, opened 2025-04-09
Convert names to NFC before translating to Haskell


edsko:
See #535 for an example use case.

edsko:
This has long since been implemented, closing.

mheinzel:
Clarification to avoid future confusion: Unicode normalization is not implemented. Some related bugs have been fixed, so there is no need for normalization. In rare instances, it would lead to prettier names, as it avoids having to escape some specific characters.

PR #2135, opened 2026-07-14
Manual: Fix references to issues

mheinzel:
The manual describes some issues that have been resolved by now. I also did a little bit of cleanup related to referencing issues.

### DIFF
--- a/manual/low-level/translation/generated-names.md
+++ b/manual/low-level/translation/generated-names.md
@@ -201,20 +201,13 @@ which `isAlphaNum` returns `False`. After escaping, this name becomes
 
 > [!NOTE]
 > We could instead _drop_ such invalid characters, but since this could result
-> in name clashes we do not do this by default.
+> in name clashes we avoid dropping characters by default.
 
 > [!NOTE]
 > For this _particular_ example we could generate a better name if we
 > transformed it to [NFC][unicode:nfc] first. Indeed, `gcc` (but not `clang`)
 > will issue a warning that this name is not in NFC.
 > <https://github.com/well-typed/hs-bindgen/issues/560>
-
-> [!NOTE]
-> C _functions_ with names that contain characters that are invalid in Haskell
-> identifiers can currently not be imported, because `ghc` applies Haskell
-> naming rules to C identifiers. This is unlikely to be an issue, but if it is,
-> we'd need to either patch ghc or generate a C wrapper.
-> <https://github.com/well-typed/hs-bindgen/issues/569>
 
 ### Capitalization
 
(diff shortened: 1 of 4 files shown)


manual/low-level/translation/generated-names.md
```diff
@@ -194,29 +193,12 @@ The name might contain invalid characters (characters that are invalid
...
-[`isAlphaNum`][hackage:base:isAlphaNum] is always invalid. This is sometimes
-counter-intuitive; for example, `isAlphaNum` returns `True` for both `'ó'` and
-`'你'`, but `"adiós"` contains a combining diacritical mark after the `o` for
-which `isAlphaNum` returns `False`. After escaping, this name becomes
-`adio'0301s` (or `Adio'0301s`) in Haskell.
+[`isAlphaNum`][hackage:base:isAlphaNum] is always invalid.
 
 > [!NOTE]
 > We could instead _drop_ such invalid characters, but since this could result
 > in name clashes we do not do this by default.
```
dschrempf:
```suggestion
> in name clashes we avoid dropping characters by default.
```

(As a side note, if there were name clashes, we would detect them with out shiny new name mangler clash-detection traversal; I don't think this is relevant here, just saying).

dschrempf (approved):
Thank you! I just left a tiny nit.

Feel free to merge after considering my suggestion, we have Edsko's OK!

manual/low-level/translation/generated-names.md
```diff
@@ -194,29 +193,12 @@ The name might contain invalid characters (characters that are invalid
...
 
-> [!NOTE]
-> For this _particular_ example we could generate a better name if we
-> transformed it to [NFC][unicode:nfc] first. Indeed, `gcc` (but not `clang`)
-> will issue a warning that this name is not in NFC.
-> <https://github.com/well-typed/hs-bindgen/issues/560>
```
mheinzel, line 211: Since [#560] is closed as completed, I thought generated names would now always be in NFC. However, looking at the related golden test case, we seem to still generate identifiers `Adio'0301s` etc., so are we not normalising after all? I'll double check and potentially revert part of my change.

mheinzel (commented):
We shouldn't merge just yet.

manual/low-level/translation/generated-names.md
dschrempf: Yes, I was wondering about this change, and checked the issue. But it was closed, so I assumed it is "fixed".

mheinzel: Unless I'm missing something we're not doing Unicode Normalization. I now re-added that note to the manual and will leave a brief clarifying comment on the closed issue as well to avoid future confusion.

Issue #2061, opened 2026-06-17
Flatten `{struct,union}`-in-`{struct,union}` by providing additional `Base.HasField`/`Compat.HasField` instances


edsko:
When we discussed adding support for implicit fields (#1649), we considered flattening where possible, but in the end decided that it would make more sense to treat everything uniformly, and never flatten. Given

```c
struct struct_with_struct {
  struct {
    int  x;
    char y;
  };
  double z;
};

struct struct_with_union {
  union {
    int  x;
    char y;
  };
  double z;
};

struct union_with_struct {
  struct {
    int  x;
    char y;
  };
  double z;
};

union union_with_union {
  union {
    int  x;
    char y;
  };
  double z;
};
```

we [currently](40387be91b857b68b5b8374e1905aea02eca54e) generate the following code.

<details>
Focusing only on the Haskell fields, omitting `HasField` for `Ptr` and `HasCField` instances:

```hs
{-------------------------------------------------------------------------------
  Struct with struct
-------------------------------------------------------------------------------}

data Struct_with_struct = Struct_with_struct{
    x :: Struct_with_struct_x
  , z :: RIP.CDouble
  }

data Struct_with_struct_x = Struct_with_struct_x{
    x :: RIP.CInt
  , y :: RIP.CChar
  }

-- HasField instances all generated by ghc, not hs-bindgen

instance RIP.HasField "x" Struct_with_struct   Struct_with_struct_x
instance RIP.HasField "z" Struct_with_struct   RIP.CDouble

instance RIP.HasField "x" Struct_with_struct_x RIP.CInt
instance RIP.HasField "y" Struct_with_struct_x RIP.CChar

{-------------------------------------------------------------------------------
  Struct with union
-------------------------------------------------------------------------------}

data Struct_with_union = Struct_with_union{
    x :: Struct_with_union_x
  , z :: RIP.CDouble
  }

newtype Struct_with_union_x = Struct_with_union_x{
    unwrap :: RIP.ByteArray
  }

-- HasField instances generated by ghc

instance RIP.HasField "x" Struct_with_union Struct_with_union_x
instance RIP.HasField "z" Struct_with_union RIP.CDouble

-- Setters/getters generated by hs-bindgen

get_struct_with_union_x_x :: Struct_with_union_x -> RIP.CInt
get_struct_with_union_x_y :: Struct_with_union_x -> RIP.CChar

set_struct_with_union_x_x :: RIP.CInt  -> Struct_with_union_x
[...]

PR #2137, opened 2026-07-14
`MangleNames`: support non-local mangling of scoped names

jorisdral:
Preparation for #2061

Scoped names are historically only referenced locally. For example, the name of struct/union field is only referenced by the field itself. This makes mangling of scoped names a purely local computation. However, as part of #2061, we will need to reference field names from elsewhere in the C AST. To support non-local referencing of scoped names, we have to make sure that the mangling of scoped names also happens non-locally. Currently the `MangleNames` pass creates Haskell names for C scoped names locally and immediately replaces them with a `ScopedNamePair`. This PR changes this so that we first create a Haskell name for each C scoped name while leaving all references to C scoped names unresolved. Only after all names have been chosen do we resolve each C scoped name reference to the `ScopedNamePair` that we created for it before. This follows the approach used for mangling `DeclId`: we first create a Haskell name for each `DeclId`, and only then we do resolve each `DeclId` reference to the `DeclIdPair` that was created for it.

### PR checklist

- [x] Did you update the manual?

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/CreateNames.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/CreateNames.hs
@@ -14,6 +14,7 @@ import Control.Applicative ((<|>))
 import Control.Monad.Except (ExceptT, MonadError (..), liftEither, runExcept,
                              runExceptT)
 import Control.Monad.Reader (Reader, asks, runReader)
+import Control.Monad.State (StateT (runStateT), modify)
 import Data.Foldable qualified as Foldable
 import Data.Map qualified as Map
 import Data.Proxy
@@ -44,15 +45,23 @@ import HsBindgen.Util.Tracer (WithCallStack, withCallStack)
 
   This pass is local to name mangling: it is /not/ exposed as a frontend
   artefact and not selectable via the CLI. It is the result of the first of the
-  three name-mangling traversals ("create names"): it has all within-declaration
-  names minted (as annotations and 'ScopedNamePair' slots), but all references
-  (the 'Id' slots) are still unresolved 'DeclId's. The final 'resolveNames'
-  traversal rewrites these into 'DeclIdPair's, producing 'MangleNames'.
+  three name-mangling traversals ("create names").
+
+  'CreateNames' creates top-level declaration names (the 'Id' slots), scoped
+  names ('ScopedName'), and all local auxiliary names such as auxiliary type or
+  data constructor names.
+
+  'CreateNames', however, does not resolve names within declarations; names that
+  can be referred to from other declarations are stored in the 'NameMap'
+  (top-level names and scoped names). Names that are used purely locally are
+  stored in annotations (local auxiliary type and data constructor names).
+
+  The final 'resolveNames' traversal rewrites 'C.DeclId's into 'DeclIdPair's and
+  'C.ScopedName's into 'ScopedNamePair's, producing 'MangleNames'.
 
   Note that @'Ann' ix 'CreateNames'@ and @'Ann' ix 'MangleNames'@ reduce to the
-  same bundle types, and 'ScopedName' is 'ScopedNamePair' in both. The
-  annotations and scoped names can therefore be carried across directly during
-  'resolveNames'.
+  same bundle types. The annotations can therefore be carried across directly
+  during 'resolveNames'.
 -------------------------------------------------------------------------------}
 
 -- | Create names pass (local to name mangling)
@@ -75,7 +84,7 @@ instance PassId CreateNames where
   type Id CreateNames = C.DeclId
 
 instance PassScopedName CreateNames where
-  type ScopedName CreateNames = ScopedNamePair
+  type ScopedName CreateNames = C.ScopedName
 
 instance PassMacro CreateNames where
   type MacroId         CreateNames = C.DeclId
@@ -102,9 +111,10 @@ instance PassMsg CreateNames where
   CoercePass: ResolveBindingSpecs → CreateNames
 
   Used by the "create names" traversal to reindex the parts of a declaration
-  that do not change representation (types, comments, enclosing references). The
-  'Id' (and hence 'MacroId') and 'ExtBinding' representations agree between the
-  two passes, so these coercions are pure reindexing.
+  that do not change representation (types, comments, scoped names, enclosing
+  references). The 'Id' (and hence 'MacroId'), 'ScopedName', and 'ExtBinding'
+  representations agree between the two passes, so these coercions are pure
+  reindexing.
 -------------------------------------------------------------------------------}
 
 instance CoercePassId               ResolveBindingSpecs CreateNames
@@ -177,11 +187,13 @@ mangleCandidate mc _ cName =
   Traversal 1: Create names
 
   For each declaration we first choose its top-level name (step 1a,
-  'nameForDecl') and, when that succeeds, immediately mint all its
-  within-declaration names (step 1b, 'createDecl').
+  'nameForDecl'). When that succeeds, we immediately create all its
+  within-declaration names (step 1b, 'createDecl'). This includes creating
+  scoped names and purely local names.
 
   The top-level name feeds the 'NameMap' regardless of what happens in 1b: a
   squashed or otherwise un-emitted declaration must still be resolvable by name.
+  Scoped names are also fed into the 'NameMap'.
 -------------------------------------------------------------------------------}
 
 createNames ::
@@ -193,13 +205,15 @@ createNames ::
   -> ( [C.Decl l CreateNames]
      , [(C.DeclId, Hs.Name Hs.NsTypeConstr, TypedefAnalysis.Squash)]
      , NameMap
+     , Maybe (NonEmpty DupAssign)
      , [MangleNamesFailure]
      , [AnnMsg MangleNames]
      )
 createNames td mc strategy decls = (
-      map snd successes
+      map (snd . snd) successes
     , squashes
     , nameMap
+    , nameMapDups
     , failures
     , messages
     )
@@ -217,17 +231,17 @@ createNames td mc strategy decls = (
       , fieldNamingStrategy = strategy
       }
 
-    results :: [CreateNamesResult (C.Decl l CreateNames)]
+    results :: [CreateNamesResult ([ScopedNamePair], C.Decl l CreateNames)]
     messages :: [AnnMsg MangleNames]
     (results, messages) = second concat $ unzip $ map perDecl decls
 
     perDecl ::
          C.Decl l ResolveBindingSpecs
-      -> (CreateNamesResult (C.Decl l CreateNames), [AnnMsg MangleNames])
+      -> (CreateNamesResult ([ScopedNamePair], C.Decl l CreateNames), [AnnMsg MangleNames])
     perDecl decl = (,msgs) $ case nameResult of
         CnFailure err      -> CnFailure err
         CnSquashed nC nH s -> CnSquashed nC nH s
-        CnMangled name _   -> createNamesWithin env name decl
+        CnMangled name ()  -> createNamesWithin env name decl
       where
         nameResult :: CreateNamesResult ()
         msgs       :: [AnnMsg MangleNames]
@@ -235,13 +249,20 @@ createNames td mc strategy decls = (
 
     failures  :: [MangleNamesFailure]
     squashes  :: [(C.DeclId, Hs.Name Hs.NsTypeConstr, TypedefAnalysis.Squash)]
-    successes :: [(DeclIdPair, C.Decl l CreateNames)]
-    (failures, squashes, successes) = partitionCreateResults results
+    successes :: [(DeclIdPair, ([ScopedNamePair], C.Decl l CreateNames))]
+    (failures, squashes, successes) = partitionCreateNamesResults results
 
     nameMap :: NameMap
-    nameMap = fromDeclIdPairs $
-         map fst successes
-      ++ map (\(cN, hN, _) -> DeclIdPair cN (Hs.demoteNs hN)) squashes
+    nameMapDups :: Maybe (NonEmpty DupAssign)
+    (nameMap, nameMapDups) = fromNames $
+         map fromSuccess successes
+      ++ map fromSquash squashes
+      where
+        fromSuccess (declIdPair, (scopedNamePairs, _)) = (declIdPair, scopedNamePairs)
+        fromSquash (cN, hN, _)                         = (declIdPair, scopedNamePairs)
+          where
+            declIdPair = DeclIdPair cN (Hs.demoteNs hN)
+            scopedNamePairs = []
 
 nameForDecl ::
      HasCallStack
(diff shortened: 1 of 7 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/CreateNames.hs
```diff
@@ -44,15 +45,16 @@ import HsBindgen.Util.Tracer (WithCallStack, withCallStack)
...
+  'ScopedName' slots) are still unresolved 'C.ScopedName's. The final
+  'resolveNames' traversal rewrites 'C.DeclId's into 'DeclIdPair's and
+  'C.ScopedName's into 'ScopedNamePair's, producing 'MangleNames'.
 
   Note that @'Ann' ix 'CreateNames'@ and @'Ann' ix 'MangleNames'@ reduce to the
-  same bundle types, and 'ScopedName' is 'ScopedNamePair' in both. The
-  annotations and scoped names can therefore be carried across directly during
-  'resolveNames'.
+  same bundle types. The annotations can therefore be carried across directly
+  during 'resolveNames'.
```
dschrempf, line 64: I tried to come up with a more understandable comment:

This pass is local to name mangling: it is /not/ exposed as a frontend artefact
and not selectable via the CLI. It is the result of the first of the three
name-mangling traversals ("create names").

'CreateNames' mangles top-level declaration names (the 'Id' slots), scoped names
('ScopedName'), and all local auxiliary names such as auxiliary type or data
constructor names.

'CreateNames', however, does not resolve names within declarations; instead, it
stores names either in the 'NameMap' (top-level names and scoped names), or in
annotations (local auxiliary type and data constructor names).

The final 'resolveNames' traversal rewrites 'C.DeclId's into 'DeclIdPair's and
'C.ScopedName's into 'ScopedNamePair's, producing 'MangleNames'.

Note that @'Ann' ix 'CreateNames'@ and @'Ann' ix 'MangleNames'@ reduce to the
same bundle types. The annotations can therefore be carried across directly
during 'resolveNames'.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/CreateNames.hs
```diff
@@ -178,10 +180,12 @@ mangleCandidate mc _ cName =
 
   For each declaration we first choose its top-level name (step 1a,
   'nameForDecl') and, when that succeeds, immediately mint all its
-  within-declaration names (step 1b, 'createDecl').
+  within-declaration names (step 1b, 'createDecl'). Scoped names are collected
```
dschrempf:
```suggestion
  within-declaration names (step 1b, 'createDecl'). Scoped names are mangled/minted/created/collected
```
?

Not your fault, I already introduced a second and a third name for "mangling", namely "minting" and "creating".

I had to come up with a "CreateNames" because "MangleNames" is the complete pass. So maybe we stick to "create" here, and remove references to "mint" and "mangle"? Arguably, we "mangle" names in the "CreateNames" traversal, doh. Ideas?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/CreateNames.hs
dschrempf: Coming back to this. I found it inconsistent, that we use the name map _and_ annotations. However, there is clear distinction:

The name map contains names that can be referred to from other declarations.

Annotations are purely local.

Maybe we should state this somewhere?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/CreateNames.hs
```diff
@@ -366,22 +393,23 @@ mkName ns candidate = do
...
 
-data CreateNamesResult a =
-      CnFailure   MangleNamesFailure
-    | CnSquashed  C.DeclId (Hs.Name Hs.NsTypeConstr) TypedefAnalysis.Squash
-    | CnMangled   DeclIdPair a
+-- | Result of creating within-declaration names
+data CreateNamesWithinResult a =
+      CnwFailure   MangleNamesFailure
+    | CnwSquashed  C.DeclId (Hs.Name Hs.NsTypeConstr) TypedefAnalysis.Squash
+    | CnwMangled   DeclIdPair [ScopedNamePair] a
```
dschrempf: I don't completely understand why we need both, `CreateNamesResult` and `CreateNamesWithinResult`. The list of scoped name pairs can be empty, then they are isomorphic, right? Do you think we can reduce complexity here?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/Error.hs
```diff
@@ -44,7 +44,9 @@ instance IsTrace Level MangleNamesError where
     MangleNamesCreationError CreateNamesSquashMustBeTypeConstr{} -> Bug
     MangleNamesCreationError{}                                   -> Warning
     MangleNamesCollisionError{}                                  -> Warning
-    MangleNamesResolutionError{}                                 -> Warning
+    MangleNamesResolutionError e -> case e of
+      ResolveNamesUnderlyingDeclNotMangled{} -> Warning
+      ResolveNamesScopedNameNotMangled{}     -> Bug
```
dschrempf: What if name mangling of a scoped name fails for some reason, then it is not a bug that the scoped name has not been mangled, right?

dschrempf (commented):
Looks good, thank you.

I left some comments and questions. In general, it would be great if we reduced complexity.

The only directions I can think of at the moment:
- Use a flat name map (no namespace distinction on the type-level). I don't really like that.
- Remove replace `CreateNamesResult` with `CreateNamesWithinResult`, and use an empty list of scoped names for top level declarations. Actually, I don't completely understand why `CreateNamesWithinResult` is "within". It still contains the mangled top-level name, right?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/CreateNames.hs
jorisdral: I think your new version of the comment make sense! An indeed it might be good to explicitly state that we use annotations for names that are purely local, and the name map contains names that can be referred to from other declarations

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/CreateNames.hs
jorisdral: I can try to use "create" everywhere. Fits with the name of the sub-pass :)

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/CreateNames.hs
jorisdral: Sure they are isomorphic but simply using an empty list in `CreateNamesResult` is confusing because it makes an impossible case representable. Maybe I can change `CreateNamesResult` a bit though so we have only one type but without making it possible to create scoped names in `nameForDecl`

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/Error.hs
jorisdral: I think in the `CreateNames` sub-pass we are sure to create a `ScopedNamePair` for every `ScopedName`, so my thinking here was that if it is missing from the `NameMap` during resolution then we have a bug in the he `CreateNames` sub-pass. Could the `ScopedNamePair` be missing from the `NameMap` for other reasons than that, for example due to `DetectClashes`?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/DetectClashes.hs
```diff
@@ -115,13 +115,23 @@ derivedNames strategy owner loc = \case
...
+          Just hsName -> [ (Field, fieldScope, field.info.loc, hsName) ]
+
 
-    enumConstantName :: C.EnumConstant CreateNames -> (NameRole, Scope, SingleLoc, Hs.SomeName)
-    enumConstantName constant = (Constructor, GlobalScope, loc, constant.info.name.hsName)
+    enumConstantName :: C.EnumConstant CreateNames -> [(NameRole, Scope, SingleLoc, Hs.SomeName)]
+    enumConstantName constant =
+        case lookupScopedName owner constant.info.name nameMap of
+          Nothing -> []
+          Just hsName -> [ (Constructor, GlobalScope, constant.info.loc, hsName) ]
```
jorisdral: @dschrempf since we create names for every `ScopedName`, it should actually be "impossible" to have a lookup fail here. How should we handle a failing lookup like that? Should we emit a bug-level trace message?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/Names.hs
```diff
@@ -64,26 +66,43 @@ data NameMap = NameMap {
...
+
+    aux2 :: DeclIdPair -> NameMap -> ScopedNamePair -> NameMap
+    aux2 pair nameMap scopedPair =
+        nameMap & #scopedNames %~
+          Map.insertWith (Map.unionWith combine) pair.cName (Map.singleton scopedPair.cName scopedPair.hsName)
+      where
+        -- if we insert the same 'C.ScopedName' twice, then there is a bug in
+        -- our code. Scoped names should be unique (within a single
+        -- declaration)!
+        combine _x _y = panicPure "one scoped name is inserted twice"
```
jorisdral: I personally don't like the panics I introduced here, but `Map.insert` had me a little worried because it silently allows overwriting entries in the `NameMap` for `DeclId`s that were already inserted before. Maybe we should emit bug-level trace messages if we insert the same `DeclId` into the same `Map` twice?

jorisdral:
> I left some comments and questions. In general, it would be great if we reduced complexity.
>
> The only directions I can think of at the moment:
> 
>     * Use a flat name map (no namespace distinction on the type-level). I don't really like that.

Maybe we can do something like this?

```hs
newtype NameMap = NameMap { unwrap :: Map C.DeclId NameMapEntry }

data NameMapEntry = NameMapEntry {
    typeConstr  :: Maybe (Hs.Name Hs.NsTypeConstr)
  , dataConstr  :: Maybe (Hs.Name Hs.NsTypeConstr)
  , var         :: Maybe (Hs.Name Hs.NsVar)
  , scopedNames :: Map C.ScopedName Hs.SomeName
  }
```

All the lookup functions can stay the same. It's only the representation that changes

>     * Remove replace `CreateNamesResult` with `CreateNamesWithinResult`, and use an empty list of scoped names for top level declarations. Actually, I don't completely understand why `CreateNamesWithinResult` is "within". It still contains the mangled top-level name, right?

I'll try to merge the two types. I called it `CreateNamesWithinResult` because it is only returned from `createNamesWithin`

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/CreateNames.hs
dschrempf: You are right, of course. Maybe it could be polymorphic, and only structures return hte result with a list?

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/Error.hs
dschrempf: I think you are right for a single declaration. It is going to fail, if one of the name creation steps within it fail (that's the result type if I recall).

But now that other structs may refer to scoped name of a struct that may fail, I am not sure if it is a bug when you don't find the scoped name _of another struct_.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/DetectClashes.hs
dschrempf: See my comment above. I think if we fail to mangle an anonymous, indirect struct; the outer struct will refer to the non-existing fields. In that case, the lookup may fail, and it is not a "bug" in the classical sense. But maybe I am missing something.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/Names.hs
dschrempf: I think so. Bug-level messages would fit nicely here.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/CreateNames.hs
jorisdral: Right, I'm going with that

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/Error.hs
jorisdral: Ahh you're right.. If we fail to create the top-level name for a decl, then we won't create scoped names for that decl either. If a validly missing scoped name is then referred from a different decl, the lookup would fail. It shouldn't be traced as a bug-level message in that case

I realise now that if a scoped name is currently missing from the name map, it can mean three things:

* We failed to create a name for the top-level decl that the scoped name lives in
* We squashed the top-level decl that the scoped name lives in
* We have a bug where we "forget" to create a name for a scoped name

The same is kind of true for a DeclId too (but probably less likely). If a DeclId is missing from the name map, it can mean:

* We failed to create a name for the top-level decl 
* We squashed the top-level decl
* We have a bug where we "forget" to create a name for a DeclId

Should we make that difference between "create names failure" and "squashed" and "bug" explicit at some point in the `NameMap`?

dschrempf: In theory we could do so, by just storing the CreateNamesResult, right? It stores exactly this information... But we lose the namespace type safety, which I quite like.

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/DetectClashes.hs
jorisdral: I added warning messages for failing lookups

jorisdral:
> > I left some comments and questions. In general, it would be great if we reduced complexity.
> > The only directions I can think of at the moment:
> > ```
> > * Use a flat name map (no namespace distinction on the type-level). I don't really like that.
> > ```
> 
> Maybe we can do something like this?
> 
> ```haskell
> newtype NameMap = NameMap { unwrap :: Map C.DeclId NameMapEntry }
> 
> data NameMapEntry = NameMapEntry {
>     typeConstr  :: Maybe (Hs.Name Hs.NsTypeConstr)
>   , dataConstr  :: Maybe (Hs.Name Hs.NsTypeConstr)
>   , var         :: Maybe (Hs.Name Hs.NsVar)
>   , scopedNames :: Map C.ScopedName Hs.SomeName
>   }
> ```

I ended up not changing the representation to this. It didn't feel like an improvement

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/MangleNames/ResolveNames.hs
```diff
@@ -164,6 +178,15 @@ lookupVarR declId = do
           ResolveNamesUnderlyingDeclNotMangled declId (NonEmpty.singleton Hs.NsVar)
       Just hsNm -> pure hsNm
 
+lookupScopedNameR :: C.DeclId -> C.ScopedName -> ResolveE Hs.SomeName
+lookupScopedNameR declId scopedName = do
+    nameMap <- ask
+    case lookupScopedName declId scopedName nameMap of
+      Nothing ->
+        throwError $
+          ResolveNamesUnderlyingDeclNotMangled declId (NonEmpty.singleton Hs.NsVar)
```
jorisdral: Should be `ResolveNamesScopedNameNotMangled`

edsko (approved):

PR #2138, opened 2026-07-15
Replace example submodules with pinned clones

bolt12:


TravisCardwell (commented):
Looks good!  Thanks!

I notice that just the example submodules are removed, while the alternatives remain.  Perhaps we should remove the alternative submodules as well?

bolt12:
Sure, I can remove the alternatives too, I left them because I wasn't sure if I should remove all submodules.

jorisdral:
Maybe we can check out tags rather than commits, if applicable? That was a conscious decision for submodules in the examples I created

bolt12:
Good point! On it!

examples/c-rpm/generate-and-run.sh
```diff
@@ -11,6 +11,13 @@ RPMBUILD_DIR="$SCRIPT_DIR/rpm/_build/build/"
 BINDING_SPEC_DIR="$SCRIPT_DIR/binding-specs"
 HS_OUTPUT_DIR="$SCRIPT_DIR/hs-project/src/"
 
+# RPM Package Manager, pinned so the example is reproducible. We fetch it on
+# demand rather than as a git submodule: a submodule is cloned by cabal for
+# every project that depends on hs-bindgen via source-repository-package, even
+# though only hs-bindgen-runtime is needed.
+RPM_REPO="https://github.com/rpm-software-management/rpm.git"
+RPM_COMMIT="2530a42e4618f41fcc6cd21f0e70ca851b24ddc2"
```
TravisCardwell, line 19: Is there a reason to not use a tag for `rpm`?  ([tags](https://github.com/rpm-software-management/rpm/tags))

alternatives/cabal.project
```diff
@@ -3,5 +3,13 @@ packages:
   experiment/inline-c
   experiment/hsc2hs
   experiment/c2hs
-  tools/hsc2hs
-  tools/c2hs
+
+source-repository-package
+  type: git
+  location: https://github.com/haskell/hsc2hs.git
+  tag: b1fcea0bf0fbd22f00729cf6479129e8f78ce0d7
```
TravisCardwell, line 10: Is there a reason to not use a tag for `hsc2hs`?  ([tags](https://github.com/haskell/hsc2hs/tags))

alternatives/cabal.project
```diff
@@ -3,5 +3,13 @@ packages:
...
+
+source-repository-package
+  type: git
+  location: https://github.com/haskell/c2hs.git
+  tag: 6f699f611b8123219ec521f04e1902d4f0f12711
```
TravisCardwell, line 15: Is there a reason to not use a tag for `c2hs`?  ([tags](https://github.com/haskell/c2hs/tags))

TravisCardwell (commented):
This has been merged, so I pulled and updated my example/exercise projects.  Building is now very fast (with no more clones from the WAN) and no longer has lots of git noise.  Thank you!

examples/c-rpm/generate-and-run.sh
bolt12: The repo wasn't using any tag before, so I pinned to the version they were using with git submodules. I didn't check the repo, might as well do it for all

bolt12: hmm I did notice that all except `rpm` don't have an up to date tag so I will update the tag on rpm and leave the others as a pinned commit

PR #2145, opened 2026-07-20
Fix tarball generation: remove mention of non-existant files

dcoutts:
`hs-bindgen-test-runtime` has no changelog, so don't mention it in the .cabal file.

Presumably due to a previous package split, and the new one doesn't get its own changelog file.

This fixes doing:
```
cabal install hs-bindgen
```
in the repo root directory. This is because `cabal install` for local packages works by first creating a .tar.gz file for each local package, and installing that. So when that breaks, install breaks.

Since the alpha releases are quite old at this stage, this is handy for new users to install `hs-bindgen`.


dschrempf:
As @dcoutts pointed out in private communications, we should add `cabal sdist all` to our CI pipeline. It is failing on `main`.

dschrempf:
Superseded by #2146, which cherry picks this commit! Thank you!

PR #2146, opened 2026-07-20
More preparations for release

dschrempf:
This mostly affects the Nix code (we now have a single source of truth for source-repository-package stanzas, and they are handled automatically; also, we directly store the derivations, and do not run `callCabal2nix`).

Also, fix creation of `sdist` (thank you @dcoutts, I cherry-picked your change, fixed the formatting, and added `cabal sdist all` to CI).


dschrempf:
I am putting this on the queue (it only contains Nix changes and minor CI changes).

PR #2164, opened 2026-07-23
Zero values for structs

### LINKED ISSUE #2121: Support `zero` values for structs
After #2061, users of generated bindings will be able to get and set indirect fields of `struct`s and `union`s without having to bother with anonymous `struct`s/`union`s (and their arguably confusing names) directly. However, these getters and setters do not help with constructing new `struct` values. For `union` values, we have the `zero` value in the `IsUnion` class, after which the user can set fields as they see fit. For `struct` values, we do not yet have such a `zero` value.

jorisdral:
Resolves #2121

Since #2061, we generate bindings for indirect fields. This makes it easier for users to access and modify anonymous structs/unions without having to use the awkward names that get assigned to them (and implicit fields). For unions we also have the `IsUnion` class with a `zero` value, which helps make it easy to *construct* an anonymous union. This PR adds an analogous `IsStruct` class with a `zero` value, and `hs-bindgen` now generates `IsStruct` instances for structs. This allows users to initialise a `zero` struct value and set its fields one by one without having to bother with anonymous structs/unions directly.

See also the commit messages.

### Example

```c
struct S {
  struct {
    int x;
    char y;
  };
};
```
```hs
-- generated bindings
data S_anon'x = S_anon'x { s_anon'x_x :: CInt, s_anon'x_y :: CChar }
data S = S { s_anon'x :: S_anon'x }

-- custom application: use anonymous structs/unions directly
withoutZeroExample :: S
withoutZeroExample = S (S_anon'x 13 17)

-- custom application: use indirect fields on a zero value
withZeroExample :: S
withZeroExample = setField @"x" (setField @"y" zero 13) 17
```

---


edsko (approved):

For the record: an example actually using rebindable syntax will be a separate PR.

hs-bindgen-runtime/src/HsBindgen/Runtime/Marshal.hs
```diff
@@ -1,7 +1,21 @@
...
+--   is a way to compute the missing C fields, or unless writes are allowed to
+--   be partial.
+--
+-- [Write-only instances] might be necessary if the Haskell struct is larger
+--   than the C struct. In this case we can write it but not read it, unless
+--   there is a way to compute the missing Haskell fields.
+--
+-- For more details, see https://github.com/well-typed/hs-bindgen/issues/649.
 --
 -- This module is intended to be imported qualified.
```
jorisdral, line 26: @edsko what do you think about this elaboration?

edsko: I think it's good, but I think it's useful if we mention `Tm` as an example, as that was really the motivation behind the split i nthe first place.

PR #2167, opened 2026-07-24
Make ansi color setting configurable

### LINKED ISSUE #2166: Client option to force use of color (or avoid use of color)
At the moment, the client does not expose any features related to ANSI terminal color management.

`hs-bindgen` automatically determines whether a handle supports colors, or not.

The [`hs-bindgen-playground`](https://github.com/dschrempf/hs-bindgen-playground) needs the colors of diagnostics, but runs `hs-bindgen` in a non-interactive shell. Only using a pseudoterminal (PTY) makes colors available. This is suboptimal because PTYs are a limited resource, which is potentially exhausted in a webserver with multiple users.

We have all configuration options to enable/disable ANSI colors available, and only need to set those, exposing a flag.

dschrempf:
Closes #2166 

---

```hs
-- TODO <link to issue>
-- Brief description
```

### DIFF
--- a/hs-bindgen/CHANGELOG.md
+++ b/hs-bindgen/CHANGELOG.md
@@ -81,6 +81,9 @@
 
 ### New features
 
+* A new CLI option `--color WHEN` controls ANSI colors in diagnostics, where
+  `WHEN` is `always`, `auto` (default, unchanged behavior: detect terminal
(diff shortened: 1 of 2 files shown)


dschrempf:
I have tested this locally and via the playground; merging.

PR #2169, opened 2026-07-24
Support bit-fields in unions

### LINKED ISSUE #1253: Support bit-fields in unions

jorisdral:
Resolves #1253

Now that we generate bindings for indirect fields, it's probably not too uncommon that someone would try to generate bindings for an *indirect* bit-field in a union. This PR adds support for bit-fields in unions, which turned out to be not too much work!

---

### DIFF
--- a/examples/feature-tests/c/Makefile
+++ b/examples/feature-tests/c/Makefile
@@ -19,6 +19,8 @@ clean:
 	rm -f *.o *.so *.dylib *.dll
 
 libfeature-tests$(TARGET_EXT): \
+		bit-fields/structs.o \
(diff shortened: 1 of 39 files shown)


jorisdral:
Moved the PR back to draft because I'm updating the manual

jorisdral:
Ready for review now

edsko (approved):
Approving modulo changes discussed, including some tests that actually peek and poke these bitfields.

PR #2191, opened 2026-08-06
Report the number of macros that failed to translate

### LINKED ISSUE #2185: Inform the user about the number of macro failures
Some C libraries contain many macros. By default, `hs-bindgen` does not emit traces when it fails to translate a macro (but see [1] and [2]).

The default behavior may lead to users thinking that `hs-bindgen` does not translate many macros, or does not translate macros at all. For example, see https://old.reddit.com/r/haskell/comments/1v960p9/ann_sdl3bindgensys_machinegenerated_lowlevel/

I suggest adding a feature facilitating the detection of many macro failures: collect the number of macro failures, and report them at then end! Then, we can say, for example,

> 14 macros dropped; use --log-enable-macro-warnings for details

[1]: These traces can be enabled using, for example, `--log-enable-macro-warnings`.
[2]: What *is* visible by default are selected, but failing, reverse transitive dependencies: `[select]` emits `Transitive dependency unusable` at `Warning`.

bolt12:
Closes #2185

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
@@ -30,6 +30,7 @@ import HsBindgen.Frontend.Analysis.UseDeclGraph qualified as UseDeclGraph
 import HsBindgen.Frontend.DeclMeta
 import HsBindgen.Frontend.Pass.AdjustTypes.IsPass
 import HsBindgen.Frontend.Pass.ConstructTranslationUnit.IsPass
+import HsBindgen.Frontend.Pass.Parse.Msg
 import HsBindgen.Frontend.Pass.Select.IsPass
 import HsBindgen.Frontend.Predicate
 import HsBindgen.Frontend.TranslationUnit qualified as C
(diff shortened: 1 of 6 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
```diff
@@ -520,6 +538,21 @@ getBugMsgsNotSelected = concatMap (uncurry aux) . DeclIndex.toList
...
+-- | Does this selection-root message represent a macro that hs-bindgen dropped?
+--
+isDroppedMacro :: SelectMsg -> Bool
+isDroppedMacro = \case
+    SelectUnusable reason -> case reason of
+      UnusableMacroTypecheckFailure  _ -> True
+      UnusableMacroResolutionFailure _ -> True
+      -- A parse failure counts as a dropped macro iff it is a macro parse
+      -- failure.
+      UnusableParseFailure p           -> "macro" `List.isInfixOf` (getTraceId p).id
```
bolt12: Pattern match on DelayedParseMsg for the specific macro constructors.

edsko (approved):
Feel free to merge once the change we discussed is done :)

PR #2193, opened 2026-08-06
`Manual`: move around some code related to generated names

jorisdral:
---


manual/generate-and-run.sh
```diff
@@ -23,21 +23,31 @@ echo "# "
...
+fi
+
+mkdir -p "$SPECS_DIR"
+
+GENERATED_DIR=hs/manual/generated
+if [ -d "$GENERATED_DIR" ]; then
+  rm -r "$GENERATED_DIR"
+fi
+
+mkdir -p "$GENERATED_DIR"
```
jorisdral, line 38: This is a small change that ensures that old files are removed before we generate new ones

PR #2196, opened 2026-08-11
Add `--toposort` flag to `info include-graph`

### LINKED ISSUE #2080: Expose the topological header order

bolt12:
Closes #2080


hs-bindgen/src-internal/HsBindgen.hs
```diff
@@ -182,12 +182,13 @@ hsBindgenEMacroLang
 -- | Write the include graph to @STDOUT@ or a file.
 writeIncludeGraph ::
      Boolean Regex
-  -> Bool
+  -> Bool      -- ^ Show paths instead of @#include@ arguments
+  -> Bool      -- ^ Topologically sorted list instead of a Mermaid graph
```
bolt12: Avoid boolean blindness for these flags

edsko (approved):
Free free to merge once those `Bool`s are a proper datatype.

PR #2202, opened 2026-08-13
Add `withFunPtrAs`

bolt12:
This PR adds a nice combinator to hs-bindgen-runtime that was found useful by #2123 . Since #2123 is going to be canceled in favour of moving the high-level library to https://github.com/well-typed/binding-combinators. This PR only contains the hs-bindgen related changes from #2123

### DIFF
--- a/examples/c-qrcode/generate-and-run.sh
+++ b/examples/c-qrcode/generate-and-run.sh
@@ -44,19 +44,18 @@ cabal run hs-bindgen-cli -- \
     "$SCRIPT_DIR/QR-Code-generator/c/qrcodegen.h"
 
 echo "# "
-echo "# Updating cabal.project.local"
+echo "# Writing cabal.project.local"
 echo "# "
 
-LINE=$(
-    cat <<-EOF
+# cabal.project.local is gitignored and machine-specific, so rewrite it from
+# scratch each run rather than appending (which would accumulate stale paths).
+cat > "$SCRIPT_DIR/hs-project/cabal.project.local" <<EOF
 package c-qrcode
     extra-include-dirs:
         $SCRIPT_DIR/QR-Code-generator/c/
     extra-lib-dirs:
         $SCRIPT_DIR/QR-Code-generator/c/
 EOF
-)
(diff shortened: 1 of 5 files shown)


examples/c-qrcode/generate-and-run.sh
```diff
@@ -44,28 +44,27 @@ cabal run hs-bindgen-cli -- \
...
 echo "# "
-echo "# Updating cabal.project.local"
+echo "# Writing cabal.project.local"
 echo "# "
 
-LINE=$(
-    cat <<-EOF
+# cabal.project.local is gitignored and machine-specific, so rewrite it from
+# scratch each run rather than appending (which would accumulate stale paths).
+cat > "$SCRIPT_DIR/hs-project/cabal.project.local" <<EOF
```
jorisdral, line 52: I use `cabal.project.local` when working on examples to hold my local configuration. This would delete my local configuration when I run the script

More importantly, the `examples` workflow also uses `cabal.project.local` to set CI-specific options. This change now overwrites the file contents and this will probably cause problems even if they are not immediately visibile

Also, we're not changing the paths very often, so the stale path argument isn't really true

examples/libpcap/generate-and-run.sh
```diff
@@ -31,18 +42,18 @@ echo "# "
...
+cat > "$SCRIPT_DIR/hs-project/cabal.project.local" <<EOF
 package libpcap
     extra-include-dirs:
         $LIBPCAP_DIR
     extra-lib-dirs:
         $LIBPCAP_DIR
 EOF
-)
-grep -qxF "$LINE" "$SCRIPT_DIR/hs-project/cabal.project.local" || echo "$LINE" >> "$SCRIPT_DIR/hs-project/cabal.project.local"
 cat "$SCRIPT_DIR/hs-project/cabal.project.local"
```
jorisdral, line 57: Same here, we should probably keep appending

examples/c-qrcode/generate-and-run.sh
jorisdral: Fixed in #2213

PR #2222, opened 2026-08-20
Emit TH string literals without relying on `-O`

dschrempf:
`[| s |]` for `s :: String` only produces `StringL` together with the `TH:liftString` rewrite rule. At `-O0` it corresponds to a lit of `Char` literals, which leads to differences in golden tests in TH mode.


dschrempf:
Minor change with no fixture changes, putting on the queue.

dschrempf:
(I just wanted to get #2222).

PR #2224, opened 2026-08-23
Fix hs-bindgen nix build

KovalevDima:
Resolves #2223

This change is a result of running:
```bash
sh ./nix/generate.sh
```

PR #2227, opened 2026-08-24
Manage git dependency pins

### LINKED ISSUE #2223: hs-bindgen-cli derivation build failure
Running
```sh
nix build .#hs-bindgen-cli
```
on the current main branch fails with:
```sh
[ 47 of 214] Compiling HsBindgen.IR.C.Naming ( src-internal/HsBindgen/IR/C/Naming.hs, dist/build/internal/HsBindgen/IR/C/Nam>
src-internal/HsBindgen/IR/C/Naming.hs:332:19: error: [GHC-88464]
    Variable not in scope: clang_hashCursor :: CXCursor -> m CUInt
    |
332 |       hash     <- clang_hashCursor curr
```
because `./nix/generated/libclang-bindings.nix` pins an older revision of `libclang-bindings` that does not provide `clang_hashCursor`

The problem also occurs when using the Nix overlay in a flake

A newer version of `libclang-bindings` is pinned in `cabal.project.base`, but the corresponding Nix package expression appears to be out of sync

dschrempf:
Cheery picks the commit by @KovalevDima in #2224, superseding that PR.

Cabal reads git pins from the `source-repository-package` stanzas in
`cabal.project.base`; Nix reads its own copy of them from `nix/generated/`. The
two can drift apart, because the version numbers (but not the revisions) still
match.

Details:

- `scripts/lib/git-pins.sh`: shared stanza parser. Fixes the previous inline
  version, which dropped every stanza but the last.
- `scripts/ci/check-nix-pins.sh`: fails when a pin and its generated expression
  disagree. The "Check filesystem" workflow can run it on every pull request.
- `scripts/update-git-dependency.sh`: bumps a stanza's `tag` and regenerates the
  affected expressions. Replaces `scripts/update-well-typed-dependency.sh`,
  removed in 5e8459f70.
- `nix/generate.sh`: accepts package arguments, pins every Hackage version (an
  unpinned `cabal://pkg` resolved to the newest one on Hackage, unrelated to
  `index-state`), and writes atomically rather than truncating the target before
  `cabal2nix` runs.
- `dev/dependencies.md`: version pins and how to change them.

---

```hs
-- TODO <link to issue>
-- Brief description
```


dschrempf:
I am putting this on the queue because it only affects the Nix toolchain.

PR #2228, opened 2026-08-24
Do not label Template Haskell traces as errors

### LINKED ISSUE #2216: Warnings are printed as errors in TH mode
This is a split off from #2212.

In TH mode, warnings read as errors. For example,

```
Template Haskell error: [Warning] [HsBindgen] [select] 'RGFW_getPhysicalDevicePresentationSupport_Vulkan' at "include/RGFW.h 2757:19":
  Could not select declaration (direct selection predicate match):
    ...
```

### Details

`outputConfigTH` routes warnings and errors to GHC through `runQ`:

```haskell
-- hs-bindgen/src-internal/HsBindgen/Util/Tracer.hs:389-393
report level _ = case level of
  Warning -> runQ . reportWarning
  Error   -> runQ . reportError
  _level  -> putStr
```

But the pipeline runs in `IO`:

```haskell
-- hs-bindgen/src-internal/HsBindgen/TH/Internal.hs:107
(deps, decls) <- liftIO $ do
    hsBindgenMacroLang …
```

So `runQ` resolves to `instance Quasi IO`, whose `qReport` prefixes `"Template Haskell error: "` for *both* severities.

dschrempf:
`outputConfigTH` routed warnings and errors through `runQ . reportWarning` and `runQ . reportError`, but the pipeline runs in `IO`, so `runQ` picked `instance Quasi IO`, whose `qReport` prefixes "Template Haskell error: " for both severities.

Template Haskell mode now uses the plain handle output configuration: all traces go to `stderr` (previously the other log levels went to `stdout`), without ANSI colours. The `[Warning]`/`[Error]` label the formatter prepends is now the only label.

See #2216.

### DIFF
--- a/hs-bindgen/CHANGELOG.md
+++ b/hs-bindgen/CHANGELOG.md
@@ -372,6 +372,10 @@
 * Make `UnnamedId`s unique by including a hash of the unnamed declaration's AST.
   This prevents accidental `UnnamedId` collisions on all supported `llvm`
   versions, even on `llvm < 19.1.0`. See [issue #2210][is-2210].
+* In Template Haskell mode, traces are no longer prefixed with "Template
+  Haskell error: ", which made warnings look like errors. All traces now go to
+  `stderr` (previously, only warnings and errors did). See [issue
+  #2216][is-2216].
 
 [is-1225]: https://github.com/well-typed/hs-bindgen/issues/1225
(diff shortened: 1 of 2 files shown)


dschrempf:
Minor PR, cleared with Edsko.

PR #2238, opened 2026-08-27
`Manual`: elaborate section on (non-)portability

### LINKED ISSUE #1720: Manual section: why we do not generate portable bindings
We talk about this here and there, but it would be good to have a single authoritative part of the manual where we document once and for all why we consider the generated bindings to be build artifacts, and all the ways in which they reflect machine and configuration specific properties, and why they do that (as just one example: the `foreign import`/`HasFFIType` stuff).

### LINKED ISSUE #1729: Record assumptions about `FFIType` and `Elem` type families
Instances of these type families should match the types used in C code

### `FFIType`

For example, if we have this C code:

```c
typedef int T;
```

And let's say the size of an `int` is 32 bits on the target system, then we will generate Haskell bindings like so:

```hs
newtype T = T Int32
  deriving newtype HasFFIType -- FFIType T = Int32
```

Even if `T` is replaced with a custom type using an external binding spec, then `FFIType T = Int32` should still hold on the same target system. This assumption should be recorded somewhere to inform users of `hs-bindgen`

### `Elem`

For example, if we have this C code:

```c
typedef char T [];
```

Then we will generate Haskell bindings like so:

```hs
newtype T = T (IncompleteArray CChar)
  deriving newtype IsArray -- Elem T = CChar
```

Even if `T` is replaced with a custom type using an external binding spec, then `Elem T = CChar` should still hold. This assumption should be recorded somewhere to inform users of `hs-bindgen`

#### Open question

If we have this C code:

```c
typedef char S;
typedef S T[];
```

Then we will generate Haskell bindings like so:

```hs
newtype S = S CChar
newtype T = T (IncompleteArray S)
  deriving newtype IsArray -- Elem T = S
```

If `T` is replaced with a custom type using an external binding spec, then should `Elem T = CChar` hold or `Elem T = S`? Is it fine if `Elem T = a` where `a` is some type that is coercible to `CChar`? We should answer this question, and this assumption should be recorded somewhere to inform users of `hs-bindgen`

jorisdral:
Resolves #1720
Resolves #1729

---


manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+
+> The bindings generated by `hs-bindgen` should be regarded as build artifacts.
+
+In this manual section, we aim to answer three questions:
+
+* [Causes][t:causes]: what causes non-portable bindings?
+* [Consequences][t:consequences]: how does non-portability materialize in
+  generated bindings?
+* [Solutions and mitigations][t:solutions]: how do we mitigate or solve
+  non-portability of generated bindings?
```
jorisdral, line 17: When starting to extend this manual section, I found that I was often repeating myself about these three questions, and the information was spread over the document in different places. So adding three separate subsections is my attempt to be a little more structured about it

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+## Preliminary: reasonably designed C libraries
+
+In theory, the public API of a C library may vary wildly based on the platform,
+specially since [conditional compilation][t:conditional-compilation] can almost
+arbitrarily modify C source code. *Unreasonably designed* C libraries are
+therefore likely not portable (or even very useful). We should not expect to
+generate portable bindings for such libraries, or to have sensible guidelines
+for solving/mitigating non-portable bindings. Throughout this manual section, we
+expect C libraries to be *reasonably designed* with at least some degree of
+portability in mind.
```
jorisdral, line 37: This sub-section is a little verbose, but it's important to set expectations/boundaries

manual/low-level/usage/portability.md
jorisdral, line 17: This may be a little philosophical, but I'm not super happy with the terminology "cause" vs. "consequence" because it's a bit of a gray area. Are generated bindings non-portable because `libclang` reports platform-dependent information, or because `hs-bindgen` decides it in a way that makes the generated bindings non-portable? For example, we use FFI types to implement `foreign import` declarations, which is purely a Haskell concern, but I list it in "consequences" instead of "cause"

Thoughts?

jorisdral: @edsko once you review this PR, the structuring into sections is something I'd appreciate your input on in particular. 

My thinking here is: C libraries can vary such much, it's impossible to cover everything with examples and solutions. So my hope is that with these three sections we can give broad ideas about what can go wrong, how it affects the bindings, and how to work around it

manual/low-level/usage/portability/portable-types.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+> [!Note]
+> We have ideas about making generated `Storable` instances more portable right
+> out of the box. See [issue #1818][issue-1818].
+
+#### `HasFFIType`
+
+Portable types should get a `HasFFIType` instance (if applicable) where the
+associated type `FFIType` matches the FFI type that `hs-bindgen` would deduce.
```
jorisdral, line 55: Of course, here we should give users an idea of how `hs-bindgen` deduces FFI types

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+This manual section is not necessarily exhaustive. If you experience portability
+issues that are not (satisfactorily) described here, please let us know by
+creating an issue on the [issue tracker][github:issue-tracker].
+
+There are ongoing efforts to improve portability: see [issue #1746][issue-1746].
+[Cross-compilation][manual:cross-compilation] may also be useful when one needs
+to generate bindings for one specific platform, but this does not solve the
+portability problem by itself.
+
+## Preliminary: reasonably designed C libraries
```
edsko, line 28: I'm a little surprised at seeing this proviso here. You started by saying that the bindings generated hs-bindgen are not in general portable, indeed they are usually not; and that they are considered build artefacts. So this 'reasonably designed" doesn't seem a requirement here at all; indeed, we're pretty proud of the fact that hs-bindgen can handle almost anything you throw at it. It seems to me this section belongs elsewhere, maybe in the chapter on mitigation?

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+therefore likely not portable (or even very useful). We should not expect to
+generate portable bindings for such libraries, or to have sensible guidelines
+for solving/mitigating non-portable bindings. Throughout this manual section, we
+expect C libraries to be *reasonably designed* with at least some degree of
+portability in mind.
+
+## Causes
+[t:causes]: #causes
+
+`hs-bindgen` relies on `libclang` for all decisions, such as the [resolution of
```
edsko, line 42: "for all decisions" is rather overstating the case :) We make plenty of decisions that are nothing to do with libclang, of course.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+C preprocessor (CPP) conditionals][t:conditional-compilation]. This means that
+the shape of the generated bindings and the corresponding binding specifications
```
edsko, line 44: "This means that.." establishes an implication that doesn't follow, at least not trvially. "libclang makes the decision therefore the bindings are platform specific" isn't particularly clear, even if it's perhaps true.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+depends on the platform where `hs-bindgen` runs. In general, this causes
+generated bindings to be non-portable across platforms, or even different
+configurations of the same library.
```
edsko, line 47: I'm also not terribly sure what distinction you're drawing between the second statemnt and the third: "the shape is platform dependent ... which cuases generated bindings to be non-portable"; that doesn't feel like it's quite what you want to say?

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+libraries, but leaves it up to an implementation to decide on many of the
+details, for example:
+
+* What is the size of an `int`?
+* Is the `char` type signed?
+* How much unnamed padding should be included between struct fields?
+* How should adjacent bit-fields be packed?
+* How is `size_t` implemented?
+
+These compiler decisions may be platform-dependent. For example, a compiler
```
edsko, line 64: May be useful to say explicitly that different compilers can make different decisions here (re-emphasizing the "up to an implementation"), and then from there: but the same compiler may also make different decisions on different platforms.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+implementation may decide on one platform that an `int` is 32 bits, and 64 bits
+on another. If the size of an `int` is used in the generated bindings, then the
+bindings will differ depending on the platform.
```
edsko, line 67: I realize that this is a bit of a chicken-and-egg problem, but "If the size of an int is used in the generated bindings" is a bit vague. If we have a function signature that has an int, does that mean "the size of an int is used"? In a way, yes: the size of the int will dictate how that int can be passed (which registers? where in the register?); but we still don't think of that case as being non-portable (though answering "why not?" is actually not that easy now that I think about it). Perhaps we could sidestep the issue here by saying: "If the generated bindings depend on the size of an `int`, for example in a `Storable` instance, then..".

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+
+In some cases `libclang` may even disagree with other C compilers on some
+information, though `clang` strives to be [compatible][clang:compatibility] with
+other C compilers. This is relevant for Haskell bindings as well: `hs-bindgen`
```
edsko, line 71: Would it be useful to mention `gcc` specifically here? I guess that's the one that most people would worry about.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+uses `libclang` to generate bindings, `ghc` uses `clang` or `gcc` to compile
+Haskell bindings, and (external) C libraries could be compiled with the same or
```
edsko, line 73: What does "compile Haskell bindings" mean? If this means "compile the generated Haskell code", then this doesn't really involve the C compiler at all; I don't mean to be nitpicking but I think it's good to be precise about these issues, because it's quite a confusing topic.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+another C compiler. We have yet to find examples where discrepancies between
+compilers have been a problem, however.
```
edsko, line 75: I think we should be a bit more careful here: we haven't found any differences between `gcc` and `clang`; "between compilers" seems to suggest we've tested more than that, and we haven't.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+
+## Conditional compilation
+[t:conditional-compilation]: #conditional-compilation
+
+Conditional compilation uses CPP conditionals to arbitrarily modify C
+declarations, or even leave them out entirely. Some conditional compilation
+happens without user intervention, for example as part of the standard libraries
+implementation. It can also be intended to be explicitly configured by the user.
```
edsko, line 83: "It can also be intended to be explicitly configured by the user" -- I had to read that a few times before I understood what you meant; I think you mean, some libraries need some CPP flags to be set to configure them, I guess? It might be good to add a few more words to this sentence to clarify it.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+`libclang` resolves these CPP conditionals before `hs-bindgen` gets to traverse
+them, so `hs-bindgen` cannot detect the difference between alternatives.
```
edsko, line 85: "between alternatives" I think makes the problem sound easier than it is ("just explore both"), and also doesn't match what you said at the start of the section ("arbitrarily modify declarations"); so perhaps we should simply say that `hs-bindgen` only sees the processed file, with all conditionals already resolved.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+by the public headers:
+
+```c
+/* generated by ./configure */
+#define HAVE_IPV6 1
+#define SIZEOF_LONG 8
+```
+
+Nothing in the public header itself signals that it is platform-specific, and
+`hs-bindgen` has no way to distinguish configure-generated content from
```
edsko, line 121: "has no way" I guess isn't strictly true.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+hand-written content. The generated bindings will simply reflect whichever
+`#define`s were in effect at the time.
+
+## Consequences
+[t:consequences]: #consequences
+
+Generated bindings can be platform-dependent because `libclang` makes
+platform-dependent decisions.
```
edsko, line 129: You just spent a lot of text explaining lots of different ways in which bindings can be platform specific, beyond what just clang does. "There are therefore many reasons why ..." or something like that might be better.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+
+What are some common ways in which non-portability materializes in generated
+bindings?
+
+### Memory layout of types
+[t:memory-layout]: #memory-layout-of-types
+
+C types such as structs and unions have a memory layout that is
+platform-dependent. The memory layout includes information about the type and
```
edsko, line 138: "includes information about " -> "is affected by" ?

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+its fields, such as size, alignment, offset, and (bit-)width. `hs-bindgen`
+generates bindings that hard-code these values, for example to implement class
+instances like `Storable`, `HasCField`, `HasCBitField`, and `HasField` (for
+union bit-fields). As such, bindings generated on one platform may not hard-code
```
edsko, line 142:
```suggestion
union bit-fields). As such, bindings generated on one platform may not use
```

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+the right values for a different platform.
+
+### Functions and FFI types
+[t:ffi-types]: #functions-and-ffi-types
+
```
edsko, line 147: Again, chicken and egg, I get it, but you introduced reasons for non-portability, but nothing so far has explained FFI types; I'm not sure what the fix is here, whether this should stay in consequences or not; maybe this deserves a page of its own, I'm not sure. But if we leave it here, then I think we should restructure this section a bit. _First_ state how hs-bindgen compiles functions, the FFI types thing, even if we omit rationale here (maybe a reference somewhere else), and then say that _therefore_ we get other non-portability.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+The precise reason why we use FFI types in function binding implementations is
+quite technical, but chiefly it ensures that the bindings that `hs-bindgen`
+generates are *compositional*. See the "Full example" dropdown below for a
+detailed explanation.
+
+<details>
+<summary>Full example</summary>
+
+This example is copied from [a Reddit comment by
+Edsko][reddit:ffi-types-example].
```
edsko, line 167: Maybe this should be that separate page I mentioned, and if so, it should be cleaned up a bit; no need to cite me or Reddit, just turn it into a proper manual page of its own.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+* `hs-bindgen-cli preprocess --hash-define USE_EXTENDED 1 ... foo.h`
+* `hashDefine "USE_EXTENDED" "1"` in Template Haskell
+
+A root directive is stated once and reaches every C stage: both the headers
+`libclang` parses and the generated C wrapper source that GHC compiles. This
+matters, because a header parsed under one set of macros and compiled under
+another gives missing symbols or, worse, silently mismatched `struct` layouts.
+Note that a `-D` passed via `--clang-option` or `BINDGEN_EXTRA_CLANG_ARGS` does
+*not* have this property: it affects binding generation alone. See [C
+stages][manual:c-stages].
```
edsko, line 348: This paragraph doesn't belong here at all; it's an explanation of compilation, binding generation, the scope of macros; it's a distraction here.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+## Solutions and mitigations
+[t:solutions]: #solutions-and-mitigations
+
+There are several strategies for managing non-portable bindings. In some cases,
+portability may be improved by user intervention.
+
+### Define macros explicitly
+
+State the macros the headers expect as `#define` root directives, so that the
+CPP resolution matches the intended target:
```
edsko, line 337: This whole page is about "Non-portability", so I was expecting a "solution" to be "how to make it portable". That's not what "matches the intended target" does, of course; not a problem in and of itself, but I think we should be clearer about the ways we work with non-portability: (1) accept it, and make sure that when we generate bindings, we set things up for the target pltform, (2) try to actually improve things so they they are more portable.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+
+* `hs-bindgen-cli preprocess --hash-define USE_EXTENDED 1 ... foo.h`
+* `hashDefine "USE_EXTENDED" "1"` in Template Haskell
```
edsko, line 340: You introduce `USE_EXTENDED`, but we're quite a few sections down already, and I had already completely forgotten that this was an example above in your struct; and others might jump straight to this section without reading that struct example at all. A back reference here I think would be good.

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+Macros that select an *implementation*, as in a header-only library, are the
+exception: they must be confined to a single translation unit. See [Header-only
+libraries][manual:c-stages-header-only].
+
+### Treat binding generation as part of the build process
+
+Do not assume that generated binding are portable across platforms. On each
+target platform, regenerate the bindings using that platform's headers and
+configuration (perhaps using `SetupHooks`). At present `hs-bindgen` does not yet
+provide any explicit support for making this process easier.
```
edsko, line 359: This isn't really mitigation at all, is it?

manual/low-level/usage/portability.md
```diff
@@ -0,0 +1,408 @@
...
+### Gate platform-specific packages
+
+If distributing generated bindings (e.g. on Hackage), gate them to the
+platform(s) they were generated on (or
+[cross-compiled][manual:cross-compilation] for). For example, in a `.cabal`
+file:
+
+```cabal
+if !(os(linux) && arch(x86_64))
+  buildable: false
```
edsko, line 370: Could we instead show an example that actually provides different implementations for different platforms, rather than just refusing to build on the one intended one?

manual/low-level/usage/portability/portable-types.md
```diff
@@ -0,0 +1,91 @@
+# Portable types
+
+`hs-bindgen` generates datatype bindings that are platform-dependent. For
+example, their implementation may reference platform-dependent [memory
```
edsko, line 4:
```suggestion
example, their implementation may depend on platform-specific [memory
```

manual/low-level/usage/portability/portable-types.md
```diff
@@ -0,0 +1,91 @@
...
+is portable. This manual section includes guidelines for writing portable types.
+
+> [!IMPORTANT]
+> Even with portable types, function bindings are currently not portable
+> due to [FFI types][manual:portability-ffi-types]. We are planning on
+> improving this. See [issue #1748][issue-1748].
+
+This manual section is not necessarily exhaustive. If there is missing
+information that you need, please let us know by creating an issue on the [issue
+tracker][github:issue-tracker].
```
edsko, line 17: undefined reference

manual/low-level/usage/portability/portable-types.md
```diff
@@ -0,0 +1,91 @@
...
+      return S{x, y}
+
+  poke s S{x, y} = do
+      (#poke struct S, x) s x
+      (#poke struct S, y) s y
+```
+
+> [!Note]
+> We have ideas about making generated `Storable` instances more portable right
+> out of the box. See [issue #1818][issue-1818].
```
edsko, line 50: It might be good to mention that `c2hsc` generates `hsc` files ; credit where credit is due and all that.

manual/low-level/usage/portability/portable-types.md
```diff
@@ -0,0 +1,91 @@
...
+Consider this example `typedef`:
+
+```c
+typedef int T;
+```
+
+Since the size of `int` is platform-dependent, `FFIType T` should also have a
+platform-dependent implementation. For example, if `int` is 32 bits on one
+platform, then `FFIType = Int32` should hold, and if `int` is 64 bits, then
+`FFIType = Int64` should hold, etc.
```
edsko, line 66: This is very difficult to understand without a deep understanding of FFI types. I think we really need to write that page.

manual/low-level/usage/portability/portable-types.md
```diff
@@ -0,0 +1,91 @@
...
+
+#### `IsArray`
+
+Portable types should get an `IsArray` instance (if applicable) where the
```
edsko, line 70: This needs more explanation also.

edsko (commented):
@jorisdral Sorry, I've left quite a few comments, don't be disheartened :) This is a good start, but I think needs more work. This is a tricky subject and an important one, I think it's worth spending some time on trying to get this as clear as we can.

manual/low-level/usage/portability.md
jorisdral: I get your point, we should move this proviso further down. But I think this proviso could already be relevant for the "Consequences" section. It's good to set boundaries *before* we start discussing how bindings are affected by platform-dependent information -- specially since conditional compilation can effect the bindings so much (in ways we might not expect)

manual/low-level/usage/portability.md
jorisdral: I suppose what I'm trying to say here is this: if any part of the source code of the generated Haskell bindings "hard-codes" the size of an `int` on the value or type level, then the bindings are non-portable. So for example, if we use `Int32` in a type or the value `32` in a function definition. Register passing and all that is not a concern for portability in this case, right? That should be handled all by the GHC/C compilers. 

Anyway, I'm trying to convey in broad terms how platform-specific decisions matter for portability. There are various places in the generated bindings where this can be a concern, but I don't want to make any elaborate statements just yet about how bindings are affected. The "example" is mainly intended as a kind of spoiler for the next "Consequences" section, where I intended the specifics to be discussed

I could either make the example a little more specific, or I could remove the examples and delay discussing how bindings are affected to the "Consequences" section. I may do the latter

manual/low-level/usage/portability.md
jorisdral: Makes sense

manual/low-level/usage/portability.md
jorisdral: Maybe my understanding is a bit limited here about how GHC and the FFI work during compilation. But part of the Haskell bindings are also the C wrappers and included headers, and these can be compiled as part of GHC using GCC/Clang, right?

manual/low-level/usage/portability.md
jorisdral: Right, make sense. The current statement is too strong.

manual/low-level/usage/portability.md
jorisdral: I can see how one would call this mitigation: `hs-bindgen` generates non-portable code, so to make that less painful, integrate `hs-bindgen` into your build plan so that it always generates the code for the platform you're using

manual/low-level/usage/portability.md
jorisdral: I wonder if we can write such an example without too much detail, but we can give it a try. I do think simply using `buildable : False` is a valid option, but maybe not the most useful one

manual/low-level/usage/portability.md
jorisdral: Maybe we should also try to pin down more precisely what "portability" means for the generated bindings... Maybe something along the lines of:

> The source code of the generated Haskell bindings "hard-codes" information that is platform-specific. 

I'm not sure if this is correct. Is it really only the source code we care about, or are there are more low-level concerns that we're forgetting about?

manual/low-level/usage/portability.md
jorisdral: Done

manual/low-level/usage/portability.md
jorisdral: No longer relevant: the structure of the document changed

manual/low-level/usage/portability.md
jorisdral: This section is now  removed, but it might be good to revive (parts) of it somewhere in the new version

manual/low-level/usage/portability.md
jorisdral: In V2, I emphasize that there are multiple sources of non-portability, but that non-portability overall is often caused by interplay between them. So the cause vs. consequence distinction kind of goes away

manual/low-level/usage/portability/portable-types.md
jorisdral: No longer relevant since FFI types are refactored

manual/low-level/usage/portability.md
jorisdral: I rephrased this in V2

manual/low-level/usage/portability.md
jorisdral: Removed in V2

manual/low-level/usage/portability.md
jorisdral: Removed in V2

manual/low-level/usage/portability.md
jorisdral: Done

manual/low-level/usage/portability.md
jorisdral: V2 still says something about "compiling Haskell bindings", but feel free to take it out if you think it's wrong

manual/low-level/usage/portability.md
jorisdral: In V2 I make it more concrete how a compiler decision can affect the generated bindings using an example

manual/low-level/usage/portability.md
jorisdral: Removed this in V2

manual/low-level/usage/portability.md
jorisdral: Done

manual/low-level/usage/portability.md
jorisdral: Rephrased this in V2

manual/low-level/usage/portability.md
jorisdral: Removed in V2

manual/low-level/usage/portability.md
jorisdral: Removed in V2

manual/low-level/usage/portability.md
jorisdral: We have a dedicated FFI types section now, but the Portability section does not link to it yet

manual/low-level/usage/portability.md
jorisdral: Done

manual/low-level/usage/portability.md
jorisdral: Removed in V2

manual/low-level/usage/portability.md
jorisdral: Removed in V2

manual/low-level/usage/portability.md
jorisdral: Removed in V2

manual/low-level/usage/portability/portable-types.md
jorisdral: Removed in V2

manual/low-level/usage/portability/portable-types.md
jorisdral: Removed in V2

manual/low-level/usage/portability/portable-types.md
jorisdral: Done

manual/low-level/usage/portability.md
jorisdral: Done in V2

manual/low-level/usage/portability.md
jorisdral: V2 still says "hard-code" rather than "use". I'm undecided about whether "use" is better than "hard-code". I don't think "use" conveys that the bindings specifically use the value, whereas "hard-code" does that better IMO

manual/low-level/usage/portability.md
jorisdral: This distinction you make between (1) and (2) is not represented in V2 of the binding yet, but I think it is useful to talk about both.

manual/low-level/usage/portability/portable-types.md
jorisdral: We should still add this to V2

manual/low-level/usage/portability/portable-types.md
jorisdral: This is not yet included in V2

Issue #2219, opened 2026-08-19
Dependency order depends on `Digraph` vertex-insertion history


dschrempf:
I think this is a known limitation @TravisCardwell which may become a problem if we decide to move to a declaration database as single source of truth (#1580).

### Current limitation

At the moment, `toDecls` establishes the ordering of declarations in the frontend:

```haskell
-- hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclUseGraph/Query.hs:34
toDecls index declUseGraph =
    mapMaybe (`DeclIndex.lookup` index) . Digraph.sort $
      Digraph.filterEdges (== NeedsShape) declUseGraph.graph
```

`Digraph.sort` breaks ties by internal vertex index (`Data/Digraph.hs:442`), and internal indices are assigned in *insertion* order:

```haskell
-- hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclUseGraph/Construction.hs:66
-- We first insert all declarations, so that they are assigned vertices. For
-- successfully parsed declarations, we do this in source order. This
-- ensures that we preserve source order as much as possible in 'toDecls'
-- (modulo dependencies).
verticesGraph = foldl' (flip Digraph.insertVertex) Digraph.empty $
  map (.info.id) sortedSuccessfulDecls ++ Set.toList failedDeclIds
```

So the order in the frontend is a function of the *history* of the graph value, not of the graph itself.

This may be problematic, because:

- Two passes mutate the graph afterwards: `ReparseMacroExpansions.updateDeps` deletes and re-inserts a declaration's edges (`Pass/ReparseMacroExpansions.hs:106-112`), and `ResolveBindingSpecs` deletes dependencies and reverce dependencies (`Pass/ResolveBindingSpecs.hs:70-73`). Calling `toDecls` again after those may not reproduce the original order.

- The `DeclDB` design (#1580) turns the declaration list into a view computed on demand, and we run into the above reproducibiliyt problem.

### Suggested fix

Make `dependencyOrder` a function of the graph data only:

```haskell
dependencyOrder :: (C.DeclId -> SortKey) -> DeclUseGraph -> [C.DeclId]
dependencyOrder sortKey g =
    Digraph.sortBy (comparing sortKey) $
      Digraph.filterEdges (== NeedsShape) g.graph
```

and remove the pre-sorting during construction mentioned above.

[...]

Issue #742, opened 2025-06-14
Tooling to assist users for creating bindings for multiple C headers


edsko:
Many C libraries consist of multiple C headers, with a potentially complicated dependency graph. External bindings allows users to process these headers one by one, but this must be done in the right order: if `A` depends on `B`, `B` must be done first; see detailed discussion at https://github.com/well-typed/hs-bindgen/issues/75#issuecomment-2702297396 .  It would be nice to provide users with some tooling to make this process easier.

edsko:
As of #1085 users can get a include graph with `info include-graph`.

edsko:
See also @TravisCardwell 's detailed discussion of how he currently does this by hand at https://github.com/well-typed/hs-bindgen/issues/91#issuecomment-3051325770 . @dschrempf , this may also be useful for the tutorial you are writing.

TravisCardwell:
Note that we can now output the include graph, using `hs-bindgen-cli info include-graph`.

edsko:
Going to do this for the first release, since we're waiting for some people to come back from holidays anyway, and @bolt12 already has some logic doing exactly this, but as a script that calls bindgen rather than a proper bindgen feature.

TravisCardwell:
Yesterday, @bolt12 and I discussed some `HashIncludeArg` complications.  We can use the include graph to determine the order that modules need to be created, *but* main headers must be specified as `BracketInclude` `HashIncludeArg`s.  What should we do when a header only has `QuoteInclude` `HashIncludeArg`s?

We could group headers included via `QuoteInclude`s into single modules.  The (unfortunate) result would be that it would likely group more aggressively than users desire.  It is common for libraries to use `QuoteInclude`s internally, so many libraries would end up being translated to single modules, defeating the purpose of the tooling.

Alternatively, we can try to translate all `Include`s to `BracketInclude`s.  Since all edges from the root are `BracketInclude`s, we could calculate a `BracketInclude` using any traversal from the root to the target.

```mermaid
graph TD;
  A["(root)"];
  B["/path/to/include/widget.h"];
  C["/path/to/include/widget/core.h"];

  A-->|"BracketInclude widget.h"| B;
  B-->|"QuoteInclude widget/core.h"| C;
```

```haskell
>>> normalise $ takeDirectory "widget.h" </> "widget/core.h"
"widget/core.h"
```

```mermaid
graph TD;
  A["(root)"];
  B["/path/to/include/widget/legacy/widget.h"];
[...]

PR #2239, opened 2026-08-28
Update `Digraph` API

TravisCardwell:
This API is internal, not user-facing.

This PR includes two changes:

* The `sortBy` convenience function is removed.  See https://github.com/well-typed/hs-bindgen/issues/2219#issuecomment-5446897664
* A new `sortSccs` function is exposed, for use in #742.  This is a trivial change, just a matter of factoring `concat` out of `sort`.

### DIFF
--- a/hs-bindgen/src-internal/Data/Digraph.hs
+++ b/hs-bindgen/src-internal/Data/Digraph.hs
@@ -28,7 +28,6 @@ module Data.Digraph (
   , neighbors
   , reaches
   , sort
-  , sortBy
   , dfs
   , dff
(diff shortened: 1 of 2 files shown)


edsko:
After discussing this with @TravisCardwell again, we decided to close this PR. While an "ordered graph" could be useful concept in its own right, it's not really the right concept for bindgen, as we want different orderings at different times. Therefore, `reIndex` should be considered mostly an internal function, and `sortBy` should be the main driver for traversing the graph in a _call-site_ determined order (not construction site).

PR #2240, opened 2026-08-28
Add `TranslateTypes` pass infrastructure

### LINKED ISSUE #1599: Introduce "type translation" frontend pass
We want to do instance resolution in the front end, so that

* we can issue warnings (#1009, #1604)
* we can take advantage of the pass architecture, so that instance resolution becomes a separate concern

However, since instance resolution relies on _Haskell_ types, it makes sense to translate types (at _use sites_, not declaration sites) to Haskell types as a separate frontend pass, to run prior to instance resolution.

We can take advantage of this pass in other places too; for example, @jorisdral 's work on foreign import wrappers currently needs to make a choice between basing decisions on C types (which would be a more natural match but would require re-doing the translation logic to Haskell types) or Haskell types (which results in awkward unnecessary decisions because now there are "impossible cases" to consider); having the translation available in the AST would resolve this problem.

jorisdral:
Part of #1599

The pass infrastructure is in place, but we do not yet translate C types to Haskell types in the frontend.

---

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/Deps.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/Deps.hs
@@ -9,6 +9,8 @@ module HsBindgen.Frontend.Analysis.Deps (
   , depsOfDeclParsedMacro
   ) where
 
+import Data.Proxy (Proxy (..))
+
 import HsBindgen.Frontend.Analysis
 import HsBindgen.Frontend.Pass.ReparseMacroExpansions.IsPass
 import HsBindgen.Frontend.Pass.TypecheckMacros.IsPass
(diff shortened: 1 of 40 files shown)


hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/Deps.hs
```diff
@@ -35,9 +37,9 @@ depsOfDeclWith depsOfMacro = \case
     C.DeclOpaque{}                 -> []
     (C.DeclMacro m)                -> depsOfMacro m
     (C.DeclFunction function)      ->
-      C.depsOfType function.res ++
-      concatMap (\arg -> C.depsOfTypeFunArg arg.argTyp) function.args
-    (C.DeclGlobal global)          -> C.depsOfType global.typ
+      C.depsOfType (cType (Proxy @p) function.res) ++
+      concatMap (\arg -> C.depsOfType (cType (Proxy @p) arg.typ)) function.args
```
jorisdral: C.depsOfType (C.typeOfFunction function)

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/Deps.hs
```diff
@@ -109,23 +111,23 @@ depsOfRegularField ::
...
   -> [(Id p, Dependency)]
-depsOfRegularField field = C.depsOfType field.typ
+depsOfRegularField field = C.depsOfType (cType (Proxy @p) field.typ)
 
 depsOfImplicitField ::
      forall p. IsPass p
   => C.ImplicitField p
   -> [(Id p, Dependency)]
-depsOfImplicitField field = C.depsOfType field.typ ++ concatMap depsOfIndirectField field.indirect
+depsOfImplicitField field = C.depsOfType (cType (Proxy @p) field.typ) ++ concatMap depsOfIndirectField field.indirect
```
jorisdral: wrap

PR #2248, opened 2026-08-31
hs-bindgen-runtime: raw macro type

### LINKED ISSUE #2242: Raw macros should be a dedicated runtime type, not `[String]`
The `Raw` macro language (`HsBindgen.Macro.Raw`) renders every macro as a `[String]` of token spellings:

```haskell
-- #define AV_VERSION_INT ...
aV_VERSION_INT :: [String]
aV_VERSION_INT = [...]
```

In my opinion, `[String]` is not expressive enough because:

- A consumer cannot tell a raw macro from any other `[String]` binding.
- `[String]` cannot express the object-like/function-like distinction (see issue #2243).
- The original spelling cannot be reconstructed, because `[String]` loses whether adjacent tokens were separated by white space. <- I am not sure if we actually need to reconstruct the whitespace.

### Proposal

Add a type to `hs-bindgen-runtime`. Sketch:

```haskell
-- HsBindgen.Runtime.RawMacro
data RawMacro =
    ObjectLike   { tokens :: [Token] }
  | FunctionLike { parameters :: [Text], tokens :: [Token] }
```

with `Token` carrying at least the spelling. We may also add information to re-render the macro body including whitespace.

Either add `render :: RawMacro -> Text` or a `Show`/`Pretty` instance.

### Notes

- The issue "function-like macros are mangled" (#2243) needs this type; the two should probably be implemented together.
- Any per-macro macro fallback (#368) will share this renderer.

### LINKED ISSUE #2243: `Macro.Raw` splices the parameter list into the body of function-like macros
The `Raw` macro language (`HsBindgen.Macro.Raw`) renders every macro as its list of token spellings. This is also true for function-like macros:

```haskell
-- #define INCR(x) x + 1
iNCR :: [String]
iNCR = ["(", "x", ")", "x", "+", "1"]

-- #define ADD(x, y) x + y
aDD :: [String]
aDD = ["(", "x", ",", "y", ")", "x", "+", "y"]
```

A user cannot recover the macro from this. `["(", "x", ")", "x", "+", "1"]` is indistinguishable from the token list of an object-like macro whose body happens to begin with a parenthesized expression.

### Cause

`parseRaw` (`hs-bindgen/src-internal/HsBindgen/Macro/Raw.hs:56`) takes the first token as the name and *everything else* as the body:

```haskell
parseRaw (t:ts) = Right $ Macro.Unresolved $ RawMacro{
      name   = t.tokenSpelling.getTokenSpelling
    , tokens = ts
    }
```

For a function-like macro, `ts` begins with the parameter list.

### Expected

The raw representation should distinguish object-like from function-like macros and keep the parameters separate from the body, e.g.

```haskell
iNCR = functionLike ["x"] ["x", "+", "1"]
aDD  = functionLike ["x", "y"] ["x", "+", "y"]
```

The exact rendering depends on the representation chosen in issue #2242 (replacing `[String]` with an ADT); these two should probably be implemented together.

### How to detect function-likeness

libclang exposes `clang_Cursor_isMacroFunctionLike`, the binding is available but that module is an `other-module`, so the function is not reachable from `hs-bindgen`. We'd need to expose it from `Clang.LowLevel.Core` upstream.

We could also implement this locally: test whether the token following the name is `(` with no whitespace between the two.

### Notes

- This also affects any future per-macro fallback (#368), which will share the raw renderer.
- `#define FOO()` (function-like, no parameters) must render differently from `#define FOO` (object-like, empty body).

dschrempf:
Add `HsBindgen.Runtime.Macro`: `Raw`, the type of a C macro kept as its token spelling.

Raw macros can be object-like and function-like forms.

The function `render` renders a macro back to its `#define` directive (module whitespace).

Refs #2242 #2243.

Based on #2251.

---

```hs
-- TODO <link to issue>
-- Brief description
```


hs-bindgen/src-internal/HsBindgen/Macro/UniqueExpansion.hs
```diff
@@ -91,16 +116,25 @@ parseInvocation inv =
...
+-- In particular, /a macro can be ambiguous even when it is defined only once/.
+-- For example,
 --
 -- @
 -- #define A Foo
 -- #define B A
 -- #define A Bar
--- #define B A
+-- B x;
 -- @
```
jorisdral, line 132: The point of the original example was to emphasise that even if we have two definitions for `B` that have the same body, they may expand to different things. It may be useful to revive it in some form in addition to the new example (which I think is a better first example to show)

hs-bindgen/src-internal/HsBindgen/Macro/UniqueExpansion.hs
```diff
@@ -161,6 +195,13 @@ cachedIsExpansionUnique cache pInv =
 -- | Collect the names of all macros that are defined more than once, and all
 -- macros that (transitively) depend on macros that are defined more than once.
 --
+-- We seed on the /number/ of definitions, without comparing their bodies. That
+-- is a deliberate over-approximation: @\#define B 5@ twice is reported as
+-- ambiguous, although the expansion is stable. Comparing bodies would be more
+-- precise, but it is not obviously the cheaper or the safer choice — two
+-- token-identical definitions can still expand differently, since the tokens in
+-- a replacement list are resolved at the invocation site.
```
jorisdral, line 203: Wild idea: If we see only `CXToken_Literal` in a macro body, we could check the token spellings for equality. If all token spellings are equal, then the macro definitions must be equal (and non-ambiguous). I'm not sure if it is worth it to implement soon -- maybe we should only do it if someone asks for it explicitly?

https://clang.llvm.org/doxygen/group__CINDEX__LEX.html#gaf63e37eee4280e2c039829af24bbc201

jorisdral: @dschrempf also noted that a much better algorithm for checking unique expansions would be to expand macros recursively and check if the expansions are equal. However, we can't invoke the Clang preprocessor easily to do this, so it doesn't sound doable at the moment, but it would be good to make note of it here

hs-bindgen/src-internal/HsBindgen/Macro/UniqueExpansion.hs
dschrempf: I will copy this comment to the UniqueExpansion issue #2264, and fix it there.

hs-bindgen/src-internal/HsBindgen/Macro/UniqueExpansion.hs
dschrempf: I will come back to this when implementing #2264. It will not be forgotten.

manual/terminology.md
```diff
@@ -391,10 +391,12 @@ per [regular field][t:regular-field] of a struct; [unnamed][t:unnamed-field]
...
 
-The pluggable `hs-bindgen` component that parses, typechecks and translates
-[macro definitions][t:macro-definition]: the [macro name][t:macro-name], the
-[parameters][t:macro-parameter], and the [replacement list][t:replacement-list].
-See [macro languages][manual:translation/macros-macro-languages].
+The pluggable `hs-bindgen` component that parses, typechecks and translates the
+[replacement list][t:replacement-list] of a [macro
+definition][t:macro-definition]. The [macro name][t:macro-name] and the
+[parameters][t:macro-parameter] are not its concern: they are split off the same
+way for every macro language. See [macro
```
dschrempf: This is wrong. On this branch, the macro language still parses the complete macro definition.

edsko (approved):

PR #2249, opened 2026-08-31
Switch include graph and binding specs to RealPath

### LINKED ISSUE #2236: External binding specs are matched on path spellings, so a header reached under another spelling does not match
An external binding specification is ignored when a header (compat/legacy.h below) reaches one of its
dependency header (core.h below) by a different `#include` spelling ("../core.h" instead of "core.h") than the one it was generated under.

Reproduced on `main`.

## Reproduction

```c
// include/widget/core.h
struct widget_t { int id; };

// include/widget/compat/legacy.h
#include "../core.h" // Different spelling, i.e., not "core.h"
typedef struct widget_t widget_legacy_t;
```

Generate the dependency, then the dependent, passing the specification:

```console
$ hs-bindgen-cli preprocess -I include --module W.Widget.Core \
    --gen-binding-spec core.yaml  widget/core.h

$ hs-bindgen-cli preprocess -I include --module W.Widget.Compat.Legacy \
    --external-binding-spec core.yaml  widget/compat/legacy.h

[Warning] [select] 'widget_legacy_t' at "include/widget/compat/legacy.h 4:25":
  Could not select declaration (direct selection predicate match):
      'struct widget_t' at "include/widget/compat/../core.h 5:8"
```

No module is written.

With `--enable-program-slicing`, it is silent:

```console
$ hs-bindgen-cli preprocess -I include --enable-program-slicing \
    --module W.Widget.Compat.Legacy --external-binding-spec core.yaml \
    widget/compat/legacy.h
$ echo $?
0
```

Nothing is printed. `W.Widget.Compat.Legacy` does not import `W.Widget.Core`, and
defines its own `Widget_t`. There are now two distinct Haskell types for one C type.

## Why

A binding specification says "this type lives in the C header x.h". To apply it, hs-bindgen
has to decide whether the declaration it just parsed came from x.h, and it decides that
by comparing file paths.

But clang does not report a canonical path for a C header. It reports whichever
`#include` spelling first opened the file, and keeps using that name for the rest of
the translation unit. So the same file gets different paths depending on how it was
reached:

- generating `core.h`, it is opened by `#include <widget/core.h>` and reported as
  `include/widget/core.h`;
- generating `legacy.h`, it is opened by `legacy.h`'s `#include "../core.h"` and
  reported as `include/widget/compat/../core.h`.

The specification's key resolves to the first path. The declaration is at the second.
The comparison fails, so the specification is ignored.

The same thing can happen when a header is reached through a symbolic link, for example
an include directory that is a symlink to a sibling.

bolt12:
Closes #2236


edsko (commented):
After a brief discussion, we'll go with clang_File_tryGetRealPathName returning a `RealPath`, no sum type, and throw an exception when its precondition (file on disk) is violated. Then hs-bindgen side we will use `RealPath` for identity comparisons.

hs-bindgen/src-internal/HsBindgen/Frontend/ProcessIncludes.hs
```diff
@@ -194,26 +202,28 @@ toGetMainHeaders f = fmap fst . f
...
+--   on-disk path, since @#include@ targets are always real files)
+-- * 'inRoot'  will be 'True' when the directive is in the root header
 --
--- The full 'HsBindgen.Clang.HighLevel.Types.SourcePath's are constructed by @libclang@, and depend on factors
--- such as @-I@ command line arguments, environment variables such as
+-- Both paths depend on how @libclang@ resolves headers, which is affected by
+-- @-I@ command line arguments, environment variables such as
 -- @C_INCLUDE_PATH@, etc.
 data IncDir = IncDir {
       from    :: SourcePath
```
bolt12, line 216: Maybe we don't need to hold the root header in the include graph. Look into where it is used

hs-bindgen/src-internal/HsBindgen/Frontend/ProcessIncludes.hs
```diff
@@ -194,26 +202,28 @@ toGetMainHeaders f = fmap fst . f
...
 -- Then
 --
--- * 'from'    will be @/full/path/to/a.h@
--- * 'include' will be @#include "b.h"@ (exact path as in source)
--- * 'to'      will be @/full/path/to/b.h@
--- * 'inRoot'  will be 'True' if the include is in the root header
+-- * 'from'    will be @/full/path/to/a.h@ (a 'SourcePath' type, since the source
+--   of the directive can be the root header, which is an in-memory file)
+-- * 'include' will be @#include "b.h"@ (the path exactly as written in source)
+-- * 'to'      will be @/full/path/to/b.h@ (a 'RealPath' type, the canonical
```
bolt12, line 208: mention that this is also so that the include graph doesn't have redundant vertices pointing to the same files

hs-bindgen/src-internal/HsBindgen/Frontend/ProcessIncludes.hs
```diff
@@ -227,10 +237,9 @@ processInclude unit curr = do
   Internal auxiliary
 -------------------------------------------------------------------------------}
 
-getIncludeTo :: MonadIO m => CXCursor -> m SourcePath
-getIncludeTo curr = do
-    file <- clang_getIncludedFile curr
-    SourcePath <$> clang_getFileName file
+getIncludeTo :: (MonadIO m, HasCallStack) => CXCursor -> m RealPath
```
bolt12, line 240: Maybe rename this to `getIncludeTarget`

hs-bindgen/src-internal/HsBindgen/Resolve.hs
bolt12, line 90:
```suggestion
  -> IO (Map C.HashIncludeArg RealPath)
```

hs-bindgen/src-internal/HsBindgen/Resolve.hs
```diff
@@ -135,8 +135,7 @@ resolveHeaders tracer args headers =
...
           headerList !? (singleLocLine sloc - 1)
-        path <- clang_getFileName =<< clang_getIncludedFile curr
-        return $
-          if Text.null path
-            then Left header
-            else Right (header, SourcePath path)
+        mRealPath <- HighLevel.clang_tryGetRealPath =<< clang_getIncludedFile curr
+        return $ case mRealPath of
+          Nothing -> Left header
+          Just rp -> Right (header, realPathToSourcePath rp)
```
bolt12, line 141: This is going to return a value for everything except for the root header so we should throw here, i.e. use `clang_getRealPath`

edsko:
@TravisCardwell What kind of path should be even have for the `headers:` field in a binding spec? Perhaps we should discuss next week.

hs-bindgen/test-artefacts/headers/golden/binding-specs/include_spelling/sub/consumer.yaml
```diff
@@ -0,0 +1,24 @@
+version:
+  hs_bindgen: 0.1.0
+  binding_specification: '1.0'
+hsmodule: M
+ctypes:
+- headers: binding-specs/include_spelling/core.h
```
bolt12, line 6: @edsko should be HashIncludeArgs (?)

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Haddock/Translation.hs
```diff
@@ -332,13 +332,13 @@ convertInline = \case
...
 --
 -- See #966.
-updateSingleLoc :: PathStyle -> C.SingleLoc -> C.SingleLoc
+updateSingleLoc :: PathStyle -> C.SingleLoc C.RealPath -> C.SingleLoc C.RealPath
 updateSingleLoc Short C.SingleLoc{..} =
   C.SingleLoc {
-    singleLocPath = C.SourcePath
+    singleLocPath = C.RealPath
                   . Text.pack
                   . takeFileName
```
bolt12: This should be the a bit more than just a file name (e.g HashIncludeArg)

hs-bindgen/src-internal/HsBindgen/Clang/Macros/UniqueExpansion/Parse/Infra.hs
bolt12, line 1: Try out using RealPath for every Token usage and macro stuff

hs-bindgen/src-internal/HsBindgen/Frontend/Analysis/DeclUseGraph/Construction.hs
```diff
@@ -148,10 +148,7 @@ annSortKey order decl = SortKey{
     , sortColNo  = singleLocColumn decl.info.loc
     }
   where
-    path = singleLocPath decl.info.loc
-    -- Unlike a trace message, a declaration whose source is not in the include
-    -- graph means the frontend is broken; there is nothing to salvage.
-    sortPathIx = case IncludeGraph.lookupIncludeOrder order path of
+    sortPathIx = case IncludeGraph.lookupIncludeOrder order (singleLocPath decl.info.loc) of
```
bolt12: restore the comment

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/Macro.hs
```diff
@@ -18,6 +18,9 @@ getReparseInfo = \curr -> do
...
       Just macroExpansions -> do
-        unit <- getTranslationUnit
-        ReparseNeeded <$>
-          HighLevel.clang_tokenize unit extent <*> pure macroExpansions
+        unit     <- getTranslationUnit
+        rawRange <- fromRange unit extent
+        tokens   <- HighLevel.clang_tokenize unit rawRange
+        case tokens of
+          [] -> pure ReparseNotNeeded
+          _  -> pure $ ReparseNeeded tokens macroExpansions
```
bolt12: Remove the extra case here

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -78,17 +78,12 @@ parseDeclTopLevel ::
...
-    -- directive in it has no main header and 'withHeaderInfo' would fail the
-    -- parse (and it shouldn't have one). We consider root directives
-    -- configuration, not API. Do not attempt to parse it.
-    if RootHeader.isRootHeaderPath (singleLocPath declLoc) then foldContinue else do
+    -- Check for root-header cursors via the low-level API before building a
+    -- SingleLoc: clang_getCursorLocation' throws for virtual files.
+    loc <- clang_getCursorLocation curr
+    (file, _line, _col, _off) <- clang_getExpansionLocation loc
+    path <- clang_getFileName file
+    if RootHeader.isRootHeaderPath (SourcePath path) then foldContinue else do
```
bolt12: this doesn't look right.

hs-bindgen/src-internal/HsBindgen/Frontend/ProcessIncludes.hs
bolt12, line 1: Remember to ask @TravisCardwell about this once the PR is approved

hs-bindgen/src-internal/HsBindgen/Frontend/RootHeader.hs
```diff
@@ -72,8 +73,8 @@ content (RootHeader directives) = C.renderRootDirectives directives
 isRootHeaderPath :: SourcePath -> Bool
 isRootHeaderPath = (== name)
 
--- | Check if the specified location is in the root header
-isInRootHeader :: MultiLoc -> Bool
+-- | Check if the given location is in the root header
```
bolt12, line 79: Add a comment saying this takes a SourcePath because it is used with Clang Diagnostics.

hs-bindgen/src-internal/HsBindgen/Clang.hs
```diff
@@ -151,8 +150,7 @@ instance PrettyForTrace ClangMsg where
                 Just header ->
                   "unable to resolve #include <" <> header
                     <> "> (must specify header relative to directory in C include search path)"
-                Nothing     ->
-                  Text.stripStart $ Text.dropWhile (/= ' ') diag.diagnosticFormatted
+                Nothing     -> diag.diagnosticFormatted
```
bolt12: this doesn't look right

hs-bindgen/src-internal/HsBindgen/Resolve.hs
```diff
@@ -123,20 +105,22 @@ resolveHeaders tracer args headers =
...
-          headerList !? (singleLocLine sloc - 1)
-        path <- clang_getFileName =<< clang_getIncludedFile curr
-        return $
-          if Text.null path
-            then Left header
-            else Right (header, SourcePath path)
+          headerList !? (fromIntegral line - 1)
+        rp <- MaybeT
+            $ HighLevel.clang_tryGetRealPath =<< clang_getIncludedFile curr
+        return (header, rp)
```
bolt12: This is wrong, for HashIncludeArgs that can't be resolved because they don't exist anywhere (e.g. come from a bindingspec)

hs-bindgen/clang-ast-dump/Main.hs
```diff
@@ -92,9 +92,9 @@ clangAstDump opts = do
           loc <- clang_getPresumedLocation =<< clang_getCursorLocation cursor
           case loc of
             (file, _, _)
-              | opts.sameFile && SourcePath file /= src -> foldContinue
-              | not opts.builtin && isBuiltIn file      -> foldContinue
-              | otherwise                               -> foldDecls opts cursor
+              | opts.sameFile && file /= getRealPathText src -> foldContinue
+              | not opts.builtin && isBuiltIn file             -> foldContinue
+              | otherwise                                      -> foldDecls opts cursor
```
bolt12: align this code

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl/Macro.hs
```diff
@@ -18,6 +19,6 @@ getReparseInfo = \curr -> do
     case macroExpansionsMay of
       Nothing -> pure ReparseNotNeeded
       Just macroExpansions -> do
-        unit <- getTranslationUnit
-        ReparseNeeded <$>
-          HighLevel.clang_tokenize unit extent <*> pure macroExpansions
+        unit   <- getTranslationUnit
+        tokens <- HighLevel.clang_tokenize unit getRealPathText extent
+        pure $ ReparseNeeded tokens macroExpansions
```
bolt12: Try to minimize the diff here vs the original code

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -250,11 +251,11 @@ macroDefinition macroLang _enclosing ctx info = \curr -> do
 
     getMacroTokens ::
          CXCursor
-      -> ParseDecl (Range MultiLoc, [Token TokenSpelling])
+      -> ParseDecl (Range (MultiLoc RealPath), [Token SourcePath TokenSpelling])
     getMacroTokens curr' = do
         unit'  <- getTranslationUnit
         range  <- HighLevel.clang_getCursorExtent curr'
-        (range,) <$> HighLevel.clang_tokenize unit' (multiLocExpansion <$> range)
+        (range,) <$> HighLevel.clang_tokenize unit' getRealPathText (fmap multiLocExpansion range)
```
bolt12:
```suggestion
        (range,) <$> HighLevel.clang_tokenize unit' getRealPathText (multiLocExpansion <$> range)
```

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Parse/Decl.hs
```diff
@@ -472,13 +473,13 @@ macroExpansion = \curr -> do
   where
     getTokens ::
          CXCursor
-      -> ParseDecl (Range MultiLoc, [Token TokenSpelling])
+      -> ParseDecl (Range (MultiLoc RealPath), [Token SourcePath TokenSpelling])
     getTokens curr' = do
         unit'  <- getTranslationUnit
         range  <- HighLevel.clang_getCursorExtent curr'
-        (range,) <$> HighLevel.clang_tokenize unit' (multiLocExpansion <$> range)
+        (range,) <$> HighLevel.clang_tokenize unit' getRealPathText (fmap multiLocExpansion range)
```
bolt12:
```suggestion
        (range,) <$> HighLevel.clang_tokenize unit' getRealPathText (multiLocExpansion <$> range)
```

hs-bindgen/src-internal/HsBindgen/Frontend/Pass/Select.hs
```diff
@@ -589,11 +588,13 @@ data MsgSortKey =
...
-      , msgLine   = singleLocLine   loc
-      , msgColumn = singleLocColumn loc
-      }
+    Just loc ->
+      let source = IncludeGraph.lookupIncludeOrder order (singleLocPath loc)
+      in  MsgAt {
+              msgSource = source
+            , msgLine   = singleLocLine   loc
+            , msgColumn = singleLocColumn loc
+            }
```
bolt12:
```suggestion
    Just loc -> MsgAt {
        msgSource = IncludeGraph.lookupIncludeOrder order (singleLocPath loc)
      , msgLine   = singleLocLine   loc
      , msgColumn = singleLocColumn loc
      }
```

hs-bindgen/src-internal/HsBindgen/Macro/Parse.hs
bolt12: SourcePath is correct. `clang_tokenize` in `libclang-bindings` returns `Token SourcePath` because `clang_getFileName` gets called internally. Switching to toRangeRealPath could throw `ClangRealPathException` for virtual files.

PR #2250, opened 2026-09-02
`Golden`: set trace verbosity to `Debug`

jorisdral:
So that we can write trace predicates on debug-level trace messages.

This doesn't affect existing tests because default trace predicates mark debug-level messages as `Tolerated`. There are two exceptions where a custom trace predicate would mark debug-level messages as `Unexpected`, but these were easily fixed.

I was racking my brain for quite a while over why I couldn't write trace predicates over debug-level trace message in golden tests, and the verbosity setting was why. Hopefully by changing the verbosity to `Debug` now, others won't run into the same issue in the future.

---

### DIFF
--- a/hs-bindgen/test/common/Test/Common/HsBindgen/Trace/Predicate.hs
+++ b/hs-bindgen/test/common/Test/Common/HsBindgen/Trace/Predicate.hs
@@ -201,7 +201,9 @@ withTraceConfigPredicate report (TracePredicate customLogLevel predicate) action
         tracerConfig :: TracerConfig l a
         tracerConfig = def {
             customLogLevel = customLogLevel
-          , verbosity      = Verbosity Info
+            -- The verbosity is @Debug@ so that trace predicates can match on
+            -- trace messages of /every/ log level.
+          , verbosity      = Verbosity Debug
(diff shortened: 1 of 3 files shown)


dschrempf:
The only drawback I can see is that when a test fails, it floods the buffer with tons of debug messages.

But you are right, of course. We _must_ be able to match debug messages.

Maybe we should filter messages before actually printing them (when we fail). This filter should also be configurable, on CI you may want to print Debug messages on failures.

jorisdral:
> The only drawback I can see is that when a test fails, it floods the buffer with tons of debug messages.

> Maybe we should filter messages before actually printing them (when we fail). This filter should also be configurable, on CI you may want to print Debug messages on failures.

Good point! I should check if this happens. If so, then we should do some filtering as you suggested

jorisdral:
Actually, this filtering is already applied by the `TestCase`'s trace predicate: when a test fails, only the traces that are marked as `Unexpected` are traced. And these do not include debug-level traces

We could implement a feature for the golden tests that allows users to configure whether they want test failures to print more trace messages than just the `Unexpected` ones. But that should then probably be a separate issue

PR #2251, opened 2026-09-02
`UniqueExpansions`: support variadic macros

### LINKED ISSUE #2245: `UniqueExpansion` silently misreads variadic macros as object-like
`parseParams` in `HsBindgen.Clang.Macros.UniqueExpansion.Parse.hs` only accepts
identifiers:
```haskell
parseParams :: Parser [Name]
parseParams = parens $ parseName `sepBy` comma
```

In particular, it fails on `...`,

```c
#define LOG(fmt, ...) fmt
```

`sepBy` stops after `fmt`, the closing `)` is never reached, and the `functionLike` branch fails. `parseDefinition` uses `choice [try functionLike, objectLike]`, so it backtracks and parses the macro as **object-like** instead.

The GNU named form `#define F(args...) ...` fails the same way.

## Effect

A variadic macro can be marked ambiguous for the wrong reason.

Reproduce with two headers that differ only in `...`:

```c
/* w.h */
#define A int
#define A int
#define ID(A, ...) A
typedef ID(int, 0) t;
```

```c
/* w2.h */
#define A int
#define A int
#define ID(A, B) A
typedef ID(int, 0) t;
```

```
cabal run hs-bindgen-cli -- internal frontend --pass final -v4 w.h
```

`w.h` reports

```
[Info] [select-prepare-reparse] 't':
  During prepare-reparse:
    Can not expand macro invocations unambiguously, which may cause reparsing
    to fail. At "w.h 4:1" : "typedef ID ( int , 0 ) t;"
```

`w2.h` does not. `A` is ambiguous in both files, but only in `w.h` does that ambiguity wrongly reach `ID`, because there `A` is seen as a free variable rather than a parameter.

## Also affected

`HsBindgen.Macro.Raw` is meant to share this parser once #2243 is fixed. A variadic macro would then be rendered as object-like.

## What to do

- Parse `...` and `name...` in a macro parameter list.
- Decide how to represent variadic-ness in the parsed macro.
- Add test headers. There is currently no `...` in any header under `hs-bindgen/test-artefacts/headers/golden/macros/`.

## Note

`c-expr-dsl` (0.1.0.1 pinned version) knowingly doesn't support neither form either. Its parameter parser has the same shape:

```haskell
formalParams :: Parser [Identifier]
formalParams = parens $ parseIdentifier `sepBy` comma
```

and `Macro` stores `macroParams :: Vec ctx Identifier`, so the AST cannot record variadic-ness.

The behavior differs, though. `c-expr` parses the body as a real expression, and `...` is not an expression, so its object-like fallback fails as well and the macro is dropped with a visible parse error.

jorisdral:
Resolves #2245

The fix is twofold: 
* first, I make sure that macros are parsed as either object-like or function-like (not both)
* then, I make sure that function-like macro parsing supports variadic macros

---

### DIFF
--- a/hs-bindgen/src-internal/HsBindgen/Clang/Macros/UniqueExpansion/Parse.hs
+++ b/hs-bindgen/src-internal/HsBindgen/Clang/Macros/UniqueExpansion/Parse.hs
@@ -10,8 +10,8 @@ module HsBindgen.Clang.Macros.UniqueExpansion.Parse (
 
 import Control.Monad (guard, unless)
 import Data.Maybe (catMaybes)
-import Text.Parsec (anyToken, choice, lookAhead, many, manyTill, parserFail,
-                    sepBy, try)
+import Text.Parsec (anyToken, choice, lookAhead, many, manyTill, option, sepBy,
+                    try, unexpected)
 
 import Clang.Enum.Simple (fromSimpleEnum)
 import Clang.HighLevel.Types (MultiLoc (multiLocExpansion),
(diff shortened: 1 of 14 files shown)


hs-bindgen/src-internal/HsBindgen/Clang/Macros/UniqueExpansion/Parse.hs
```diff
@@ -35,32 +35,57 @@ import HsBindgen.Clang.Macros.UniqueExpansion.Types (Definition (..),
...
+-- followed immediately by a ( character without any whitespace in between. See
+-- 'lparen'.
+--
+-- @isFunctionLike@ does not consume input.
+isFunctionLike ::
+     -- | Source location of the macro definition's name
+     Range MultiLoc
+  -> Parser Bool
+isFunctionLike prevRange =
+    --  is function-like if the macro name
```
dschrempf: I think this comment was trimmed accidentially.

hs-bindgen/src-internal/HsBindgen/Clang/Macros/UniqueExpansion/Parse.hs
```diff
@@ -35,32 +35,57 @@ import HsBindgen.Clang.Macros.UniqueExpansion.Types (Definition (..),
...
+    lookAhead (option False (True <$ try (lparen prevRange)))
 
-parseBody :: [Name] -> Parser [Var]
-parseBody params = catMaybes <$> many (choice [parseVar, parseNonVar])
+parseBody :: [Name] -> Bool -> Parser [Var]
```
dschrempf, line 64: We do much more here than just parsing the tokens. Why? Is there a need to store the local params in the body etc?

We only compare the NAME of the macro(s) and, for function-like macros NAME(PARAMS).

jorisdral: > We do much more here than just parsing the tokens. Why? Is there a need to store the local params in the body etc? 

Free variables are used by `mkDependentsGraph` to track dependencies between macro definitions. The local parameters are ignored here. Whether a parameter is ambiguous depends on the invocation site. 

I simply store the local parameters in the body because it does not cost us anything right now

> We only compare the NAME of the macro(s) and, for function-like macros NAME(PARAMS).

I'm not sure which comparison you refer to

jorisdral: @dschrempf and I talked about this: the point is that we can move the analysis of local params vs. free vars *out of* the parser. Then `c-expr` can simply be passed a list of tokens for the body (when the parsers are unified), and `isUniqueExpansion` can do the local param analysis

jorisdral: @dschrempf will apply the suggested change in a follow-up PR

jorisdral:
I forgot to add a changelog entry :facepalm:

Issue #2041, opened 2026-06-04
Generalise macro reparsing beyond `typedef`-like macros


jorisdral:
(we are ignoring macro *values* in this issue because they are standalone: they do not affect bindings for non-macro declarations)

The macro infrastructure currently only supports macros that behave just like `typedef`s. In the future we may want to allow more general reparsing where macros can represent more than just `typedef`s, and/or can modify components of declarations other than just types. Some ideas of what we could support:

* Haskell-style polymorphic types: such as a polymorphic array like `#define MyArray(x) x[]` , used in something like `typedef MyArray(int) MyIntArray`
* Type qualifiers: such as `#define MyConst const`, used in something like `MyConst int x;`
* Function arguments: such as `#define MyFunArgX int x`, used in something like `void foo (MyFunArgX);`

The first idea of Haskell-style polymorphic types is a promising and achievable one. For the other ideas it is (IMO) unclear *if* and *how* we can support them: can we represent macros like `MyConst` and `MyFunArgX` meaningfully in Haskell bindings, or is it only their expansion that is interesting? That is something to consider for this issue as well.

### Assumptions

Currently there are a number of places in the code that assume that macro types behave like `typedefs`. If we generalise the reparser, then we should consider how that affects this code. Places that we know of:

* Since #2021 -- `ReparseMacroExpansions` pass: we skip reparsing a declaration as soon as we notice that it has nested declarations, because such nesting can cause confusing error messages. This skipping is only correct under the assumption that macro types behave like `typedef`. It is not clear if we can still perform skipping if reparsing is generalised, but it may depend on the degree of generalisation.
* Since #1979 and #2015 -- `Zip` pass: the AST before and after reparsing is zipped together to produce a new AST. The goal of the pass is twofold: 
  - Sanity check the post-AST with respect to the pre-AST: the pre-AST and post-AST should semantically still be the same
  - Combine information from the pre-AST and post-AST: the post-AST is sometimes missing information that we have to re-fill in using the pre-AST

  The zipping relies on the fact that the two ASTs only differ at single nodes in the tree. Specifically, at single nodes in the C type AST where a sub-tree could be replaced by a "leaf" (a reference to a macro type). If macro types behave like `typedef`s, then this is possible. It is not clear if we can still perform zipping if reparsing is generalised, but it may depend on the degree of generalisation whether we can still zip but with a different implementation.

PR #2252, opened 2026-09-02
Manual: terminology and macros

dschrempf:
Group terminology by C input / generated Haskell bindings / `hs-bindgen` translation, and cite C standard clauses.

Fix the preprocessor entries: an invocation is replaced by the replacement list, not by the definition; arguments are substituted for parameters. Add some entries, including those for selection, naming and configuration.

[WIP] macro chapter.

---

- [ ] I have updated the changelog. I have included a migration hint if this is a breaking change.

- [ ] I have updated the manual.

- [ ] I have removed all mentions of closed tickets in the code.

- [ ] I have associated each new TODO item with a ticket, using the following syntax:

```hs
-- TODO <link to issue>
-- Brief description
```

### DIFF
--- a/manual/low-level/translation/structs/nesting.md
+++ b/manual/low-level/translation/structs/nesting.md
@@ -14,6 +14,14 @@ struct-in-struct case looks like, but note that it works the same for any
 nesting of structs and unions, in any order, and even recursively. See also the
 [Unions/Nesting][manual:unions/nesting] manual section.
 
+Nesting does not introduce a [scope][creference:scope] of its own. If the nested
+struct has a [tag][manual:terminology-tag], that tag is visible from its
+declaration onwards with the same scope the enclosing struct is declared in; for
+declarations at [file scope][manual:terminology-file-scope], this is the rest of
+the [translation unit][manual:terminology-translation-unit]. An
+[untagged][manual:terminology-untagged-structunionenum] nested struct cannot be
+referred to elsewhere.
+
 ## Example A
 
 The most straightforward way to declare nested structs is by declaring `door`
@@ -239,10 +247,15 @@ supported. A warning-level trace message will be emitted in this case.
 
 <!-- sources and references -->
 
+[creference:scope]: https://en.cppreference.com/w/c/language/scope.html
 [creference:struct]: https://en.cppreference.com/w/c/language/struct.html
 [is-2178]: https://github.com/well-typed/hs-bindgen/issues/2178
 [manual:binding-specifications]: ../../usage/binding-specifications.md
 [manual:pointer-manipulation-api]: ../pointer-manipulation.md
 [manual:record-dot-syntax]: ../record-dot-syntax.md
+[manual:terminology-file-scope]: ../../../terminology.md#file-scope
+[manual:terminology-tag]: ../../../terminology.md#tag
+[manual:terminology-translation-unit]: ../../../terminology.md#translation-unit
+[manual:terminology-untagged-structunionenum]: ../../../terminology.md#untagged-structunionenum
 [manual:unions/nesting]: ../unions/nesting.md
 [manual:usage/binding-specs]: ../../usage/binding-specifications.md
--- a/manual/terminology.md
+++ b/manual/terminology.md
@@ -1,13 +1,155 @@
 # Terminology
 
-## Public
+* [C](#c) — terms describing the C input
+* [Haskell](#haskell) — terms describing the generated Haskell bindings
+* [`hs-bindgen`](#hs-bindgen) — terms describing how `hs-bindgen` translates the
+  C input into Haskell bindings
 
-### Anonymous struct/union
+References of the form (6.10.5) point at the corresponding clause of the [C
+standard][c-standard].
+
+## C
+
+### Preprocessor
+
+#### Function-like macro
+[t:function_like]: #function-like-macro
+
+A [macro definition][t:macro_def] of the form
+
+```c
+#define NAME(a, b) replacement-list
+```
+
+defines a *function-like* macro with [parameters][t:macro_param] `a` and `b`
+(6.10.5). An occurrence of `NAME` immediately followed by its arguments
+enclosed in `(` and `)` (similar to C function calls) is an
+[invocation][t:macro_inv] and is therefore replaced; a bare `NAME` is left
+alone. The [arguments][t:macro_arg] of the invocation are substituted for the
+corresponding parameters in the [replacement list][t:replacement_list]
+(6.10.5.1).
+
+#### Macro
+[t:macro]: #macro
+
+A *macro* is a [macro name][t:macro_name] associated with a [replacement
+list][t:replacement_list]. Macros are either [object-like][t:object_like] or
+[function-like][t:function_like].
+
+#### Macro argument
+[t:macro_arg]: #macro-argument
+
+The preprocessing tokens supplied at a [function-like macro][t:function_like]
+[invocation][t:macro_inv], enclosed in parentheses and separated by commas
+(6.10.5). Arguments are substituted for [parameters][t:macro_param].
+
+#### Macro definition
+[t:macro_def]: #macro-definition
+
+A macro *definition* is a `#define` preprocessing directive. It binds a [macro
+name][t:macro_name] to a [replacement list][t:replacement_list].
+
+Macro definitions are one kind of *control line* (6.10.1), the group of
+directives that also contains `#include`, `#undef` and `#pragma`.
+
+#### Macro expansion
+[t:macro_expansion]: #macro-expansion
+
+The process by which a C preprocessor replaces [invocations][t:macro_inv] with
+the corresponding [replacement lists][t:replacement_list], substituting
+[arguments][t:macro_arg] for [parameters][t:macro_param], applying the `#` and
+`##` operators, and [rescanning][t:rescanning] the result.
+
+Macro expansion operates on *preprocessing tokens*, not on text. The C standard
+specifically refers to the substitution of invocations with their replacement
+lists as "macro replacement" (6.10.5), and the overall process as "macro
+expansion" (6.10.5.1); we use *macro expansion* throughout.
+
+#### Macro invocation
+[t:macro_inv]: #macro-invocation
+
+A macro *invocation* references a [macro definition][t:macro_def] by
+[name][t:macro_name], so that it can be replaced by the [replacement
+list][t:replacement_list]. An [object-like][t:object_like] macro is invoked by
+its name alone; a [function-like][t:function_like] macro is invoked by its name
+followed by a parenthesised argument list.
+
+#### Macro name
+[t:macro_name]: #macro-name
+
+The identifier immediately following `define` in a [macro
+definition][t:macro_def] (6.10.5).
+
+Macro names have their own name space (6.10.5), separate from the name spaces
+of label names, [tags][t:tag], struct/union [fields][t:field], attributes and
+ordinary identifiers (6.2.3). A macro and a tag can therefore share a
+spelling.
+
+#### Macro parameter
+[t:macro_param]: #macro-parameter
+
+One of the identifiers in the identifier list of a [function-like
+macro][t:function_like] definition (6.10.5). Its scope ends at the new-line
+terminating the `#define`. Contrast [macro argument][t:macro_arg].
+
+#### Object-like macro
+[t:object_like]: #object-like-macro
+
+A [macro definition][t:macro_def] of the form
+
+```c
+#define NAME replacement-list
+```
+
+defines an *object-like* macro. Every subsequent occurrence of `NAME` is
+replaced by the [replacement list][t:replacement_list] (6.10.5).
+
+#### Replacement list
+[t:replacement_list]: #replacement-list
+
+The sequence of preprocessing tokens making up the remainder of a [macro
+definition][t:macro_def]. White space preceding or following the list is not
+part of it (6.10.1, 6.10.5).
+
+A replacement list is a token sequence rather than text, and it need not form a
+well-formed C construct. This is what makes generating bindings for macros hard;
+see [macros][manual:translation/macros].
+
+#### Rescanning
+[t:rescanning]: #rescanning
+
+After argument substitution and `#`/`##` processing, the resulting token
+sequence is scanned again for further macro names to replace (6.10.5.4). This
+is how macros chain: `A` expanding to `B` expanding to `C`.
+
+Rescanning does not recurse. If the name of the macro currently being replaced
+turns up during its own expansion it is *not* replaced, and it is never eligible
+for replacement again.
+
+#### Scope of a macro definition
+[t:macro_scope]: #scope-of-a-macro-definition
+
+A [macro definition][t:macro_def] lasts from its `#define` until a matching
+`#undef`, or until the end of the preprocessing translation unit (6.10.5.5).
+This is independent of block structure, and unrelated to the four kinds of scope
+(1) function, (2) [file][t:file_scope], (3) block and (4) function prototype
+scope (6.2.1).
+
+### Structs, unions and enums
+
+#### Anonymous struct/union
 [t:anon]: #anonymous-structunion
 
-A [field][t:field] that declares a nested [untagged][t:untagged] struct or union
-can optionally omit the [field name][t:named_field]. In this case, the [nested
-struct/union][t:nested] is called an *anonymous struct/union*.
+An [unnamed][t:unnamed_field] [field][t:field] whose type is an
+[untagged][t:untagged] struct or union is called an *anonymous struct* or
+*anonymous union* (6.7.3.2). Its [named fields][t:named_field] become
+[indirect fields][t:indirect_field] of the [enclosing
+struct/union][t:enclosing].
+
+The C standard applies the adjective to the *field*; we also apply it to the
+[nested][t:nested] struct or union type that the field declares. The two name
+the same construct. Our name for the field itself is [implicit
+field][t:implicit_field].
 
 <details>
 <summary>Notice</summary>
@@ -29,7 +171,6 @@ struct S {
   };
 };
 ```
-
 </details>
 
 <details>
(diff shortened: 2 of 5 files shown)


dschrempf:
@jorisdral this draft contains substantial changes to the terminology chapter of the manual.

The macro chapter is still an early draft, but I don't plan any more significant changes to the terminology part. I think it would be great if you could give it a read.

manual/terminology.md
```diff
@@ -1,13 +1,155 @@
 # Terminology
 
-## Public
+* [C](#c) — terms describing the C input
+* [Haskell](#haskell) — terms describing the generated Haskell bindings
+* [`hs-bindgen`](#hs-bindgen) — terms describing how `hs-bindgen` translates the
+  C input into Haskell bindings
```
jorisdral: To stay consistent with the manual style guide, we should use `[t:]` style references

manual/terminology.md
```diff
@@ -1,13 +1,155 @@
...
 
-### Anonymous struct/union
+References of the form (6.10.5) point at the corresponding clause of the [C
+standard][c-standard].
```
jorisdral, line 9: These section numbers are subject to change, and some information may even may be moved from one section to another. See https://en.cppreference.com/c/preprocessor/replace#References for example. Maybe we should do something similar: mention the specific C standard version and section

manual/terminology.md
jorisdral: Very useful to be more structured about what the terminology applies to :smile: 

My only concern is that some terminology does not really fall nicely into the C vs. bindings vs. `hs-bindgen` distinction. For example, we may call both an anonymous C struct and its generated Haskell binding an "anonymous struct". And "macro type" is now in the `hs-bindgen` section, though we may call the Haskell binding itself also a macro type. Maybe it's fine as long as we try to clear up confusing terminology "on demand"? Thoughts?

manual/terminology.md
```diff
@@ -1,13 +1,155 @@
...
+alone. The [arguments][t:macro_arg] of the invocation are substituted for the
+corresponding parameters in the [replacement list][t:replacement_list]
+(6.10.5.1).
+
+#### Macro
+[t:macro]: #macro
+
+A *macro* is a [macro name][t:macro_name] associated with a [replacement
+list][t:replacement_list]. Macros are either [object-like][t:object_like] or
+[function-like][t:function_like].
```
jorisdral: After reading this, it's still not so clear to me what "macro" by itself means. Is it just the name?

manual/terminology.md
```diff
@@ -1,13 +1,155 @@
...
+(6.10.5). Arguments are substituted for [parameters][t:macro_param].
+
+#### Macro definition
+[t:macro_def]: #macro-definition
+
+A macro *definition* is a `#define` preprocessing directive. It binds a [macro
+name][t:macro_name] to a [replacement list][t:replacement_list].
+
+Macro definitions are one kind of *control line* (6.10.1), the group of
+directives that also contains `#include`, `#undef` and `#pragma`.
```
jorisdral: "control line" feels like unusual terminology to me. I can see it is used in the C reference,  but only as a the LHS of a production rule in the grammar. I'd probably just call these preprocessing directives and nothing more

(and maybe this information is not necessary to be mentioned)

manual/terminology.md
```diff
@@ -1,13 +1,155 @@
...
+#define NAME(a, b) replacement-list
+```
+
+defines a *function-like* macro with [parameters][t:macro_param] `a` and `b`
+(6.10.5). An occurrence of `NAME` immediately followed by its arguments
+enclosed in `(` and `)` (similar to C function calls) is an
+[invocation][t:macro_inv] and is therefore replaced; a bare `NAME` is left
+alone. The [arguments][t:macro_arg] of the invocation are substituted for the
+corresponding parameters in the [replacement list][t:replacement_list]
+(6.10.5.1).
```
jorisdral: IMO the terminology section should be succinct as long as its get the point across. This is to reduce clutter, so that it's quick to browse the terminology page. Rigorously defining the syntax/details can then be left to separate manual sections, or by simply referencing the C reference.

For example, something like this gets the point across IMO:

```md
A *function-like* macro is a macro that is [defined] using a list of named
[macro parameters]. When a function-like macro is [invoked] using [macro
arguments], those arguments are substituted for the corresponding parameters in
the [replacement list].
```

And if we do want to add details, then we could do so in a `<details></details>`. For example, the `NAME` example could go in there.

That said, I don't really object to the text as is, but that's my philosophy for the terminology section. 

Maybe we should try to get reach agreement on the design of the terminology page with the whole team?

manual/terminology.md
```diff
@@ -1,13 +1,155 @@
...
+A macro *invocation* references a [macro definition][t:macro_def] by
+[name][t:macro_name], so that it can be replaced by the [replacement
+list][t:replacement_list]. An [object-like][t:object_like] macro is invoked by
+its name alone; a [function-like][t:function_like] macro is invoked by its name
+followed by a parenthesised argument list.
+
+#### Macro name
+[t:macro_name]: #macro-name
+
+The identifier immediately following `define` in a [macro
```
jorisdral:
```suggestion
The identifier immediately following `#define` in a [macro
```

manual/terminology.md
```diff
@@ -1,13 +1,155 @@
...
+definition][t:macro_def] (6.10.5).
+
+Macro names have their own name space (6.10.5), separate from the name spaces
```
jorisdral: I see, so the C reference uses "name space" rather than "namespace"? That's unsual :stuck_out_tongue: . I wouldn't mind if we call it "namespace" regardless, but good catch!

manual/terminology.md
```diff
@@ -1,13 +1,155 @@
...
+of label names, [tags][t:tag], struct/union [fields][t:field], attributes and
+ordinary identifiers (6.2.3). A macro and a tag can therefore share a
+spelling.
+
+#### Macro parameter
+[t:macro_param]: #macro-parameter
+
+One of the identifiers in the identifier list of a [function-like
+macro][t:function_like] definition (6.10.5). Its scope ends at the new-line
+terminating the `#define`. Contrast [macro argument][t:macro_arg].
```
jorisdral: Nit: do we need to mention the scope here? I think it's not too important

manual/terminology.md
```diff
@@ -138,33 +284,237 @@ Unnamed bit-fields are not translated to fields in the corresponding Haskell
...
+The pluggable `hs-bindgen` component that parses, typechecks and translates the
+[replacement lists][t:replacement_list] of [macro definitions][t:macro_def].
+
+#### Macro type
+[t:macro_type]: #macro-type
+
+A [macro][t:macro] whose [replacement list][t:replacement_list] parses as a C
+*type* expression, and which is therefore translated to a Haskell type using a
+[newtype wrapper][t:newtype] around the translation of that C type. For example,
+`#define YEAR int`.
```
jorisdral: If we in the future support macro types that can take arguments, would it be more appropriate to say 

```md
A macro whose [definition] parse as a C type expression
```

manual/terminology.md
```diff
@@ -138,33 +284,237 @@ Unnamed bit-fields are not translated to fields in the corresponding Haskell
...
+
+#### Macro value
+[t:macro_value]: #macro-value
+
+A [macro][t:macro] whose [replacement list][t:replacement_list] parses as a C
+*expression*, and which is therefore translated to a Haskell value binding. For
+example, `#define EPSILON 0.1`.
```
jorisdral: Same here: not only the replacement list is parsed, the parameters as well

manual/terminology.md
```diff
@@ -138,33 +284,237 @@ Unnamed bit-fields are not translated to fields in the corresponding Haskell
...
+
+#### Parsable macro
+[t:parsable_macro]: #parsable-macro
+
+A [macro definition][t:macro_def] whose [replacement list][t:replacement_list]
+the active [macro language][t:macro_lang] can parse, either as a [macro
+type][t:macro_type] or as a [macro value][t:macro_value].
```
jorisdral: Same here: definition in stead of replacement list

manual/terminology.md
```diff
@@ -138,33 +284,237 @@ Unnamed bit-fields are not translated to fields in the corresponding Haskell
...
+#### Reparsing
+[t:reparsing]: #reparsing
+
+The second parse of those declarations that contain macro
+[expansions][t:macro_expansion], recovering the [macro types][t:macro_type] that
+`libclang` had already expanded away; see [macros][manual:translation/macros].
+
+### Selection
+
+#### Extern declaration
```
jorisdral: Should this be "external declaration"? `extern` is a C keyword

manual/terminology.md
```diff
@@ -138,33 +284,237 @@ Unnamed bit-fields are not translated to fields in the corresponding Haskell
...
+[t:extern]: #extern-declaration
+
+A declaration listed in an external [binding specification][t:binding_spec].
+`hs-bindgen` does not generate bindings for it, and instead imports the
+declaration from the specified module instead.
```
jorisdral:
```suggestion
`hs-bindgen` does not generate bindings for it, and imports the
declaration from the specified module instead.
```

manual/terminology.md
```diff
@@ -1,13 +1,155 @@
...
+#### Scope of a macro definition
+[t:macro_scope]: #scope-of-a-macro-definition
+
+A [macro definition][t:macro_def] lasts from its `#define` until a matching
+`#undef`, or until the end of the preprocessing translation unit (6.10.5.5).
+This is independent of block structure, and unrelated to the four kinds of scope
+(1) function, (2) [file][t:file_scope], (3) block and (4) function prototype
+scope (6.2.1).
+
+### Structs, unions and enums
```
jorisdral, line 133: I don't agree with some of the rephrasings in this section, so maybe we should discuss those together. Some original phrasings were deliberate

jorisdral (commented):
I've only looked at the Terminology page for now

manual/terminology.md
dschrempf: I share your concerns, but I still think it is much better to have some sort of sectioning instead of none. We could merge the Haskell and `hs-bindgen` sections. If we use the same term for a C and a Haskell concept, we may better be explicit about it. Do you agree?

manual/terminology.md
dschrempf: Absolutely. I have linked the exact version, but don't write it in text, fixing now.

manual/terminology.md
dschrempf: Fixed.

manual/terminology.md
dschrempf: I agree. I think a "macro" is the concept behind a "macro definition", which is the "`#define ....`" line(s).

So maybe we should just drop "macro" and only ever refer to "macro definition"?

Or, we say a "macro" is the thing/concept/entity a "macro definition" defines (a bit recursive, but not entirely).

manual/terminology.md
dschrempf: Good point. I had "control-line" in the macro section briefly, that's why I clung on to it. It is unused terminology.

manual/terminology.md
dschrempf: Ah yes. That was actually my fault. Let's use namespace throughout.

manual/terminology.md
dschrempf: That's actually a good question. I don't know.

The definition is the "#define ... " line, which it cannot be.

But the replacement list need not be a type either, only the replacement list after substituting the provided arguments needs to be a type.

Not sure if we want this detail now. Maybe we should just state that we do not support function-like macro types at the moment?

manual/terminology.md
jorisdral: Right, yes maybe we should just add notes in places where it's relevant that a term has an overloaded meaning.

manual/terminology.md
dschrempf: Hm, maybe

```
- An [object-like-macro][...] whose [replacement list][t:replacement-list], or
- A [function-like-macro][...] whose [replacement list][...] with arguments subsituted
parses as a C *expression*, and which is therefore translated to a Haskell value binding.
```

(the same could be done for macro types).

manual/terminology.md
dschrempf: Right, this is actually a mistake. The macro language does not only parse the replacement list, it also parses the name (albeit not the complete definition).

Thinking about this, I am not sure if we can ever support function-like macro types. Because people can provide ANYTHING as arguments, even things that don't typecheck at all. So we can never translate it meaningfully to a type.

Maybe in the high-level bindings, we can say: This function-like macro is always supposed to be a macro type?

dschrempf: I have fixed the mistake (macro language parses name and replacement list).

manual/terminology.md
jorisdral: Indeed, I agree with you that "macro" is the "thing that a macro definition defines", which is like... the instruction to replace all successive occurrences of that macro's name by the replacement list?

From https://en.cppreference.com/c/preprocessor/replace:

> The #define directives define the identifier as a macro, that is they instruct the compiler to replace all successive occurrences of identifier with replacement-list, which can be optionally additionally processed. If the identifier is already defined as any type of macro, the program is ill-formed unless the definitions are identical. 

From wikipedia:

> In [computer programming](https://en.wikipedia.org/wiki/Computer_programming), a macro (short for "macro instruction"; from [Greek](https://en.wikipedia.org/wiki/Greek_language) μακρο- 'long, large'[[1]](https://en.wikipedia.org/wiki/Macro_(computer_science)#cite_note-1)) is a rule or [pattern](https://en.wikipedia.org/wiki/Pattern) that specifies how a certain input should be [mapped](https://en.wikipedia.org/wiki/Map_(mathematics)) to a replacement output.

So maybe we should say that the "macro" is a "replacement rule" or something along those lines?

Aside: I suppose it also depends a lot on context. For example, with struct terminology the name "struct" can refer to the struct declaration in one context, and a value of that struct type in another context. Maybe "macro" can refer to the definition in some context, or the replacement rule in another context

Language is messy :stuck_out_tongue: and maybe I'm being too nit-picky

manual/terminology.md
dschrempf: We will have a meeting on Friday about this.

manual/terminology.md
dschrempf: Yes, I was a bit sloppy here. I think the terminology chapter should be self-contained, but not go overboard, and instead refer to the manual sections.

I will change this definition, in particular, but also check other definitions if they could be shortened.

manual/terminology.md
jorisdral: > The definition is the "#define ... " line, which it cannot be.

I'd probably just call both `#define A ...` and `A ...` a macro definition, because I'm not sure the distinction matters in practice. It's arguably an implementation detail that the macro language parser only sees the tokens after `#define`. And then the terminology stays well-defined with respect to function-like macro types if we support them in the future.

> But the replacement list need not be a type either, only the replacement list after substituting the provided arguments needs to be a type.

I think what you mean here is supporting function-like macro types at *invocation* sites, but there are ideas for supporting function-like macro types at *definition* sites (see #2041).

> Not sure if we want this detail now. Maybe we should just state that we do not support function-like macro types at the moment?

Regardless of the wording here, it would be good to mention that function-like macro types are not supported -- maybe in a `<details>` dropdown like we do elsewhere too.

manual/terminology.md
jorisdral: I think we should say "definition" instead of "replacement list". This is related to https://github.com/well-typed/hs-bindgen/pull/2252#discussion_r3934929706

And also the part about substitution doesn't seem right. We parse macro value *definitions*, not invocations where argument substitution applies, right?

manual/terminology.md
jorisdral: Related to https://github.com/well-typed/hs-bindgen/pull/2252#discussion_r3934929706

manual/low-level/translation/structs/nesting.md
```diff
@@ -14,6 +14,14 @@ struct-in-struct case looks like, but note that it works the same for any
 nesting of structs and unions, in any order, and even recursively. See also the
 [Unions/Nesting][manual:unions/nesting] manual section.
 
+Nesting does not introduce a [scope][creference:scope] of its own. If the nested
+struct has a [tag][manual:terminology-tag], that tag is visible from its
+declaration onwards with the same scope the enclosing struct is declared in; for
+declarations at [file scope][manual:terminology-file-scope], this is the rest of
+the [translation unit][manual:terminology-translation-unit]. An
+[untagged][manual:terminology-untagged-structunionenum] nested struct cannot be
+referred to elsewhere.
```
jorisdral, line 23: We should also add this text to `unions/nesting`

jorisdral, line 23: We should also decide how/if we want to link to the terminology page. I can see the advantage that it helps us a bit to keep the terminology up to date, but many links may also make it harder to write and maintain manual pages

manual/terminology.md
```diff
@@ -1,13 +1,150 @@
...
+
+#### Anonymous struct/union
+[t:anonymous-structunion]: #anonymous-structunion
+
+An [unnamed][t:unnamed-field] [field][t:field] whose type is an
+[untagged][t:untagged-structunionenum] struct or union is called an *anonymous
+struct* or *anonymous union* (6.7.3.2). Its [named fields][t:named-field] become
+[indirect fields][t:indirect-field] of the [enclosing
+struct/union][t:enclosing-structunion].
```
jorisdral:
```suggestion
A [field][t:field] that declares a nested [untagged][t:untagged] struct or union
can optionally omit the [field name][t:named_field]. In this case, the [nested
struct/union][t:nested] is called an *anonymous struct/union* (6.7.3.2). See also [indirect fields][t:indirect-field].
```

Note: I tried not to repeat the definition of *indirect fields* in this text. The link to `t:indirect-field` should be sufficient, I think

manual/terminology.md
```diff
@@ -1,13 +1,150 @@
...
+
+The C standard applies the adjective to the *field*; we also apply it to the
+[nested][t:nested-structunion] struct or union type that the field declares. The
+two name the same construct. Our name for the field itself is [implicit
+field][t:implicit-field].
```
jorisdral:
```suggestion
The C standard calls the whole *field* an anonymous struct/union; we only apply that terminology to the
[nested][t:nested-structunion] struct or union type. The terminology for the field itself is [implicit
field][t:implicit-field].
```

manual/terminology.md
```diff
@@ -29,107 +166,105 @@ struct S {
...
-### Field
+#### Field
 [t:field]: #field
 
-A *field* is described in the C reference as a "member" of a struct/union.
-Fields are declared using variable declarations or [bit-field][t:bit_field]
-declarations.
+A *field* is what the C standard calls a *member* of a struct or union
+(6.7.3.2). Fields are declared using variable declarations or
+[bit-field][t:bit-field] declarations.
```
jorisdral: Because clang calls them fields

manual/terminology.md
```diff
@@ -29,107 +166,105 @@ struct S {
...
+[t:implicit-field]: #implicit-field
 
-### Implicit field
-
-An *implicit* field* is a [direct][t:direct_field], [unnamed][t:unnamed_field]
-field that references an [anonymous struct/union][t:anon].
+An *implicit* field is a [direct][t:direct-field], [unnamed][t:unnamed-field]
+field whose type is an [anonymous struct/union][t:anonymous-structunion]. The C
+standard has no separate name for it, since it calls the field itself the
+anonymous struct or union.
```
jorisdral:
```suggestion
field whose type is an [anonymous struct/union][t:anonymous-structunion]. 

The C standard has no separate name for the field, since it calls the field itself the
anonymous struct or union.
```

manual/terminology.md
```diff
@@ -29,107 +166,105 @@ struct S {
...
-struct/union. Such fields are still fields of the anonymous struct/union as
-well.
+This applies recursively: if an anonymous struct/union has indirect fields, then
+the enclosing struct/union has its own indirect fields for those fields as well.
 
-This also works recursively: if an anonymous struct/union has indirect fields,
-then the enclosing struct/union has its own indirect fields for those fields as
-well.
+*Indirect field* is a term coined by Clang (`IndirectFieldDecl`), not by the C
+standard.
```
jorisdral: Emphasise the "accessed as if" part, not that they are fields

manual/terminology.md
```diff
@@ -138,33 +273,253 @@ Unnamed bit-fields are not translated to fields in the corresponding Haskell
...
+
+#### Unnamed field
+[t:unnamed-field]: #unnamed-field
 
 A [field][t:field] without a name is called an *unnamed field*.
-[bit-fields][t:bit_field] are allowed to be unnamed. Fields that declare an
-[anonymous struct/union][t:anon] are unnamed by definition.
+[Bit-fields][t:bit-field] are allowed to be unnamed, and fields that declare an
+[anonymous struct/union][t:anonymous-structunion] are unnamed by definition
+(6.7.3.2).
```
jorisdral: This is good phrasing

manual/terminology.md
jorisdral: Meeting was held on a Monday in the end :stuck_out_tongue_winking_eye:

manual/terminology.md
```diff
@@ -29,107 +166,105 @@ struct S {
...
-declarations.
+A *field* is what the C standard calls a *member* of a struct or union
+(6.7.3.2). Fields are declared using variable declarations or
+[bit-field][t:bit-field] declarations.
 
-### File scope
-[t:file_scope]: #file-scope
+We say "field" rather than "member" because the Haskell side generates [record
+fields][t:record-field], and because the C standard itself already uses field in
+*bit-field*.
```
jorisdral:
```suggestion
We say "field" rather than "member" because `clang` calls them fields, and because the C standard itself already uses field in *bit-field*.
```

manual/terminology.md
jorisdral, line 212:
```suggestion
An *implicit* field is a [direct][t:direct-field], [unnamed][t:unnamed-field]
field whose type is an [anonymous struct/union][t:anonymous-structunion].

The C standard has no separate name for the field, since it calls the field itself the
anonymous struct or union.
```

manual/terminology.md
```diff
@@ -29,107 +166,105 @@ struct S {
...
+#### Indirect field
+[t:indirect-field]: #indirect-field
 
-### Indirect field
-[t:indirect_field]: #indirect-field
+The [named fields][t:named-field] of an [anonymous
+struct/union][t:anonymous-structunion] are fields of the [enclosing
+struct/union][t:enclosing-structunion] as well (6.7.3.2), and can be accessed as
+such. With respect to the enclosing struct/union, such a field is called an
+*indirect field*. It remains a field of the anonymous struct/union too.
```
jorisdral:
```suggestion
The [named fields][t:named-field] of an [anonymous
struct/union][t:anonymous-structunion] can be [accessed as if they were fields][dr-449] of the [enclosing
struct/union][t:enclosing-structunion] as well (6.7.3.2). With respect to the enclosing struct/union, such a field is called an
*indirect field*. It is only a true a field of the anonymous struct/union

[dr-449]: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2396.htm#dr_499
```

manual/terminology.md
```diff
@@ -138,33 +273,253 @@ Unnamed bit-fields are not translated to fields in the corresponding Haskell
...
+`typedef` named `date` can coexist.
 
-A struct or union or enum with a [tag][t:tag] is called *tagged*.
+#### Tagged struct/union/enum
+[t:tagged-structunionenum]: #tagged-structunionenum
 
-### Unnamed field
-[t:unnamed_field]: #unnamed-field
+A struct or union or enum with a [tag][t:tag] is called *tagged*. The C standard
+says "with a tag"; we use *tagged*.
```
jorisdral: Maybe we should consistently put comments about differences with the C standard in a separate paragraph. What do you think?
 
```suggestion
A struct or union or enum with a [tag][t:tag] is called *tagged*. 

The C standard says "with a tag"; we use *tagged*.
```

manual/terminology.md
```diff
@@ -138,33 +273,253 @@ Unnamed bit-fields are not translated to fields in the corresponding Haskell
...
-[anonymous struct/union][t:anon] are unnamed by definition.
+[Bit-fields][t:bit-field] are allowed to be unnamed, and fields that declare an
+[anonymous struct/union][t:anonymous-structunion] are unnamed by definition
+(6.7.3.2).
+
+#### Untagged struct/union/enum
+[t:untagged-structunionenum]: #untagged-structunionenum
+
+A struct or union or enum without a [tag][t:tag] is called *untagged*. The C
+standard says "without a tag"; we use *untagged*.
```
jorisdral:
```suggestion
A struct or union or enum without a [tag][t:tag] is called *untagged*. 

The C
standard says "without a tag"; we use *untagged*.
```

jorisdral (commented):
@dschrempf and I had a call about some of the changes to the sections on existing terminology

manual/terminology.md
dschrempf: This was changed in a recent commit, resolving.

manual/terminology.md
dschrempf: I think what we should be careful about is that the definition includes
- the macro name,
- optional parameter list (maybe variadic),
- replacement list.

Here, we are only referring to the replacement list. For object-like macros, we can find out if the replacement list is type-like or value-like.
 
We can never reliably typecheck a function-like macro. Because _any_ strings can be used as arguments, making the replacement list a bogus type/value.

manual/terminology.md
dschrempf: Maybe we need Edsko to decide here. I amm still favoring "replacement list", which is part of the definition, but not the complete definition.

dschrempf:
Thanks @jorisdral for discussing this and improving the terminology page! I have implemented all changes we agreed on. Some discussions are still open, but we can always revise later.

I think the two open discussions items are:
- macro definition vs macro replacement list
- macro (the concept) vs a macro (as in: a specific "macro") vs macro definition

Both are subtle and I am sure we find a common denominator so to speak :-).

PR #2254, opened 2026-09-03
Add `known-issues.md`

edsko:
Draft until I've triaged everything that still needs triaging.


hs-bindgen/known-issues.md
```diff
@@ -0,0 +1,108 @@
+# Known Issues
+
+> Current version: 0.1.0.
```
edsko, line 3: Don't list version. When bundled in a .cabal file as part of a release, the version is implicit anyway. In the repo, the version will be whatever version we're next planning to release.

hs-bindgen/known-issues.md
```diff
@@ -0,0 +1,108 @@
...
+
+Whenever we release `hs-bindgen`, our goal is of course to release a product
```
edsko, line 5: Clarify that this file is intended to describe issues in _releases_ only; not interim states in the repo.

hs-bindgen/known-issues.md
```diff
@@ -0,0 +1,108 @@
...
+```c
+struct S { const int x; int y; };
+struct S f1();
+```
+
+## Return-by-value of const-struct results in invalid code
+
+> https://github.com/well-typed/hs-bindgen/issues/1514
+
+Related to #1515, but even more obscure: we generate code for funcitons that
```
jorisdral, line 84:
```suggestion
Related to #1515, but even more obscure: we generate code for functions that
```

Issue #2257, opened 2026-09-03
preprocess command --hash-define is abnormal/incorrect


TravisCardwell:
The arguments to the `preprocess` command have been changed because we now group `#include` arguments and `#define`s into `UncheckedRootDirective`s.  The `parseInputs` function is changed to parse both of these, and the result is the following:

```
$ cabal run hs-bindgen-cli -- preprocess --help
Usage: hs-bindgen-cli preprocess [--fblocks] [--builtin-include-dir MODE]
                                 ...
                                 [--create-output-dirs] [--overwrite-files]
                                 (HEADER | --hash-define NAME VALUE)...
```

This is not a good CLI.  Previously, the parser clearly showed that at least one header must be specified, and the parser provided an error if the user did not do so (which should exit with exit code 2 btw).  Now, the CLI shows that at least one header *or* `--hash-define` must be specified.  Indeed, just passing a `--hash-define` is sufficient, and a "no `#include` root directive" warning is traced later (with exit code 1).

```
$ cabal run hs-bindgen-cli -- preprocess \
    --hs-output-dir ex-src \
    --hash-define FOO bar
[Warning] [HsBindgen] [unique-id] empty unique identifier ('UniqueId', '--unique-id'):
  C uses a global namespace.
  We encourage using a unique identifier to avoid duplicate symbol names.
  For example, use and adapt 'com.example.package'.
[Warning] [HsBindgen] [root-header] no #include root directive: nothing to generate bindings for
[Warning] [HsBindgen] [select] No declarations matched the selection predicate
[Notice ] [HsBindgen] [artefact-no-bindings-multiple] No output generated (all binding categories with base module name "Generated" are empty)
```

The parser should parse `--hash-define` options as options, *before* parsing arguments.  Zero or more may be specified.  The parser should parse headers as arguments, *after* all options.  One or more must be specified.  The parse results can then be combined to produce the `UncheckedRootDirective`s.

edsko:
After discussing this with @TravisCardwell again, closing this as wontfix; @dschrempf added this quite intionally alongside the header, because the order matters:

```c
#include "a.h"
#define FOO 1
#include "b.h"
```

PR #2258, opened 2026-09-03
Make CLI parse errors use exit code 2

TravisCardwell:
In my culture, this is a must.  I noticed that we do not do this when I found #2257.  It is a trivial change, so I figure that creating this PR is the best way to propose the change.  If the team explicitly wants to not use meaningful exit codes, feel free to close the PR.

### DIFF
--- a/hs-bindgen/CHANGELOG.md
+++ b/hs-bindgen/CHANGELOG.md
@@ -114,6 +114,8 @@
         wrapper translation unit includes every main header.
       * Declarations in the root header (i.e., the `#define`s) are not parsed: a
         root directive is configuration.
+* CLI exit codes are changed so that exit code 2 is used for CLI usage errors.
+  See the invocation section of the manual for details.
 
 ### New features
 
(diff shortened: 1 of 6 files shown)


TravisCardwell:
While configuring the exit code, I noticed the following:

```haskell
header "hs-bindgen - generate Haskell bindings from C headers"
```

This results in the following output:

```
$ cabal run hs-bindgen-cli -- --help
hs-bindgen - generate Haskell bindings from C headers

Usage: ...

Available options:
...
```

Our use of `header` is very old, so I doubt we did it on purpose.  Normal behavior for such commands is for the program description to go below the usage, which is done using `progDesc` (not `header`).  The output looks like this:

```
$ cabal run hs-bindgen-cli -- --help
Usage: ...

  Generate Haskell bindings from C headers

Available options:
...
```

I am tacking on a commit to fix this.  If something else is preferred, let me know.

edsko (approved):
I'm fine with this but @dschrempf is currently using these exit codes in a different way; we should coordinate this.

dschrempf:
@TravisCardwell and I had a meeting! We use these exit codes in our end-to-end integration tests, but the actual codes don't matter at all. I also think we should follow the UNIX conventions. Thanks Travis for making me aware of that!

PR #2263, opened 2026-09-08
`Manual`: new section on FFI types

jorisdral:
Technically part of #2238, but I decided to split it out into its own manual section

---


manual/low-level/translation/ffi-types.md
```diff
@@ -0,0 +1,377 @@
...
+type to use for `CFoo` in the `foreign import` declaration. So, `hs-bindgen`
+uses the FFI type corresponding to the *underlying C type* of `CFoo`. Clearly,
+`hs-bindgen` can also not check whether these two FFI types match after the
+fact.
+
+> [!WARNING]
+>
+> When using external types, the onus is on the user to ensure that its
+> `HasFFIType` instance matches its underlying C type. `hs-bindgen` can not
+> check whether these two FFI types match.
```
jorisdral, line 271: TODO: we should advise users on how they can make sure these two FFI types match. It's a tricky subject, and I haven't formulated a satisfying advice just yet

jorisdral: no longer relevant

manual/low-level/translation/ffi-types.md
```diff
@@ -0,0 +1,508 @@
+# FFI types
+
+For a `foreign import` declaration to be valid its type must be a [*foreign
```
edsko, line 3: I think the intro is good, but a little dense. I think start perhaps with the situation in _Haskell_ in general, never mind bindgen; give an example of a valid and an invalid foreign import, and refer to the manual; then two sentences about the consequences for newtypes. I think this can be a short section, but it sets the stage for everything else. 

Importantly, that then also distinguishes between the definitions that bindgen makes and its approach from what is just generally true in Haskell.

manual/low-level/translation/ffi-types.md
```diff
@@ -0,0 +1,508 @@
...
+type signature by converting function arguments to their FFI type, invoking the
+`foreign import`, and converting the function result back from its FFI type. We
+will show [Examples][t:examples] of what this looks like later.
+
+To generate such bindings, `hs-bindgen` essentially needs to decide on three
+pieces of information about a type such as `T`:
+
+1. What is its FFI type `F`?
+2. What module `M` needs to be imported?
+3. How do we convert between `T` and `F`?
```
edsko: I think this would be cleaner if phrased as _two_ things; after all, the _definition_ of an FFI type tightly couples `F` and `M`, so the decoupling here reads a bit strange. We just need to know to things really: what is it's FFI type `M.F`, and how do we translate back and forth?

manual/low-level/translation/ffi-types.md
```diff
@@ -0,0 +1,508 @@
...
+
+In general, `hs-bindgen` assigns FFI types to `newtype`s that it generates based
+on *inheritance*. Whenever `hs-bindgen` generates a `newtype` `U` that wraps a
```
edsko: I can see where you're coming from here, but I think we should avoid the term "inheritance". It might remind readers of approaches to OOP in Haskell, which is of course not at all what you mean.

manual/low-level/translation/ffi-types.md
```diff
@@ -0,0 +1,508 @@
...
+themselves as "base" FFI types[^2], such as `CInt` and `PtrVoid`. Without
+external binding specifications[^3], any `newtype` that `hs-bindgen` generates
+(transitively) wraps such a base type, and is therefore also assigned a base FFI
+type, with straightforward conversion functions.
+
+Alternatively, external binding specifications can be used to instruct
+`hs-bindgen` to use external Haskell types instead of generating new bindings.
+These external types may be generated by `hs-bindgen`, or they may implemented
+by hand, in which case the implementer may have picked a custom FFI type (that
+should satisfy all [necessary conditions][t:ffi-type-def]). Such
```
edsko: I'm not sure there _are_ necessary conditions; at least, not beyond it being "an FFI type". The definition _is_ the conditions?

manual/low-level/translation/ffi-types.md
```diff
@@ -0,0 +1,508 @@
...
+
+If the field is omitted from an external binding specification, then
+`hs-bindgen` will default to assigning a "base" FFI type again based on `T`'s
+*underlying C type*: the C type corresponding to `T` as reported by `libclang`.
+
+> [!NOTE]
+>
+> It is not recommended to omit the `ffitype` field in external binding
+> specifications, because the behaviour of assigning FFI types may be confusing
+> in such cases.
```
edsko: I think we should either say what the confusing behaviour _is_, or just say that you should not omit them.

manual/low-level/translation/ffi-types.md
```diff
@@ -0,0 +1,508 @@
...
+
+Finally, `hs-bindgen` uses the `HasFFIType` run-time class to take care of the
+conversions to and from FFI types in the generated bindings. Any Haskell type
+`T` may be associated with an FFI type `F` by giving it an instance of the
+`HasFFIType` class, including functions to convert from `T` to `F` and back.
```
edsko, line 113: I think this is a nice paragraph, and fits well with my "two, not three" comment above: (1) an FFI-type; through a binding spec, and (2) the convresion,through instances of the `HasFFIType` typeclass.

manual/low-level/translation/ffi-types.md
```diff
@@ -0,0 +1,508 @@
...
+as "base" FFI types. `hs-bindgen` generates instances for `HasFFIType` whenever
+applicable using `newtype`-deriving, which matches the inheritance-based
+strategy of assigning FFI types. The `toFFIType` and `fromFFIType` functions are
+used in the generated bindings.
+
+> [!WARNING]
+>
+> If an external type is used in a function binding and it is missing a
+> `HasFFIType` instance, `hs-bindgen` will currently generate code that fails to
+> compile. See [issue #2253][issue-2253].
```
edsko: I don't really think we should link to an issue here; or if we do, at least say that this is a _user bug_; _you_ told of what the FFI type is, _you_ are responsible for providing the translation. Not much we can do here.

manual/low-level/translation/ffi-types.md
```diff
@@ -0,0 +1,508 @@
...
+>
+> FFI types are currently not supported in prescriptive binding specifications.
+> See [issue #2271][issue-2271].
+
+## Examples
+[t:examples]: #examples
+
+In this subsection, we will use a series of examples to show how `hs-bindgen`
+generates bindings that involve FFI types. We will use a running example of C
+code in header `ffi-types/int.h`:
```
edsko, line 138: I like this series; it's the same one you showed me in person last week I think? A reader might find it a bit easier to follow if you give a brief enumeration of the variations that we'll consider: (1) single module (2) multiple modules, all generated (3) module multiples, hadnwritten binding for A with custom FFI type, (4) bespoke representation

manual/low-level/translation/ffi-types.md
```diff
@@ -0,0 +1,508 @@
...
+  toFFIType = id
+  fromFFIType = id
+```
+
+Here we *choose* to map `A` to itself as its FFI type. Meaning that whenever we
+generate a `foreign import` declaration for a C function that involves `A`, we
+would use `Example3.A.A` there. Recall also the [definition of FFI
+types][t:ffi-type-def], which prescribes that `A` is only an FFI type if it is
+valid in a foreign import when `Example3.A` is imported. For this to be true, we
+should also re-export `CInt` and its constructors.
```
edsko, line 354: It might be worth a note saying why bindgen didn't have to export CInt above (import from `Foreign.C`).

manual/low-level/translation/ffi-types.md
```diff
@@ -0,0 +1,508 @@
...
+
+foreign import ccall safe "foo_wrapper" foo_wrapper ::
+    Example3.A.A -> IO ()
+
+foo :: Example3.B.B -> IO ()
+foo = \x0 -> foo_wrapper (toFFIType x0)
+```
+
+As always, it uses `B`'s FFI type in the `foreign import`. It can only do this
+because the FFI types is specified in `B`'s external binding specification.
```
edsko, line 441: Would it make sense to have _two_ functions, one on `A` and one on `B`? (or even a single one that uses both). Just to show that an FFI type can then really appear in a foreign import, because _none_ of your exmaples show that.

manual/low-level/translation/ffi-types.md
edsko: Maybe just "based on newtype deriving"; then uses down below "is inherited from A" I think can still work; or perhaps "is adopted from A" if you want to avoid it.

edsko (approved):
Nice, I think this looks a lot better now! Really glad you did this. Some minor proofreading remarks only.

manual/low-level/translation/ffi-types.md
jorisdral: Right, maybe the wording is not great here. I wanted to emphasise that a custom FFI type has to satisfy the definition of FFI types, and it's up to the implementer to make sure that is the case. I'll rephrase

manual/low-level/translation/ffi-types.md
jorisdral: I use "newtype-deriving" and "adopted" now

manual/low-level/translation/ffi-types.md
jorisdral: Removed the second half of the sentence

manual/low-level/translation/ffi-types.md
jorisdral: Hmm I actually included this warning here not because users might make a mistake in defining their external type and binding spec, but because `hs-bindgen` currently does the wrong thing, which is generate incorrect code rather than tracing a message and dropping the declaration.

manual/low-level/translation/ffi-types.md
jorisdral: Indeed, I showed you this last week but there are fewer examples now

manual/low-level/translation/ffi-types.md
edsko: but there is nothing bindgen _can_ do here? we'd have to inspect whatever code the user wrote..?

manual/low-level/translation/ffi-types.md
jorisdral: > Just to show that an FFI type can then really appear in a foreign import, because none of your exmaples show that.

I'm not sure I understand -- did you mean to say:

> Just to show that a type can map to itself as an FFI type in a foreign import, ...

Specifically about `A` mapping to `A`?

manual/low-level/translation/ffi-types.md
jorisdral: If an external binding spec lists that an external type has a `HasFFIType` instance but in reality the Haskell implementation has none, then yes we can't do anything about that because the user made a mistake. 

I'm talking here about the case where an external binding does *not* list that it has a `HasFFIType` instance and also has no instance in the Haskell implementation. `hs-bindgen`  will generate invalid code in that case

jorisdral: Maybe I should just remove the warning... it may just confuse readers more than it helps. And it's a `hs-bindgen` bug anyway, not a user bug

jorisdral: Removed the warning

manual/low-level/translation/ffi-types.md
jorisdral: I'll put the PR on the merge queue for now, but we can still change this later

PR #2267, opened 2026-09-14
Overhaul of FFI types

### LINKED ISSUE #1565: Implement a generalisation of the `HasFFIType` class
The `HasFFIType` class comes with a set of default instances for primitive types. Custom instances can only be derived using `newtype` deriving. We only generate foreign imports at FFI types, and all functions that wrap FFI types can only use types that are (informally) coercible to FFI types. The downside is that this is restrictive as it only allows a small set of types to appear in generated bindings. The upside is that all function bindings that we generate are correct/safe, even dynamic wrappers and dynamic imports.

For example, custom instances can only be defined like this:

```hs
newtype S = S CInt
  deriving newtype HasFFIType
```

The goal of this issue is to generalise this approach so that users can define arbitrary mappings between their custom datatypes and FFI types, for example:

```hs
data T = T Word64 (Word64 -> String)

instance HasFFIType T where
  type FFIType T = CInt
  toFFIType (T x _) = fromIntegral x
  fromFFIType x = T (fromIntegral x) show
``` 

For non-"wrapper"/non-"dynamic" foreign imports, this should be possible without sacrificing safety/correctness. We'll keep the fixed set of `FFIType` targets, which is defined in `HsBindgen.Runtime.Internal.FFIType`, but the class will otherwise be open.

For foreign import declarations that are a dynamic wrapper or a dynamic import, we still need a more restrictive version of the `HasFFIType` class. For example, in this example:

```hs
foreign import ccall "wrapper" foo :: (CInt -> IO ()) -> IO (FunPtr (CInt -> IO ())

fooS :: (S -> IO ()) -> IO (FunPtr (S -> IO ()))
fooS = \fun -> fmap castFunPtrFromFFIType  (foo (toFFIType fun))

fooT :: (T -> IO ()) -> IO (FunPtr (T -> IO ()))
fooT = \fun -> fmap castFunPtrFromFFIType  (foo (toFFIType fun))
```

`fooS` is safe, but `fooT` is not! `fooT` is essentially unsafely coercing `foo`.

So, we will need a class a la:

```hs
class CastableHasFFIType a 
```

We might be able to use a `Coercible a (FFIType a)` constraint in the superclass constraints, but we should check that we don't run into the "constructor import issue" again if we do (see #1282)

### LINKED ISSUE #1745: Reconsider `HasFFIType` class hierarchy
placeholder (@jorisdral should write some text here)

### LINKED ISSUE #1747: Use portable types for C primitive types
Given 

```c
int f(int, int);
```

we currently generate

```hs
foreign import ccall safe "f_wrapper" f_wrapper ::
     Int32 -> Int32 -> IO Int32
```

This introduces unnecessary non-portable code. For the _specific_ case of C primitive types, and C primitive types _only_, we can do better. 

See https://www.reddit.com/r/haskell/comments/1r0z0nc/hsbindgen_release_preview_automatic_binding/ , and specifically https://www.reddit.com/r/haskell/comments/1r0z0nc/comment/o6eqp6k/  , for detailed discussion.

### LINKED ISSUE #1748: Make user-defined "portable" types possible
Suppose someone wanted to design a type (by hand), which is intended to be portable. Something like this perhaps:

```hs
{-# LANGUAGE CPP #-}

module MyPortableType where

import Data.Word

newtype CFoo = CFoo
#if ..
  Int32
#else
  Int64
#endif
```

(think `CSize`, but user-defined). Ideally if we then have a binding spec that maps

```c
// some_header.h
#if ..
typedef ... foo;
#else
typedef ... foo;
#endif
```

to `CFoo`, something like

```yaml
ctypes: 
  - cname: struct foo
    header: some_header.h
    hsname: CFoo
hstypes:
  - hsname: CFoo
    ..
```

then if we process another, something like

```c
// another_header.h
#include <some_header.h>
void f(foo x);
```

ideally we would generate a _portable_ binding for `foo`. Currently we don't; we'd generate something like

```hs
module AnotherHeader where

import SomeHeader qualified

foreign import ... "f" f_wrapper :: Int64 -> IO ()

f :: SomeHeader.Foo -> IO ()
f = fromFFIType f_wrapper
```

which is _almost_ what we want, but not quite: the implementation of `f` reflects the machine that we happen to generate these bindings on. 

So while users could write "portable" types, they're essentially unusable in `hs-bindgen` as it stands. Can we do better?

jorisdral:
Resolves #1565
Resolves #1745
Resolves #1747
Resolves #1748

---


hs-bindgen/src-internal/HsBindgen/BindingSpec/Private/V1.hs
```diff
@@ -396,6 +412,12 @@ encodeYaml' = Data.Yaml.Pretty.encodePretty yamlConfig
       "constructor"           -> 18
       -- HsRecordRep:2, HsNewtypeRep:2
       "fields"                -> 19
+      -- HsNewtypeRep:3
+      "ffitype"               -> 20
+      -- HsFFIType: 1
+      "module"                -> 21
+      -- HsFFIType: 2
+      "type"                  -> 22
```
jorisdral:
```suggestion
```

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
```diff
@@ -232,144 +239,198 @@ foreignImportDynamicDec sizeofs name hsType origin =
...
+  | ImportForNewtype {
+        args   :: [Hs.Type]
+      , res    :: Hs.Type
+      , newtyp :: Hs.Newtype
+      }
+
+mapToFFI :: ImportFor -> Size ctx -> SHs.SExpr ctx -> SHs.SExpr ctx
+mapToFFI dynFor size funExpr = case dynFor of
+    ImportForFunction args res -> forFunction args res
+    ImportForNewtype args res nt -> forNewtype args res nt
```
jorisdral: comment

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
```diff
@@ -232,144 +239,198 @@ foreignImportDynamicDec sizeofs name hsType origin =
...
+          , funExpr = eBindgenGlobal HasField_getField `ETypeApp` fieldLit `EApp` funExpr
+          , size    = size
+          })
+      where
+        fieldLit = SHs.translateType $ Hs.StrLit $ Hs.nameToStr nt.field.name
+
+mapFromFFI :: ImportFor -> Size ctx -> SHs.SExpr ctx -> SHs.SExpr ctx
+mapFromFFI dynFor size funExpr = case dynFor of
+    ImportForFunction args res -> forFunction args res
+    ImportForNewtype args res nt -> forNewtype args res nt
```
jorisdral: comment

hs-bindgen/src-internal/HsBindgen/Backend/Hs/Translation/ForeignImport.hs
```diff
@@ -232,144 +239,198 @@ foreignImportDynamicDec sizeofs name hsType origin =
...
+      Hs.PtrArrayElem {}      -> yes Hs.FFIPtrVoid
+      Hs.PtrConstArrayElem {} -> yes Hs.FFIPtrVoid
+      Hs.Ptr{}                -> yes Hs.FFIPtrVoid
+      Hs.FunPtr{}             -> yes Hs.FFIFunPtrVoid
+      Hs.PtrConst{}           -> yes Hs.FFIPtrVoid
+      Hs.IO{}                 -> no
+      Hs.Fun{}                -> no
+      Hs.ExtBinding _ref _cSpec hsSpec t' ->
+        case BindingSpec.hsSpecFFIType hsSpec of
+          -- TODO <https://github.com/well-typed/hs-bindgen/issues/1599>: Ideally we'd warn if a type does not have an FFI type
```
jorisdral: wrap

hs-bindgen/src-internal/HsBindgen/Backend/SHs/Translation/MapFunction.hs
```diff
@@ -0,0 +1,148 @@
+-- | Map a function 'SHs.SExpr' from one type another by converting its
+-- arguments and result
```
jorisdral:
```suggestion
-- | Map a function 'SHs.SExpr' from one type to another by converting its
-- arguments and result
```

hs-bindgen/src-internal/HsBindgen/Backend/SHs/Translation/MapFunction.hs
```diff
@@ -0,0 +1,148 @@
...
+--
+-- * @fun@ is the name of the function we are converting
+-- * @convArg@ is the expression we apply to arguments
+-- * @convRes@ is the expression we apply to the result
+--
+-- Then the result of @mapFunctionExpr@ looks roughly like:
+--
+-- >  \x0 -> \x1 -> \x2 -> \x3 ->
+-- >    convRes (fun (convArg x0) (convArg x1) (convArg x2) (convArg x3))
+--
```
jorisdral: Example: toFFIType and fromFFIType

hs-bindgen/src-internal/HsBindgen/Backend/SHs/Translation/MapFunction.hs
```diff
@@ -0,0 +1,148 @@
...
+applyConvRes :: ConvRes -> Hs.Type -> SHs.SExpr ctx -> SHs.SExpr ctx
+applyConvRes (ConvRes k) = k
+
+
+-- local type synonyms
+type ArgType = Hs.Type
+type ResType = Hs.Type
+type FunExpr ctx = SHs.SExpr ctx
+
+-- | Map a function 'SHs.SExpr' from one type another by converting its
```
jorisdral:
```suggestion
-- | Map a function 'SHs.SExpr' from one type to another by converting its
```

hs-bindgen/src-internal/HsBindgen/Backend/SHs/Translation/Common.hs
```diff
@@ -79,3 +90,119 @@ asNaryTApp = go []
...
+  DeBruijn utilities
+--------------------------------------------------------------------------------}
+
+-- | Reverse an 'Env'
+reverseEnv :: forall ctx a. Env ctx a -> Env ctx a
+reverseEnv = \env -> go (lzeroAdd (sizeEnv env)) EmptyEnv env
+  where
+    go ::
+         forall ctx1 ctx2 ctx3.
+         -- | Proof that the output environment 's size is the same as the the
```
jorisdral:
```suggestion
         -- | Proof that the output environment's size is the same as the the
```

hs-bindgen/src-internal/HsBindgen/Backend/SHs/Translation/Common.hs
```diff
@@ -79,3 +90,119 @@ asNaryTApp = go []
   where
     go acc (TApp f x) = go (x : acc) f
     go acc t          = (t, acc)
+
+{-------------------------------------------------------------------------------
+  Shift
```
jorisdral: comments

PR #2268, opened 2026-09-16
Macro languages parse macro bodies only

dschrempf:
- [x] Based on #2248 which needs to be merged first.

- [x] Involves a `c-expr` bump; see PR https://github.com/well-typed/c-expr/pull/40
  which also needs to be merged first.
Closes #2246 
Related: #2264 #2260

---

- [x] I have updated the changelog. I have included a migration hint if this is a breaking change.

- [x] I have updated the manual. <- Will be a separate PR see #812 

- [x] I have removed all mentions of closed tickets in the code.

- [x] I have associated each new TODO item with a ticket, using the following syntax:

```hs
-- TODO <link to issue>
-- Brief description
```

PR #2272, opened 2026-09-21
Manual: macros

### LINKED ISSUE #812: Manual: macros

dschrempf:
Based on #2268.

Adapts the manual section about macros to accomodate the new macro language changes and the unified macro parse infrastructure in `hs-bindgen`.

Closes #812.

---

```hs
-- TODO <link to issue>
-- Brief description
```


edsko (approved):

dschrempf:
Merging this into #2268, so we only have one merge into `main`.

